PR #75, opened 2024-04-13
fixed language server only running for filetypes with tree-sitter Language

parkerbedlan:
I found that even when I setup lspconfig with something like this:
```lua
lspconfig.harper_ls.setup {
    filetypes = { "markdown", "text", "lua" },
}
```
that the harper linter would only run for lua files, not markdown or text files. I realized that this is because if `new_from_language_id` returns `None`, the linter simply wouldn't run. So I instead had the wildcard pattern return an arbitrary tree-sitter Language, and it fixed the problem.

### DIFF
--- a/harper-ls/src/tree_sitter_parser.rs
+++ b/harper-ls/src/tree_sitter_parser.rs
@@ -30,7 +30,7 @@ impl TreeSitterParser {
             "csharp" => tree_sitter_c_sharp::language(),
             "toml" => tree_sitter_toml::language(),
             "lua" => tree_sitter_lua::language(),
-            _ => return None
+            _ => tree_sitter_toml::language()
         };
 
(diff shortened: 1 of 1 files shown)


elijah-potter:
Thanks for bringing your issue to my attention, unfortunately, I cannot merge this request in its current state.

The lack of support for the `text` filetype is not a bug--there simply isn't any support for it. 
Its somewhat difficult to pin down how to treat those files since they are often not actually English.
Your solution to this seems to be to treat `text` files as `.toml` files, which isn't a robust enough solution for the main repository.

I replaced my `harper-ls` config with your snippet and was unable to replicate your issue. Would it be possible for me to get a copy of your whole Neovim config? I'd love to find the root problem.

elijah-potter:
Are you still having this issue on the latest official builds of Harper?

PR #83, opened 2024-05-06
It is better to use the repository field

szabgab:
To allow [Crates.io](https://crates.io/) , [lib.rs](https://lib.rs/) and the [Rust Digger](https://rust-digger.code-maven.com/) to link to it. See also [the manifest](https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field).


elijah-potter:
Looks perfect to me! This will get placed in the next release.

PR #92, opened 2024-06-07
fix: configuration and legacy numerics

lukasmwerner:
Make it such that `precommit.sh` no longer errors


elijah-potter:
This has already been merged as part of #93.

PR #100, opened 2024-08-05
feat: add LaTeX parser

grantlemons:


lervag:
Just a random thought, if anyone would find it of interest: I've already written a little bit about things like grammar checkers and LSPs in the VimTeX docs. If and when this PR is finished and merged, I think it could be of interest to the community that I also mention Harper as a possible tool for spell and grammar checking (for the English language). I think it should be mentioned under `:help vimtex-af-lsp` and `:help vimtex-grammar`. I would be happy to consider a PR, but also to write something myself. The latter would be easier if someone could help me with providing some good relevant links to how a Vim/Neovim user could "get started" with Harper.

grantlemons:
Update: This is mostly functional, but does not treat macros in paragraphs correctly.
Additionally, it lints package import configuration, which is non-optional.

grantlemons:
To anyone waiting on this, sorry for the lack of progress, LaTeX is such a weird language, and nvim-treesitter itself sometimes has trouble.

The biggest issue is considering the inputs to macros as separate sentences, and giving lints about capitalization.

lervag:
No worries, take your time. It's no surprise, really! 😅

grantlemons:
Note, this isn't actually closed, but I moved it to a branch of the main repository instead of my fork. Not sure I can edit the PR to reflect that though.

pinpox:
> Note, this isn't actually closed, but I moved it to a branch of the main repository instead of my fork

I'm trying to use harper with [typst](https://typst.app/docs/) and would consider contributing support for it. Could your work for latex (which has a similiar use-case) be expanded to that? If not, please link the new PR/branch, maybe I can figure out how to build it from there.

pinpox:
> @pinpox The typst syntax is very different than latex syntax, though, isn't it?

Yes, it is more similar to markdown. typst has treesitter and LSP support though, which if I understand correctly are used for Latex support implementation (correct me if I'm wrong).

I suppose it would be a matter of just interfacing with a different treesitter grammar to extract text parts, but I might be wrong.

grantlemons:
> > @pinpox The typst syntax is very different than latex syntax, though, isn't it?
> 
> Yes, it is more similar to markdown. typst has treesitter and LSP support though, which if I understand correctly are used for Latex support implementation.
> 
> I suppose it would be a matter of just interfacing with a different treesitter grammar to extract text parts, but I might be wrong.

Thankfully, Typst is written in Rust, so the folks over there have already done the work of creating a parsing crate that we can interface with, which is a lot simpler than using the treesitter implementation.

LaTeX in particular is tough because of macros, which are expanded at compile-time, so we can't know how a piece of text is expanded just through grabbing every text node from treesitter.

Issue #116, opened 2024-08-22
VS Code Support


mcecode:
Hi! Previously, I've been using [Grammarly for VS Code](https://github.com/znck/grammarly) as my grammar checker because I couldn't find a better alternative. Recently, it stopped working and I was so happy that I was able to find Harper.

I saw that you created a VS Code extension in `packages`, but haven't published it yet. With some changes, I was able to package it using `vsce` and install it locally. It works great already, gives good feedback and I'm able to do quick fixes.

One suggestion I have is to either bundle it with the latest version of `harper-ls` when publishing it or to set it so that it automatically downloads, updates to, and uses the latest version of `harper-ls` from GitHub releases. I just thought that would make it more convenient for users since they don't have to download it manually themselves. Users who want to opt-out of that can just provide the path to their `harper-ls` installation in the settings and the extension would use that instead.

Something I did notice from the logs though is that even if it says it can't load the dictionary file, it is loading it and is applying the list of words I put there. Here's a screenshot of the error:

![harper-error](https://github.com/user-attachments/assets/2feaf255-492d-41c0-94e4-34c8972b339d)

All in all, this is just my general feedback on the extension and a thanks for creating this project. I'd be willing to help with getting the VS Code extension up and running. I'm sure previous users of Grammarly for VS Code like me would be happy if this got published. I can upstream the changes I made so that it's packageable by `vsce` or I can help with implementing my suggestion above if you're down with that.

elijah-potter:
Wow, I'm seriously impressed you were able to do all that! It's great to hear it's going so smoothly. 

I'm curious what you've changed. Feel free to open a pull request and we'll get `vsce` hooked into CI.

> One suggestion I have is to either bundle it with the latest version of harper-ls when publishing it or to set it so that it automatically downloads, updates to, and uses the latest version of harper-ls from GitHub releases

I've considered both of those options, but I won't have time in the short term to get that implemented myself. If you're interested, feel free to toy around. If you end up getting something working, open a new pull request (or add it onto the other one) and we'll get the extension published.

The "errors" you've found in the logs aren't super self explanatory. They aren't actually errors, just a product of how `harper-ls` manages those dictionaries. I'll remove those logs so as to not confuse anyone else.

mcecode:
> I'm curious what you've changed.

Not much tbh, mostly just adding a `publisher`  field in the `package.json` file and fixing where `main` was pointing.

> Feel free to open a pull request

[...]

PR #117, opened 2024-08-23
feat: make vscode extension packageable by vsce

mcecode:
Related: #116


elijah-potter:
This mostly looks good to me. My only concern is that you've added a new license file to the repo. Would it be possible to use the existing one?

mcecode:
> My only concern is that you've added a new license file to the repo. Would it be possible to use the existing one?

I did worry about this too, but `vsce` doesn't have a flag for setting the path to a license file so I ended up adding one. I've pushed some changes to circumvent this. I used `just` to package the extension, adding a step of copying the license file at the root of the repo to `packages/vscode-plugin` before running `yarn package-extension`, which is just `vsce package`.

Since you previously mentioned hooking `vsce` into the CI, I thought an upside of using `just` is that it'd now be easier to setup an action that can deploy the extension to the Visual Studio Marketplace, as well as to the Open VSX Registry for editors that support the VS Code Extension protocol like VSCodium and Theia to be able to install and use it too.

Here's a possible `deploy_vscode_plugin.yml` workflow I drafted up for doing just that. Everytime a release tag get pushed, say `harper-vscode@0.0.1`, the action deploys the extension using the [Publish VS Code Extension GitHub Action](https://github.com/HaaLeo/publish-vscode-extension):

```yaml
name: Deploy VS Code Plugin

on:
  push:
    tags: [ 'harper-vscode*' ]

jobs:
  deploy:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v4
      - name: Package Extension
        id: package-extension
        uses: extractions/setup-just@v2
        run: |
          just package-vscode
          echo "file=packages/vscode-plugin/harper-$(jq -r .version packages/vscode-plugin/package.json).vsix" >> "$GITHUB_OUTPUT"
      - name: Publish to Visual Studio Marketplace
        uses: HaaLeo/publish-vscode-extension@v1
        with:
          registryUrl: https://marketplace.visualstudio.com
          pat: ${{ secrets.VS_MARKETPLACE_TOKEN }}
          extensionFile: ${{ steps.package-extension.outputs.file }}
      - name: Publish to Open VSX Registry
        uses: HaaLeo/publish-vscode-extension@v1
        with:
          pat: ${{ secrets.OPEN_VSX_TOKEN }}
          extensionFile: ${{ steps.package-extension.outputs.file }}
```

mcecode:
I updated the `precommit.yml` workflow to specify the Node.js version to use since what broke the workflow is `vsce` expecting Node.js version 20 or higher. I think specifying the version is a good idea in general since `actions/setup-node` [recommends this](https://github.com/actions/setup-node#usage):

> The `node-version` input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.

Making the version be inline with the [Active LTS](https://nodejs.org/en/about/previous-releases) would also probably be a good idea since this or higher is what libraries usually expect. Maybe we can document this somewhere with a `.node-version` file or something so it's more clear for new contributors what version to use.

elijah-potter:
This looks good to me. There are a couple minor changes I'd like to make, but I can do those after this PR is merged. If you're curious, keep an eye on the commit log.

Issue #132, opened 2024-09-03
Codeblocks inside rust comments should not be spellchecked


tyilo:
Steps to reproduce:
- Create a new rust project:
```
$ cargo new foo
$ cd foo
```
- Open `neovim` and change the contents of `src/main.rs` to:
```rust
/// ```
/// println!("Test");
/// ```
fn main() {
    println!("Hello, world!");
}
```
- `harper-ls` thinks tries to spellcheck the code inside the doc comment:
![image](https://github.com/user-attachments/assets/6b60f2df-769e-49b6-bf63-c839b4d899ad)

elijah-potter:
Thanks for catching this! This is a somewhat embarrassing issue. I've got a quick fix committed and should be available in the next release (which should be soonish). I'm going to leave this issue open for now, since there is a more robust solution that I don't have time to implement for now.

tyilo:
> In the meantime, [@tyilo](https://github.com/tyilo) could you confirm that the fix is working on your end?

Yes, it seems to work 🙂

PR #136, opened 2024-09-04
fix: remove dbg

StevenLove:
One day ago, [a commit](https://github.com/elijah-potter/harper/commit/bd21b53530469d2b18283ab0aeabdedf8a3342ed) addressing issue #132 introduced a debug statement that pollutes CLI output when running `just dogfood`


elijah-potter:
Thanks for bringing this up. I'll also configure `clippy` to stop this from happening again.

PR #145, opened 2024-09-06
feat: allow to configure file dictionary path

philipp-tailor:
Adds `fileDictPath` configuration, akin to `userDictPath`.

The implementation and documentation mirror the existing implementation of `userDictPath`, and hence only make the smallest set of changes required to provide a configurable file dictionary path.


elijah-potter:
We might have to revisit this once I start work on #147, but it looks great as is!

PR #168, opened 2024-09-17
fix: break words on hyphens for the `an_a` lint

alex-ozdemir:
The sentence "This is not an ML-based system." is grammatical, because the an/a vowel rule applies to "ML" (an initialism).

Another approach would be to add a new document iterator (similar to `iter_words`) that splits on hyphens too, and use that iterator.


elijah-potter:
Looks great! Thanks for fixing it up.

PR #180, opened 2024-09-24
feat: pin nightly version of latest stable rust and add a `just check-rust` command

fisherdarling:
I generally use latest stable rust and found that the project compiles fine with it. However, it looks like harper is using nightly rustfmt arguments. I installed the latest nightly and then found that clippy checks failed. I think it would be useful to pin a specific nightly version using a `rust-toolchain.toml` file.

To fix this, this commit adds a `rust-toolchain.toml` file which installs the nightly version of the latest stable. With the toolchain file, we still get nightly rustfmt features, but everyone uses the same nightly version.

I got the version by going to https://releases.rs and selecting the date at which 1.81 was branched from master.

Also, this added a simple `check-rust` command.

### DIFF
--- a/.github/workflows/build_harper_ls.yml
+++ b/.github/workflows/build_harper_ls.yml
@@ -54,6 +54,7 @@ jobs:
       - name: Build binary
         uses: houseabsolute/actions-rust-cross@v0
         with:
+          toolchain: nightly # we use the nightly version of the latest stable rustc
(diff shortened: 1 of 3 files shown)


elijah-potter:
I appreciate your work on this. My concern is that this sets the Rust version to nightly for _all_ tools in the project. I'd rather set it to the latest stable version and not use the nightly features of `rustfmt`.

PR #186, opened 2024-09-27
Move to stable rustc

fisherdarling:
This PR adds a `rust-toolchain.toml` and switches the repo to use stable rust. This also adds a `just check-rust` command which runs the rust-based lints.

Previous discussion in #180

(commits are just for semantic separation, we should squash merge, or I can squash myself).


elijah-potter:
Looks good. It should shake out once it's been merged and formatted once more.

Issue #166, opened 2024-09-16
LanguageTool comparision?


juanolon:
Hi, this is a great alternative for grammar checkers! Until now, I was using LanguageTool. I would like to know, what are the differences to LanguageTool? not only in speed, but also does it use the same rules? or does it work differently? 
Thanks!

anarcat:
i think having a comparison with other spell checkers like grammalecte and hunspell would be great as well.. From what i gathered, harper imported dictionaries from the latter, but doesn't work in french like the former (nor non-US english, see #113). Not sure about language tool, but i would assume it uses different rules and works differently as well. Certainly languagetool doesn't have LSP support, AFAIK...

elijah-potter:
Sorry it took so long to get a response to you. It looks like @lukasmwerner wrote up a [table](https://github.com/elijah-potter/harper/blob/master/COMPARISON.md) that (might) answer your questions. 

If there's more information that you think needs to be added, say so.

elijah-potter:
I think we can safely close this.

Issue #133, opened 2024-09-03
Mention in the README and website that only American English is supported


tyilo:

elijah-potter:
Looks like @grantlemons handled this. Sorry for the delay!

Issue #113, opened 2024-08-18
Support non-US English


ccjmne:
Hey! 👋

Thanks for this excellent tool!

I'd like to request a way to switch "language flavour", without implementing my dictionaries myself.

I understand that committing to maintaining the built-in dictionaries in one language is already pretty great, and the purpose of this tool was always to be an "English grammar checker", that's its scope and it's very respectable! I have no intention of even suggesting any different!

However, I **wish it would allow me to type `colour`** the way most of the English-speaking world (all but the US) would.
As far as I can tell, the main difference in the grammar (so, not talking about pronunciation, idioms, culture... 😉) between the different English "flavours" lies in the spelling of a few patterns (`ou`/`o`,  `ise`/`ize`, `ogue`/`og`...).

Some words are outright different (`trousers`/`pants`, `university`/`college`), but these are a lot more local than they are global: I'm fairly certain the Singaporeans, New Zealanders, Indians and other English-speaking communities have their own specific overrides and additions there, and I do not suppose that this should be something Harper's built-in dictionaries would need to look into!

To get back to the original topic, I also sorta wish it'd suggest to me:

> Hey, you spelt `subsidized` here whereas you probably meant `subsidised`

So my ideal wouldn't be a mere addition to the default dictionary, but some substitution. I mean, effectively a substitution, I do not know how the solution would work in practice. Is it realistically possible to compute the differences between the two spellings with a set of rules?

I'd like to help if there's anything that can be done! 😄
Alternatively, how would I proceed? I could also just add the differently-spelt variants to my personal dictionaries as I encounter them.

Thanks in advance, and thanks again for Harper!

elijah-potter:
You're in luck! I've been working on significantly enhancing how Harper's dictionary works under the hood. I reckon I can just add flags to different variations of words, as well as an option to set your locale. That way we can continue using the same dictionary, just modifying it to include a bit more information about what version of English each word should be used in.

I'll update here once I've got something working.

anarcat:
I speak french too: are there plans to support entirely different dictionaries and languages here? How about conjugation and grammar? Should I file another issue for this?

hippietrail:
I would like to additionally suggest the ability to base a language on another language.

For instance, Australian, Canadian, and New Zealand English are much more like British than US English, but are not the same and differ from each other. For instance, the rubber part of a wheel is a "tyre" in UK, Australian, and New Zealand English, but is a "tire" in US and Canadian English.

[...]

PR #191, opened 2024-10-01
fix: add blurb on language support and change count of ways to use

grantlemons:
Several issues (#133, #166, #113) I have read refer to the fact that Harper only supports American English, this will add a small blurb about language support and encourage contribution.


elijah-potter:
Looks great! Thanks for addressing those!

Issue #160, opened 2024-09-11
Feature: allow 'misspellings' that are identifiers


dvdsk:
Context: 

I have a doc comment naming an identifier (in this case an parameter of the function). It has an underscore and harper highlights that as a spelling mistake.

Benefits:
- no erroneous warnings
- keeps the docs in sync with the underlying code. Change the name of the parameter of the function and you get a warning :tada:.

Details:
Should probably not allow all identifiers (maybe a config option), function names and parameter names might be a good default.

related: https://github.com/elijah-potter/harper/issues/29

elijah-potter:
This is actually already implemented for identifiers that aren't either snake_case or chain-case. I'll look into getting this done sometime early next week. Sorry for the delay!

elijah-potter:
It looks like @grantlemons is looking into this. Grant: I'd suggest creating a new Harper `Parser` that wraps another parser. Have it collapse any token strings that match the pattern `Word` `Underscore` `Word` or `Word` `Hyphen` `Word` AND exist in a provided identifier dictionary. Let me know if you have any questions.

elijah-potter:
We're actually going to reopen this issue until we have a release out that @dvdsk can confirm no longer has the problem.

dvdsk:
Thank you all very much for working on this! Ill go try it out immediately and let you know how it works after the weekend :+1:

dvdsk:
Worked in most cases :tada:. I ran into one issue thats still there but I think it has a different cause. So great work thanks you all very much! I will make a new report for the new issue.

dvdsk:
nvmd it all works perfectly! No other issues on my end :partying_face:

dvdsk:
Apologies if this is spamy. But this feature just really helped my out. I refactored an API and forgot to update the docs. Harper gave me spell errors for words that where identifiers of the previous API. 

Very very nice

elijah-potter:
Not spamy at all. That actually makes us so happy. Glad Harper is doing something good.

PR #199, opened 2024-10-03
feat: Collapse identifiers seperated by hyphens or underscores if they exist in the file dictionary

grantlemons:


grantlemons:
Tests need to be fixed.

harper-core/src/token.rs
```diff
@@ -242,6 +242,11 @@ impl TokenKind {
         matches!(self, TokenKind::Punctuation(Punctuation::At))
     }
 
+    pub fn is_case_separator(&self) -> bool {
```
grantlemons, line 245: This should probably be two separate functions, but it's probably OK for now. If someone really wants it I can figure out how to do OR in pattern matching.

harper-core/src/parsers/cases.rs
```diff
@@ -0,0 +1,247 @@
...
+
+thread_local! {
+    static WORD_OR_NUMBER: Lrc<SequencePattern> = Lrc::new(SequencePattern::default()
+                .then_any_word()
+                .then_one_or_more(Box::new(SequencePattern::default()
+        .then_case_separator()
+        .then_any_word())));
+}
+
+impl Parser for Cases {
```
lukasmwerner, line 34: I'm currently getting a compile time error when building for `harper-cli`:
```rust
error[E0277]: `Rc<MergedDictionary<FullDictionary>>` cannot be shared between threads safely
  --> harper-core/src/parsers/cases.rs:34:17
   |
34 | impl Parser for Cases {
   |                 ^^^^^ `Rc<MergedDictionary<FullDictionary>>` cannot be shared between threads safely
   |
   = help: within `Cases`, the trait `Sync` is not implemented for `Rc<MergedDictionary<FullDictionary>>`, which is required by `Cases: Sync`
note: required because it appears within the type `Cases`
  --> harper-core/src/parsers/cases.rs:12:12
   |
12 | pub struct Cases {
   |            ^^^^^
note: required by a bound in `parsers::Parser`
  --> harper-core/src/parsers/mod.rs:15:26
   |
15 | pub trait Parser: Send + Sync {
   |                          ^^^^ required by this bound in `Parser`

error[E0277]: `Rc<MergedDictionary<FullDictionary>>` cannot be sent between threads safely
  --> harper-core/src/parsers/cases.rs:34:17
   |
34 | impl Parser for Cases {
   |                 ^^^^^ `Rc<MergedDictionary<FullDictionary>>` cannot be sent between threads safely
   |
   = help: within `Cases`, the trait `Send` is not implemented for `Rc<MergedDictionary<FullDictionary>>`, which is required by `Cases: Send`
note: required because it appears within the type `Cases`
  --> harper-core/src/parsers/cases.rs:12:12
   |
12 | pub struct Cases {
   |            ^^^^^
note: required by a bound in `parsers::Parser`
  --> harper-core/src/parsers/mod.rs:15:19
   |
15 | pub trait Parser: Send + Sync {
   |                   ^^^^ required by this bound in `Parser`

For more information about this error, try `rustc --explain E0277`.
```

lukasmwerner (changes requested):
Works great in harper-ls. Just a small issue for harper-cli

harper-core/src/parsers/cases.rs
grantlemons: This was an issue I was having with tests when running it in the `harper-core` repository, it has to do with the `Lrc` definition, but I'm not quite sure how to fix it.

grantlemons: I implemented a fix... but obviously no warnings doesn't mean I didn't screw something up performance-wise. I'll need to get elijah to look at it to figure out what's going on with the concurrent feature.

harper-core/src/token.rs
elijah-potter: It's probably fine to leave this here, but if you do, add a doc-comment that explains what a "case separator" is and why it's needed.

I would much rather create an OR `Pattern` and use that though, since we need one anyway. It's up to you which you decide to do (laziness is totally fine in this case).

harper-core/src/parsers/cases.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// This implementation is quite gross to look at, but it works.
```
elijah-potter, line 35: Looks like this got copied from `Markdown`. I'd toss it.

harper-core/src/parsers/cases.rs
```diff
@@ -0,0 +1,247 @@
...
+
+use itertools::Itertools;
+
+use super::{Parser, TokenKind};
+use crate::patterns::{PatternExt, SequencePattern};
+use crate::{Dictionary, FullDictionary, MergedDictionary, Span, Token, VecExt, WordMetadata};
+
+/// A parser that wraps any other parser to collapse token strings that match
+/// the pattern word_word or word-word.
+pub struct Cases {
```
elijah-potter, line 12: Is there a more descriptive name for `Cases`? Maybe something like `CollapseIdentifiers`?

harper-core/src/parsers/mod.rs
```diff
@@ -1,14 +1,23 @@
...
 
 use blanket::blanket;
+pub use cases::Cases;
 pub use markdown::Markdown;
 pub use mask::Mask;
 pub use plain_english::PlainEnglish;
 
 pub use crate::token::{Token, TokenKind, TokenStringExt};
 
+#[cfg(not(feature = "concurrent"))]
```
elijah-potter, line 14: I'm surprised this wasn't already here. How was `harper-cli` able to compile?

harper-core/src/parsers/cases.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::parsers::{PlainEnglish, StrParser};
+
+    use super::*;
+
+    #[test]
+    fn no_collapse() {
```
elijah-potter, line 82: I adore these test cases. Great work.

elijah-potter:
Overall, this is really great work @grantlemons. I've left a few (nitpicky) comments on some bits. They should be pretty straightforward changes.

Other than those, would you mind changing your commit messages to match [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)? Feel free to force-push to the feature branch.

I would also mention issue #160 in them:

```
fix(#160): yadda yadda yadda
```

harper-core/src/parsers/mod.rs
grantlemons: Idk, I just checked, and it turns out `mask::Masker` uses Sync too and doesn't have a concurrent cfg.

PR #200, opened 2024-10-03
VS Code Support

### LINKED ISSUE #116: VS Code Support

mcecode:
Hi, I finally got around to finishing things and making this PR. You may have already seen some of these changes since I saw that you starred my fork a few days ago, but here's a breakdown of what I've done and some notes about them.

## Extension

- The extension now expects `harper-ls` to be bundled with it.
- Users can now manually restart `harper-ls` if they want.
- When a user changes any setting contributed by the extension, `harper-ls` is now automatically restarted so those changes are reflected right away.
- Added icon for the Marketplace.

## Tests

I've added some integration tests, mostly to serve as a smoke test to ensure that the extension works with the current state of `harper-ls`. In the process, I removed unused scripts and dependencies in `package.json` and cleaned up the files in `.vscode` a bit.

## `justfile`

- Added `test-vscode` recipe.
- Updated `package-vscode` recipe, mostly so it can be used by the workflow discussed below.
- Opted to use `test-vscode` instead of `package-vscode` in both `precommit`, because it's a better gauge if the extension is working properly, and `setup`, because testing downloads extra resources that packaging doesn't.

## Docs

- Added an "Unreleased" section in `CHANGELOG.md` for the extension changes I've done.
- Updated `README.md` for the Marketplace. I intentionally left the "Installation" section empty for when you're ready to release or have released the extension as people might be misled when they get an error trying to install `elijah-potter.harper` and get an error.
- Added `development-guide.md` to make it easier for those who want to contribute.

## Publishing Workflow

I created a workflow to assist in publishing the extension, basing on `build_harper_ls.yml`, it compiles `harper-ls` for every platform currently supported, packages it with the extension, then publishes the extension for each platform. It can be triggered by pushing a tag prefixed with `harper-vscode` or manually in the GitHub Actions UI. For the workflow to work, you need to [create the `elijah-potter` publisher](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#create-a-publisher), [get a Personal Access Token (PAT)](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#get-a-personal-access-token), and create a [secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions) named `VSCE_PAT` set to the value of the PAT you previously obtained.

My idea with this is that the extension would have the same version as `harper-ls` so it's easy for users (and us if they file an issue) to know that extension version X is also `harper-ls` version X. So when publishing with this workflow, the main manual work that needs to be done before running it is to update the `version` in `package.json` and maybe change the "Unreleased" section in `CHANGELOG.md`, if there is any, to the new version. That said, I'm open to other ideas, maybe you prefer a different way of publishing the extension or you want to reduce the manual steps or maybe you want to test the extension first before it's packaged to be sure it works.

Note, you may notice that I added `"endOfLine": "auto"` to `.prettierrc` and that's because Prettier throws a fit in Windows without this. Here's an example of a run where [everything is okay with macOS and Linux but not Windows](https://github.com/mcecode/harper/actions/runs/10961513550).

Closes #116


.github/workflows/publish_vscode_plugin.yml
```diff
@@ -0,0 +1,75 @@
+name: Publish VS Code Plugin
```
elijah-potter, line 1: I'd love to make this actions workflow simply build the plugin artifact and make it available on every push to master.

That will make it easy to publish manually with everything else, and make it easier to review changes to `harper-ls` and the plugin when pull requests are made.

I'd change the naming from "publish" to "package" and remove the actual publish step. See my other comment.

justfile
```diff
@@ -41,17 +41,59 @@ build-obsidian:
 
   zip harper-obsidian-plugin.zip manifest.json main.js
 
+test-vscode:
```
elijah-potter, line 44: This is great.

packages/vscode-plugin/.vscode/settings.json
```diff
@@ -1,13 +1,6 @@
...
 {
-	"files.exclude": {
-		"out": false, // set this to true to hide the "out" folder with the compiled JS files
-		"dist": false // set this to true to hide the "dist" folder with the compiled JS files
-	},
-	"search.exclude": {
-		"out": true, // set this to false to include "out" folder in search results
-		"dist": true // set this to false to include "dist" folder in search results
-	},
+	"search.exclude": { "build": true },
```
elijah-potter, line 3: I'm not familiar with VSCode settings like this. I'm going assume they make sense. Maybe someone like @lukasmwerner has an opinion?

packages/vscode-plugin/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
```
elijah-potter, line 3: This looks good. We'll just have to remember to swap this out when bump the version. I might write a script for this.

packages/vscode-plugin/README.md
```diff
@@ -1,30 +1,38 @@
...
+
+## Features
+
+### Commands
+
+| Command                         | Id                              | Description         |
+| ------------------------------- | ------------------------------- | ------------------- |
+| Harper: Restart Language Server | `harper.languageserver.restart` | Restart `harper-ls` |
+
+### Settings
```
elijah-potter, line 15: Is it necessary to include linter descriptions in the README? I would imagine they should be self-explanatory enough to stand alone in the settings page. I'm not a huge fan of the duplication with `package.json`.

packages/vscode-plugin/development-guide.md
```diff
@@ -0,0 +1,48 @@
+# Development Guide
```
elijah-potter, line 1: 10/10. Fantastic.

packages/vscode-plugin/package.json
```diff
@@ -81,13 +92,13 @@
 					"scope": "resource",
 					"type": "boolean",
 					"default": true,
-					"description": "Make sure you close you quotation marks."
+					"description": "Make sure you close your quotation marks."
```
elijah-potter, line 95: Oof. That's embarrassing on my end.

elijah-potter:
I would like to test this out on a couple systems other than my own. @lukasmwerner can you confirm that this works on ARM Mac? Make sure you remove your existing `harper-ls` before testing.

lukasmwerner:
Confirmed working on M1 running Sonoma 14.0! Loads the bundled `harper-ls` build perfectly. No issues, even after adding my local install of `harper-ls` back, very impressive!

packages/vscode-plugin/.vscode/settings.json
lukasmwerner: Yeah these all look reasonable to me.

packages/vscode-plugin/README.md
mcecode: I understand. I'd also prefer not to duplicate them, but I just thought it might help people who find out about Harper from the VS Code Marketplace first to get an overview of what exactly it checks for. However, I can remove it if you think it is unnecessary.

servusdei2018 (approved):

elijah-potter:
I've done some testing and it looks like our current CI setup for Windows is broken. See #208. I'd like to get this fixed before merging.

elijah-potter:
Sorry this has takes so long to get through. I'm going to merge it and cut a release as soon as possible. Thank you so much for all you work.

Issue #198, opened 2024-10-03
harper-ls --stdio using too many threads


acampove:
Hi,

I just downloaded this with Mason and I am using it with neovim. I suddenly found out that I cannot spawn threads and when I check with htop I see that harper-ls is using about 150 theads:

![image](https://github.com/user-attachments/assets/7f555a7c-fade-4c49-8d89-20ecff3a8b89)

Why is this happening? How can I configure this tool to use only  one thread?

Cheers.

grantlemons:
This should be configured by tokio to be the number of cores on your machine, how many do you have?

acampove:
> This should be configured by tokio to be the number of cores on your machine, how many do you have?

I probably have around 128 cores. Do you know how to configure this using? I definitely do not think the program should be allowed to use all the cores. It should be limited to 2, at most 3. I am seeing the same problem with pyright, BTW.

elijah-potter:
Ah. This should be a simple fix. We'll set `harper-ls` to cap out at 4 threads. That should be more than enough.

elijah-potter:
@grantlemons drafted a fix and it has been merged. I'm going to spend some time with Harper this afternoon to make sure everything checks out before creating a release. At that point, you can confirm that the problem is fixed.

acampove:
@grantlemons Thanks, could you please make another tag, 0.12.0? I am using Mason and this seems to download the latest tag.

elijah-potter:
Sorry it took so long to get back to you on this. The Mason maintainer is extremely unresponsive when it comes to PRs. You should be able to get a new version now. Would you mind confirming whether your issue has been fixed?

elijah-potter:
I'm assuming we're good here. Thanks for letting us know about the problem!

acampove:
Hi @elijah-potter thanks. I am having other issues with neovim, this time memory issues, but I think I haven't really seen any problem with this plugin lately, so I assume the update fixed the issue.

elijah-potter:
I'm sorry about the other issues, but I'm glad Harper's working. Thanks again!

PR #202, opened 2024-10-04
fix(#198): limit tokio threads to 4 (+ main)

grantlemons:
[Tokio has two different types of threads](https://github.com/tokio-rs/tokio/discussions/3858#discussioncomment-869878):
- Worker threads: for async things
- Blocking threads: main and spawn_blocking

I limited the number of worker threads to 4, but the actual number of threads should hover around 5.
I'm not certain whether we spawn any blocking threads other than main, but if we do it should be minimal.

### DIFF
--- a/harper-ls/src/main.rs
+++ b/harper-ls/src/main.rs
@@ -31,7 +31,9 @@ struct Args {
     stdio: bool,
 }
 
-#[tokio::main]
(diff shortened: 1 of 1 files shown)


elijah-potter:
Looks good to me. Unless `tower-lsp` spawns any blocking threads, there isn't anywhere where we do.

Issue #138, opened 2024-09-05
Use FST as dictionary?


PSeitz:
Just wanted to ask if you considered using FST as the dictionary. It allows efficient fuzzy queries with levenshtein distance for spelling correction. Look ups are also quite fast, and it provides pretty good compression. Iteration would be slower than a vec, but not sure why you would need that with the features it provides.

FST dicitonary
https://docs.rs/fst/latest/fst/
FST fuzzy queries
https://docs.rs/levenshtein_automata/latest/levenshtein_automata/

elijah-potter:
I did not consider using an FST, although it looks exceptionally cool. I'll toy around with it as soon as I have time. It should be pretty trivial to get it integrated into Harper (and hopefully speed up those spellcheck times).

Thanks for bringing it up!

grantlemons:
This would be very cool, but I started looking into this, and it doesn't look like a drop-in replacement to me. The biggest differences are these:
>  1. Once constructed, a Map can never be modified.
>  2. Maps must be constructed with lexicographically ordered byte sequences. There is no restricting on the ordering of values.

The existence of MergedDictionary might make this feasible.

Fuzzy finding would be nice, but the documentation has this to say:
> Warning: Levenshtein automatons use a lot of memory
> 
> The construction of Levenshtein automatons should be consider “proof of concept” quality. Namely, they do just enough > to be correct. But they haven’t had any effort put into them to be memory conscious.
> 
> Note that an error will be returned if a Levenshtein automaton gets too big (tens of MB in heap usage).

I'm not 100% sure how big of an issue this would be, but my understanding is that one of Harper's motivations is to have a small memory footprint (compared to LanguageTool).

grantlemons:
I'm going to do a proof of concept implementation regardless, and just see how everything works out.

PSeitz:
You probably should use https://crates.io/crates/levenshtein_automata for fuzzy search, as the fst one is broken. I don't think memory consumption will be an issue.

grantlemons:
Can you elaborate on what you mean by broken?

PSeitz:
https://github.com/BurntSushi/fst/issues/38

grantlemons:
From what I read there, it looks like it is only broken for japanese characters (or maybe complex unicode in general), but considering Harper only supports english for the forseable future, I'll just make a note of it and continue with Fst in my proof of concept.

PR #258, opened 2024-11-02
feat: FST-based Curated Dictionary Spellchecking

grantlemons:
Created a new dictionary `FstDictionary` that uses a finite-state transducer (FST)-based map underneath. This precomputed map is built based on the dictionary file at compile time, and allows for extremely fast edit-distance calculations used in the spellchecking logic.

`FstDictionary` is ideal for the curated dictionary, but is immutable, so the current `FullDictionary` implementation is still used for user and file dictionaries. 

Using this new dictionary results in much faster spellchecking (about 10x faster when not yet cached per benchmark below).
image

One downside to this PR is that it moves the hunspell parsing stuff out into a new crate `harper-dictionary-parsing`, which makes some of the imports a little weird. For instance, `harper-core` now imports `Span` from `harper-dictionary-parsing`.

~Also, there is a noted bug in the `fst` library with its handling of japanese characters, but for the time being this should not pose an issue as multi-language support is nowhere near being implemented. (#138, [fst/#38](https://github.com/BurntSushi/fst/issues/38))~


grantlemons:
Made it more correct (better utf-8 edit-distance support using the levenshtein-automata crate mentioned in #138) and a tiny bit faster by switching levenshtein automata implementations and storing the DFA builders thread-locally 💪

image

harper-dictionary-parsing/src/lib.rs
```diff
@@ -1,24 +1,30 @@
 mod affix_replacement;
```
elijah-potter, line 1: Would you mind writing up a `README.md` for this crate? If so, make sure you link it in this file the same way they are for the other crates.

harper-dictionary-parsing/src/lib.rs
```diff
@@ -1,24 +1,30 @@
...
 use self::word_list::parse_word_list;
 pub use self::word_list::MarkedWord;
 
 pub fn parse_default_word_list() -> Result<Vec<MarkedWord>, Error> {
-    parse_word_list(include_str!("../../../dictionary.dict"))
+    parse_word_list(include_str!("../dictionary.dict"))
 }
 
 pub fn parse_default_attribute_list() -> AttributeList {
     let human_readable: HumanReadableAttributeList =
```
elijah-potter, line 26: What are your thoughts on just moving all the dictionaries into this crate? That way, people (including us) can consume them on their own, without any of the rest of the Harper baggage.

harper-core/build.rs
```diff
@@ -0,0 +1,47 @@
...
+    attr_list.expand_marked_words(word_list, &mut word_map);
+
+    let mut words: Vec<CharString> = word_map.iter().map(|(w, _)| w.clone()).collect();
+    // This may seem weird, but it ensures that the indexes from enumerate below match with those
+    // of the words array in FullDictionary.
+    words.sort_unstable();
+    words.dedup();
+    words.sort_by_key(|w| w.len()); // DO NOT MAKE THIS UNSTABLE.
+
+    // Using u64 shouldn't pose any issues since I don't think the English
```
elijah-potter, line 30: I don't think this comment makes sense. Was it `u32` previously?

harper-dictionary-parsing/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+name = "harper-dictionary-parsing"
+version = "0.12.0"
+edition = "2021"
+description = "The language checker for developers."
+license = "Apache-2.0"
+repository = "https://github.com/elijah-potter/harper"
+
+[dependencies]
+blanket = "0.4.0"
+hashbrown = { version = "0.15.0", features = ["serde"] }
```
elijah-potter, line 11: `cargo +nightly udeps` suggests that `blanket`, `itertools`, `ordered-float`, `paste`, `unicode-blocks`, and `unicode-width` aren't needed.

harper-dictionary-parsing/src/char_string.rs
```diff
@@ -16,4 +17,7 @@ impl CharStringExt for [char] {
 
         out
     }
+    fn to_string(&self) -> String {
```
elijah-potter, line 20: Would it be possible to implement `Display` instead?

harper-dictionary-parsing/src/span.rs
```diff
@@ -11,9 +11,7 @@ pub struct Span {
```
elijah-potter, line 11: I don't love the idea of putting `Span` in this crate. It feel somewhat arbitrary. Would it be possible to put it back in `harper-core` or perhaps a new crate?

harper-dictionary-parsing/src/span.rs
```diff
@@ -11,9 +11,7 @@ pub struct Span {
 
 impl Span {
     pub fn new(start: usize, end: usize) -> Self {
-        if start > end {
```
elijah-potter, line 14: Let's keep this the way it is in `master`. You might be able to cherry-pick it.

harper-dictionary-parsing/src/char_string.rs
```diff
@@ -6,6 +6,7 @@ pub type CharString = SmallVec<[char; 12]>;
 
 pub trait CharStringExt {
     fn to_lower(&self) -> CharString;
```
elijah-potter, line 8: Similar to `Span`: I don't think this should be defined in this crate. I think it would perfectly OK to define it in `harper-core` and include `harper-core` as a dependency.

harper-core/src/parsers/collapse_identifiers.rs
```diff
@@ -1,24 +1,24 @@
...
 use crate::patterns::{PatternExt, SequencePattern};
-use crate::{Dictionary, FullDictionary, MergedDictionary, Span, Token, VecExt};
+use crate::{Dictionary, Lrc, Span, Token, VecExt};
 
 /// A parser that wraps any other parser to collapse token strings that match
 /// the pattern `word_word` or `word-word`.
 pub struct CollapseIdentifiers {
     inner: Box<dyn Parser>,
-    dict: Lrc<MergedDictionary<FullDictionary>>,
+    dict: Arc<dyn Dictionary>,
```
elijah-potter, line 14: Love this change.

harper-core/build.rs
```diff
@@ -0,0 +1,47 @@
...
+use fst::MapBuilder;
+use harper_dictionary_parsing::{
+    parse_default_attribute_list, parse_default_word_list, CharString,
+};
+use hashbrown::HashMap;
+use itertools::Itertools;
+
+fn main() {
+    let wtr = io::BufWriter::new(File::create("dictionary.fst").unwrap());
+    let mut build = MapBuilder::new(wtr).unwrap();
```
elijah-potter, line 13: If we move the dictionaries into `harper-dictionary-parsing`, could we use `FstDictionary::new()` instead of building it manually?

harper-core/src/spell/full_dictionary.rs
```diff
@@ -62,7 +67,7 @@ impl FullDictionary {
 
     /// Create a dictionary from the curated dictionary included
     /// in the Harper binary.
-    pub fn curated() -> Lrc<Self> {
+    pub(super) fn curated() -> Arc<Self> {
```
elijah-potter, line 70: Rather than making it private to the module, add a comment warning people which situations `FstDictionary` is a better choice.

harper-core/src/spell/full_dictionary.rs
```diff
@@ -172,18 +167,101 @@ impl Dictionary for FullDictionary {
...
+    fn fuzzy_match(
+        &self,
+        word: &[char],
+        max_distance: u8,
+        max_results: usize,
+    ) -> Vec<(&[char], u8, WordMetadata)> {
+        let misspelled_charslice = seq_to_normalized(word);
+
+        let shortest_word_len =
+            std::cmp::max(1, misspelled_charslice.len() - max_distance as usize);
```
elijah-potter, line 184: This modification to the algorithm results in integer underflow for short words.

harper-core/src/spell/mod.rs
```diff
@@ -212,31 +174,248 @@ mod tests {
...
+        );
+
+        dbg!(&results);
+
+        assert!(results.iter().take(3).contains(&"hello".to_string()));
+    }
+
+    #[test]
+    fn fst_spellcheck_capital() {
+        let results = suggest_correct_spelling_str(
```
elijah-potter, line 370: I don't see the utility of this test.

harper-core/src/spell/mod.rs
```diff
@@ -212,31 +174,248 @@ mod tests {
...
+            &FstDictionary::curated(),
+        );
+
+        dbg!(&results);
+
+        assert!(results.iter().take(3).contains(&"hello".to_string()));
+    }
+
+    #[test]
+    fn full_spellcheck_match() {
```
elijah-potter, line 355: This is testing a situation that should never occur.

harper-core/src/spell/mod.rs
```diff
@@ -212,31 +174,248 @@ mod tests {
...
+            &FullDictionary::curated(),
+        );
+
+        dbg!(&results);
+
+        assert!(results.iter().take(3).contains(&"hello".to_string()));
+    }
+
+    #[test]
+    fn spellchecking_is_deterministic() {
```
elijah-potter, line 397: Love it.

elijah-potter (changes requested):
Grant, this is some really great work. There are a couple of test cases and organizational things that need to be looked at, but overall it's looking very good, particularly for the WASM target.

Let me know if you have any questions about my comments.

harper-dictionary-parsing/src/char_string.rs
grantlemons: Only traits defined in the crate can be implemented on types outside of the crate, so I don't think so, no.

harper-dictionary-parsing/src/span.rs
grantlemons: Putting it in a new crate is probably best, but we don't want to make too many crates! Maybe `harper-lib`?

harper-dictionary-parsing/src/char_string.rs
grantlemons: The issue is that it gets used in harper-core's build script, so in doing so we create a circular dependency. A new crate would be best, but we already have quite a few.

harper-dictionary-parsing/src/span.rs
grantlemons: Did that.

harper-core/build.rs
grantlemons: can you elaborate on what you mean by this?

grantlemons: Nevermind, I think I get what you're saying. Yeah, I think we can do that.

harper-dictionary-parsing/src/lib.rs
grantlemons: I just started trying this, but the issue is dependencies again. Because the build script depends on the parsing stuff, and the crate contains this, the `dictionary.fst` file has to be generated in `harper_core`, and since `FstDictionary` depends on `dictionary.fst` at compile-time, it has to be in `harper_core` as well.

harper-dictionary-parsing/src/span.rs
grantlemons: I don't remember making this change, which makes me assume clippy told me to do it. Regardless, I believe they have pretty much identical behavior, but the new one is better style imo.

Why do you want it the original way?

elijah-potter: Use the `panic!` method. It makes it _so_ much easier to debug related backtraces.

harper-core/Cargo.toml
```diff
@@ -21,9 +22,10 @@ smallvec = { version = "1.13.2", features = ["serde"] }
 thiserror = "1.0.65"
 unicode-blocks = "0.1.9"
 unicode-width = "0.2.0"
+levenshtein_automata = { version = "0.2.1", features = ["fst_automaton"] }
 
 [dev-dependencies]
-criterion = "0.5.1"
+criterion = { version = "0.5.1", features = ["plotters", "cargo_bench_support"], default-features = false}
```
elijah-potter, line 28: Unless this directly effects our ability to run the benchmarks, let's make these changes in a separate PR.

harper-core/src/parsers/collapse_identifiers.rs
```diff
@@ -1,24 +1,24 @@
...
 }
 
 impl CollapseIdentifiers {
-    pub fn new(inner: Box<dyn Parser>, dict: &Lrc<MergedDictionary<FullDictionary>>) -> Self {
+    pub fn new(inner: Box<dyn Parser>, dict: Box<Arc<dyn Dictionary>>) -> Self {
```
elijah-potter, line 18: Is there a reason this is an `Arc` instead of an `Lrc`?

harper-core/src/spell/full_dictionary.rs
```diff
@@ -27,7 +32,7 @@ pub struct FullDictionary {
 
 /// The uncached function that is used to produce the original copy of the
 /// curated dictionary.
-fn uncached_inner_new() -> Lrc<FullDictionary> {
+fn uncached_inner_new() -> Arc<FullDictionary> {
```
elijah-potter, line 35: Why did you swap this with an `Arc`?

harper-core/src/spell/full_dictionary.rs
```diff
@@ -103,11 +113,13 @@ impl FullDictionary {
     /// list. NOTE: This function will sort the original word list by its
     /// length. If the word list's order is changed after creating the
     /// lookup, it will no longer be valid.
-    fn create_len_starts(words: &mut [CharString]) -> Vec<usize> {
-        words.sort_by_key(|a| a.len());
+    fn create_len_starts(words: &[CharString]) -> Vec<usize> {
```
elijah-potter, line 116: I'm not directly opposed to making the argument here immutable, I'm just curious your reasoning.

harper-core/src/spell/full_dictionary.rs
```diff
@@ -116,6 +128,13 @@ impl FullDictionary {
 
         word_len_starts
     }
+
+    pub fn get_word(&self, index: usize) -> &CharString {
```
elijah-potter, line 132: I'm not a huge fan of making the underlying indices of the word list part of the API for this struct. Is there a way to avoid this?

harper-core/src/spell/full_dictionary.rs
```diff
@@ -125,29 +144,9 @@ impl Default for FullDictionary {
...
-            self.word_len_starts[len + 1]
-        };
-
-        Box::new(self.words[start..end].iter().map(|v| v.as_slice()))
-    }
-
     fn get_word_metadata(&self, word: &[char]) -> WordMetadata {
         let normalized = seq_to_normalized(word);
-        let lowercase: CharString = normalized.iter().flat_map(|c| c.to_lowercase()).collect();
+        let lowercase: CharString = normalized.to_lower();
```
elijah-potter, line 149: Same comment as for `get_word`.

harper-core/src/spell/merged_dictionary.rs
```diff
@@ -1,62 +1,42 @@
...
-#[derive(Clone, PartialEq)]
-pub struct MergedDictionary<T>
-where
-    T: Dictionary + Clone,
-{
-    children: Vec<Lrc<T>>,
-    merged: HashMap<CharString, WordMetadata>,
+#[derive(Clone)]
+pub struct MergedDictionary {
+    children: Vec<Arc<dyn Dictionary>>,
```
elijah-potter, line 12: Same question: why `Arc` instead of `Lrc`?

harper-core/src/spell/dictionary.rs
```diff
@@ -2,17 +2,26 @@ use blanket::blanket;
...
+        max_distance: u8,
+        max_results: usize,
+    ) -> Vec<(&[char], u8, WordMetadata)>;
+    /// Gets best fuzzy match from dictionary
+    fn fuzzy_match_str(
+        &self,
+        word: &str,
+        max_distance: u8,
+        max_results: usize,
+    ) -> Vec<(&[char], u8, WordMetadata)>;
```
elijah-potter, line 24: Can we make this tuple its own struct? Maybe `FuzzyMatchResult`?

elijah-potter (changes requested):
I've got some more requests for changes. Still looks good. Let me figure out the performance problem.

harper-core/src/spell/full_dictionary.rs
grantlemons: This is because the `FstDictionary` needs to index into `FullDictionary`. I guess we can make it `pub(super)`?

harper-core/src/parsers/collapse_identifiers.rs
grantlemons: I don't remember exactly why, but there were issues with `MergedDictionary` and using `Lrc`. [Here's the commit in question if you want to find another way](https://github.com/elijah-potter/harper/pull/258/commits/ee77e5d9ab66bfdbcf35fdeba8a90892be8f9145)

grantlemons: If I remember correctly, it had to do with the `flat_map` in `MergedDictionary`'s `words_iter()` function.

harper-core/src/spell/full_dictionary.rs
grantlemons: I feel like it just makes the logic of sorting by length more explicit in the constructor. I had trouble realizing that was occuring because I expected this function not to have side-effects.
This is especially important because if the way that the words array is sorted changes in any way, it needs to match the sorting when creating the precomputed map.

harper-core/src/spell/mod.rs
```diff
@@ -12,96 +10,126 @@ pub use self::merged_dictionary::MergedDictionary;
...
+mod hunspell;
 mod merged_dictionary;
 
-fn order_suggestions(matches: Vec<(&[char], u8, WordMetadata)>) -> Vec<&[char]> {
-    let mut found: Vec<(&[char], u8, WordMetadata)> = Vec::with_capacity(matches.len());
+/// Suggest a correct spelling for a given misspelled word.
+/// [`misspelled_word`] is assumed to be quite small (n < 100).
+/// [`max_edit_dist`] relates to an optimization that allows the search
+/// algorithm to prune large portions of the search.
+pub fn suggest_correct_spelling<'a>(
```
grantlemons: Why are we overriding the logic from the individual dictionaries and just using the `FullDictionary` implementation?

grantlemons: And not even the current one, this is the original implementation from `master` iirc?

grantlemons: Spoke offline, this was unintentional.

harper-core/src/spell/full_dictionary.rs
elijah-potter: Alright. Well, if you change the behavior of the function, make sure you update that function's comment (which claims the function mutates the provided word list).

harper-core/src/spell/full_dictionary.rs
elijah-potter: It did not need to. I fixed it.

harper-core/src/spell/mod.rs
```diff
@@ -2,132 +2,118 @@ use std::borrow::Cow;
...
-    // Let common words bubble up.
-    found.sort_by_key(|v| {
-        if dictionary.get_word_metadata(v).common {
-            0
-        } else {
-            1
-        }
-    });
+    // Let common words bubble up, but do not prioritize them over all else.
+    found.sort_by_key(|fmr| fmr.edit_distance + if fmr.metadata.common { 0 } else { 1 });
```
elijah-potter, line 62: Why did you do this?

harper-core/src/spell/mod.rs
```diff
@@ -212,31 +188,200 @@ mod tests {
...
+        assert_eq!(edit_distance(&source, &target), 1);
+        assert_eq!(edit_distance(&target, &source), 1);
+
+        let target: Vec<_> = "Hello".chars().collect();
+        assert_eq!(edit_distance(&source, &target), 1);
+        assert_eq!(edit_distance(&target, &source), 1);
+    }
+
+    #[test]
+    fn common_words_first() {
```
elijah-potter, line 246: I'm a little confused. What is this test doing? It feels really complicated and difficult to read.

harper-core/src/spell/mod.rs
grantlemons: I found we got better results this way. Before, any common word is prioritized from the hundred results, which doesn't make much sense.

If there are three common words out of the hundred results, then all of them would be prioritized over an uncommon word of edit distance 1.

This way, being uncommon just acts as if it were one further away.

I'm fine with getting rid of this if you want, iirc it was something I was trying when I was working on the `I'm` test, but it had good results so I kept it.

harper-core/src/spell/mod.rs
grantlemons: This just checked that the algorithm prioritizes common words by getting both common and uncommon words and then running the algorithm over them.

I agree it is a poorly written test though.

elijah-potter: Would you mind adding a bit of documentation that makes it clearer? I'm sorry for all my nitpicks, but this should be the last one!

harper-core/src/spell/mod.rs
elijah-potter: Leave it as-is. It's on my radar and it's honestly low-priority.

elijah-potter (approved):
LFG

grantlemons:
Update: For correctness reasons it is now only 80% faster.

harper-core/src/spell/mod.rs
```diff
@@ -58,8 +59,8 @@ fn order_suggestions(matches: Vec<FuzzyMatchResult>) -> Vec<&[char]> {
         found.swap(0, 2);
     }
 
-    // Let common words bubble up, but do not prioritize them over all else.
-    found.sort_by_key(|fmr| fmr.edit_distance + if fmr.metadata.common { 0 } else { 1 });
+    // Make commonality relevant
+    found.sort_by_key(|fmr| if fmr.metadata.common { 0 } else { 1 });
```
grantlemons, line 63: Why did you remove the edit distance portion?

PR #266, opened 2024-11-04
Language ID Fixes

mcecode:
- Similar and related to #241, use standard language ids for Python (`python`), Shellscript/Bash (`shellscript`), and Git Commit (`git-commit`) files.
- Support Nix (`nix`) files as per #244.
- Add Plaintext (`plaintext`) support. It's not in [the standard](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem) but it's what VSCode uses to indicate `.txt` files.
- Base `documentSelector` on `activationEvents` so we only need to maintain one list of languages.
- Add tests to make sure that the languages we support work in VSCode.


elijah-potter:
Love your work here. Unrelated to this PR, I'd like to have a chat with you privately. Is there a way I can contact you?

mcecode:
Hi, I sent an email to the email you use in your commits so we can set up something there.

elijah-potter:
Great work! I'll get back to you soon. In the meantime, this is `merged`.

Issue #253, opened 2024-10-29
repeated word rule doesn't fire


notuntoward:
Repeated words aren't detected, although I have this rule turned on.  Below, "accurate" is repeated.

![image](https://github.com/user-attachments/assets/dc97b292-21f0-4e33-a82b-44709b7e5ad6)

![image](https://github.com/user-attachments/assets/9b7c3297-2192-4b97-a5c3-6ca5f5b17709)

elijah-potter:
Hey @notuntoward,

Sorry for the delay in getting back to you on this. Things have been a little hectic.

We've just [released Harper `v0.13.0`](https://github.com/Automattic/harper/releases/tag/v0.13.0), which incorporates some significant improvements to this rule, thanks to @grantlemons. 
Would you mind testing to make sure you don't get any more false-negatives like this so we can close issue?

notuntoward:
@elijah-potter , Thanks.  Unfortunately, it still doesn't work for me in this example:

> this paper shows that, while the method may be more accurate accurate, the turnout overestimate suggests that self-selection bias is not sufficiently reduced.

"accurate" is typed twice but the repetition isn't detected.  

But I should note that I had difficulties updating -- the install procedure stalled on Harper.  When I restarted obsidian, then I got a message saying something like "Harper cannot be restarted," After uninstall and then reinstall, Harper did restart, and appeared to be the newest version (1.3.0).  But maybe it's not, really?

grantlemons:
Ah, the obsidian plugin seemed to have had a large problem on 0.13.0 (#338), but @elijah-potter has pushed out 0.14.0 which should hopefully resolve the issue.

In my local tests I found that harper linted your example correctly, so if you update the obsidian plugin to the 0.14.0 version, it will probably work as intended.

elijah-potter:
@notuntoward, would you mind confirming that the issue has been fixed on your end so we can close it out? Thanks again for letting us know about it!

notuntoward:
With 0.15.0, it still doesn't detect the repeat.  This is on Obsidian.

elijah-potter:
Hmm, that's odd. I'm not able to replicate your issue. Are you by chance on MacOS? Would you mind opening the developer tools and giving us a screenshot of any errors you see in the "Console" tab? To do that, you can press `cmd + option+i` on MacOS or `ctrl + shift + i` otherwise.

Here's what it looks like on my system.

[...]

PR #277, opened 2024-11-12
feat(#253): remove requirement that word be in list for repetition

grantlemons:
There are some cases where words are repeated legitimately, but I feel these are rare enough that a false-positive isn't that bad.

~~The more interesting case is when homographs appear next to each other, I don't know of any way we can correct for these, but here are some examples ([source](https://english.stackexchange.com/a/37835)):~~

~~- to record record profits~~
~~- to address address problems~~
~~- to refuse refuse charges~~


mcecode:
Just a question since I saw the VSCode smoke test failed, so I can update it if needed. Is the change in the span length/range returned intended? For example, in the sentence "I like like this." It now seems to return the range from the space after the first like to the end of the second like (" like") instead of the whole "like like" previously.

grantlemons:
@mcecode My starting point for this change was to rollback to before we used patterns for this lint, which apparently didn't have the same span behavior as the most recent version.
I've changed it such that the matching span should cover both words.

harper-core/src/word_metadata.rs
```diff
@@ -113,6 +113,41 @@ impl WordMetadata {
         )
     }
 
+    /// Checks if the word is a homophone
+    /// This is based off of metadata, so it is not guaranteed to be correct
+    pub fn is_homophone(&self) -> bool {
+        [
```
elijah-potter, line 119: I'd like to backtrack something I said earlier. Change the name of this function to `is_likely_homophone` to indicate that, unlike the other `is_&` functions, this is not definitive.

harper-core/src/word_metadata.rs
```diff
@@ -113,6 +113,41 @@ impl WordMetadata {
...
+                    is_plural: None,
+                    is_possessive: None,
+                    is_pronoun: None
+                })
+            ),
+            matches!(
+                self.verb,
+                Some(VerbData {
+                    is_linking: None,
+                    tense: None
```
elijah-potter, line 133: It feels like implementation is likely to result in an inconsistency with the model for `WordMetadata`. Is there a way to make this throw an error if we modify `WordMetadata` but don't update the code here?

harper-core/src/linting/repeated_words.rs
```diff
@@ -71,6 +47,12 @@ mod tests {
...
     #[test]
     fn catches_basic() {
-        assert_lint_count("I wanted the the banana.", RepeatedWords::new(), 1)
+        assert_lint_count("I wanted the the banana.", RepeatedWords::default(), 1)
+    }
+
+    #[test]
+    fn does_not_lint_homophones() {
+        assert_lint_count("To address address problems.", RepeatedWords::default(), 0);
+        assert_lint_count("To record record profits.", RepeatedWords::default(), 0);
```
elijah-potter, line 56: Make each of these lines a separate test.

harper-core/src/linting/repeated_words.rs
```diff
@@ -1,66 +1,42 @@
...
+pub struct RepeatedWords;
+
+impl Linter for RepeatedWords {
+    fn lint(&mut self, document: &Document) -> Vec<Lint> {
+        let mut lints = Vec::new();
+
+        for chunk in document.iter_chunks() {
+            let mut iter = chunk.iter_word_indices().zip(chunk.iter_words()).peekable();
+
+            while let (Some((idx_a, tok_a)), Some((idx_b, tok_b))) = (iter.next(), iter.peek()) {
```
elijah-potter, line 15: Love it.

elijah-potter (changes requested):
Looks good, just some bookkeeping and future-proofing needed before merging.

harper-core/src/word_metadata.rs
elijah-potter: I would also like to point out that these are homographs, not homophones

harper-core/src/word_metadata.rs
grantlemons: Hmm, not sure about that. I can't think of any way.

elijah-potter: I found a way. We're good here.

elijah-potter (approved):

PR #278, opened 2024-11-12
feat: Create lint for 'that that' -> 'that which'

grantlemons:
- **feat(that-which): add then_exact_word_or_lowercase to pattern matching**
- **feat(that-which): add lint for 'that that' -> 'that which'**


harper-core/src/linting/that_which.rs
```diff
@@ -0,0 +1,68 @@
...
+    pattern: Box<dyn Pattern>,
+}
+
+impl Default for ThatWhich {
+    fn default() -> Self {
+        let mut pattern = WordPatternGroup::default();
+
+        pattern.add(
+            "that",
+            Box::new(
```
grantlemons, line 20: Used twice -> refactor into own variable.

grantlemons:
Needs some tests

harper-core/src/linting/that_which.rs
elijah-potter: You can solve this without using `Clone`. A smart-pointer for example.

elijah-potter (approved):
LGTM

PR #292, opened 2024-11-27
Improve config change handling

mcecode:
I noticed that config updates don't get reflected in the editor unless `harper-ls` is restarted, at least for VSCode. Hence, my original solution to just restart it, but that feels like such a brute force way and it clears all the diagnostics within the workspace. So, I thought it'd be better if the new config is reflected and the diagnostics are updated when `did_change_configuration` is called.

P.S. This is my first time writing any Rust so I may have made a mistake somewhere.

### DIFF
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -47,8 +47,8 @@ describe('Integration >', () => {
 	it('updates diagnostics on configuration change', async () => {
 		const config = workspace.getConfiguration('harper-ls.linters');
 		await config.update('repeated_words', false, ConfigurationTarget.Workspace);
-		// Wait for `harper-ls` to restart and send new diagnostics
-		await sleep(1000);
+		// Wait for `harper-ls` to update diagnostics
+		await sleep(200);
 
 		compareActualVsExpectedDiagnostics(
 			getActualDiagnostics(markdownUri),
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -48,7 +48,7 @@ describe('Integration >', () => {
 		const config = workspace.getConfiguration('harper-ls.linters');
 		await config.update('repeated_words', false, ConfigurationTarget.Workspace);
 		// Wait for `harper-ls` to update diagnostics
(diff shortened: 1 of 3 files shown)


packages/vscode-plugin/src/tests/suite/integration.test.ts
```diff
@@ -47,8 +47,8 @@ describe('Integration >', () => {
 	it('updates diagnostics on configuration change', async () => {
 		const config = workspace.getConfiguration('harper-ls.linters');
 		await config.update('repeated_words', false, ConfigurationTarget.Workspace);
-		// Wait for `harper-ls` to restart and send new diagnostics
-		await sleep(1000);
+		// Wait for `harper-ls` to update diagnostics
+		await sleep(250);
```
grantlemons, line 51: You said you had to increase this so that precommit would not fail. What are the consequences if this sleep is too short?

elijah-potter: After cloning and running it with a couple different timings, it just looks like it might be a (very minor) performance issue (which is out of scope for this test). Since config reloads are not something I care is fast, I think we're alright. At the very least, it's on my radar now.

elijah-potter:
Looks great @mcecode. I'm impressed that you were able to root through that (admittedly quite messy) file, especially for your first time writing Rust. Merging...

packages/vscode-plugin/src/tests/suite/integration.test.ts
mcecode: Just to chime in, these `sleep`s in the VSCode tests are there because editor actions, language server communication, etc. that affect diagnostics are not instantaneous, so they're needed to account for that latency. You could sort of treat them as benchmarks, albeit not good ones because they don't just measure how fast `harper-ls` is, they measure how fast VSCode can detect changes + how fast it can communicate with `harper-ls` + how fast `harper-ls` can process requests + how fast VSCode can show updates. Additionally, they're very much affected by how beefy the machine the tests are running on. So, that `250` milliseconds is the max time needed by the GitHub runner to run everything so that the diagnostic updates for this particular test are reflected after a config change. For my machine, it's only `150`, that's why it's hard for me to nail on the first try exactly what the right number to use is when running these on CI. Practically speaking, unless the user has a lot of files open or they're using a really old, slow computer, this latency should be negligible.

PR #294, opened 2024-11-29
ci: disable packaging and building workflows on draft PRs

grantlemons:
These workflows fail on draft PRs anyway, and I believe they will get run when the draft status is removed.

I still think it's important to have the precommit workflow run, but these others don't seem too important for a draft.


grantlemons:
I was wrong about the packaging ones failing, this was because the `Cargo.lock` was out of line with master, but I still don't think these need to run. Feel free to close if you disagree @elijah-potter, I don't have a strong opinion on this.

PR #302, opened 2024-12-04
feat: Typst Language Support

grantlemons:
Re-opening of https://github.com/Automattic/harper/pull/289 from my fork instead of a branch


grantlemons:
Re-opening of #289 from my fork instead of a branch

grantlemons:
Mostly just needs more tests

harper-core/src/parsers/typst.rs
```diff
@@ -0,0 +1,540 @@
...
+                TokenKind::Word(_),
+                TokenKind::Word(_),
+                TokenKind::Punctuation(Punctuation::Quote { .. }),
+                TokenKind::Word(_),
+                TokenKind::Punctuation(Punctuation::Quote { .. }),
+                TokenKind::Word(_),
+                TokenKind::Number(OrderedFloat(2019.0), None),
+            ]
+        ))
+    }
```
Andrew15-5, line 491: Going from intuition, the quotes should not count as actual quotes, because it's just how you construct a string in Typst.

The second problem is the question of whether the custom var/func names should be checked. Ideally this should be configurable.

On one hand, it would be nice to check the correctness of the variable names etc., but most IDEs already can do that and you can't use commas and periods, for example, as well as the name can be lowercase or uppercase. So I don't see any good reason to count those in by default.

Plus, it will mess up the sentence structure and will give false positives (like "the word should start with a capital letter" while it's a variable name). So the dictionary names definitely should not by checked (by default).

Going deeper, the problem with Typst is that you can construct text from code mode like `#range(5).map(_ => "a")` which will result in "aaaaa". There is a `lorem()` function that can output a ton of Latin text, and so can do user functions. This is probably an edge case that is unsolvable, unless you get the resolved AST or string, meaning all the code part is gone and the raw text (maybe with some styling) is left. I have yet to learn all the insides of Typst, especially since several previous rewrites (of some parts). So I can't be sure about the Rust's side of Typst in this PR.

In general, maybe it's not a great idea to check anything in code mode, only in markup. But I think this can be resolved if harper supports isolated text. So if you have 3 plain text paragraphs (and nothing else), then it would count as 1 isolated text chunk. If you also have some separate text, e.g., an image caption, then it would be a second text chunk. And since these chunks are isolated, you won't get any false positives from checking them together.

In a table with many cells you would have a separate text block for each cell, because they can have different content that isn't a sentence. If this is possible, then you can also safely check dictionary value etc.

grantlemons: > On one hand, it would be nice to check the correctness of the variable names etc., but most IDEs already can do that and you can't use commas and periods, for example, as well as the name can be lowercase or uppercase. So I don't see any good reason to count those in by default.
>
> Plus, it will mess up the sentence structure and will give false positives (like "the word should start with a capital letter" while it's a variable name). So the dictionary names definitely should not by checked (by default).

Harper has some degree of support for identifiers already for code comments, I don't think it would be too complicated to extend that here, we'd just need to extract all the identifiers to build the file dictionary.

grantlemons: > Going deeper, the problem with Typst is that you can construct text from code mode like #range(5).map(_ => "a") which will result in "aaaaa". There is a lorem() function that can output a ton of Latin text, and so can do user functions. This is probably an edge case that is unsolvable, unless you get the resolved AST or string, meaning all the code part is gone and the raw text (maybe with some styling) is left. I have yet to learn all the insides of Typst, especially since several previous rewrites (of some parts). So I can't be sure about the Rust's side of Typst in this PR.

Yes, this whole thing was the biggest issue when I was trying to support LaTeX as well. I agree that it is probably unsolvable. I think we could theoretically expand things using typst's libraries, but I assume it would have horrible performance.

harper-core/src/parsers/typst.rs
```diff
@@ -471,21 +480,52 @@ mod tests {
...
+                TokenKind::Word(_),                            // key 2
+                TokenKind::Number(OrderedFloat(2019.0), None), // value 2
+            ]
+        ))
+    }
+
+    #[test]
+    fn str_parsing() {
+        let source_with_quotes = r#"#let ident = "This is a string""#;
+        let source_no_quotes = r#"#let ident = This is a string"#;
```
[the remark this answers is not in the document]
grantlemons: Ah, I should really read the quick start guide huh haha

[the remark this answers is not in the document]
Andrew15-5: Yes, the friendly manual is waiting for you! :)

I guess the most essential for this PR would be https://typst.app/docs/reference/syntax/, but of course, by reading everything else you would be able to make any examples you want.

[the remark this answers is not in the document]
grantlemons: Very helpful, that's pretty much exactly what I needed!

grantlemons:
Turns out this doesn't work for unicode at all, since the typst crate uses byte spans, and we use char spans.

Andrew15-5:
What exactly about Unicode did you want to test? Typst source files must be a valid Unicode sequence of bytes. The example has next AST (if you need one):

```
Markup: 17 [
    Text: "group’s",
    Space: "\n",
    Text: "writing",
]
```

grantlemons:
> What exactly about Unicode did you want to test? Typst source files must be a valid Unicode sequence of bytes. The example has next AST (if you need one):
> 
> ```
> Markup: 17 [
>     Text: "group’s",
>     Space: "\n",
>     Text: "writing",
> ]
> ```

The issue I found was that the typst_syntax crate gives the spans of certain textual elements in terms of bytes, while we use byte spans, so any expressions after a unicode character would have their span shifted by the difference between the number of unicode codepoints and the number of characters.
In the test above, the apostrophe is three unicode codepoints, so when we parse the following expressions (the space and text) the spans are offset, appearing like so:
image

I've implemented a fix that seems to resolve it.

Andrew15-5:
> the typst_syntax crate gives the spans in terms of bytes, while we use byte spans

Previously, you've said char spans and not byte spans. For me, "spans of bytes" and "byte spans" are the same thing. And what do "chars" represent, Unicode code points? But also I haven't looked into spans and other nerdy stuff, so I can't comment on this particular topic.

elijah-potter:
> For me, "spans of bytes" and "byte spans" are the same thing. And what do "chars" represent, Unicode code points?

When we talk about a "span", we are referring to a data structure with two elements, a start and an end:

```rust
struct Span {
    start: usize,
    end: usize
}
```

Harper uses Spans to denote slices into [character](https://www.unicode.org/glossary/#unicode_scalar_value) arrays (without storing a reference to that underlying array), whereas Typst uses similar structures to denote slices into byte arrays. When we do interop with Typst, we have to convert between the two formats, since Unicode characters aren't necessarily a single byte.

grantlemons:
> > the typst_syntax crate gives the spans in terms of bytes, while we use byte spans
> 
> Previously, you've said char spans and not byte spans. For me, "spans of bytes" and "byte spans" are the same thing. And what do "chars" represent, Unicode code points? But also I haven't looked into spans and other nerdy stuff, so I can't comment on this particular topic.

Oops I mistyped, I meant char spans.

`chars` are utf-8 characters. In utf-8, characters can be encoded in more than one byte.

A Span is just a range in a document where something occurs. It has a start and an end, and everything in between is being represented by it. A byte span would represent the range of bytes in the document where something occurs (such as a word), while a char span would represent the range of characters in the document where that same thing occurs. If a character is represented by more than one character (because of utf-8's variable encoding) before or during the part of the document the span is intended to represent, a char span and a byte span will have different start and end points because of it.

Andrew15-5:
Thanks.

harper-core/Cargo.toml
```diff
@@ -24,6 +24,7 @@ thiserror = "2.0.9"
 unicode-blocks = "0.1.9"
 unicode-width = "0.2.0"
 levenshtein_automata = { version = "0.2.1", features = ["fst_automaton"] }
+typst-syntax = "0.12.0"
```
elijah-potter, line 27: Typst is pretty big, and not applicable to some applications (i.e. Obsidian). Would it be possible to put Typst support behind a feature flag?

harper-cli/src/main.rs
```diff
@@ -142,16 +157,16 @@ fn main() -> anyhow::Result<()> {
...
 
-    let mut parser: Box<dyn harper_core::parsers::Parser> =
-        if let Some("md") = file.extension().map(|v| v.to_str().unwrap()) {
-            Box::new(Markdown)
-        } else {
-            Box::new(
+    let parser: Box<dyn harper_core::parsers::Parser> =
+        match file.extension().map(|v| v.to_str().unwrap()) {
+            Some("md") => Box::new(Markdown),
+            Some("typ") => Box::new(Typst),
```
elijah-potter, line 163: With the exception of adding the file extension here, the changes in this file aren't really relevant to the PR. I like them. Would you mind putting them in a different PR?

harper-tree-sitter/src/lib.rs
```diff
@@ -91,7 +91,7 @@ impl TreeSitterMasker {
 }
 
 impl Masker for TreeSitterMasker {
-    fn create_mask(&mut self, source: &[char]) -> Mask {
+    fn create_mask(&self, source: &[char]) -> Mask {
```
elijah-potter, line 94: Why did you remove mutability? I remember discussing it with you, but forgot why.

harper-core/Cargo.toml
elijah-potter, line 27: Typst is pretty big and completely unused for certain applications (i.e. Obsidian). Would you mind putting Typst support behind a feature flag?

harper-core/src/mask/mod.rs
```diff
@@ -8,7 +8,7 @@ use crate::Span;
 /// This is primarily used by [`crate::parsers::Mask`] to create parsers for
 /// things like comments of programming languages.
 pub trait Masker: Send + Sync {
-    fn create_mask(&mut self, source: &[char]) -> Mask;
+    fn create_mask(&self, source: &[char]) -> Mask;
```
elijah-potter, line 11: Why remove mutability here? Not against it, just want to know your reasoning.

elijah-potter (changes requested):
Looks good, but I have some questions.

harper-core/src/document.rs
```diff
@@ -284,6 +284,12 @@ impl Document {
                     }
 
                     let child_tok = &copy[cursor];
+
+                    // Only condense adjacent spans
+                    if start_tok.span.end != child_tok.span.start {
```
elijah-potter, line 289: Add this change to `condense_newlines` as well.

grantlemons: I did that originally, but it broke a test so I reverted it.

elijah-potter: The test might be wrong then.

harper-core/Cargo.toml
elijah-potter: Alternatively, you could make the Typst parser its own crate.

harper-core/src/mask/mod.rs
grantlemons: Because of this, `parse()` needs `&mut self`, and because of that, every parser needs to be a mutable reference, so it cannot be cloned or copied, even though they are almost all unit structs. This creates some issues using them in closures, where ownership is passed to the closure.

harper-core/Cargo.toml
grantlemons: How would making it a seperate crate help? It would still be a dependency...

I feel like feature flag is the way to go. I assume we would have it enabled for harper-ls and harper-cli?

harper-core/src/document.rs
grantlemons: Ah, it's because newlines in comments need to be merged for paragraph breaks.

```
// I am a comment
//
// My span starts after the //
```
There is no paragraph break in between because the newline on line 2's span isn't contiguous with the newline at the end of line 1.

harper-core/src/mask/mod.rs
elijah-potter: Alright. That makes sense.

harper-core/Cargo.toml
elijah-potter: There are a couple reasons.

The intention is eventually, and especially with `harper-tree-sitter`, that `Parsers` may become a community endeavor. They should be able to be built without great understanding of Harper internals. Such parsers should be able to stand on their own without official support from me or Automattic.

If people build applications (other than `harper-ls`) using Harper, they should be able to build and pull in third-party parsers as they need. Having many parsers in `harper-core` sets the expectation that all parsers must be officially supported and maintained by us. 

Eventually, I hope this to be true for `Linter`s as well, but that's a little ways off.

In fact, if you wish to move the Typst `Parser` to its own repository and make this PR simply import it.

harper-core/tests/run_tests.rs
```diff
@@ -36,13 +39,18 @@ macro_rules! create_test {
...
+create_test!(obsidian_links.md, &Markdown, 2);
+create_test!(issue_267.md, &Markdown, 0);
+create_test!(proper_noun_capitalization.md, &Markdown, 2);
+create_test!(amazon_hostname.md, &Markdown, 0);
+create_test!(issue_159.md, &Markdown, 1);
+
+#[cfg(feature = "typst")]
+create_test!(complex_typst.typ, &Typst, 0);
+#[cfg(feature = "typst")]
+create_test!(typst_spelling_mistakes.typ, &Typst, 4);
```
mattfbacon, line 56: Why is this line incomplete?

grantlemons: Made a mistake just now when merging through github interface, just pushed a fix.

harper-core/src/parsers/typst.rs
```diff
@@ -0,0 +1,721 @@
...
+};
+
+use super::{Parser, PlainEnglish};
+use crate::{
+    parsers::StrParser,
+    patterns::{PatternExt, SequencePattern},
+    ConjunctionData, Lrc, NounData, Punctuation, Token, TokenKind, VecExt, WordMetadata,
+};
+
+/// A parser that wraps the [`PlainEnglish`] parser allowing one to parse Typst files.
```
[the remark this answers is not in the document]
grantlemons: [done](https://github.com/Automattic/harper/pull/302/commits/ada56a0cf08082dc5d4113a630f8a5f70e6f1651)

grantlemons:
I decided to move typst parsing into its own crate, since it simplifies the feature dependency and establishes good precident that can be followed for unofficial/external parsers in the future.

harper-ls/Cargo.toml
```diff
@@ -25,3 +26,7 @@ resolve-path = "0.1.0"
 open = "5.3.0"
 futures = "0.3.31"
 serde = { version = "1.0.214", features = ["derive"] }
+
+[features]
```
elijah-potter, line 30: Similar to `harper-cli`. I don't think we're going to be distributing binaries _without_ Typst support, so I don't think we need this feature flag.

harper-ls/src/backend.rs
```diff
@@ -206,17 +211,15 @@ impl Backend {
...
-            } else if language_id == "git-commit" || language_id == "gitcommit" {
-                Some(Box::new(GitCommitParser))
-            } else if language_id == "html" {
-                Some(Box::new(HtmlParser::default()))
-            } else if language_id == "mail" || language_id == "plaintext" {
-                Some(Box::new(PlainEnglish))
-            } else {
-                None
-            };
+            }
```
elijah-potter, line 214: Much better syntax. I approve.

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+use itertools::Itertools;
+use typst_syntax::{
+    ast::{AstNode, Markup},
+    Source,
+};
+
+/// A parser that wraps the [`PlainEnglish`] parser allowing one to parse Typst files.
+pub struct Typst;
+
+thread_local! {
```
elijah-potter, line 21: I would toss this in a constructor for the parser, but I'm not going to let that stop us merging this.

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+
+        // Recurse through AST to create tokens
+        let parse_helper = TypstTranslator::new(&typst_document);
+        let mut tokens = typst_tree
+            .exprs()
+            .filter_map(|ex| parse_helper.parse_expr(ex, OffsetCursor::new(&typst_document)))
+            .flatten()
+            .collect_vec();
+
+        // Consolidate conjunctions into single tokens
```
elijah-potter, line 45: The `Document` takes care of contractions. Is there something specific to `Typst` that makes this necessary?

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,391 @@
...
+#[cfg(test)]
+mod tests {
+    use itertools::Itertools;
+    use ordered_float::OrderedFloat;
+
+    use super::Typst;
+    use harper_core::{parsers::StrParser, NounData, Punctuation, TokenKind, WordMetadata};
+
+    #[test]
+    fn conjunction() {
```
elijah-potter, line 107: [*contraction](https://en.wikipedia.org/wiki/Contraction_(grammar))

elijah-potter (commented):
There's some stuff I didn't see last time. Sorry again, but I just need some clarification.

elijah-potter (approved):

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+                TokenKind::Word(WordMetadata {
+                    noun: Some(NounData {
+                        is_possessive: Some(true),
+                        ..
+                    }),
+                    ..
+                })
+            )
+        }))
+    }
```
Andrew15-5, line 63: Uhh, why is contraction identified a "possessive"? To me, it looks obviously wrong.

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+                TokenKind::Unlintable, // Ident
+                TokenKind::Punctuation(Punctuation::Comma),
+                TokenKind::Space(1),
+                TokenKind::Word(_), // and
+                TokenKind::Space(1),
+                TokenKind::Space(2),
+                TokenKind::Word(_),
+            ]
+        ))
+    }
```
Andrew15-5, line 214: So how would harper work in this case? Will it complain that there are 2 consecutive commas in the "sentence"? (And also that there is not 2, but 3 consecutive spaces, even though the 1st of 3 spaces won't go directly next to the last 2 spaces.) Or are `Unlintable` tokens split lintable tokens into isolated sentences/blocks?

```
, , and   bob
```

And also, are 2 spaces intentional?

P.S. the `// bob` comment is missing.

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+
+        assert!(matches!(
+            &token_kinds.as_slice(),
+            &[
+                TokenKind::Word(_),
+                TokenKind::Newline(1),
+                TokenKind::Word(_)
+            ]
+        ))
+    }
```
Andrew15-5, line 238: 2 concerns:
1. Are all the leading spaces in `"= Header\n                       Paragraph"` string ignored by the harper?
2. The LF between two text lines are treated as spaces (`Space: "\n"`) in Typst's AST, which are replaced by whitespaces so that you can wrap your long paragraphs after 80's column. But in this case, the first line is a heading, which means that there is `heading()` + `par()` and `Newline(1)` seems appropriate in this specific use case. So will it still print `Newline` for wrapped long paragraph? And most importantly, what exactly does `Newline` mean to harper?

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+            &token_kinds.as_slice(),
+            &[
+                TokenKind::Word(_),
+                TokenKind::Newline(1),
+                TokenKind::Unlintable,
+                TokenKind::Newline(1),
+                TokenKind::Word(_),
+            ]
+        ))
+    }
```
Andrew15-5, line 280: Similar to above, the LF after the label won't add any new line to the output, as labels by themselves aren't visible (usually they are attached to some element in the markup mode). So this again raises concern if `Newline` token is really a hard newline that I'm thinking of (that breaks paragraphs into separate paragraphs or something similar).

harper-typst/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+                TokenKind::Word(_),
+                TokenKind::Punctuation(Punctuation::Period),
+            ]
+        ))
+    }
+
+    #[test]
+    fn smart_apostrophe_newline() {
+        let source = r#"group’s
+writing"#;
```
Andrew15-5, line 317: Why this is written differently from previous examples? Why not write:

```rs
        let source = r#"
            group's
            writing
        "#;
```

There is `dedent` crate that can prettify this into `"group's\nwriting"`, but that's a new dependency.

harper-typst/tests/test_sources/complex_typst.typ
Andrew15-5, line 1: I'm not sure if an overly complex example is the intent here, so if it isn't, then I would like to simplify it and make it both: use best practices and be more practical. The way the sugar syntax for `emph` is used, is really cursed and isn't the intended way. And in addition, fix some bugs that were not caught because the visual output isn't present, so not audited.

<details><summary>Simplified example</summary>

```typ
#let template(
  title: "Default Title",
  authors: ("Author 1", "Author 2"),
  abstract: [*This is content*],
  body,
) = {
  set par(justify: true)
  set page(
    paper: "us-letter",
    columns: 2,
    number-align: top,
    numbering: (..n) => if n.pos().first() > 1 {
      n.pos().map(str).join(" of ") + h(1fr) + title
    },
  )

  place(
    top + center,
    float: true,
    scope: "parent",
    clearance: 2em,
  )[
    #show heading: set text(17pt)
    = #title

    #let authors-line = if authors.len() > 3 {
      // "et al." isn't parsed properly, but this isn't the fault of the Typst
      // parser.
      // authors-max3.push("et al.")
      authors => authors.join(", ")
    } else {
      authors => authors.join(", ", last: ", and ")
    }
    #emph(authors-line(authors.slice(0, calc.min(authors.len(), 3))))

    #par(justify: false)[
      *Abstract* \
      #abstract
    ]
  ]

  body
}

#show: template.with(
  title: "A fluid dynamic model for glacier flow",
  authors: ("Grant Lemons", "John Doe", "Jane Doe"),
  abstract: lorem(80),
)

= Introduction
#lorem(300)

= Related Work
#lorem(200)
```

</details>

Andrew15-5 (commented):
Eyyy! It was already merged, kewl. I guess I'm late to the party, but I still have a few things I want to discuss about the final patch.

P.S. I was really busy and only now found time to look into this.

elijah-potter:
> Eyyy! It was already merged, kewl. I guess I'm late to the party, but I still have a few things I want to discuss about the final patch.

Sorry about that! I'm sure @grantlemons would be happy to open a new PR once he has time to look at your comments. Thanks for catching the stuff you've pointed out! I've been quite busy with web stuff, so I admittedly rushed this one. 

I see that you're a significant contributor on Typst. I'm impressed. It would be cool to get Harper plugin made for upstream.

harper-typst/src/lib.rs
grantlemons: Yes, the unlintable tokens split it up because our matching patterns don't expect them.

The two spaces are intentional, they are included because of some changes I was making to how Space Spans are condensed. Really, I don't think it's relevent to typst though, so I'm going to remove it from the test

harper-typst/src/lib.rs
grantlemons: > Are all the leading spaces in `"= Header\n                       Paragraph"` string ignored by the harper?
I believe leading spaces are ignored, yes. @elijah-potter correct me if I'm wrong.

> what exactly does Newline mean to harper?
`TokenKind::Newline(count)` represents `count` `\n` newline characters.

I believe they act to break up sequences of tokens between lines. In this case, for instance, you don't want Harper to consider the text to be `Header Paragraph`, you want it to be `Header` and `Paragraph` seperetely.

harper-typst/src/lib.rs
grantlemons: This is because this test exists to test the offsetting functionality, converting between byte spans and char spans. I wanted to keep the string as similar as possible to the string that had issues, hence why I didn't change the formatting at all.

Andrew15-5:
> Sorry about that!

If you thought it was a negative "Eyyy!", then it wasn't, the hint is "kewl", so it's fine, no need to apologize. I also wouldn't want to delay the merge just because I'm busy and can't reply for a long time.

> I'm sure @grantlemons would be happy to open a new PR once he has time to look at your comments. 

Actually, perhaps I can open the new PR. I'll add the Typst fixes, then anyone can request changes or open the PR into my PR and I will accept it (considering we agree on the changes).

> Thanks for catching the stuff you've pointed out!

You're welcome! As a perfectionist, this is what I do, hehe. And I'm really interested in both Typst and a tool that can deliver the best spellcheck experience into mine editor. I use Neovim btw.

> I see that you're a significant contributor on Typst. I'm impressed.

Thank you! 💚 Actually, I'm not _really_ a significant contributor yet (most of the changes so far were docs changes), but I'm intending to fix that. I have some things that are piling up, so I just need to finish other stuff so I can focus on making Typst even better! 🚀

> It would be cool to get Harper plugin made for upstream.

Huh? plugin? What do you mean by that? Harper is already a "plugin" by itself that just needs to be added to an IDE. I'm open to collaboration.

P.S. The biggest blocker for me for the long time was finishing the NixOS config so I can ditch the Pop!_OS 22.04 (with utterly broken GNOME, at times). So this is my top priority right now.

harper-typst/tests/test_sources/complex_typst.typ
grantlemons: Yes, the intent was to make it rather complicated, to see if any issues would arise from nesting statements and such. If you'd like to add more examples that demonstrate a more idiomatic typst document, that would be fantastic.

harper-typst/src/lib.rs
grantlemons: You're right, that must have happened when I was removing the contraction logic from the typst parser itself. I'm actually going to remove the contraction tests entirely, since that's document logic.

grantlemons:
@Andrew15-5 I've created a new PR #391, let me know if you have any further suggestions. (Idiomatic tests would probably be best suited by you creating a new pr though)

PR #303, opened 2024-12-04
feat: Span visualization command

grantlemons:
Added a command to harper-cli and the justfile that visually displays spans on a parsed file.

It can be painstaking to verify that a parser is working correctly when writing one, but this command makes it easier to verify that the correct spans are being chosen.

I created this to assist me in implementing the typst parser (#302), but am spinning it off into it's own PR since it is pretty distinct.

The output is rather cluttered because it prints the span numbers as well. I'm split on this one, it can be helpful because you can cross-reference with the output of `just parse` to check what each span corresponds to, but it also makes the output less readable (or at least a lot longer vertically).

Example usage:
image

Note: it alternates colors to make adjacent spans stand out.

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -77,7 +77,7 @@ fn main() -> anyhow::Result<()> {
             let report = report_builder.finish();
             report.print((&filename, Source::from(source)))?;
 
-            std::process::exit(1);
+            Ok(())
         }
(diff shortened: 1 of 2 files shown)


elijah-potter:
Love the work. Since `harper-cli` continues to primarily be a debugging tool, this is fine. Be aware that this might be removed if that changes.

PR #304, opened 2024-12-05
feat(harper-ls): remove deleted file/s diagnostics

mcecode:
I noticed that even after files are deleted, their diagnostics stay. This fixes that.

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -17,12 +17,14 @@ use tower_lsp::jsonrpc::Result;
 use tower_lsp::lsp_types::notification::PublishDiagnostics;
 use tower_lsp::lsp_types::{
     CodeActionOrCommand, CodeActionParams, CodeActionProviderCapability, CodeActionResponse,
-    Command, ConfigurationItem, Diagnostic, DidChangeConfigurationParams,
+    Command, ConfigurationItem, DeleteFilesParams, Diagnostic, DidChangeConfigurationParams,
     DidChangeTextDocumentParams, DidCloseTextDocumentParams, DidOpenTextDocumentParams,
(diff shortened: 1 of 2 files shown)


harper-ls/src/backend.rs
```diff
@@ -407,6 +435,34 @@ impl LanguageServer for Backend {
 
     async fn did_close(&self, _params: DidCloseTextDocumentParams) {}
 
+    async fn did_delete_files(&self, params: DeleteFilesParams) {
```
elijah-potter, line 438: It looks like you're just listening for files that have been deleted by the editor. If a file is deleted separately (i.e. via File Explorer or bash), this won't pick it up. 

Would you mind adding listeners to watch for other kinds of deletions? I'd be happy to make it happen, just won't be able to for a bit.

packages/vscode-plugin/src/tests/suite/integration.test.ts
```diff
@@ -61,4 +61,23 @@ describe('Integration >', () => {
 		// Set config back to default value
 		await config.update('repeated_words', true, ConfigurationTarget.Workspace);
 	});
+
+	it('updates diagnostics when files are deleted', async () => {
```
elijah-potter, line 65: Love the extensive test coverage you've been building out.

elijah-potter (changes requested):
I've got a concern or two, but otherwise it looks great! Thanks for noticing the issue.

harper-ls/src/backend.rs
mcecode: Yeah, this only works if a file is deleted within the editor. I'll look for a way so it'd also pick up deletions that happen outside of the editor. I'll turn this PR into a draft until it's ready.

mcecode:
Took me a while to get back to this but I think it's ready. It now also picks up deletions that happen outside of the editor.

Issue #306, opened 2024-12-06
Feat: Configure diagnostic severity per linter


ficd0:
Currently, the `diagnosticSeverity` configuration option applies globally to all of Harper's output. However, it would be nice to have an option to configure a different severity for each linter. For example, I would like to set `spell_check` and `sentence_capitalization` to the `error` severity, while some other things like `spaces` and `linking_verbs` would make more sense as `warning` or `info`. For reference, [ltex-ls](https://github.com/valentjn/ltex-ls) implements this feature.

elijah-potter:
Hey @ficcdaf,

I like this idea. I don't imagine this would be difficult to implement. It is somewhat lower priority, so it might be a bit before I can get to it. It looks like @grantlemons has taken it up, though. He might be able to provide a status on it.

Thanks for raising the issue!

grantlemons:
Status was that it was supported by harper-ls, but still needed integration with harper-wasm, and the vscode plugin.

I'll need to revisit it to ensure nothing was broken by all the changes to master recently.

ficcdaf:
No problem and no rush. Thanks for all your hard work!

PR #312, opened 2024-12-11
feat(#306): individual linter diagnostic severities

grantlemons:
Edits config to allow the user to set diagnostic severities for each lint, as well as a default severity.

This is a breaking change config-wise.


grantlemons:
Yeah this would need to be restarted from scratch lol.

PR #324, opened 2024-12-22
refactor: remove raw hashes that are unnecessary

### LINKED ISSUE #323: There are some unnecessary `#`s
There are some unnecessary `#`s that we can remove to improve code readability and maintainability.

For example, we can simplify

```rust
r#"example.com"#,
r#"example.com"#,
r#"example.com"#,
r#"and.subdomains.example.com"#,
r#"example.com"#,
r#"example.com"#,
r#"example"#,
r#"s.example"#,
r#"example.org"#,
r#"example.org"#,
r#"example.org"#,
r#"strange.example.com"#,
r#"example.org"#,
r#"example.org"#,
```

to

```rust
r"example.com",
r"example.com",
r"example.com",
r"and.subdomains.example.com",
r"example.com",
r"example.com",
r"example",
r"s.example",
r"example.org",
r"example.org",
r"example.org",
r"strange.example.com",
r"example.org",
r"example.org",
```

hamirmahal:
Closes #323

### DIFF
--- a/harper-core/src/language_detection.rs
+++ b/harper-core/src/language_detection.rs
@@ -106,7 +106,7 @@ mod tests {
     #[test]
     fn detects_python_fib() {
         assert_not_english(
-            r#"
+            r"
 def fibIter(n):
     if n < 2:
         return n
@@ -115,7 +115,7 @@ def fibIter(n):
     for _ in range(2, n):
         fibPrev, fib = fib, fib + fibPrev
     return fib
-        "#,
+        ",
         );
     }
 
--- a/harper-core/src/lexing/email_address.rs
+++ b/harper-core/src/lexing/email_address.rs
@@ -109,23 +109,23 @@ mod tests {
 
     fn example_local_parts() -> impl Iterator<Item = Vec<char>> {
         [
-            r#"simple"#,
-            r#"very.common"#,
-            r#"x"#,
-            r#"long.email-address-with-hyphens"#,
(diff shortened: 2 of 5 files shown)


hamirmahal (commented):
`cargo test` passes locally.

hamirmahal:
I came across this repository when looking for alternatives to Grammarly. Thanks for making this open source.

grantlemons (commented):
I don't personally feel this is important. I kinda like the consistency of using raw strings even if they may not be needed.

I also think this standard is hard to maintain. If there was a clippy lint that would help us be consistent, I would be much more on board with it.

Also, this will need to be expanded to open PRs as well, so if this gets merged we'll need to look at all the open PRs to make appropriate changes.

hamirmahal:
Oh, there's a Clippy lint for it, `clippy::needless_raw_string_hashes`.

elijah-potter:
I'm with @grantlemons on this. I appreciate your effort on this, but this PR as it stands right now is pedantic and doesn't really address the underlying problem. That said, if you were able to configure our build system (which runs `cargo check`) to catch the problem, I would merge it.

hamirmahal:
I've been looking into this. We can modify `justfile` to catch this issue outside of tests.

If we want to catch the issue when it appears inside of tests, too, I think we'd need to do `cargo clippy --all-targets` in `justfile`, which had a lot of output when I ran it.

PR #325, opened 2024-12-22
fix(core): updates to SpelledNumbers

grantlemons:
- Fix comments
- Change from suggesting if <= 10 to < 10
- Expand spelling function (110 would give stack overflow before)
- Update tests accordingly

I wasn't able to find an "Automattic style guide" as referenced in recent
commits on this, but it is rather standard to spell out numbers under but not
including 10. This is what APA does, and I believe MLA as well.

Recent commits also didn't update doc comments to reflect new behavior, so I
fixed that.

One thing to keep in mind, but I don't think is implemented yet, is that
numbers beginning sentences should always be spelled out, so it would be best
to expand behavior to include this in further enhancements.

### DIFF
--- a/harper-core/src/linting/spelled_numbers.rs
+++ b/harper-core/src/linting/spelled_numbers.rs
@@ -121,4 +121,12 @@ mod tests {
     fn does_not_correct_ten() {
         assert_suggestion_result("There are 10 pigs.", SpelledNumbers, "There are 10 pigs.");
     }
+
+    /// Check that the algorithm won't stack overflow or return `None` for any numbers within the specified range.
+    #[test]
+    fn services_range() {
+        for i in 0..1000 {
+            spell_out_number(i).unwrap();
+        }
+    }
 }
--- a/harper-core/src/linting/spelled_numbers.rs
+++ b/harper-core/src/linting/spelled_numbers.rs
@@ -1,7 +1,7 @@
 use crate::linting::{LintKind, Linter, Suggestion};
 use crate::{Document, Lint, TokenStringExt};
 
-/// Linter that checks to make sure small integers (< one hundred) are spelled
+/// Linter that checks to make sure small integers (< 10) are spelled
 /// out.
 #[derive(Default, Clone, Copy)]
 pub struct SpelledNumbers;
@@ -14,14 +14,14 @@ impl Linter for SpelledNumbers {
             let (number, _suffix) = number_tok.kind.number().unwrap();
             let number: f64 = number.into();
 
-            if (number - number.floor()).abs() < f64::EPSILON && number <= 10. {
+            if (number - number.floor()).abs() < f64::EPSILON && number < 10. {
                 lints.push(Lint {
                     span: number_tok.span,
                     lint_kind: LintKind::Readability,
                     suggestions: vec![Suggestion::ReplaceWith(
                         spell_out_number(number as u64).unwrap().chars().collect(),
                     )],
-                    message: "Try to spell out numbers less than a hundred.".to_string(),
+                    message: "Try to spell out numbers less than ten.".to_string(),
                     priority: 63,
                 })
             }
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/spelled_numbers.rs
```diff
@@ -64,14 +70,18 @@ fn spell_out_number(num: u64) -> Option<String> {
...
-            let child = num % 10;
+            let n = 10u64.pow((num as f32).log10() as u32);
+            let parent = (num / n) * n; // truncate
+            let child = num % n;
 
             format!(
-                "{} {}",
+                "{}{}{}",
                 spell_out_number(parent).unwrap(),
+                if num <= 99 { '-' } else { ' ' },
```
grantlemons, line 84: This is because numbers between twenty-one and ninety-nine should be hyphenated.

harper-core/src/linting/spelled_numbers.rs
```diff
@@ -64,14 +70,18 @@ fn spell_out_number(num: u64) -> Option<String> {
...
-        100 => "one hundred".to_string(),
+        hundred if hundred % 100 == 0 => {
+            format!("{} hundred", spell_out_number(hundred / 100).unwrap())
+        }
         _ => {
-            let parent = (num / 10) * 10;
-            let child = num % 10;
+            let n = 10u64.pow((num as f32).log10() as u32);
+            let parent = (num / n) * n; // truncate
+            let child = num % n;
```
grantlemons, line 79: This modification allows the truncation to extend to n digits, instead of just working for two digits as in the past.

This used to cause a stack overflow for `n = 110`. (and probably `101` to `109` as well)

harper-core/src/linting/spelled_numbers.rs
```diff
@@ -64,14 +70,18 @@ fn spell_out_number(num: u64) -> Option<String> {
         70 => "seventy".to_string(),
         80 => "eighty".to_string(),
         90 => "ninety".to_string(),
-        100 => "one hundred".to_string(),
+        hundred if hundred % 100 == 0 => {
+            format!("{} hundred", spell_out_number(hundred / 100).unwrap())
+        }
```
grantlemons, line 75: This logic should probably be expanded to include more suffixes for powers of 10 up to `10^19` so it will work for large numbers at the beginning of sentences.

(`10^19` is the largest power of 10 a u64 could represent)

harper-core/src/linting/spelled_numbers.rs
```diff
@@ -14,14 +14,14 @@ impl Linter for SpelledNumbers {
...
-            if (number - number.floor()).abs() < f64::EPSILON && number <= 10. {
+            if (number - number.floor()).abs() < f64::EPSILON && number < 10. {
                 lints.push(Lint {
                     span: number_tok.span,
                     lint_kind: LintKind::Readability,
                     suggestions: vec![Suggestion::ReplaceWith(
                         spell_out_number(number as u64).unwrap().chars().collect(),
                     )],
-                    message: "Try to spell out numbers less than a hundred.".to_string(),
+                    message: "Try to spell out numbers less than ten.".to_string(),
```
grantlemons, line 24: It is kinda funny that this is spelled out when under my change it shouldn't be.

I've always worked on the principle of:
- Spelled out for < 10
- Numerals for > 10
- Whatever feels best for 10 in the moment

harper-core/src/linting/spelled_numbers.rs
elijah-potter: Toss this info in a comment in the code.

elijah-potter:
> One thing to keep in mind, but I don't think is implemented yet, is that
numbers beginning sentences should always be spelled out, so it would be best
to expand behavior to include this in further enhancements.

I am aware. This was planned.

> I wasn't able to find an "Automattic style guide" as referenced in recent
commits on this

It's quite closely based on the Chicago Style, with a few changes.

Overall, this PR looks great! Merging...

PR #332, opened 2024-12-26
feat(#331): Haskell Comment Support

grantlemons:
Had to use a weird old version of tree-sitter-haskell for dependencies to work out.

PR #336, opened 2024-12-30
feat(vscode-plugin): Add Haskell Support

mcecode:
Could supersede or be merged after #332.


elijah-potter:
Looks pretty straightforward. 

@mcecode, just a heads up: we'll be significantly speeding up the cadence of releases, so don't worry about any other PRs getting in to the next one.

PR #342, opened 2024-12-31
docs: Fixup README links for language server integration

claydugo:
Can revert first commit if you want to point to the `harper-ls` subdirectory initially. I was just looking for the website.


elijah-potter:
Thanks Clay! Mergining...

PR #343, opened 2024-12-31
feat: Add technical words to dictionary

claydugo:
Here are some "missing" words that are flagged in my projects that seem to fit with the other defined dictionary words.

Let me know if you would like to remove any.


harper-core/dictionary.dict
```diff
@@ -20894,6 +20897,8 @@ downpour/14MS
 downrange/5
 downright/5
 downriver/5
+downsample/
```
claydugo, line 20900: I could use some direction on how to specify the affixes here. 

I found the [affixes.json](https://github.com/automattic/harper/blob/301215ef91b3caaf29a6/harper-core/affixes.json) file but I am struggling a bit at defining them myself. 

Is there something I should read before jumping in like this?

elijah-potter: Yeah. I just wrote a quick page about this. Mind taking a look and letting me know if there's anything it doesn't clear up?

https://writewithharper.com/docs/contributors/dictionary

claydugo: Thanks for writing this up!

I went through and redid each commit using `just addnoun <word>`.

It did end up flagging when I attempted to duplicate `subclass` so I think my environment is setup correct.

I am reading the `justfile` and it seems my below concerns are expected/designed but I wanted to flag in case it is worth adding to this new documentation page. 

I had assumed there was some order enforced to the dictionary as some of it is alphabetical but this seemed to just append entries to the bottom. Is that expected?

All words added seem to have been assigned the affix `/SM` when I would have assumed the `ing` words to just be handled through this.

elijah-potter: > I had assumed there was some order enforced to the dictionary as some of it is alphabetical but this seemed to just append entries to the bottom. Is that expected?

Order does not matter. It happens to be ordered that way because we actually started with an existing Hunspell dictionary that happened to be sorted at the time.

> All words added seem to have been assigned the affix /SM when I would have assumed the `ing` words to just be handled through this.

The `S` affix creates expansions for plural nouns and the `M` affix creates expansions for ownership. For example, `lamp` can be expanded to `lamps` and `lamp's`.

I would remove the entry for your `-ing` words and instead add the `G` affix to the base word.

elijah-potter:
Looks great. Thank you so much!

Issue #331, opened 2024-12-26
Can I add language support to harper at runtime?


andremarianiello:
If I wanted to use harper on Haskell files, is there a way to integrate harper with the tree-sitter-haskell grammar at runtime? Perhaps I could point harper to my installation of the parser for dynamic loading? If not, what is the process of getting language support statically added to harper?

grantlemons:
I don't think it's possible to integrate it at runtime, but adding support for parsing comments should be pretty trivial since it is just parsing comments.

elijah-potter:
Hey @andremarianiello, we've added very rough initial support for Haskell. Give it whirl and let us know if there are any issues. 

@grantlemons, would you mind looking into whether structured comment support [as described here](https://writewithharper.com/docs/integrations/visual-studio-code) is applicable to Haskell?

grantlemons:
> Hey [@andremarianiello](https://github.com/andremarianiello), we've added very rough initial support for Haskell. Give it whirl and let us know if there are any issues.
> 
> [@grantlemons](https://github.com/grantlemons), would you mind looking into whether structured comment support [as described here](https://writewithharper.com/docs/integrations/visual-studio-code) is applicable to Haskell?

Are you sure this is the right link? I don't see anything about structured comment support.

andremarianiello:
My very basic test with Haskell worked nicely, and will definitely be useful going forward in my Haskell projects. Unfortunately I wasn't specific enough in my initial description, because I initially chose harper to help me grammar check blog posts that I write with _literate_ haskell. As expected, I get no hints from harper on literate haskell (.lhs) files. Would you accept a PR that adds literate haskell support? I saw it was removed in https://github.com/grantlemons/harper/commit/c100c6d9955c108526281069340a9579751481a8

grantlemons:
> My very basic test with Haskell worked nicely, and will definitely be useful going forward in my Haskell projects. Unfortunately I wasn't specific enough in my initial description, because I initially chose harper to help me grammar check blog posts that I write with _literate_ haskell. As expected, I get no hints from harper on literate haskell (.lhs) files. Would you accept a PR that adds literate haskell support? I saw it was removed in [grantlemons@c100c6d](https://github.com/grantlemons/harper/commit/c100c6d9955c108526281069340a9579751481a8)
> 
[...]

PR #344, opened 2024-12-31
feat(#331): Literate Haskell Support

grantlemons:
- **feat(#331): create literate haskell masker and parser**
- **tests(#331): change tests to have spelling issues**
- **tests(#331): create literate haskell masker unit tests**
- **refactor(#331): simplify ident dict code**

The parser does not support parsing the actual code beyond extracting identifiers, but I don't know how that would be implemented considering the vastly different contexts.
If we were able to, using CommentParser would be great. It is a little weird to have comments in your code when it's literate programming anyway, but I can see it happening.

Will need to be revisited before/after merging #302 because of the changes to the `Parser` and `Masker` traits.


harper-ls/src/backend.rs
```diff
@@ -204,6 +206,30 @@ impl Backend {
                 } else {
                     Some(Box::new(ts_parser))
                 }
+            } else if language_id == "lhaskell" {
+                let source = text.chars().collect_vec();
+                let parser = LiterateHaskellParser;
+
+                if let Some(new_dict) = parser.create_ident_dict(&source) {
```
elijah-potter, line 213: Could this be generalized to the other languages?

elijah-potter (changes requested):
Looks and seems to work great! Mostly just looking for enhanced documentation. Why did you make the decisions you did?

PR #349, opened 2025-01-02
refactor: Harper-ls Error-Handling Improvements

grantlemons:
This should make it easier to track issues in `harper_ls`. This changes appropriate functions to return `anyhow::Result`s, and in functions that can't, the anyhow error is logged in a concise way. These functions also add context to the anyhow error to make it easier to trace.

Anyhow also ships with a macro `bail!()`, which simplifies forceable returns of new anyhow errors. For instance:
```rust
return Err(anyhow::format_err!("forceStable must be a boolean value."));
```
Becomes:
```rust
bail!("forceStable must be a boolean value.");
```


harper-ls/src/backend.rs
```diff
@@ -51,58 +50,68 @@ impl Backend {
...
+            .or(Ok(FullDictionary::new()))
     }
 
     async fn save_file_dictionary(&self, url: &Url, dict: impl Dictionary) -> anyhow::Result<()> {
-        Ok(save_dict(
+        save_dict(
             self.get_file_dict_path(url)
                 .await
-                .ok_or(anyhow!("Could not compute dictionary path."))?,
+                .context("Unable to get file path.")?,
```
elijah-potter, line 95: "*the file path"

harper-ls/src/backend.rs
```diff
@@ -51,58 +50,68 @@ impl Backend {
...
             dict,
         )
-        .await?)
+        .await
+        .context("Unable to save dictionary to path.")
```
elijah-potter, line 99: "*the dictionary"

harper-ls/src/config.rs
```diff
@@ -73,32 +72,30 @@ pub struct Config {
...
         let Value::Object(value) = value else {
-            return Err(anyhow::format_err!("Settings must be an object."));
+            bail!("Settings must be an object.");
         };
 
         let Some(Value::Object(value)) = value.get("harper-ls") else {
-            return Err(anyhow::format_err!(
-                "Settings must contain a \"harper-ls\" key."
-            ));
+            bail!("Settings must contain a \"harper-ls\" key.");
```
elijah-potter, line 83: Love this stuff. I didn't know `bail!()` was a thing.

harper-ls/src/dictionary_io.rs
```diff
@@ -1,27 +1,34 @@
...
-use tokio::io::{self, AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt, BufReader, BufWriter};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt, BufReader, BufWriter};
 
 /// Save the contents of a dictionary to a file.
 /// Ensures that the path to the destination exists.
-pub async fn save_dict(path: impl AsRef<Path>, dict: impl Dictionary) -> io::Result<()> {
+pub async fn save_dict(path: impl AsRef<Path>, dict: impl Dictionary) -> Result<()> {
     if let Some(parent) = path.as_ref().parent() {
-        fs::create_dir_all(parent).await?;
+        fs::create_dir_all(parent)
```
elijah-potter, line 12: I would leave this function to return an `io::Result` and make the caller responsible for handling it in a readable way.

We generally want to try to only use `anyhow` stuff as close to the entry point as possible. When we are doing more specific tasks, we should use more specific error types.

harper-ls/src/dictionary_io.rs
```diff
@@ -35,16 +42,20 @@ async fn write_word_list(dict: impl Dictionary, mut w: impl AsyncWrite + Unpin)
     Ok(())
 }
 
-pub async fn load_dict(path: impl AsRef<Path>) -> io::Result<FullDictionary> {
-    let file = File::open(path.as_ref()).await?;
+pub async fn load_dict(path: impl AsRef<Path>) -> Result<FullDictionary> {
```
elijah-potter, line 45: Same feedback as `save_dict` above.

elijah-potter (changes requested):
I've got some pedantic things I'd like to see changed. Feel free to push back if you disagree.

PR #353, opened 2025-01-03
feat: Add support for cmake files

alcroito:
This adds support for cmake script files.

It's using 0.4.1 instead of 0.5.0, because the latter depends on tree-sitter 0.22, which is not compatible with the 0.20 that is currently used.


elijah-potter:
Looks good. Before I merge it, would you mind commenting on [this](https://github.com/neovim/nvim-lspconfig/pull/3534) issue to get `cmake` added to `lspconfig`? 

Would you also add a couple tests with example `cmake` files to `harper-comments/tests/language_support.rs` to make sure it all works properly?

alcroito:
The linked issue is about haskell. Did you perhaps mean to open a separate issue for that project about adding cmake?
I can try, but i'll have to look up their contribution guidelines.

I'll try to add some tests next week.

elijah-potter:
> The linked issue is about haskell. Did you perhaps mean to open a separate issue for that project about adding cmake?

I was hoping @grantlemons would be able to combine the `cmake` and `haskell` changes. We'll likely get it figured out by the time the tests are in.

grantlemons:
I won't be back at my computer until this evening, but I'd be happy to make the change of the PR doesn't get merged first. If it does, there's always the Typst PR.

grantlemons:
> I won't be back at my computer until this evening, but I'd be happy to make the change if the PR doesn't get merged first. If it does, there's always the Typst PR.

The change has been made, I can also add vimscript if y'all figure out a way to do it without a git repository.

elijah-potter:
Actually, I've changed my mind. The `cmake` comment syntax is simple enough that I don't need you to do test files. It looks good on my end, so I'm going to merge it.

PR #354, opened 2025-01-03
feat: Add support for vim script

alcroito:


harper-comments/Cargo.toml
```diff
@@ -27,6 +27,7 @@ tree-sitter-lua = "0.0.19"
 tree-sitter-bash = "0.20.0"
 tree-sitter-java = "0.20.0"
 tree-sitter-nix = "0.0.1"
+tree-sitter-vim = { git = "https://github.com/tree-sitter-grammars/tree-sitter-vim.git", tag = "v0.3.0" }
```
elijah-potter, line 30: We can't depend on `git` repositories. They are unfortunately not allowed by `crates.io`.

harper-comments/src/comment_parser.rs
```diff
@@ -77,6 +78,7 @@ impl CommentParser {
             "go" => "go",
             "c" => "c",
             "cpp" => "cpp",
+            "cmake" => "cmake",
```
elijah-potter, line 81: Looks like this leaked in from your other PR.

elijah-potter (changes requested):

harper-comments/Cargo.toml
alcroito: That's a shame.

Please check my comment in the PR description. None of the published crates are currently buildable with tree-sitter 0.20, which is the current requirement for harper.

Unless there is some other way, I guess this PR can be closed, and re-opened whenever the tree-sitter version is bumped.

harper-comments/src/comment_parser.rs
alcroito: Thanks, removed.

alcroito:
I'll close the PR for now, because i'm not sure how to get rid of the git dependency.

PR #360, opened 2025-01-08
feat(vscode-plugin): Add CMake Support

mcecode:
Add CMake support to the VSCode extension as per #353.

As an aside, I was about to update the changelog but I noticed that it was removed in ace693c00fe7128919c545ffbd22ec0b176b2fcc with the reasoning that it doesn't get updated. I just wanted to ask, what exactly wasn't being updated? Since, as far as I can recall, it was updated every time the extension was. Did it mean that the `## Unreleased` section wasn't being updated to the correct version every release? If so, that can be remedied by adding a command to also bump it in the `justfile` `bump-versions` recipe. I can add that to this PR, if you think it's a good idea.

### DIFF
--- a/packages/vscode-plugin/package.json
+++ b/packages/vscode-plugin/package.json
@@ -32,6 +32,7 @@
 	],
 	"activationEvents": [
 		"onLanguage:c",
+		"onLanguage:cmake",
 		"onLanguage:cpp",
 		"onLanguage:csharp",
 		"onLanguage:git-commit",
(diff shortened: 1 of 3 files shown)


elijah-potter:
> I just wanted to ask, what exactly wasn't being updated? 

I wasn't clear enough, sorry about that. 

The bigger issue was that there would inevitably be duplicate information (once in the CHANGELOG.md and once on the actual release page in GitHub. Updating both is redundant. 

We can return the `CHANGELOG.md`, but direct readers to the release page. Would that work?

mcecode:
Okay, I understand. It's just that the release pages are for Harper in general, while the `CHANGELOG.md` was VSCode-specific. It was primarily for people who want to view changes that only applied to VSCode, plus it's viewable in the extension's page within the editor so users don't need to go elsewhere to view them. Pointing users to the release pages may cause some confusion. For example, the 0.14.0 release was primarily for `harper.js` and tooling, so it didn't have much to do with the VSCode extension.

If you'd really rather not have the duplication, I think it's alright to not have the changelog. Albeit less convenient, I think most users who would want to view it can probably find their way to the release pages and parse them on their own.

PR #361, opened 2025-01-08
docs: added Scoop instructions

elijah-potter:
@EphemeralDev, I've added instructions for installing Harper through Scoop. Since I'm not familiar with how Scoop works, can you confirm that this looks right?


EphemeralDev:
The main bucket is available by default, you could probably remove that unless you want to keep it for odd cases where someone removed it. The install command is fine, it could be shortened to `scoop install harper`. I think specifying the bucket is only necessary if multiple buckets have manifests with the same name.

PR #367, opened 2025-01-10
fix the nvim-lspconfig link

wadsaek:
the link in the neovim documentation leads to a moved page. this pr fixes this by providing the right link


elijah-potter:
Good catch! Merging...

PR #376, opened 2025-01-14
feat(core): create new linter for dollar sign placement

elijah-potter:
I recently wrote a blog post with the title "The Best $25 I Ever Spent". While I was sharing it with a friend, I accidentally placed the dollar sign _after_ the number, which is improper. This PR makes sure that doesn't happen again, at least in Harper-enabled buffers.


hippietrail:
This is actually the rule for all currency symbols in English such as £€¥₩฿₭ etc. Not necessarily for ones written in letters though such as MYR etc.

harper-core/src/currency.rs
```diff
@@ -0,0 +1,57 @@
...
+            '€' => Self::Euro,
+            '₽' => Self::Ruble,
+            '₺' => Self::Lira,
+            '£' => Self::Pound,
+            '¥' => Self::Yen,
+            '฿' => Self::Baht,
+            '₩' => Self::Won,
+            '₭' => Self::Kip,
+            _ => return None,
+        };
```
ccoVeille, line 39: Maybe you would consider adding them all

https://en.wikipedia.org/wiki/Currency_Symbols_(Unicode_block)

harper-core/src/linting/currency_placement.rs
```diff
@@ -0,0 +1,135 @@
+use crate::{
+    patterns::{EitherPattern, Pattern, SequencePattern},
+    Punctuation, Token, TokenKind, TokenStringExt,
+};
+
+use super::{Lint, LintKind, PatternLinter, Suggestion};
+
+static DESCRIPTION: &str =
+    "Currency symbols should always come before the quantity being described.";
```
ccoVeille, line 9: Be careful with this, this is the Emglish notation

But things are complicated

https://fastspring.com/blog/how-to-format-30-currencies-from-countries-all-over-the-world/

Maybe this rule is about making things simple. So maybe you would consider other notation

But it's common to see something like this written

```
$1 = 0.97 €
```

Each part respecting the local convention

The same way you can see

```
Eva Longoria and Tony Parker
Miss Longoria and M. Parker

Vanessa Paradis and Johnny Depp
Mme Paradis and Mr. Depp
```

Because in such cases the nationality of each of them implies the role.

ccoVeille (approved):

PR #395, opened 2025-01-16
docs: added Homebrew as an installation method

elijah-potter:
I just discovered that someone has create a Homebrew formulae for Harper. What joy! I've added it to the installation documentation on the website. 

@lukasmwerner does this all look right? Is there anything that should be added?


lukasmwerner:
Can confirm install works and looks great!

PR #409, opened 2025-01-17
fix(core): indexing problems

elijah-potter:
@lukasmwerner sent me a file that revealed two problems:

- If Harper attempted to make a chunk of text title case (that had a previously lower-case word at the end of input) it would try to access memory out of bounds.
- If the 'UnitedOrganization` rule encountered a correct capitalization of `United States`, it would create a false positive.

This PR solves both problems.

@lukasmwerner, would you mind compiling this branch and confirming that it solves those issues?

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -49771,3 +49771,4 @@ Harper/SM
 a8c/SM
 a11n/1
 a12s/9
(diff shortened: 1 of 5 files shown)


lukasmwerner:
Looks fixed to me!

Issue #416, opened 2025-01-19
feat: Flag copulae/verbs agreeing with complement/object instead of subject


hippietrail:
I've been noticing this in spoken English a lot on YouTube but I don't think it's new and I don't think it's limited to native or non-native English speakers:

Verbs, including auxiliary verbs and copulae (am, are, is, was, were, etc.) agree with the subject, so normally the nomonal phrase to its left. But under some circumstances some people get confused and make it agree with the complement of the copula or object of the verb:

- The main difference are the metadata fields.

  → The main difference is the metadata fields.
  → The main differences are the metadata fields.

hippietrail:
(Trying to submit a PR to fix this where I found it in the docs but damned if git will play along.)

hippietrail:
I'll maintain a list here of examples I find in the wild:
- Honestly, the only real existential **weakness** that I see right now **are** the water desalination plants

hippietrail:
This is one of my biggest pet peeves. Reopening.

hippietrail:
🌲

PR #417, opened 2025-01-19
fix agreement error in docs

hippietrail:
I truly despise git but I finally got it to let me submit a fix in the docs for the problem addressed in issue 416 on agreement with complement or object instead of subject: https://github.com/Automattic/harper/issues/416


ccoVeille (approved):

grantlemons (changes requested):
I think "The main difference**s** are" would make more sense because it's a plural difference in what metadata fields are used.

hippietrail:
It says I closed this myself:
image

That must've been an accident. I just rediscovered it while exploring the source thinking it was a second case of the same mistake. I'll see if I can change it to "differences are" but I made the PR directly through the github website.

ccoVeille (approved):

PR #419, opened 2025-01-19
chore: update description of Oxford rule

ccoVeille:
Recently, the code was updated to suggest the "Oxford comma" with an "or-enumeration".
This commit updates the description to reflect this addition.


grantlemons (changes requested):
If you're changing this anyway, I'm not a fan of the "Here, we make sure that" bit.

I feel it would be better stated as "Enabling this lint checks that", because it makes it feel more like an option.

ccoVeille:
I made the changes @grantlemons 

I also removed the "we" that was making things personal and strange

grantlemons (approved):

PR #421, opened 2025-01-19
feat(harper-core): Mark shebangs as unlintable

### LINKED ISSUE #65: feat: ignore shebangs
Harper should not lint shebangs found at the start of files.

Examples include:

```ruby
#! /usr/bin/ruby
```

```bash
#! /bin/bash
```

sohunpatel:
Closes #65.

## Changes

- Adds check to see if the content is a shebang, in the parse function.
- Add test to comment parser to make sure that comments with shebang are ignored

I am not sure if this is a great solution since this should only apply to comment parsers, but I could not find an elegant solution for that.

### DIFF
--- a/harper-comments/src/comment_parser.rs
+++ b/harper-comments/src/comment_parser.rs
@@ -108,3 +108,23 @@ impl Parser for CommentParser {
         self.inner.parse(source)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{CommentParser, MarkdownOptions, Parser};
+    use harper_core::TokenKind;
+
+    #[test]
+    fn shebang_unlintable() {
+        let source = String::from("#!/usr/bin/env python3")
+            .chars()
+            .collect::<Vec<char>>();
+        let markdown_options = MarkdownOptions::default();
+
+        let parser = CommentParser::new_from_language_id("python", markdown_options).unwrap();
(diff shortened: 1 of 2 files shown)


grantlemons:
I think a better solution would be to mask them out in harper-tree-sitter. See fd5ebae0ddf016fe096bfc16c0985d4a8c9309a1.

elijah-potter:
@grantlemons, @sohunpatel I think the best way to do this would be to create a new implementaion of [the `Masker` trait](https://docs.rs/harper-core/latest/harper_core/trait.Masker.html) that ignores lines with shebangs. From there, it'd wrap it around the inner `Masker` [that `CommentParser` uses internally.](https://github.com/Automattic/harper/blob/567d5143a8c4486b97fa22dec59b72a7d5fdb4d5/harper-comments/src/comment_parser.rs#L12)

Let me know if you have any questions about any of that.

holmanb:
> Let me know if you have any questions about any of that.

@elijah-potter How strongly do you feel about this approach? I'm currently testing out a one-line change which modifies harper-comments/src/masker.rs and I think this would be a bit simpler than creating a new `Masker` implementation, but I suppose what you describe would have cleaner separation.

elijah-potter:
@holman, that depends on what the one-line change looks like. Why don't you open a draft PR and we can take a look?

PR #391, opened 2025-01-15
Typst Test Fixes

grantlemons:
Thanks to @Andrew15-5 for suggestions on #302.

PR #442, opened 2025-01-20
Typst Corrections

grantlemons:
There are many issues with Typst support.

Based on #391, so merge that one first.


harper-typst/tests/run_tests.rs
```diff
@@ -37,3 +37,4 @@ macro_rules! create_test {
 create_test!(complex_document.typ, 0);
 create_test!(simplified_document.typ, 0);
 create_test!(complex_document_with_spelling_mistakes.typ, 4);
+// create_test!(issue_399.typ, 3);
```
elijah-potter, line 40: If you're going to disable it (rather than remove it), you should add a comment here that explains why.

elijah-potter (changes requested):
Give us a brief summary (for posterity) of what problems this PR fixes. I'd like to test your changes locally, but I'm not sure what to look for.

grantlemons:
It changes the newlines after some textual elements, such as headers to paragraph breaks, and marks references in text as unlintable.

PR #445, opened 2025-01-21
fix(core): the Oxford comma applies to `nor` as well

### LINKED ISSUE #420: Oxford comma, aka Serial comma with "nor"
Oxford comma not only applies for `and` and `or`, but also for `nor`

https://www.umaryland.edu/media/umb/oaa/campus-life/writing-center/documents/Commas.pdf

elijah-potter:
Resolves #420

### DIFF
--- a/harper-core/src/linting/oxford_comma.rs
+++ b/harper-core/src/linting/oxford_comma.rs
@@ -81,7 +81,7 @@ impl Linter for OxfordComma {
     }
 
     fn description(&self) -> &str {
-        "The Oxford comma is one of the more controversial rules in common use today. Enabling this lint checks that there is a comma before `and` or `or` when listing out more than two ideas."
+        "The Oxford comma is one of the more controversial rules in common use today. Enabling this lint checks that there is a comma before `and`, `or`, or `nor` when listing out more than two ideas."
     }
 }
 
--- a/harper-core/src/linting/oxford_comma.rs
+++ b/harper-core/src/linting/oxford_comma.rs
@@ -21,7 +21,7 @@ impl OxfordComma {
                 ))
                 .then_noun_phrase()
                 .then_whitespace()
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/oxford_comma.rs
```diff
@@ -150,4 +150,13 @@ mod tests {
             "They enjoy playing soccer, basketball, or tennis.",
         );
     }
+
+    #[test]
+    fn nor_vegetables() {
+        assert_suggestion_result(
+            "I like carrots, kale nor broccoli.",
+            OxfordComma::default(),
+            "I like carrots, kale, nor broccoli.",
```
ccoVeille, line 159: this example seems to come out your childhood 🤣

harper-core/src/linting/oxford_comma.rs
```diff
@@ -21,7 +21,7 @@ impl OxfordComma {
                 ))
                 .then_noun_phrase()
                 .then_whitespace()
-                .then(Box::new(WordSet::all(&["and", "or"])))
+                .then(Box::new(WordSet::all(&["and", "or", "nor"])))
```
ccoVeille, line 24: The description function around line 80 also need to be updated.

And I like the fact you will have to use an Oxford list to write it 😁

harper-core/src/linting/oxford_comma.rs
grantlemons, line 159: I feel like these should have neither at the beginning, but it's gramatically correct as is.

elijah-potter: It does. I still don't like broccoli.

ccoVeille (approved):

ccoVeille:
About https://github.com/Automattic/harper/commit/458566126d5e5d0b1cd2d2b95de5bd6f92f19432, I would like to suggest using co-authored-by tag 😁

Issue #443, opened 2025-01-21
`harper-ls`: Getting "No such file or directory" when editing a file that has no file dictionary


mcecode:
Editing a file that has no corresponding file dictionary gives the following logs:

```text
[2m2025-01-21T05:56:23.135358Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:24.428769Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:25.275200Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:26.253019Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:27.470153Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:29.189194Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:30.675969Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
[2m2025-01-21T05:56:37.179787Z[0m [32m INFO[0m [2mharper_ls::backend[0m[2m:[0m No such file or directory (os error 2)
```

Once I add a word to the file dictionary with code actions, therefore creating one for the file I'm currently editing, the logs go away. It's not a big deal since everything works fine, but it might be confusing or misleading for others who stumble upon these logs.

This happens in both the current master and 0.16.0 builds.

elijah-potter:
I believe this is due to @grantlemons recent changes in #349. I'll draft a fix.

elijah-potter:
Alright, I've got a fix in. Mind checking to make sure everything's good on your end?

mcecode:
Yup. Everything's all good on my end. Thanks for the fix!

PR #447, opened 2025-01-21
Clean Up `harper-ls` logs

elijah-potter:
We've been seeing a lot of erroneous logs being made (see #443), cluttering buffers and making debugging more challenging. This PR aims to reduce this clutter by:

1. Reducing the number of logs `harper-ls` produces.
2. Reducing the noise within each log.


grantlemons (approved):
Looks good, fyi we only have one call at warn level.

grantlemons (approved):

PR #449, opened 2025-01-21
 feat(harper.js): added ability to explicitly set config to the default

elijah-potter:
In editors like Obsidian, we need to know what actual values the default config holds. This PR takes a stateful approach to the matter, allowing you to call a method on a `Linter` to set its configuration to the global default.

This PR also adds a button that will copy a Neovim-specific version of this default configuration to the clipboard for use in Lua configuration of `harper-ls`.

If SveltePress/sveltepress#310 is implemented, we will replace the button with the more reader-friendly alternative.

### DIFF
--- a/harper-wasm/src/lib.rs
+++ b/harper-wasm/src/lib.rs
@@ -105,6 +105,10 @@ impl Linter {
         self.lint_group.config.serialize(&serializer).unwrap()
     }
 
+    pub fn fill_lint_config_with_default(&mut self) {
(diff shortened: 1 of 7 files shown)


elijah-potter:
I've elected to use a stateless alternative. It's much easier downstream in integrations like Obsidian.

grantlemons (approved):
LGTM, is there a way we can add site previews on PRs without spending extra? (created #451)

PR #452, opened 2025-01-21
feat(core): created linter for #426

### LINKED ISSUE #426: Flag "let's us do" or "let's me do"
Some people think "let's" an expression, and don't get that "let's" states for "let us"

So they write:
- let's us do
- let's me do

Examples
https://github.com/search?q=(%22let%27s%20us%22%20OR%20%22let%27s%20me%22)&type=code

elijah-potter:
This PR introduces a rule that searches for incorrect usage of the contraction `let's`. 

@ccoVeille or @hippietrail, before merging I'd like to add more test cases. If have some more, just shoot 'em below.

Closes #426.


ccoVeille:
Let's me
Let's you
Let's them
Let's him
Let's her
Let's us
Lets me
Lets you

hippietrail:
Lets go and check if this lint let's us catch this class of errors.

harper-core/src/linting/lint_group.rs
```diff
@@ -187,7 +188,8 @@ create_lint_group_config!(
     PluralConjugate => false,
     OxfordComma => true,
     PronounContraction => true,
-    CurrencyPlacement => true
+    CurrencyPlacement => true,
+    LetUsRedundancy => true
 );
```
ccoVeille, line 193: Does Rust support tge following notation as other languages do.


```suggestion
    CurrencyPlacement => true,
    LetUsRedundancy => true,
);
```

It could avoid updating unrelated line on each rule addition

harper-core/src/linting/let_us_redundancy.rs
```diff
@@ -0,0 +1,55 @@
...
+            lint_kind: LintKind::Repetition,
+            suggestions: vec![Suggestion::Remove],
+            message: "`let's` stands for `let us`, so including another pronoun is redundant."
+                .to_owned(),
+            priority: 31,
+        }
+    }
+
+    fn description(&self) -> &'static str {
+        "Many are not aware that the contraction `let's` is short for `let us`. As a result, many will incorrectly use it before a pronoun, such as in the phrase `let's us do`."
```
ccoVeille, line 41: Please take a look at

https://github.com/Automattic/harper/issues/426#issuecomment-2606719099

harper-core/src/linting/lint_group.rs
elijah-potter: Rust does, but this is actually on of our macros. I don't think it would be worth the effort to add that in.

elijah-potter:
I've revised it to cover more cases, like those discussed in #470. Let me know if there is anything else that should be added to this rule before I can merge it.

harper-core/src/linting/lint_group.rs
grantlemons: turns out it's really easy

add `$(,)?` to the end of the macro signature

```rust
    ($($linter:ident => $default:expr),* $(,)?) => {
```

PR #453, opened 2025-01-21
feat(#422): add `same than` -> `same as` matcher trigger

### LINKED ISSUE #422: Flag "same than" used instead of "same as"
Pretty common error

https://github.com/search?q=%22same%20than%22&type=code

grantlemons:
Closes #422.

### DIFF
--- a/harper-core/src/linting/matcher.rs
+++ b/harper-core/src/linting/matcher.rs
@@ -186,7 +186,11 @@ impl Matcher {
             "to", "towards" => "towards",
             "though", "process" => "thought process",
             "the", "this" => "that this",
-            "take", "a", "decision" => "make a decision"
+            "take", "a", "decision" => "make a decision",
+            "same", "than" => "same as",
+            "Same", "than" => "same as",
+            "same", "then" => "same as",
+            "Same", "then" => "same as"
         };
 
         triggers.push(Rule {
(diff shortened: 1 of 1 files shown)


elijah-potter:
I would revise this into its own rule by implementing `PatternLinter` with [`SequencePattern::aco`](https://docs.rs/harper-core/latest/harper_core/patterns/struct.SequencePattern.html#method.aco).

See #452 for a good model of what this looks like.

harper-core/src/linting/matcher.rs
```diff
@@ -186,7 +186,11 @@ impl Matcher {
             "to", "towards" => "towards",
             "though", "process" => "thought process",
             "the", "this" => "that this",
-            "take", "a", "decision" => "make a decision"
+            "take", "a", "decision" => "make a decision",
+            "same", "than" => "same as",
+            "Same", "than" => "same as",
+            "same", "then" => "same as",
+            "Same", "then" => "same as"
```
ccoVeille, line 193: Oh if it's as simple as this, I can contribute, I know how to copy-paste !! :rofl:  (see [this](https://github.com/Automattic/harper/pull/455#pullrequestreview-2565863984) for context)

I will open my first PR then, I have another example that would fit here

ccoVeille: Oh, I just found this comment and it seems a refactoring is needed before that

https://github.com/Automattic/harper/pull/453#issuecomment-2605916867

I tried to set up the dev env, and I'm facing rust oddities ...

I will open an issue with the errors I know

ccoVeille: Done https://github.com/Automattic/harper/issues/456

grantlemons: > Oh, I just found this comment and it seems a refactoring is needed before that
> 
> [#453 (comment)](https://github.com/Automattic/harper/pull/453#issuecomment-2605916867)
> 
> I tried to set up the dev env, and I'm facing rust oddities ...
> 
> I will open an issue with the errors I know

Don't worry about that, we'll migrate everything in matcher when the time comes (#472), so just add things to matcher for the time being.

PR #454, opened 2025-01-21
feat: wrote first draft of statistics logging

elijah-potter:
As the authors of Harper, we want to know which rules are being most utilized, so we can properly allocate our time to what is actually useful to our users. This PR sets up a __local-only__ log that keeps track of which lints are being applied and when.

This is done by creating an append-only file in the data directory (which is OS dependent).

This is not analytics. No data is being sent anywhere by `harper-ls`. That said, if you would like to submit your stats logs, it would be a huge help and we would welcome it.


grantlemons:
Could be prudent to include an analysis tool in harper-cli? Nothing too crazy, but just creating a graph from >= 1 log file. I can make an issue and push it to the backlog if so.

hippietrail:
You should also log which rules are enabled and disabled etc. I don't know what settings there are yet.

elijah-potter:
I've written up a page that should give a good description of how this all works: 

image

elijah-potter:
Additionally, I've added a prototype dashboard page (that will be improved over time).

image

packages/web/src/routes/docs/contributors/local-stats/+page.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Local Statistics
+---
+
+Harper keeps track of certain aspects of your writing.
+Things like:
+
+- What words to misspell most often?
```
hippietrail, line 8: This should probably be something like `What words are misspelled most often` or `What words do you misspell most often`

hippietrail:
> I've written up a page that should give a good description of how this all works:
> 

@elijah-potter There is no link to the page - just to the image. I really want to read up on this as I feel it's gone under the radar.

PR #455, opened 2025-01-21
feat(core): created a  new rule that resolves #414

### LINKED ISSUE #414: Optionally flag "somewhat of a"
This seems to have become completely normalized from the millennial generation onward. I only started hearing this variant a few years ago on YouTube. (There may be a regional component I'm not aware of.)

It seems to have come about due to combining "something of a <noun>" and "somewhat <adjective>".

Some organizations might want the traditional style, some writers might be striving to emulate an older style and not realize this phrasing would be an anachronism:

- This may be somewhat of a surprise. → This may be something of a surprise.

Also correct would be "This may be somewhat surprising" but that would be a difficult suggestion to generate automatically I think.

Here's one online discussion I found:
- https://brians.wsu.edu/2016/05/31/somewhat-of-a/

elijah-potter:
I've created a new rule that should cover #414. From here, we mostly just need a couple more test cases.

You'll notice that I set it to be enabled by default in the `LintGroup`. If you disagree with this decision, let me know and I'd be happy to change it.

### DIFF
--- a/harper-core/src/linting/lint.rs
+++ b/harper-core/src/linting/lint.rs
@@ -69,6 +69,11 @@ pub enum Suggestion {
 }
 
 impl Suggestion {
+    /// Variant of [`Self::replace_with_match_case`] that accepts a static string.
+    pub fn replace_with_match_case_str(value: &'static str, template: &[char]) -> Self {
+        Self::replace_with_match_case(value.chars().collect(), template)
+    }
+
     /// Construct an instance of [`Self::ReplaceWith`], but make the content match the case of the
     /// provided template.
     ///
--- a/harper-core/src/linting/lint_group.rs
+++ b/harper-core/src/linting/lint_group.rs
@@ -22,6 +22,7 @@ use super::proper_noun_capitalization_linters::{
 };
 use super::repeated_words::RepeatedWords;
 use super::sentence_capitalization::SentenceCapitalization;
+use super::somewhat_something::SomewhatSomething;
 use super::spaces::Spaces;
 use super::spell_check::SpellCheck;
 use super::spelled_numbers::SpelledNumbers;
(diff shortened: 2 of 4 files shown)


ccoVeille (approved):
LGTM*

*I have 24 years experience in copy-paste and reverse engineering, and I think I could review your PR in Rust, while I have never coded any line of Rust

image

hippietrail:
You can check against this instance in the harper codebase in architecture/page.md

`- [Parser](https://docs.rs/harper-core/latest/harper_core/parsers/trait.Parser.html): A trait that describes an object that consumes text and emits tokens. The name is somewhat of a misnomer since it is supposed to only lex English (and emit [Tokens](https://docs.rs/harper-core/latest/harper_core/struct.Token.html)), not parse it. It is called a parser since most types that implement this trait parse _other_ languages (JavaScript) to extract the English text.
`

I'm not submitting a PR to fix this one because "kids these days" almost certainly don't consider it an area. But you may of course choose to fix it if you wish (-:

----

By the way, is there an easy way to test this? I have a clone of the repo but I haven't attempted to build it yet because I hate build systems and every new project I play with these days has a new build system that entails a day or two down an unfun rabbithole when I could be doing fun OSS coding instead (-:

elijah-potter:
> By the way, is there an easy way to test this? I have a clone of the repo but I haven't attempted to build it yet because I hate build systems and every new project I play with these days has a new build system that entails a day or two down an unfun rabbithole when I could be doing fun OSS coding instead (-:

In #471, I've added a page that should explain some of this. Mind taking a look?

hippietrail:
> #471

I must be a dummy but how do I get to the file to read it? No matter how I click on any link I can find from the PR link I'm only seeing things like diffs.

Issue #118, opened 2024-08-23
Add support for "et al."


engeir:
Similar to #98, it would be nice if "et al." in citations was recognized.

grantlemons:
vs. is another one that we should support

grantlemons:
and etc.

ghorsey:
I took a screenshot of the issue:

![Image](https://github.com/user-attachments/assets/daa54365-bc28-4e6d-8937-757e768efc81)

elijah-potter:
I've drafted #473 that should resolve this issue. Let me know if it does not.

elijah-potter:
This has been resolved. Feel free to reopen if that is not the case.

PR #473, opened 2025-01-22
feat(core): condense multi-token Latin words and phrases

elijah-potter:
Seeks to address #118. As per usual, I'd love more test cases.


elijah-potter:
@grantlemons, would you confirm that this works for the documents you were referencing in the issue?

harper-core/dictionary.dict
```diff
@@ -49778,3 +49778,24 @@ intergenerational
 etc.
 vs.
 et al.
+Laravel/M
```
hippietrail, line 49781: I wonder if it might be worth having a separate dictionary for techy terms from regular English words? They'll accumulate much faster, tend to be mostly proper nouns, and will probably all be applicable across languages later on.

Or maybe have the English words at the top and the techy terms at the bottom?

elijah-potter: That's interesting thought. I've got some big changes to the dictionary system stashed on my system. It's something I'll consider when I start rolling them out.

PR #474, opened 2025-01-22
Improvements to `PluralConjugate`

elijah-potter:
I saw some issues with this rule while working on some documentation. This PR sets "you" to be potentially plural and adds some test cases.


elijah-potter:
It turns out the pronouns weren't the problem. There was an inaccuracy in the dictionary.

Issue #478, opened 2025-01-23
Difficulties following "Author a Rule" docs for Rust noobs?


hippietrail:
The current steps are:

1. Start by forking the Harper monorepo and create a new file under harper-core/src/linting called my_rule.rs.
2.Follow our guide to get your environment setup. (sic)
3. Copy in a template rule (like from that_which.rs).
4. Modify the constructor to create a pattern to look for the problematic text you have in mind.

After I do steps 1. 2. and 3. I hit this problem:

<img width="1098" alt="Image" src="https://github.com/user-attachments/assets/d92600e0-0a93-4e48-9308-769b1d3edc4d" />

The Rust source code is greyed out with the error message
```
This file is not included anywhere in the module tree, so rust-analyzer can't offer IDE services.

If you're intentionally working on unowned files, you can silence this warning by adding "unlinked-file" to rust-analyzer.diagnostics.disabled in your settings.
```
from rust-analyzer

I probably don't want to disable the diagnostic in my settings. I'm not sure what it means for a Rust source file to be included in a "module tree". I guess it's editing some other kind of unspecified file that exists... somewhere? Maybe a `.toml` file?

I have actually programmed a modest amount a Rust before but I've never seen this. Total Rust noobs will likely be even more confused when they hit this step.

I'm pretty sure a new 5th step has to inserted here?

----

**Update** After fiddling around and grepping I think the next steps are there but I'm not experienced enough with Rust to see that's what they were. The phrases "copy in a template rule", "modify the constructor", "export your rule", and "add your rule to the lint group macro call" all might assume more Rust knowledge.

My brain kind of read those lines as something like "make a file with this name" and give it a similar structure to the `that_which.rs` file.

I think at least an intro sentence along the lines of "You'll have to edit x files in y places"?

Shouldn't there also be a test step before submitting the PR?

----

I'ver persisted with this and tried to make a clone of the `that_which` by duping its main file, renaming it to `foo_bar` and doing a global replace of every "that" with "foo" and every "which" with "bar" and in every source file that references `that_which` I'm adding a line after it that does the same for `foo_bar` and then I attempt to build and run it to test if it works using this commandline
```sh
cargo run --bin harper-cli lint ~/harper-test/harper-test.md
```
Which successfully builds, but seems to ignore my new rule.

Here's a diff:
```diff
diff --git a/harper-core/src/linting/lint_group.rs b/harper-core/src/linting/lint_group.rs
index be7ef92..239e827 100644
[...]

PR #482, opened 2025-01-23
docs: rewrote instructions on how to author a rule

elijah-potter:
In #478, it became clear that a lot of information was missing from the instructions for contributing a new rule to Harper.

This PR is a complete rewrite of that documentation.

@hippietrail, does this make more sense?


packages/web/src/routes/docs/contributors/author-a-rule/+page.md
```diff
@@ -6,23 +6,162 @@ title: Author a Rule
...
+Create a new file under that directory with the name of your rule in `snake_case`.
+If you can't decide yet, just call it `my_rule.rs`.
+
+Don't put anything in this file yet, there's some bookkeeping we have to do first.
+
+## Register Your Rule
+
+Before we start describing to Harper what grammatical errors to look for, we need to register your rule within the system.
+
+First, add your rule's module to the tree by adding it to [the top of the mod file](https://github.com/Automattic/harper/blob/master/harper-core/src/linting/mod.rs).
```
hippietrail, line 47: Might I suggest something like this:

You need to add it in two places. First add a "module declaration":
```rust title="harper-core/src/linting/mod.rs"
mod an_a;
mod avoid_curses;
mod boring_words;
// [svp! df:+]mod my_rule;
```
Then a little further down add a "public use declaration":

```rust title="harper-core/src/linting/mod.rs"
pub use an_a::AnA;
pub use avoid_curses::AvoidCurses;
pub use boring_words::BoringWords;
// [svp! df:+]pub use my_rule::MyRule;
```

hippietrail:
> In #478, it became clear that a lot of information was missing from the instructions for contributing a new rule to Harper.
> 
> This PR is a complete rewrite of that documentation.
> 
> @hippietrail, does this make more sense?

Yes it's better and better. I made one suggestion. I also think it would be more helpful for non-rust-coders to include an example where I think it's now just `unimplemented()`.

But also my attempt at a new still doesn't work now that I've set up `just` and using `just lint` instead of `cargo run`. It builds and runs and checks the file but only finds the old lints.

elijah-potter:
> But also my attempt at a new still doesn't work now that I've set up just and using just lint instead of cargo run. It builds and runs and checks the file but only finds the old lints.

That's odd. Would you mind opening a draft PR so I can get a look at what's going on?

packages/web/src/routes/docs/contributors/author-a-rule/+page.md
elijah-potter: I feel that the single code block with both edits is simpler. The correct placement of the lines will be handled by `rustfmt` before we merge the PR.

PR #483, opened 2025-01-24
feat(core): allow trailing commas in the `lint_group` macro

elijah-potter:
From [the comment on #452](https://github.com/Automattic/harper/pull/452#discussion_r1927630529).


hippietrail:
I used to hate trailing commas because they're not logical. But now I love them because they make for much cleaner diffs.

grantlemons (approved):

Issue #218, opened 2024-10-11
Zig Support


elijah-potter:
@nathan-george has suggested we add Zig support to Harper. I think this is a great idea.

nuIIpointerexception:
Maybe this is beneficial?
You surely have this under your radar, but in case not:
https://github.com/tree-sitter-grammars/tree-sitter-zig.git

elijah-potter:
Related to #79

fmorroni:
Is this not planned as in zig will never be supported? Or was it just closed because it's already covered by https://github.com/Automattic/harper/issues/79?

hippietrail:
Linear was not supposed to be running around closing issues.

Issue #79, opened 2024-04-27
feat: more languages supported


elijah-potter:
I'd like to expand the number of languages supported for comment checking. There are a number of additional considerations for structured comments (like JavaDoc), but some should be pretty straightforward.

Running list:
- Bash
- YAML
- HTML

Feel free to add suggestions to this list.

elijah-potter:
Alright. At the time of writing, we've got JavaDoc, HTML, and Bash supported, as well as `gitcommit` buffers.

mostafaqanbaryan:
If you could add `php` to the list as well, it would be great

Vinzent03:
Adding `dart` would be great as well.

kai-tub:
Hey, markdown support seems pretty advanced. :+1:
But are there plans to also support other markup languages, such as [djot](https://djot.net/), with the same level of functionality?
Or will only markdown be seen as a first-class citizen in the near future?

Either way, I am thrilled about the project and being able to use a solid spell-checker LSP :partying_face: :

PS: Are you planning on adding a sponsoring option?

elijah-potter:
Coming back to this issue after a solid period of stagnation, to give an update.

#332 has added Haskell support, #354 is adding Vimscript, and #353 is adding `cmake` support. These will all be considered "rough" until someone has the time to go in and touch them up (which is a good first issue to tackle, if anyone is interested).

I plan on adding PHP and potentially PowerShell later today. 

> But are there plans to also support other markup languages, such as [djot](https://djot.net/), with the same level of functionality?

We've got an open PR for Typst support: #302. I do not _personally_ have time in the near future to add similar levels of polish to a new language as the Markdown parser. That said, we are entirely open to PRs on the topic, and I can answer any questions you may have. I'm not sure how complex djot is, but it may be as simple as importing the right Treesitter parser.

> Either way, I am thrilled about the project and being able to use a solid spell-checker LSP 🥳 

That's great to hear! I aim to please.

> PS: Are you planning on adding a sponsoring option?

[We might be a little past that.](https://techcrunch.com/2024/11/21/wordpress-com-owner-automattic-snaps-up-grammar-checker-harper)

krupan:
Emacs org-mode support would be rad

hippietrail:
I just had a peep at the supported languages and here are some I note to be missing:
- Assembly
- Elixir
- Kotlin
- Odin
- Zig

[...]

PR #495, opened 2025-01-24
Add more languages

elijah-potter:
In #79, we discussed some of the missing programming languages for `harper-ls`. This PR adds several of the requested languages.


elijah-potter:
I wanted to resolve #218 in this PR, but it looks like there are some dependency resolution problems I don't have time to tackle right now.

PR #496, opened 2025-01-24
fix(ls): potential deadlock causing #486

### LINKED ISSUE #486: VSCode: Stuck on "Getting code actions from 'Harper'"
This was mentioned in #441. Filing as an issue so we can track it.

The issue seems to be intermittent and is hard to replicate, so I tried a bunch of things and could finally replicate it using the following steps:

1. Open VSCode with the Harper extension disabled.
2. Open a file that `harper-ls` can lint like a Markdown file.
3. Enable the Harper extension.
4. Type anything in the file you opened and try to save it with `ctrl+s`.
5. Saving the file gets blocked indefinitely until you click "Skip" on the pop-up that shows up:
   Image

Some notes:

- The extension does not work (you get no diagnostics) even after you click on "Skip".
- Step 2 is important, the issue does not occur if you enable the extension while no files that `harper-ls` can lint is open.
- The `harper-ls` process that was spawned by the extension when it is activated in step 3 does not get terminated even after the extension is deactivated. The process instead, at least for my system, gets attached to systemd instead of VSCode.
- I'm not sure why this is happening or if this is the only way to replicate this issue.

elijah-potter:
I haven't been able to replicate #486 on my system using Visual Studio Code or Neovim.

That said, I've had a comb though `backend.rs` and there's a spot where a deadlock _may_ be occurring under special circumstances. @mcecode, would you mind testing this branch to see if it resolves #486?

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -289,10 +289,15 @@ impl Backend {
         url: &Url,
         range: Range,
     ) -> JsonResult<Vec<CodeActionOrCommand>> {
-        let (config, mut doc_states) = tokio::join!(self.config.read(), self.doc_state.lock());
+        let (config_guard, mut doc_states) =
+            tokio::join!(self.config.read(), self.doc_state.lock());
         let Some(doc_state) = doc_states.get_mut(url) else {
             return Ok(Vec::new());
         };
+
+        // Allow writes as soon as possible to avoid deadlocks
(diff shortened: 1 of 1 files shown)


mcecode:
Hi, sorry it took me a while to get back to you on this. Unfortunately, I can still reproduce the issue even with this fix.

elijah-potter:
This was fixed in #612.

Issue #465, opened 2025-01-22
Flag phrases used as qualifiers when they're not hyphenated


hippietrail:
This is related to https://github.com/Automattic/harper/issues/434 but is not the same. In particular, this general case is probably harder to detect.

I'll add examples I find in the wild here:
- it is forced to read in **fixed sized** chunks → it is forced to read in **fixed-sized** chunks

As with the other issue linked, hyphenation has been losing popularity for decades, so this definitely needs to be optional.

Issue #433, opened 2025-01-20
feat: Flag using a plural noun as a modifier for another noun


hippietrail:
This is very common in non-native English but I've started seeing in younger native English speakers' writing sometimes. Also there are cases where it's not an error
- Incorrect stack **variables detection** and usage on x86 → ... variable detection ...

When it is fine it's often acceptable as both singular and plural:
- There was a huge **drug bust** in San Diego.
- There was a huge **drugs bust** in San Diego.

In other cases, the plural qualifier is used to disambiguate from another sense if the singular qualifier were used instead. I'll add some examples when I spot some in the wild.

I'll start compiling a list of exceptions to the rule here as I encounter them in the wild:
- finals spot
- finals win

And here I'll make a list of instances of this mistake I spot in the wild:
- High-Resolution 3D **Assets** Generation → **Asset** Generation
- SATA uses the same basic ATA and ATAPI **commands** sets → **command** sets
- their **summits** certificates were cancelled → their **summit** certificates were cancelled
- A **shoes** company. A hotel. → A **shoe** company. A hotel.
- Add **controllers** support → Add **controller** support
- ... is a unified view for visualizing the **events** history for a file. → ... is a unified view for visualizing the **event** history for a file.
- i have to mess with tyhe **colours** scheme → i have to mess with tyhe **colour** scheme
- for Windows devs to troubleshoot dll load **dependencies** issues.→ ... dll load **dependency** issues.

hippietrail:
I've been working on this today and the edge case are way more than expected!

The intuitive pattern is: `plural_noun , noun`

But of course very many words are multiple POS at once. So this will unexpectedly match
- causes amazement

So maybe we should only match a pattern like `+plural_noun-verb , noun`
But this will unexpectedly reject:
- colours scheme
- commands sets
- events history
- shoes company
- summits certificates

ccoVeille:
@hippietrail started working on this, it should not be stalled

Issue #432, opened 2025-01-20
Flag "worth to"


hippietrail:
Very common among non-native English speakers.

Most common is "worth to verb" → "worth verbing"

But sometimes a suggestion would include "worth it".

Here's some examples from github:

- **Worth to add** a closed current loop check in getFT?
- Check if it's **worth to use** flutster for testing
- Question: is it **worth to store** the reference to a map?

PR #510, opened 2025-01-26
Apply rule from issue #465 to core document.rs

hippietrail:
when best-effort is an adjective it should by connected using a hyphen

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -233,7 +233,7 @@ impl Document {
     }
 
     /// Searches for quotation marks and fills the
-    /// [`Punctuation::Quote::twin_loc`] field. This is on a best effort
+    /// [`Punctuation::Quote::twin_loc`] field. This is on a best-effort
     /// basis.
     ///
     /// Current algorithm is basic and could use some work.


ccoVeille:
Two remarks:

- using a referencesuch as `#465` (I'm using back ticks on purpose here) in the PR title doesn't create a link between the issues: ypu have to use the description or comment

- a PR title should be meaningful. `add rule for #465` is not. Especially if you consider the PRs might at some point have the same format. Consider having such a list:

- `add rule for #465`
- `fix for #462, and add #432`
- `revert #433`
...

ccoVeille (approved):

hippietrail:
> Two remarks:
> 
> First, using a reference such as `#465` (I'm using back ticks on purpose here) in the PR title doesn't create a link between the issues: ypu have to use the description or comment
> 
> Then, a PR title should be meaningful. `add rule for #465` is not. I help to find back what it's about.
> 
> Especially if you consider the PRs might at some point have the same format. Consider having such a list:
> 
> * `add rule for #465`
> * `fix for #462, and add #432`
> * `revert #433`
>   ...

Thanks. I'm still struggling with git and github but I thought this was showing that it did create a link: 
image

This PR doesn't add any rule, it's just fixing spelling which happens to be the same spelling issue I made that PR about. Most of my spelling/typo/grammar PRs have been very short-lived so I wasn't as careful with the wording as with feature PRs.

ccoVeille:
Nothing is short lived.

Once the PR will be merged, they will stay forever

ccoVeille:
> Thanks. I'm still struggling with git and github but I thought this was showing that it did create a link: 
> image

The commit did, because you mentioned the issue in it. But not the PR

Take a look here.

You did it right



You mentioned in the description

And then in the issue, you get a mention of the PR

hippietrail:
If I'm following you, those I did that way because git & github seem to want commit messages to be very short so I tried to make them information-dense without being long. But maybe I'm second-guessing how it's supposed to work while I struggle to learn?

Issue #364, opened 2025-01-09
Unusual words


Elwood4415:
One of the few challenges I am having with Harper is spelling. I am a pastor who uses numerous theological/biblical words (Greek word "pistis") as well as proper names ("Julien") that are understandably not included in the dictionary that Harper is apparently drawing from. Unfortunately, I am given neither the option to add the word to the dictionary, nor to ignore the word. I would be okay with that, except when I change Spell Check in the options for Harper from "default" to "off" it seems to turn itself back on the next time I use Obsidian. Is there a way to add words to the dictionary or to permanently turn off the spell check option so that I can use the Obsidian spell checker instead?

Elwood

elijah-potter:
> Unfortunately, I am given neither the option to add the word to the dictionary, nor to ignore the word. I would be okay with that, except when I change Spell Check in the options for Harper from "default" to "off" it seems to turn itself back on the next time I use Obsidian.

Hey @Elwood4415, I'm sorry to hear that. That is absolutely not intended behavior. I'm looking into it now. I'll update you once I've found the problem.

elijah-potter:
Alright, I believe I've solved the problem of your settings getting reset. The fix will be available in the next release. I'll let you know when that's available so you can confirm that it works! Thanks for letting us know when you ran into this problem. I hope the rest of your day goes well :). 

At some point in the near future, we'll add the ability to manage a Harper-specific dictionary in Obsidian.

Elwood4415:
Super, thanks for looking into it. And I appreciate your doing it so quickly!

elijah-potter:
> I would be okay with that, except when I change Spell Check in the options for Harper from "default" to "off" it seems to turn itself back on the next time I use Obsidian.

We recently released version [`0.16.0` of the plugin.](https://github.com/Automattic/harper-obsidian-plugin/releases/tag/0.16.0) Would you mind checking if your problem persists with the new version?

Elwood4415:
I did download the update, and it is working. When I uncheck spell check, it stays unchecked. Thank you.

I do have a question, however. On the grammar checker, when Harper finds something it doesn't like, it will provide me with options that it thinks will resolve the issue. There are times when I don't agree with the change, but there is no way to dismiss the warning. Is this by design?

elijah-potter:
> I did download the update, and it is working. When I uncheck spell check, it stays unchecked. Thank you.

That's great to hear!

[...]

PR #529, opened 2025-01-28
Ignore Lints

elijah-potter:
In conversations I've had about Harper (including in #364), a common request has been the ability to ignore individual lint results. I've avoided tackling this problem until now, since I was concerned it would disincentivize us from fixing false-positives in our rules. Now that we're rolling out more integrations, the context has shifted enough that this is now an important feature.

## The Plan

I plan on making significant changes to the `harper-core::linting` module to make this happen.
The implementation should have a general-enough API that it can be consumed in `harper-ls` (as a code action), and consumers of `harper.js`.

We will store each lint to be ignored in a persistent manner, that is compatible with whatever client is consuming Harper.

This PR will remain a draft until completion.


elijah-potter:
To summarize what I've done:

I've introduced a structure (`IgnoredLints`) to organize, store and identify lints that have been previously marked as "ignored". I've the `Linter` type in `harper.js` (not to be confused with the Rust `Linter` type) composes `IgnoredLints` to allow JavaScript consumers to ignore lints in a privacy-respecting way, as well as store those settings as a string. 

I then added UI for `harper-ls` and the Obsidian plugin to allow users to ignore lints.

PR #533, opened 2025-01-29
feat(core): even more holidays

hippietrail:
I went through and verified every two-word "Day" holiday in the English Wiktionary. These are the ones that passed muster.


harper-core/src/linting/proper_noun_capitalization_linters.rs
```diff
@@ -114,29 +114,120 @@ create_linter_for!(
...
+                        "Mother's",
+                        "Mother’s",
                         "Mothers'",
+                        "Mothers’",
+                        "Nakba",
+                        "Nevada",
+                        "Occupation",
                         "Parents",
+                        "Patrick's",
+                        "Patrick’s",
```
elijah-potter, line 192: I see you have duplicates here with different apostrophe characters. What is the motivation for that?

hippietrail: I didn't know if we handled them but I knew there was a lint to convert " to “ and ” and I knew if they shouldn't be there we'd just remove them. I didn't actually try to look into what the logic currently does. A quick check shows ' and ’ treated identically. So remove them?

ccoVeille: The Unicode character `’` is a variation of ASCII `'` one.

The Unicode one is less common, but often recommended for the possessive in English.

Microsoft Word replaces it the `'s` by `’s` depending on your locale and setting.

https://en.wikipedia.org/wiki/Apostrophe

While I'm OK to see then caught there. I have a problem with the fact each line in this file needs to be doubled or tripled.

I think each word in the list should be listed only once and combinations should be added programmatically.

So for `foo` word, `foo's`, `foo’s`, `foo'`, `foo’` should be added to the combination.

Be careful about the classic error of "it ends with an s, so adding an apostrophe is enough", the rules are complex

https://new.sewanee.edu/academics/writing-center/resources-for-students/handouts-and-writing-resources/apostrophes/#:~:text=1.,nouns%20(cars%2C%20protestors).

elijah-potter: Remove the duplicate lines. I'll modify the implementations of `WordSet`  and `SequencePattern` to account for the difference.

ccoVeille:
> Remove the duplicate lines. I'll modify the implementations of `WordSet`  and `SequencePattern` to account for the difference. 

_Originally posted by @elijah-potter in https://github.com/Automattic/harper/pull/533#discussion_r1934549983_

Is there an issue/commit/PR for this?

Issue #536, opened 2025-01-29
Compilation error when installing with cargo


eduardz1:
Hello, whne trying to install harper with the latest release with the command `cargo install harper-ls --locked` I get the following error

```
The following warnings were emitted during compilation:

warning: tree-sitter-html@0.19.0: src/scanner.cc:2:10: fatal error: 'algorithm' file not found
warning: tree-sitter-html@0.19.0:     2 | #include <algorithm>
warning: tree-sitter-html@0.19.0:       |          ^~~~~~~~~~~
warning: tree-sitter-html@0.19.0: 1 error generated.

error: failed to run custom build command for `tree-sitter-html v0.19.0`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installGVVONI/release/build/tree-sitter-html-cd40957d7cd4904c/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=src/parser.c
  OUT_DIR = Some(/tmp/cargo-installGVVONI/release/build/tree-sitter-html-2d0d18edfd82bb7f/out)
  TARGET = Some(x86_64-unknown-linux-gnu)
  OPT_LEVEL = Some(3)
  HOST = Some(x86_64-unknown-linux-gnu)
  cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
  CC_x86_64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
  CC_x86_64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
  RUSTC_WRAPPER = None
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some(false)
  CARGO_CFG_TARGET_FEATURE = Some(fxsr,sse,sse2)
  cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
  CFLAGS_x86_64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
  CFLAGS_x86_64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=HOST_CFLAGS
[...]

PR #538, opened 2025-01-29
docs: mention #536 on installation pages

elijah-potter:
In #536, it was revealed that you may need an additional dependency on certain platforms. This PR updates the documentation to mention it.

### DIFF
--- a/packages/web/src/routes/docs/integrations/language-server/+page.md
+++ b/packages/web/src/routes/docs/integrations/language-server/+page.md
@@ -14,7 +14,7 @@ To install `harper-ls`, simply run:
 cargo install harper-ls --locked
 ```
 
-If you have issues with a missing `algorithm` file and you're on a Debian-based Linux distribution, try installing `libstdc++-14-dev`.
+If you are on a Debian-based Linux distribution, you may need to install `build-essential`.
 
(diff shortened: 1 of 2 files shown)


elijah-potter:
@eduardz1 does this look good?

eduardz1:
> @eduardz1 does this look good?

Yes! Perhaps I can try investigating if it's reproducible in a VM and if it's really the only package that needs to be installed. I just don't have too much time for it this week, sorry

eduardz1:
I can't reproduce the bug, in `Ubuntu 24.04` the package `build-essential` already bundles `libstdc++-13-dev` so I can't really see any reason why it wouldn't pick that up.

I have tried both in a `Ubuntu 24.04` and `Pop_OS! 24.04` with the `Cosmic` alpha, on `Ubuntu` I can confirm that, other than the rust tool chain, `build-essential` is required, on `Pop_OS!` it's already installed out of the box.

I would suggest only recommending installing `build-essential`

elijah-potter:
@eduardz1, sounds good. Am I free to merge?

eduardz1:
> @eduardz1, sounds good. Am I free to merge?

Yes

PR #539, opened 2025-01-29
feat: issue templates

elijah-potter:
The number of issues we've been getting is great. This commit helps newcomers structure them in a way that reduces the amount of back-and-forth needed to get them resolved.

### DIFF
--- a/.github/ISSUE_TEMPLATE/report-grammatical-error.md
+++ b/.github/ISSUE_TEMPLATE/report-grammatical-error.md
@@ -1,0 +1,20 @@
+---
+name: Report Grammatical Error
+about: If a grammatical error is not found by Harper, let us tus now
+title: ''
+labels: enhancement, harper-core, linting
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/report-grammatical-error.md
```diff
@@ -0,0 +1,20 @@
...
+
+Give a detailed description of the grammatical error Harper should be able to find.
+
+## Resources
+
+If there are any resources online we can reference that described grammatical rules, let us know.
+
+## Examples
+
+Please provide at least three examples of the grammatical error in real text.
```
hippietrail, line 20: Is it worth asking that they be real (encountered in the wild), or that some of them be real, or marking which ones were real vs contrived examples? Or adding a note that real ones can be more helpful in developing the linter than contrived examples alone?

Issue #543, opened 2025-01-30
bug: Hexadecimal numbers in code comments should not be flagged


hippietrail:
**Describe the bug**
Hexadecimal numbers are flagged as errors. They're about as common as decimal numbers in source code comments so should be treated the same way.

**To Reproduce**
1. Create a new file in a Harper-supported programming language.
2. In a code comment write some hex numbers.
3. Savour the squigglies.

**Expected behavior**
They should be skipped, just like decimal numbers are.

**Screenshots**

<img width="685" alt="Image" src="https://github.com/user-attachments/assets/63ec8f88-58ef-43c8-96d0-c818e9bce0d4" />

**Platform**
What platform has the issue?
Verified in both harper-cli and VS Code

**Additional context**
The most common hex format today is `0xabcdef` or `0x56789ABC`

Older formats are still seen in assembly and in documentation and code comments:
- `$abcdef`
- `0x56789ABCh`

The affixes are mutually exclusive. You should only ever see _one_ of `0x-` or `$-` or `-h`

I don't think `0X` with a capital `X` in the prefix or `H` with a capital as the suffix occur, but maybe we should check.

All formats allow either lowercase or uppercase for the six letters, but never a mix.

The number of digits is usually a power of two and is even more commonly even, which may or may not be useful heuristics for disambiguation if needed.

Hex numbers also occur without any affixes or sigils. This means they can look like words if they consist only of the letters `a`, `b`, `c`, `d`, `e`, and `f` so maybe we'd put a feature through some tests before deciding whether to pass these through to the spellchecker or not?

It might be worth treating text documents different to code comments for this lint.

hippietrail:
> I believe this has been fixed, no?

Correct. I thought it hadn't been merged but it has.

PR #553, opened 2025-01-31
feat(core): start support for hex numbers

hippietrail:
towards #543

based on the existing `lex_number` which cuts off when the digits finish, which is fine because a unit of measurement etc might immediately follow. But with a hex number a non-hex letter might follow and not be flagged if it's a single letter. Also if it's very wrong, the start won't be flagged if it's `0x` and at least one hex digit. Examples will help illustrate:

image

So just a draft PR to solicit thoughts and mods.

### DIFF
--- a/harper-core/src/lexing/mod.rs
+++ b/harper-core/src/lexing/mod.rs
@@ -3,6 +3,7 @@ mod hostname;
 mod url;
 
 use hostname::lex_hostname_token;
+use ordered_float::OrderedFloat;
 use url::lex_url;
 
 use self::email_address::lex_email_address;
@@ -24,6 +25,7 @@ pub fn lex_token(source: &[char]) -> Option<FoundToken> {
         lex_tabs,
         lex_spaces,
         lex_newlines,
+        lex_hex_number,
         lex_number,
         lex_url,
         lex_email_address,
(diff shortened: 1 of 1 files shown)


hippietrail:
I'm still figuring out GitHub. I made this PR a draft because it's not ready to be merged.

But I think "draft" also means "not ready for review" but I specifically want it reviewed to get feedback on what it should do with text "stuck to" the hex that is not hex. And perhaps how to change the code to handle comply to what's decided. Especially if we want to flag such "hex with appendages".

If my understanding of draft status for a PR is still wrong, please let me know.

elijah-potter (changes requested):
@hippietrail, the code looks good. Would you mind adding some test cases under `harper-core/tests/test_sources`? 

I would add a couple Markdown files that should be considered "clean" with this lexer added. Make sure you actually import them in `harper-core/tests/run_tests.rs`.

hippietrail:
> @hippietrail, the code looks good. Would you mind adding some test cases under `harper-core/tests/test_sources`?
> 
> I would add a couple Markdown files that should be considered "clean" with this lexer added. Make sure you actually import them in `harper-core/tests/run_tests.rs`.

Are you sure? Did you see my concerns about what to do regarding non-hex text immediately after the hex with no space etc between?

harper-core/src/lexing/mod.rs
```diff
@@ -186,4 +225,37 @@ mod tests {
...
+                Some(FoundToken {
+                    token: TokenKind::Number(_, None),
+                    ..
+                })
+            ));
+        }
+    }
+
+    #[test]
+    fn lexes_bad_hex() {
```
elijah-potter, line 253: I would rename this to `does_not_lex_bad_hex`. Even better would be to split these two tests into many smaller ones.

hippietrail: Went with the "even better" option.

ccoVeille:
I'm worried about possible reporting of leet speak like 0xf0rd

Then it's a matter of balance, what is the more likely to be found in text? Hexadecimal or leet speak ?

I updated #555 accordingly.

I would prefer hex rules to be merged after leet speak is handled #598

ccoVeille:
> I'm worried about possible reporting of leet speak like 0xf0rd
> 
> Then it's a matter of balance, what is the more likely to be found in text? Hexadecimal or leet speak ?
> 
> I updated #555 accordingly.
> 
> I would prefer hex rules to be merged after leet speak is handled #598
> 

@elijah-potter you merged. It's OK, but can I get a feedback on what I explained?

Issue #355, opened 2025-01-05
Serial comma


nb:
Have you considered adding an Oxford/serial comma rule? It's more and more common, also part of Chicago Manual of Style (even if optional).

trinks-slam818:
Hello, we tried to solve the issue.

This is what we did:

Implemented a new Oxford comma linter to check for proper usage of serial commas in lists. Updated the main linting module to include the new linter.

You can review changes in this commit: https://github.com/trinks-slam818/Automattic-harper-355/commit/a6446b33dbd5c00b475b166954d945f024a77aa4.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

grantlemons:
@elijah-potter are you able to block @trinks-slam818 or would you need to go to an Automattic person to do that?

elijah-potter:
@nb, I have! It should be pretty straightforward. It's on my list for today. 

@grantlemons, it looks like GitHub did it for us.

nb:
Thanks, @elijah-potter, no rush.

elijah-potter:
I've written an [initial rule](https://github.com/Automattic/harper/blob/master/harper-core/src/linting/oxford_comma.rs) for Oxford commas. For now, it only looks for noun phrases. I'll come back to it soon to enable adjective phrases as well. Until then, I'm going to leave the issue open.

hippietrail:
There should also be the opposite option. I am an Oxford comma user but there are style guides on both sides. This one probably warrants a 3-way setting: "Enforce serial commas", "Disallow serial commas", and "Off".

PR #562, opened 2025-02-02
feat(core): Lint for those who don't like Oxford commas

hippietrail:
I do use them, but many dislike them.
As I noted on issue https://github.com/Automattic/harper/issues/355#issuecomment-2591733027


hippietrail:
Since there is not yet any mutual-exclusive switch for opposing preferences, both "Oxford comma" and "No Oxford comma" can be active at once. A side-effect is any lists not flagged for either probably indicate some imperfection in the sentence parsing.

elijah-potter:
I'm quite tempted to merge this. Once I do, @hippietrail would you mind using Harper with this linter enabled for a few days to catch any issues? If you find any, we can open a new PR with fixes.

hippietrail:
> I'm quite tempted to merge this. Once I do, @hippietrail would you mind using Harper with this linter enabled for a few days to catch any issues? If you find any, we can open a new PR with fixes.

Absolutely! You may notice I filed a number of issues with other linters yesterday.

elijah-potter:
Fantastic. Merging...

hippietrail:
> Fantastic. Merging...

Try as I might I cannot get it to work now in VS Code or Windsurf. It does work with `harper-cli` though.
Could it be related to there not being an option in the settings for `NoOxfordComma`?

???

As I came to write this it has now flipped and it's only flagging the sentences with Oxford commas and not the ones without. It's almost 1am so maybe I'm just missing something. I'm going to sleep on it but would be great if somebody could test on their system.

mcecode:
> Try as I might I cannot get it to work now in VS Code or Windsurf. It does work with `harper-cli` though. Could it be related to there not being an option in the settings for `NoOxfordComma`?

It should still work as long as the default setting for `NoOxfordComma` is true in your `harper-ls` build. If you'd like there to be a setting for it within VSCode when you're trying it out in the extension host, you can run `just update-vscode-linters` to add any missing linters that's in the current master build but not in the latest release before opening the extension host.

hippietrail:
> just update-vscode-linters

I just committed what I was working on to go back to master to see if it's still happening. And I now have two pinned cores. `just update-vscode-linters` and reloading the window did not make both Oxford lints work again.

I wonder if switching branches with `harper-cli` running is causing the pinned cores, and that in turn might be causing the old `harper-cli` to still be running instead of the updated one with both the "Oxford" and "no Oxford" lints?

mcecode:
Sorry it took me a while to get back to you on this.

> I just committed what I was working on to go back to master to see if it's still happening. And I now have two pinned cores. `just update-vscode-linters` and reloading the window did not make both Oxford lints work again.

`just update-vscode-linters` is only meant to allow you to set the new `NoOxfordComma` rule within VSCode. Did the Oxford comma lints not work despite setting both of them to `true` within VSCode?

> I wonder if switching branches with `harper-cli` running is causing the pinned cores, and that in turn might be causing the old `harper-cli` to still be running instead of the updated one with both the "Oxford" and "no Oxford" lints?

So are you having problems with `harper-cli` now or did you mean `harper-ls`  when you said `harper-cli`? If it's the latter, yes, changing branches and running different/old versions of `harper-ls` may be what's causing the Oxford comma lints to not work.

hippietrail:
> Sorry it took me a while to get back to you on this.
> 
> > I just committed what I was working on to go back to master to see if it's still happening. And I now have two pinned cores. `just update-vscode-linters` and reloading the window did not make both Oxford lints work again.
> 
> `just update-vscode-linters` is only meant to allow you to set the new `NoOxfordComma` rule within VSCode. Did the Oxford comma lints not work despite setting both of them to `true` within VSCode?
> 
> > I wonder if switching branches with `harper-cli` running is causing the pinned cores, and that in turn might be causing the old `harper-cli` to still be running instead of the updated one with both the "Oxford" and "no Oxford" lints?
> 
> So are you having problems with `harper-cli` now or did you mean `harper-ls` when you said `harper-cli`? If it's the latter, yes, changing branches and running different/old versions of `harper-ls` may be what's causing the Oxford comma lints to not work.

`harper-cli` always worked for both `OxfordComma` and `NoOxfordComma`. But `harper-ls` Would only work with one until suddenly it switched to only working with the other. But now they are both working. It seemed odd that killing the `harper-ls` processes after changing branches didn't fix. I may have mixed them up when I was at my most frustrated and confused.

hippietrail:
> In VSCodium, it doesn't tell me to remove it with this option disable, but it does tell me to add it with this option enabled. Luckily I'm for Oxford commas, so I'm happy.

Until now I only had one of the Oxford comma options show up, but now I see both at last.
Though now I see some other problems... will report...

PR #592, opened 2025-02-05
fix(core): add `on` to the list of special cases

elijah-potter:
I was writing an article for my blog when I noticed that Harper wouldn't detect repetitions of the word `on`. This PR fixes that.

### DIFF
--- a/harper-core/src/linting/repeated_words.rs
+++ b/harper-core/src/linting/repeated_words.rs
@@ -16,6 +16,7 @@ impl RepeatedWords {
             special_cases: vec![
                 smallvec!['i', 's'],
                 smallvec!['a'],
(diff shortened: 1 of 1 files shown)


ccoVeille:
I'm using the following linter for code

https://github.com/Abirdcfly/dupword

It's logic is based on repetition.

I do think it could a better approach than adding them one by one

elijah-potter:
@ccoVeille, the underlying implementation of the rule checks whether the repeated repeated word has any homographs (and therefore _can_ be repeated). There are some special cases though (like `on`), where there are homographs, but still shouldn't be repeated.

ccoVeille:
OK thanks, it's clearer now I looked at all the tests in the rule and not only the one that was added.

Issue #528, opened 2025-01-28
feat: Need to handle "open compounds" with parts that are not words in their own right.


hippietrail:
English has three kinds of compound nouns. Or three ways of writing them:
- closed : "backseat"
- hyphenated : "just-in-time" (as an adjective)
- open : "can opener"

Closed compounds are never tricky, they're either in the dictionary or they're not (for English at least).

But there are some words or phrases that contain one or more "orthographic words" that won't be in the dictionary on their own, only as part of the compound or phrase:
- cul de sac / cul-de-sac

"Sac" happens to an English word. "De" is marginal in English but is listed in several English dictionaries. "Cul" is not usually listed in English dictionaries but "cul-de-sac" should be. These days fewer and fewer people use the hyphenated forms so it mostly appears as "cul de sac" now. Of course we currently flag "cul":

<img width="601" alt="Image" src="https://github.com/user-attachments/assets/a3e4cfca-975b-47e9-8d63-9a48b19a9ad1" />

(An "orthographic word" is what people would consider a "word", ie. separated by spaces from the words around it. May be opposed to "semantic word" or "dictionary word" etc. These terms are probably not universal.)

Start of a list of examples I find in the wild - seems like Latin phrases will be common:
- de facto
- twenty year **olds** - (should be hyphenated but we flag it either way)

hippietrail:
Compare with #532

hippietrail:
This is pretty much what `phrase_corrections.rs` and `phrase_set_corrections.rs` do.

PR #596, opened 2025-02-05
feat: add some words to curated dictionary

hippietrail:


harper-core/dictionary.dict
```diff
@@ -49799,3 +49799,74 @@ KSQL/1
...
+grifter/MS
+inoperability/1MS
+iterable/5
+ketamine/1M
+leaderboard/1MS
+matcher/1MS
+maximalist/1MS
+merchantability/1
+microarchitecture/1MS
+min
```
elijah-potter, line 49841: I don't think `min` should be added.

harper-core/dictionary.dict
```diff
@@ -49799,3 +49799,74 @@ KSQL/1
...
+MSNBC/2M
+neoliberalism/1M
+Nvidia/2M
+NYU/2M
+oblast/1SM
+olds
```
elijah-potter, line 49847: I'm not certain `olds` is a word.

elijah-potter (changes requested):
I have just a couple of nitpicks.

harper-core/dictionary.dict
hippietrail: The context is in phrases like "twenty year olds". Is there another way we can add words that only occur in phrases? I actually [brought it up in an issue](https://github.com/Automattic/harper/issues/528) a few days ago. At the time I could only think of Latin phrases.

elijah-potter: Excellent point. `olds` should stay.

harper-core/dictionary.dict
hippietrail: min is common in code comments, whether referring to a function or macro, or short for minimum to keep comments concise. I think I already mentioned a possibility of varying the dictionary and/or rules for code comments vs document formats. Let me know...

ccoVeille: Here we are reaching the logic cspell project faced

And why they split their dictionaries

https://github.com/Automattic/harper/issues/597

A developer would like to have `min`, while a non-technical writer won't.

Please take a look at how cspell-dicts is organized, and how they can be enabled, disabled in cspell.

It's a great feature @jason3s maintains

Also, cspell-dicts are almost all (maybe all?) with MIT license,  so they could be reused by project like yours

elijah-potter: This is something we'll handle in a different PR. Remove `min` and I'll merge this.

Issue #595, opened 2025-02-05
feat: `just`/`addnoun` needs to be smarter


hippietrail:
At the moment `just`/`addnoun` always adds the `/S` (plural -s & -es) nand `/M` (possessive -'s)suffixes:

<img width="495" alt="Image" src="https://github.com/user-attachments/assets/32591489-f594-444e-9689-27dfb435b304" />

But the problem is that there are two different kinds of nouns that will be the most common things people will want to add to the dictionary:

1. common nouns that are just regular English words with regular plurals and everything else, but missing from the dictionary
2. Proper Nouns, especially names of companies, products, and technologies. Typically these are more restricted in that it's very uncommon for them to be used in the plural. And when they are occasionally used in the plural it won't hurt to flag them. It is not uncommon for them to be used in the possessive though.

The `justfile` can probably be modified to go by the heuristic that if the noun being added starts with a capital letter it can either prompt the user if it's a proper noun, somebody's name, company name, product name, etc, or can just assume, and add only the `/M` suffix. Otherwise add `/SM` as it currently does.

I already see quite a few of type 2. in our dictionary.

Ones I'm pretty sure wouldn't be correctly used in the plural:
GitLabs, Golangs, Systemds, Joomlas, Automattics, ElasticSearches, TechCrunches, DevOpses, OpenSearches, OpenPGPs, PromQLs, OpenMetricses, NYTimeses, IntelliJs, OpenGLs, Harpers, Typsts, Luas, CMakes, and maybe Gradles - not sure about that one.

I may be wrong about a couple I don't know very well. There's lots of these I left out because I don't know them at all. I also didn't check any of the all-caps entries.

<img width="533" alt="Image" src="https://github.com/user-attachments/assets/78346b7a-7dd2-40fa-aa81-28157b7519ea" />

hippietrail:
Closing since #601 merged.

PR #601, opened 2025-02-05
feat: change `just addnoun` to have different affixes as per #595

hippietrail:
Addresses #595 to add only `/M` for capitalized words = proper nouns but continue adding `/SM` for lowercase words = common nouns.

This means all nouns will be eligible for possessive but only common nouns will be eligible for plural.

For proper nouns that do have plurals, edit the curated dictionary manually.

### DIFF
--- a/justfile
+++ b/justfile
@@ -234,7 +234,11 @@ addnoun noun:
     exit 0
   fi
 
-  echo "{{noun}}/SM" >> $DICT_FILE
+  if [[ "{{noun}}" =~ ^[A-Z] ]]; then
+    echo "{{noun}}/M" >> $DICT_FILE
+  else
+    echo "{{noun}}/SM" >> $DICT_FILE
+  fi
 
 # Search Harper's curated dictionary for a specific word
 searchdictfor word:


elijah-potter:
This feels good to me. `just addnoun` still assumes that a human will look at the change before it gets merged into `master`, so I feel comfortable with a certain amount of YOLO.

justfile
```diff
@@ -234,7 +234,11 @@ addnoun noun:
     exit 0
   fi
 
-  echo "{{noun}}/SM" >> $DICT_FILE
+  if [[ "{{noun}}" =~ ^[A-Z] ]]; then
+    echo "{{noun}}/M" >> $DICT_FILE
+  else
+    echo "{{noun}}/SM" >> $DICT_FILE
+  fi
```
ccoVeille, line 241: You should comment the code in the very same way you explained in the PR and commit

Or at least mention the PR I'd

Please note, the notation /SM or /M obscure for people

Also, I would recommend something like this


```suggestion
  modifier="SM" # a comment explaining what SM is about
  if [[ "{{noun}}" =~ ^[A-Z] ]]; then
    modifier="M" # a comment explaining how it differs
  fi
  echo "{{noun}}/$modifier" >> $DICT_FILE
```

This is pseudo code, please adapt

ccoVeille:
@elijah-potter you merged while I was writting a review on the PR

Not that's a big deal, but if you or @hippietrail could consider my suggestion, I would appreciate

By appreciate, I mean you can also decline it

PR #604, opened 2025-02-05
fix(core): currency placement with certain decimal positions

elijah-potter:
In a discussion with @sumnerevans, I was informed that the currency placement rule throws false positives when the decimal is placed in certain locations. 

I will be refactoring information about numbers into their own type and fixing this bug by encoding decimal position in Harper's lexer.


lukasmwerner:
I know this might not be the best place to mention this so I'm happy to open a new issue. When writing Euros in a European style e.g. `10,50€` aka incorrectly writing 10 Euro 50 cents it only suggests a fix for the `50€`

elijah-potter:
Yeah, @lukasmwerner, go ahead and open a new issue for this. I'm going to merge the current work so it ends up in today's release.

PR #605, opened 2025-02-06
refactor: add comments to `just addnoun` and tweak logic

hippietrail:
Addresses [ccoVeille's concerns](https://github.com/Automattic/harper/pull/601#discussion_r1943387221)


justfile
```diff
@@ -234,12 +234,17 @@ addnoun noun:
...
-    echo "{{noun}}/M" >> $DICT_FILE
-  else
-    echo "{{noun}}/SM" >> $DICT_FILE
+  # 'M', possessive -'s suffix for both common and proper nouns
+  flags='M'
+
+  # 'S', plural -(e)s suffix only for common nouns (not capitalized)
+  if [[ "{{noun}}" =~ ^[a-z] ]]; then
+    flags+='S'
   fi
```
ccoVeille, line 243:
```suggestion
  # detect nouns (not capitalized)
  if [[ "{{noun}}" =~ ^[a-z] ]]; then
   # 'S', plural -(e)s suffix
    flags+='S'
   fi
```

Maybe something like this. You can rephrase of course, but you get the point I think

elijah-potter:
I'm not sure how to resolve these conflicts. Mind taking a look @hippietrail?

hippietrail:
> I'm not sure how to resolve these conflicts. Mind taking a look @hippietrail?

Sorry it's 1.30am here and git/github are confusing me. I don't understand why there's a conflict. Shouldn't my change just replace the old code like usual in PRs or is there a change from ccoVeille there too and not just a comment? My local copy of this feature branch has
```
  # 'M', possessive -'s suffix for both common and proper nouns
  flags='M'

  # 'S', plural -(e)s suffix only for common nouns (not capitalized)
  if [[ "{{noun}}" =~ ^[a-z] ]]; then
    flags+='S'
  fi
```
But now I see a bug because the flags should include `1` for common noun and `2` for proper noun.

If I try to push another change while in it's in this state will it become more of a mess or just fix it?

elijah-potter:
> If I try to push another change while in it's in this state will it become more of a mess or just fix it?

It might fix it. Just go ahead and push the change. I'll figure it out.

hippietrail:
> > If I try to push another change while in it's in this state will it become more of a mess or just fix it?
> 
> It might fix it. Just go ahead and push the change. I'll figure it out from there.

I think I'll just update it to fix the bug I spotted and push that... in the morning.

PR #607, opened 2025-02-06
feat(harper.js): export both binary and inlinedBinary for different runtimes

Asuka109:
This pull request includes several changes to improve the Harper.js project, focusing on updating dependencies, modifying test configurations, and refactoring the code to use the new `harper-wasm` package. The most important changes include updating the `package.json` dependencies, modifying the test scripts, and refactoring the `Linter` and `LocalLinter` classes to use the new `binary` module.

Dependency and configuration updates:

* [`packages/harper.js/package.json`](diffhunk://#diff-7028a5e7b2d8690432e02f627f9c6be357aeeb651c7ae2ddaad624f50a02b2a6L20-R42): Updated dependencies to use `harper-wasm` instead of `wasm`, added new test scripts, and updated the versions of several development dependencies. Added `sideEffects` field so that all the exports can be removed by bundler treeshaking.

Refactoring and code improvements:

* `packages/harper.js/src/binary.ts`: Migrated from `src/WorkerLinter/communication.ts` and `src/loadWasm.ts`, it provides `BinaryModule` as a wrapper for the raw WebAssembly module, bringing more readable method names and additional features. It also exports `binary` and `binaryInlined` to load the WebAssembly file in different ways.
* [`packages/harper.js/src/Linter.ts`](diffhunk://#diff-8814f2f56dc35edb8e81226f9eebadefd9a9e16e894197a31743a107f7c5c7d0L1-R3): Updated imports to use `harper-wasm` and added a new `LinterInit` interface that includes the `binary` module. [[1]](diffhunk://#diff-8814f2f56dc35edb8e81226f9eebadefd9a9e16e894197a31743a107f7c5c7d0L1-R3) [[2]](diffhunk://#diff-8814f2f56dc35edb8e81226f9eebadefd9a9e16e894197a31743a107f7c5c7d0R68-R72)
* [`packages/harper.js/src/LocalLinter.ts`](diffhunk://#diff-a7df90a982d9c8dea0897d3e2259f9d91b90c5d4e466be6879b0f4964f3b17b3L1-R105): Refactored the `LocalLinter` class to use the new `binary` module for WebAssembly operations, removing the old `loadWasm` method. Now it should be initialized with passing a `BinaryModule`.
* [`packages/harper.js/src/WorkerLinter/index.ts`](diffhunk://#diff-a284a2bf6376eddccc252f187ea41f4a5ccbe3ce7f45077f5497bb1627213564L1-R33): Updated the `WorkerLinter` class to use the new `binary` module and refactored the worker communication setup.

Testing improvements:

* [`packages/harper.js/src/Linter.test.ts`](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bR4): Modified tests to use the new `binary` module when initializing `Linter` instances. [[1]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bR4) [[2]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL12-R29) [[3]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL41-R38) [[4]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL58-R55) [[5]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL69-R79) [[6]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL99-R112) [[7]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL126-R123) [[8]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL136-R133) [[9]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL153-R150) [[10]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL164-R161) [[11]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL178-R175) [[12]](diffhunk://#diff-af1de798337059413c1194239375225d618e89638a9654cd2ebd9a0149e7be3bL193-R190)


Asuka109:
@elijah-potter It seems like there is something wrong with the cases running in Firefox. I will handle it later. Could you please take a look at the changes and let me know if anything needs to be changed?

packages/harper.js/src/utils.ts
```diff
@@ -0,0 +1,5 @@
+export function invariant(condition: unknown, message?: string): asserts condition {
```
elijah-potter, line 1: I think `assert` would be a more consistent name for this, since that's the nomenclature that our Rust code uses.

packages/harper.js/src/binary.ts
```diff
@@ -0,0 +1,182 @@
...
+	procName: string;
+	/** The arguments to the procedure */
+	args: any[];
+}
+
+export function isSerializedRequest(v: unknown): v is SerializedRequest {
+	return typeof v === 'object' && v !== null && 'procName' in v && 'args' in v;
+}
+
+/** This class aims to define the communication protocol between the main thread and the worker.
```
elijah-potter, line 56: Explain your reasoning for refactoring this out of the `Worker` module. It doesn't seem necessary for the LocalLinter.

packages/harper.js/src/LocalLinter.ts
```diff
@@ -1,127 +1,107 @@
...
+	private inner: Promise<WasmLinter>;
 
-	/** Initialize the WebAssembly and construct the inner Linter. */
-	private async initialize(): Promise<void> {
-		if (!this.inner) {
-			const wasm = await loadWasm();
-			wasm.setup();
-			this.inner = wasm.Linter.new();
-		}
+	constructor(init: LinterInit) {
```
elijah-potter, line 12: Could we the argument to the constructor optional?

packages/harper.js/src/binary.ts
Asuka109: The original `loadWasm` depends on virtual modules to work, and it globally caches the first loaded WASM module. This can lead to unexpected behavior when using both binary and binaryInlined, and making it harder to write tests (as it requires clearing the cached WASM module). To support two different module loading ways, we need to ensure they can run independently without being affected by the global context. Consequently, WASM-based utils should also operate within their respective contexts.

elijah-potter: That makes a lot of sense. Sally forth!

packages/harper.js/src/LocalLinter.ts
elijah-potter: Yeah let's not make it optional then.

Asuka109: Okay, I will leave the option as it is. We can discuss it in the following PRs.

Asuka109:
@elijah-potter I think everything is ready now.

Asuka109:
@elijah-potter I found that initializing `binary` makes Node.js programs panic, even when only `binaryInlined` is used. Please hold off on reviewing this until I fix it.

Asuka109:
@elijah-potter Sorry for keeping you waiting. I've made some commits, and now it's ready for review!

Asuka109:
When exporting both `binary` and `binaryInlined`, treeshaking can only remove the unused one when using a bundler. However, I forgot that Node.js runtime doesn't execute treeshaking. It caused an error in the binary constructor when `harper_wasm.js` attempted to fetch the wasm module from the binary URL. So in the latest commit, binary has been changed to initialize lazily.

Asuka109:
https://github.com/Automattic/harper/blob/c99168ddf8a8da1bdac7b00a273f648fdaa0d62c/packages/harper.js/src/binary.ts#L9-L27

In the implementation of `loadBinary`, I had to use the `typeof process` expression to determine if it's a Node.js environment to prevent `harper_wasm.js` executing `fetch(module_or_path)`. It's hard to find a better way because I've noticed that many JavaScript libraries using wasm don't support both Node.js and web platforms. As for the `binary` parameter in `new Linter({ binary })`, I've thought of a way to make it optional, which I'll introduce in a later PR.

elijah-potter:
> It's hard to find a better way because I've noticed that many JavaScript libraries using wasm don't support both Node.js and web platforms.

Yeah, it's an unfortunate circumstance to find ourselves in. Do you think

> As for the binary parameter in new Linter({ binary }), I've thought of a way to make it optional, which I'll introduce in a later PR.

Nice! Keep me updated.

I can't take a look at this PR right now. I'll circle back later today.

Asuka109:
@elijah-potter I found the way to load binary in different ways according to the platform. Here's the demo: https://github.com/Asuka109/binary-mutiple-platform

elijah-potter:
> @elijah-potter I found the way to load binary in different ways according to the platform. Here's the demo: https://github.com/Asuka109/binary-mutiple-platform

Good idea! Let's make that part of a different PR though. I've gone ahead and fixed the WordPress plugin to use the API changes. I think this is ready to merge!

elijah-potter (approved):

PR #612, opened 2025-02-06
fix(ls): release config lock to avoid deadlocks

nyonson:
I have yet to convince myself exactly where the deadlock(s) is happening, but I suspect it is somewhere here in the `update_document` function. I am not sure if the Helix IDE is unique in this case, but it throws a lot of document change events at the LSP as you type, doesn't look like a ton of de-bouncing is happening. So very possible for some concurrent update doc requests. I initially thought there were issues with the order of when the config lock and doc lock are grabbed, but the config lock is a Rwlock, leaving me unable to think of an actual deadlock scenario.

Still, this patch shows a great improvement as far as avoiding deadlocks, although I can't be sure it completely fixes it. Not super satisfying.

Two lock updates in the patch, both attempt to lower the time locks are held while other work is being done.

1. Release the config lock after copying the necessary subsections, use those copies for the rest of the work.
2. Don't grab the doc lock until a little later. I think this could probably be further improved, but haven't been able to tease apart any of the logic covered by the lock between `update_document` and `publish_diagnostics`.

Could maybe help with #486

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 
 use anyhow::{anyhow, Context, Result};
 use harper_comments::CommentParser;
-use harper_core::linting::LintGroup;
+use harper_core::linting::{LintGroup, LintGroupConfig};
 use harper_core::parsers::{CollapseIdentifiers, IsolateEnglish, Markdown, Parser, PlainEnglish};
 use harper_core::{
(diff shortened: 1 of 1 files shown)


mcecode:
I just tried it and it seems this fixes #486. I can't reproduce it anymore in my system when using this build of `harper-ls`. Great work!

elijah-potter (approved):
Looks great! Thank you for tracking down the issue :partying_face:.

PR #621, opened 2025-02-09
bug: reverse correcting "supposed to" to "suppose to"

hippietrail:
I'm assuming this was a typo or thinko.

"Supposed to" is the correct phrase. "Suppose to" is the error.

References:
- https://grammarist.com/spelling/supposed-to/
- https://english.stackexchange.com/questions/8129/supposed-to-or-suppose-to
- https://quillbot.com/blog/definitions/supposed-to/
- https://www.reddit.com/r/grammar/comments/n171gz/suppose_to_or_supposed_to/
- https://langeek.co/en/grammar/course/1613/suppose-vs-supposed

### DIFF
--- a/harper-core/src/linting/phrase_corrections.rs
+++ b/harper-core/src/linting/phrase_corrections.rs
@@ -83,7 +83,7 @@ create_linter_for_phrase!(EnMasse, "en masse", 1);
 create_linter_for_phrase!(LetAlone, "let alone", 1);
 create_linter_for_phrase!(LoAndBehold, "lo and behold", 2);
 create_linter_for_phrase!(SneakingSuspicion, "sneaking suspicion", 3);
-create_linter_for_phrase!(SupposeTo, "suppose to", 1);
+create_linter_for_phrase!(SupposeTo, "supposed to", 1);
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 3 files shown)


harper-core/src/linting/phrase_corrections.rs
```diff
@@ -83,7 +83,7 @@ create_linter_for_phrase!(EnMasse, "en masse", 1);
 create_linter_for_phrase!(LetAlone, "let alone", 1);
 create_linter_for_phrase!(LoAndBehold, "lo and behold", 2);
 create_linter_for_phrase!(SneakingSuspicion, "sneaking suspicion", 3);
-create_linter_for_phrase!(SupposeTo, "suppose to", 1);
+create_linter_for_phrase!(SupposeTo, "supposed to", 1);
```
elijah-potter, line 86: Update the name of the generated struct as well.

hippietrail: Ah I wondered about that. I think I asked before if there is a convention of naming them after the mistake version or the fix version. I don't remember getting an answer but it seemed both were used. For instance my `DespiteOf` is named after the mistake version.

But maybe in this case there's at least a local convention in this file to stick to.

Also I did this PR directly in the GitHub website since I was in the middle of something I didn't want to commit in an intermediate state to checkout another branch... So I've got to figure out how to modify such a PR again.

PR #626, opened 2025-02-09
feat: if spellcheck only has one suggestion mention it directly

hippietrail:
When the spellchecker has suggestions you have to click on Quick Fix to see them.
First:  
Screenshot 2025-02-10 at 1 55 52 am
Second:  
Screenshot 2025-02-10 at 1 56 06 am

But when `matcher.rs` suggests a fix it tells you directly, without having to click on Quick Fix
image

So use the same text in spell_check as in matcher when there's only one suggestion.
Screenshot 2025-02-10 at 1 55 29 am

### DIFF
--- a/harper-core/src/linting/spell_check.rs
+++ b/harper-core/src/linting/spell_check.rs
@@ -76,17 +76,27 @@ impl<T: Dictionary> Linter for SpellCheck<T> {
             }
 
             let suggestions = possibilities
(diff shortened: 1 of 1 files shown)


elijah-potter:
Good idea!

hippietrail:
> Good idea!

Consistency FTW! 🚀

PR #627, opened 2025-02-09
fix(obsidian): correctly handle Unicode conversions

### LINKED ISSUE #609: Underlines are off-by-one following an emoji.
**Describe the bug**
In the Obsidian plugin, there's some misalignment between the underline and the problem target.

**To Reproduce**
1. Open a new note, and add an emoji like :smile:
2. Enter some bad grammar like "This is a a bad sentence." after the emoji.
3. All of the underlines after the emojis are off by one for each emoji. 

**Expected behavior**
I'd expect the underlined text to match the quoted text in the popup window.

**Screenshots**
Image

**Platform**
This is from the Obsidian plugin.  I'm about to install this in VScode and will test there next.

**Additional context**
Seems to only happen when there are emojis in the text.  I have not tested with images or call-out boxes or anything else, though.  I cannot replicate this without using emojis.

Also, the "Replace" functionality in the popup window is affected - the replacement occurs on the misaligned underlined text instead of the correct text.  Here's the before/after.

Before clicking "replace", the underline is under "is " in "This is":
Image

After clicking "replace", the text now reads "Thais" as the "is " is replaced by "a":
Image

elijah-potter:
Fixes #609 by correctly converting between character indexes (used by Harper) and Unicode code point indexes (used by Obsidian).

I may decide to move this logic into `harper.js` later.

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -20082,10 +20082,9 @@ diaereses/1
 diaeresis/1M
 diagnose/4DSG
 diagnosis/14M
-diagnostic/51S
+diagnostic/51MS
 diagnostically/
 diagnostician/1SM
-diagnostics/1M
 diagonal/51SMY
 diagram/14SM
 diagrammatic/5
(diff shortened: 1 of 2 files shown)


harper-core/dictionary.dict
```diff
@@ -20082,10 +20082,9 @@ diaereses/1
 diaeresis/1M
 diagnose/4DSG
 diagnosis/14M
-diagnostic/51S
+diagnostic/51MS
```
hippietrail, line 20085: `5` means proper noun, which was wrong before this change, which is otherwise right.

PR #636, opened 2025-02-10
feat: split triggers in `matcher.rs` into categories with explanations

hippietrail:
I know the logic here is up for replacement but I think this categorization will still help in various ways:

1. It will help elucidate different classes and types of replacements.

2. It will help us see which categories will be amenable to various kinds of automation.

3. It will help curators adding more lints in the meantime.

4. It also shows that some lints combine multiple kinds of problems while others don't fit well into any neat category. There is some fuzziness.

(I first sorted them alphabetically then grouped them, so the groups themselves are more in a kind of alphabetical order than any semantic order.)

### DIFF
--- a/harper-core/src/linting/matcher.rs
+++ b/harper-core/src/linting/matcher.rs
@@ -101,90 +101,152 @@ impl Matcher {
     pub fn new() -> Self {
         // This match list needs to be automatically expanded instead of explicitly
         // defined like it is now.
-        let mut triggers = pt! {
-            "spacial","attention" => "special attention",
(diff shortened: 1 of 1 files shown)


elijah-potter:
This PR isn't super helpful. I appreciate the thought, but it feels like busy work.

Your changes also significantly increase the construction cost of the `Matcher` by allocating more than the previous implementation. It's an acceptable cost under most circumstances, but it underlines why reviewing this kind of change is non-trivial.

I'll merge it, but next time include it in another PR so I can review it as a whole, rather than as a part.

PR #653, opened 2025-02-11
feat: Sample forms `jot` and `shuf`

hippietrail:
New version of `just sampleforms` that should work on systems with either `shuf` or `jot`. macOS only has the latter.

Please test.


elijah-potter:
Why don't you merge this into #646?

elijah-potter:
It seems to be working on my machine with `shuf`

hippietrail:
> Why don't you merge this into #646?

I thought there was a chance it would break it and result in a mess. Merged there now so I'll close this one. Test again to be sure (-:

PR #654, opened 2025-02-11
Remove Duplicate Spelling Suggestions

### LINKED ISSUE #624: bug: Sometimes the same suggestion appears twice in Quick Fix etc.
**Describe the bug**
If the algorithm comes up with the same suggestion two different ways, it is not deduped.

**To Reproduce**
1. Type in the wonderful Euroenglish word "semantical".
2. Notice the three suggestions: semantic, semantically, semantic
3. Notice the first and third are the same word.

**Expected behavior**
Deduplication or detection before adding each suggestion to see if it's already a suggestion, ie. just "semantic" and "semantically" suggested.

**Screenshots**

VS Code or Windsurf:  
Image

Obsidian:  
Image

**Platform**
VS Code, Windsurf, Obsidian

**Additional context**
Presumably happens for other words but I've only seen this one.

elijah-potter:
Fixes #624.

The `FstDictionary` searches for both the given capitalization ("Hwllo") and the lowercase form ("hwllo"). We were getting duplicate suggestions when those two searches were merged together and had overlaps.

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -685,7 +685,6 @@ AstroTurf/1M
 Asturias/2M
 Asuncion/M
 Aswan/2M
-At/SM
 Atacama/2M
 Atahualpa/M
 Atalanta/2M
@@ -1908,7 +1907,6 @@ Ceylonese/51
 Cezanne/M
 Cf/M
 Ch'in/2M
-Ch/N
 Chablis/21M
 Chad/21M
 Chadian/51MS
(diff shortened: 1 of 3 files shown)


hippietrail:
> Fixes #624.
> 
> The `FstDictionary` searches for both the given capitalization ("Hwllo") and the lowercase form ("hwllo"). We were getting duplicate suggestions when those two searches were merged together and had overlaps.

Great! I expected it would be something along those lines.

hippietrail:
This is still happening sometimes: 
image

PR #669, opened 2025-02-13
Regexish

### LINKED ISSUE #649: bug: Should probably not flag regex-ish things in comments
**Describe the bug**
Often in comments we mention partial- or pseudo- regexes. [Here's an example from Ghidra](https://github.com/NationalSecurityAgency/ghidra/blob/master/GhidraBuild/Skeleton/src/main/java/skeleton/SkeletonFileSystem.java): `// ([a-z0-9]+ only)` - We currently flag just the **z0** in the middle of it.

```java
@FileSystemInfo(type = "fstypegoeshere", // ([a-z0-9]+ only)
		description = "File system description goes here", factory = SkeletonFileSystem.MyFileSystemFactory.class)
```

Note that it's not a real regex since it doesn't have the `/` and the beginning and end. For starters we could just lex character ranges.

**To Reproduce**
1. Type in something very similar to a regex, like the example `([a-z0-9]+ only)`
2. Notice the oddity of what's flagged and what's left alone.

**Expected behavior**
Nothing flagged.

**Screenshots**

Image

**Platform**
`harper-cli`

**Additional context**
This issue bears some similarities to https://github.com/Automattic/harper/issues/543

hippietrail:
This resolved #649

For now it only deals with character ranges such as [a-zA-Z0-9] which is enough to prevent us from flagging zA as a spelling mistake.

I find lexers much easier to write and debug with very short variable names. Let me know if I should change them.

### DIFF
--- a/harper-core/src/lexing/mod.rs
+++ b/harper-core/src/lexing/mod.rs
@@ -102,8 +102,6 @@ pub fn lex_number(source: &[char]) -> Option<FoundToken> {
 // Often in comments we mention partial- or pseudo- regexes. Here's an example from Ghidra:
 // ([a-z0-9]+ only) - We previously flagged just the z0 in the middle of it.
 pub fn lex_regexish(src: &[char]) -> Option<FoundToken> {
-    eprintln!("** lex_regexish: {}", src.iter().collect::<String>());
-
     let l = src.len();
     let mut i = 0;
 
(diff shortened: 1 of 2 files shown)


elijah-potter:
My main concern with this PR is that a variety of perfectly valid words would get caught by this additional lexer. I'd like to set up property testing for it before merging, which I don't have time to do this week. Hang tight and we'll get this in.

hippietrail:
> My main concern with this PR is that a variety of perfectly valid words would get caught by this additional lexer. I'd like to set up property testing for it before merging, which I don't have time to do this week. Hang tight and we'll get this in.

Well I guess it would match words inside square brackets with no whitespace like `[this]` or `[even-this]` ...

elijah-potter:
> Well I guess it would match words inside square brackets with no whitespace like [this] or [even-this] ...

I suppose that's fine... 

Just be on the lookout where this could result in false-negatives in the wild. In the meantime, I'm going to merge it.

elijah-potter:
It looks like this PR is causing our `make_title_case` tests to fail. An exception may be needed in those tests.

Issue #673, opened 2025-02-14
Issue Parsing Plural Years/Decades


lukasmwerner:
**Describe the bug**
When writing a paper using `1980s` harper misunderstands the ending and treats it as if it is part of the next work or token, not entirely sure. 

**To Reproduce**
```markdown
late 1980s and early 1990s.
```
This produces the following lints in harper-ls:
```
Advice:
   ╭─[test.md:1:1]
   │
 1 │ late 1980s and early 1990s.
   │          ──┬──           ┬
   │            ╰──────────────── Replace with: "sand"
   │                          │
   │                          ╰── Did you mean to spell “s” this way?
───╯
```

Strangely enough I get different titled lints in harper-cli:
```
Advice:
   ╭─[test.md:1:1]
   │
 1 │ late 1980s and early 1990s.
   │          ──┬──           ┬
   │            ╰──────────────── It seems these words would go better together.
   │                          │
   │                          ╰── Did you mean to spell “s” this way?
───╯
```

**Expected behavior**
Ideally nothing would happen I think this should be okay to do.

**Platform**
Neovim `v0.10.4`, harper `0.21.1`

hippietrail:
I was thinking this could be tricky because we want to also deal with errors like `late 1980's` and non-errors like `1980's summer was the hottest on record`.

But upon reflection I think it's enough to have a decade lexer since dropping the apostrophe from a decade in the possessive is vanishingly rare compared to the other way around, which needs its own tricker linter.

I'll tackle this easy one right now.

PR #675, opened 2025-02-14
feat: lex 4-digit plural decades

hippietrail:
Addresses #673
Work in progress since it passes <s>all</s> test but does not fix 673's problem

@elijah-potter I think I'm going to need help with this one. It still fails with "1980s and" but I don't really know how to debug it. I would assume the "number" lexer is matching the "1980" part but "1980s" on its own passes its test... So perhaps "s and" is getting lexed first?

### DIFF
--- a/harper-core/src/lexing/mod.rs
+++ b/harper-core/src/lexing/mod.rs
@@ -25,7 +25,8 @@ pub fn lex_token(source: &[char]) -> Option<FoundToken> {
         lex_tabs,
         lex_spaces,
         lex_newlines,
-        lex_hex_number, // before lex_number, which would match the initial 0
+        lex_hex_number,  // before lex_number, which would match the initial 0
+        lex_long_decade, // before lex_number, which would match the digits up to the -s
         lex_number,
         lex_url,
         lex_email_address,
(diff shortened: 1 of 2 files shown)


harper-core/src/lexing/mod.rs
```diff
@@ -139,6 +140,33 @@ pub fn lex_hex_number(source: &[char]) -> Option<FoundToken> {
     None
 }
 
+pub fn lex_long_decade(source: &[char]) -> Option<FoundToken> {
+    // lex 4-digit decades in their plural such as: 1980s 1990s 2000s 2020s
+    if source.len() != 5 {
```
elijah-potter, line 145: This is the problem. Change `!=` to `<`.

hippietrail: > The character array contains all the text that has not already been lexed, which in the failing case would be `1980s and`, which has more than 5 characters.

I may have even known that when working on a previous lexer. I can't remember. All tests passing now. Thanks!

elijah-potter:
Great work. I'm curious, are there any new lints that could make use of `TokenKind::Decade`

hippietrail:
> Great work. I'm curious, are there any new lints that could make use of `TokenKind::Decade`?

I don't think so yet but there might be.
But Is there another way to get special patterns of characters to avoid being treated as words besides making a new token for them?

PR #685, opened 2025-02-15
feat: forms of "get rid off" -> "rid of"

hippietrail:
All inflected forms: get/gets/getting/got rid off

I can't think of any false positives.

There's a case for handling just "rid off" → "rid of" but there may be a higher chance of false positives there, and I'm not sure if it occurs very much without "get". Let me know.

Random examples from Github for testing:

Please bump axios version to get rid off npm warning # 624

Get rid off "Missing" count in table1 · Issue # 21

31 Mar 2024 — ... getting rid off of all the complexity of the different accesses method of API service providers. From the API service provider, it will get ...

22 Apr 2016 — For now we got rid off circular deps in model tree structure and it's API. The question is whether I should work more to make this graph ...

29 Jan 2016 — papped commented on Jan 30, 2016. Adding at as a runtime dependency gets rid off that error, but now it comes back with the error: java.lang ...


hippietrail:
> A lot of these changes would be better implemented as macro calls in `phrase_corrections.rs`.

I'll turn this one into a draft while I have a go at movin these over...

hippietrail:
> > A lot of these changes would be better implemented as macro calls in `phrase_corrections.rs`.
> 
> I'll turn this one into a draft while I have a go at moving these over...

This whole thing still confuses me. I won't be putting multiple features into a single PR again, even if they seem to be of the same type. Atomic is much more manageable if anything goes slightly wrong.

This is the current status of each part, as best I can tell:
- [**far/way to many/much**](https://github.com/Automattic/harper/pull/718) - stuck in limbo - I think I messed up git when you contributed to my branch on my fork and couldn't figure out how to extricate myself
- **rid off** - looks like it has been merged but without "gotten rid off", which looks like my fault for not converting when I converted the other inflected forms - I'll make a new PR to add just that one.
- **operative system** - looks like it has been merged.
- **yea/yah→yeah** - looks like I never converted it. But it's not a phrase correction anyway so not sure the best way to do it?
- Did I miss anything?

@elijah-potter If you can push/merge in your version of "far/way to many/much" and I submit "gotten rid off" - then let me know if anything else is still outstanding, and then we can get rid of this old broken PR.

elijah-potter:
> @elijah-potter If you can push/merge in your version of "far/way to many/much" and I submit "gotten rid off" - then let me know if anything else is still outstanding, and then we can get rid of this old broken PR.

I'm pretty sure this happened a while ago and I simply didn't see it.

Issue #688, opened 2025-02-15
Missing dependencies in documentation


shanecranor:
I just got the repo set up locally to run and ran into a few undocumented issues.

1. Yarn classic 1.22.1 is required, the latest version of yarn (v4) causes errors
 Ideally we could upgrade to the latest version of yarn to prevent confusion.
Otherwise the version should be specified in the documentation (and possibly pinned in the code using volta https://volta.sh/)
2. On WSL running ubuntu I needed to install the following in addition to the requirements listed in the docs.
```bash
sudo apt-get install libnss3
sudo apt-get install libasound2
```

elijah-potter:
@shanecranor, we've moved to `pnpm` which is more standard. We've also documented the use of the packages you've listed. Are we clear to close this issue?

shanecranor:
Should be good to close it, I can make a new issue if I run into any more problems

PR #689, opened 2025-02-15
Update Yarn and Better Document Dependencies

elijah-potter:
#688 informed me that Harper has been using an old (deprecated) version of Yarn. This PR updates the referenced Yarn version in our `package.json` files and modifies our tooling to work with it. 

I've also added a short blurb about certain integration test dependencies on certain Linux platforms.

@shanecranor, does this look good to you?


shanecranor:
LGTM, maybe add some stuff to the .gitignore https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

mcecode:
> oops, I sent that before just setup finished, I'm getting some errors that I wasn't before
> 
> ```
> error: Cannot find module '/home/shane/tmp_stuff/harper/packages/vscode-plugin/build/tests/runTests.js'
>     at Function._resolveFilename (node:internal/modules/cjs/loader:1405:15)
>     at defaultResolveImpl (node:internal/modules/cjs/loader:1061:19)
>     at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1066:22)
>     at Function._load (node:internal/modules/cjs/loader:1215:37)
>     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
>     at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
>     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:151:5)
>     at node:internal/main/run_main_module:33:47 {
>   code: 'MODULE_NOT_FOUND',
>   requireStack: []
> }
> Node.js v23.8.0
> error: Recipe `test-vscode` failed with exit code 1
> ```

This is because Yarn >2.x no longer automatically runs pre and post scripts outside of set [lifecycle scripts](https://yarnpkg.com/advanced/lifecycle-scripts), so the tests aren't being built before Yarn tries to run them.

We can remedy this by merging the `pretest` command with the `test` command in `packages/vscode-plugin/package.json`:

```diff
	"scripts": {
		"watch:esbuild": "node esbuild.cjs --watch",
		"watch:tsc": "tsc --noEmit --watch",
-		"pretest": "yarn lint && tsc && node esbuild.cjs",
+		"test": "yarn lint && tsc && node esbuild.cjs && node build/tests/runTests.js",
		"vscode:prepublish": "yarn lint && tsc --noEmit && node esbuild.cjs --production",
		"package": "vsce package",
		"lint": "eslint src --ext ts"
	},
```

I'd do it myself but I don't think I have access to this branch to add changes so leaving the fix here.

This should also fix the failing `precommit` workflow.

mcecode:
I just realized because of the failing test that a blocker in doing this Yarn migration is `@vscode/vsce`, which is what's used to package and publish VS Code extensions. [It only officially supports Yarn 1.x](https://github.com/microsoft/vscode-vsce?tab=readme-ov-file#usage). Though, there is a [workaround](https://github.com/microsoft/vscode-vsce/issues/517#issuecomment-1724528163) that might work for us since we also use esbuild to bundle our extension.

elijah-potter:
> I just realized because of the failing test that a blocker in doing this Yarn migration is @vscode/vsce

That's a pretty big deal. Unless someone wants to pick up implementing the workaround, I'd rather just stay on Yarn v1. @shanecranor, is there anything we can do to improve the experience for incoming contributors if we stick with Yarn v1?

shanecranor:
Looks like they support the latest npm, so we could use npm or pnpm instead of yarn

shanecranor:
If you don't want to go that route you can pin yarn v1 with https://volta.sh/ and have new contributors install volta which will install the correct yarn version for them. Volta manages NPM and Node versions as well.

elijah-potter:
This PR has been superseded by #924

Issue #407, opened 2025-01-17
feat: Flag "news" used as a plural or singular noun rather than an uncountable/mass noun


hippietrail:
This is common among non-native speakers whose first language is a Slavic language but probably others as well:

- Here is another news.
- There are many interesting news today.

Can be replaced by:
- news report(s)
- news story/stories
- report(s)
- piece(s) of news
- some news

Depending on context.

"Many" be replaced by "a lot of".

I'll maintain a list of examples I find in the wild here:
- In 2022 i read **a news** about  Nikita Uvarov → In 2022 i read **some news** about  Nikita Uvarov

hippietrail:
The news that this is unplanned are very disappointing.

hippietrail:
That this is considered stale are bad news.

PR #691, opened 2025-02-16
feat: "news" used as plural

hippietrail:
Addresses #407

Because of many lower/upper case variants this and similar might be better as a dedicated lint - or `matcher.rs` could be changed to handle case variants

### DIFF
--- a/harper-core/src/linting/matcher.rs
+++ b/harper-core/src/linting/matcher.rs
@@ -242,6 +242,36 @@ impl Matcher {
             "World","War","Ii" => "World War II"
         });
 
+        // plural 'news'
+        triggers.extend(pt! {
+            "news","are" => "news is",
+            "news","were" => "news was",
+            "many","news" => "a lot of news",
+            "Many","news" => "A lot of news",
+            "many","news","are" => "a lot of news is",
+            "Many","news","are" => "A lot of news is",
+            "many","news","were" => "a lot of news was",
+            "Many","news","were" => "A lot of news was",
+            "a","news" => "some news",
+            "A","news" => "Some news",
+            "a","news" => "a piece of news",
+            "A","news" => "A piece of news",
+            "each","news" => "each piece of news",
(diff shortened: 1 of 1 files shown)


hippietrail:
Turning into a draft so this doesn't get merged before I fix it. I didn't think of the false positive "Back in the days of ..." but too busy to switch branches.

elijah-potter:
I love the additions. Would you mind making them "real" rules in `phrase_corrections`?

For example:

```rust
create_linter_map_phrase!(
    News,
    ExactPhrase::from_phrase("many news are"), // Source pattern is mapped to the 
    "a lot of news is", // target string
    "MESSAGE",
    "DESCRIPTION"
);
```

Notice how we have a source and a destination.

hippietrail:
> I love the additions. Would you mind making them "real" rules in `phrase_corrections`?
> 
> For example:
> 
> ```rust
> create_linter_map_phrase!(
>     News,
>     ExactPhrase::from_phrase("many news are"), // Source pattern is mapped to the 
>     "a lot of news is", // target string
>     "MESSAGE",
>     "DESCRIPTION"
> );
> ```
> 
> If you'd like to go a bit further, you can optionally use the `merge_linters` macro to then merge them together.

On a tangent, why does this code
```rs
create_linter_map_phrase!(LoAndBehold, ExactPhrase::from_phrase("long and behold"), "lo and behold", "Did you mean `lo and behold`?", "Detects the exact phrase `long and behold` and suggests replacing it with the idiomatically correct `lo and behold`");
```
Not trigger the precommit error for having the error in its description? I just had to change the description of the modal-of lint to not have `should of` in the description but here `long and behold` in the description does not cause a problem 🫤

hippietrail:
> I love the additions. Would you mind making them "real" rules in `phrase_corrections`?

I just put in a bunch of time on the plural news one and it turned out to be much more combinatorial and contextual than I expected. I think a dedicated linter will be needed for it. I'll check out the others...

PR #701, opened 2025-02-17
feat: add `just newest-dict-changes`

hippietrail:
It finds the last time in the repo that the curated dictionary was edited and prints out details of all the changes.

You can supply a numeric arg to instead look at previous changes. 1 is the same as no arg.

Feel free to suggest a better name! Is hyphenated or not better?

Ideally a version of this should be in a GitHub Action for visual inspection.

This functionality could also be added to `just addnoun` or an extension of it.

A great improvement would be to call `just getforms` for each added or changed entry to show all the word variants the entry will add to the FST dictionary.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,6 +1,6 @@
 # Format entire project
 format:
-  cargo fmt
+  cargo fmt  
   cd "{{justfile_directory()}}/packages"; yarn prettier -w .
 
 # Build the WebAssembly for a specific target (usually either `web` or `bundler`)
@@ -8,7 +8,7 @@ build-wasm:
   cd "{{justfile_directory()}}/harper-wasm" && wasm-pack build --target web
 
 # Build `harper.js` with all size optimizations available.
-build-harperjs: build-wasm
+build-harperjs: build-wasm 
   #! /bin/bash
   set -eo pipefail
 
@@ -25,7 +25,7 @@ build-harperjs: build-wasm
 test-harperjs: build-harperjs
   #!/bin/bash
   set -eo pipefail
-
+  
   cd "{{justfile_directory()}}/packages/harper.js"
   yarn install -f
   yarn playwright install
@@ -51,7 +51,7 @@ dev-web:
 build-web: build-harperjs
   #! /bin/bash
   set -eo pipefail
-
+  
   cd "{{justfile_directory()}}/packages/web"
   yarn install -f
   yarn run build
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -313,3 +313,109 @@ printaffixes:
     const description = fields['#'] || '';
     description && console.log(affix + ': ' + description);
   });
+
+# Get the most recent changes to the curated dictionary.
```
elijah-potter, line 317: I'd love more information here. What does the `numCommits` argument mean?

justfile
```diff
@@ -313,3 +313,109 @@ printaffixes:
...
+  const [RED, GRN, YLW, BLU, MGN, CYN, WHT] = [1, 2, 3, 4, 5, 6, 7].map(c => `\x1b[${30+c}m`);
+
+  const argv = [...process.argv];
+
+  const showHashes = argv.includes('--show-hashes');
+  if (showHashes) argv.splice(argv.indexOf('--show-hashes'), 1);
+  const showDiff = argv.includes('--show-diff');
+  if (showDiff) argv.splice(argv.indexOf('--show-diff'), 1);
+
+  // uncomment first line to use in justfile, comment out second line to use standalone
```
elijah-potter, line 335: Let's reduce redundancy here.

elijah-potter (changes requested):
I love the work you've put in here. For now, let's keep it in the `justfile`. If the trend continues, however, we're going to have to move many of these more complex scripts into a more maintainable location.

justfile
hippietrail: Without that argument (or 1 it just does a diff before and after the most recent commit to the dictionary file. With 2 it does a diff of the commit that changed it prior to that, etc.

It never does a diff from further back to present, only between neighbouring changes.

It's really simple but I'm having trouble putting it into concise words. If you follow my explanation maybe you can suggest a decent wording and/or variable name?

hippietrail: I tried asking the copilot built in to github here:

> The numCommits argument in the just newest-dict-changes command specifies the number of commits to look back to find changes to the curated dictionary.
> 
> Here is a concise amendment to the introductory comment in the justfile:
> ```
> # Get the most recent changes to the curated dictionary.
> # numCommits: Optional argument to specify the number of commits to look back. Defaults to 1.
> newest-dict-changes *numCommits:
> ```

justfile
hippietrail: I'm not sure what you mean:
1. I can get 330 to 333 done in one line using `map`
2. Get rid of those switches? They're not accessible via `just` but they were helpful while debugging. I've noticed `git diff` in words mode can still make some drastic mistakes.
3. Remove the support for using the script on its own outside `just`?

elijah-potter: Actually, I've changed my mind.

hippietrail: Ah I see you've added it. I guess this is now good to go?

PR #712, opened 2025-02-18
feat: another exception for "back in the days before"

hippietrail:
Also the unit tests from my version of this lint I was working on yesterday.

Let me know if this is the wrong kind of pull request.
Let me know if my tests should be renamed to match yours.


elijah-potter:
Great stuff. Thanks for catching that!

elijah-potter:
> Let me know if my tests should be renamed to match yours.

I usually don't care how unit tests are named, as long as they're relatively recognizable and each example resides in its own test.

Issue #681, opened 2025-02-14
"Did you mean “would've”?" false-positive


jvoisin:
> this would of course significantly impact the performances,     ■■ Did you mean “would've”?

Nope, it's would of course, not would've.

hippietrail:
> > this would of course significantly impact the performances,     ■■ Did you mean “would've”?
> 
> Nope, it's would of course, not would've.

Oh this is one of the lints I wrote - good catch!

hippietrail:
This seems to have been fixed in #715

PR #715, opened 2025-02-19
feat: move 'could of' etc from `matcher.rs` to its own linter

hippietrail:
Addresses the false positives brought to light in #681


harper-core/src/linting/modal_of.rs
```diff
@@ -0,0 +1,222 @@
...
+        );
+    }
+
+    #[test]
+    fn test_missed_part() {
+        assert_lint_count("Ah I must of missed that part.", ModalOf::default(), 1);
+    }
+
+    #[test]
+    fn test_for_of() {
```
elijah-potter, line 147: I would replace "test" with whether the rule should correct or allow the phrase.

For example: `catches_missed_part`. That way, the intent is better conveyed.

harper-core/src/linting/modal_of.rs
```diff
@@ -0,0 +1,222 @@
...
+        assert_lint_count(
+            "im more likely to believe you might of left in the 'user' option",
+            ModalOf::default(),
+            1,
+        );
+    }
+
+    #[test]
+    fn test_missed_part() {
+        assert_lint_count("Ah I must of missed that part.", ModalOf::default(), 1);
```
elijah-potter, line 143: For test cases where the rule _should_ match, try to use `assert_suggestion_result` instead.

harper-core/src/linting/modal_of.rs
```diff
@@ -0,0 +1,222 @@
...
+use super::{Lint, LintKind, PatternLinter, Suggestion};
+
+pub struct ModalOf {
+    pattern: Box<dyn Pattern>,
+}
+
+impl Default for ModalOf {
+    fn default() -> Self {
+        // Note 1. "shan't of" is plausible but very unlikely
+        // Note 2. "had of" has trickier false positives and is less common anyway
```
elijah-potter, line 15: I like your priorities here.

elijah-potter (commented):
Some nitpicky stuff, this looks good overall. Keep up the great work!

harper-core/src/linting/modal_of.rs
hippietrail: I changed them all to reflect the variant they're checking rather than the text in the found phrases.

harper-core/src/linting/modal_of.rs
hippietrail: Also done. Thanks!

Issue #667, opened 2025-02-13
bug: Inconsistent use of underscores in VS Code settings


hippietrail:
**Describe the bug**
Multi-word options for Harper itself don't have underscores between words, but options for individual lints do.

**To Reproduce**
1. In VS Code hit ⌘-comma or control-comma to bring up the settings
2. In the "Search settings" box type "harper"
3. You will now only see Harper settings.
4. Notice some settings have spaces between words: Harper-ls › Code Actions: Force Stable, Harper-ls: Diagnostic Severity
5. But other settings have underscores between words: Harper-ls › Linters: Amazon_names, Harper-ls › Linters: An_a

**Expected behavior**
Consistency look and feel.

**Screenshots**
<img width="215" alt="Image" src="https://github.com/user-attachments/assets/c549e204-9126-43ed-8c8e-9eda64268b45" />

**Platform**
VS Code, Windsurf

elijah-potter:
I thought I already replied here. Whack. 

This is not in our control. VS Code produces the name of the setting from its key in the plugin's `package.json`. VS Code doesn't know how to handle `snake_case`. For backwards compatibility reasons, we cannot change those keys.

I remember reading an issue about this on the Visual Studio Code repo, but I can't seem to find it.

hippietrail:
> This is not in our control. VS Code produces the name of the setting from its key in the plugin's `package.json`. VS Code doesn't know how to handle `snake_case`. For backwards compatibility reasons, we cannot change those keys.
> 
> I remember reading an issue about this on the Visual Studio Code repo, but I can't seem to find it.

That sounds really unfortunate. I might try to chase it down on the VS Code side.

mcecode:
I did notice this before, but it didn't really bother me much since I set my VS Code settings directly in `settings.json` rather than in the GUI. However, I do agree that it would be nice if the setting keys would be consistent.

If we can't change it in `harper-ls` directly for backwards compatibility and VS Code can't handle snake case, then a workaround could be to set all the setting keys in `package.json` to camel case and transform what needs to be transformed into snake case before sending them to `harper-ls`. Though, I'm not really sure how to do that yet.

I wonder if it's a good idea to have a standard on whether new configs added to `harper-ls` should be snake case or camel case, so at least any new configs added would be uniform.

hippietrail:
[...]

PR #726, opened 2025-02-19
Overhaul

elijah-potter:
There are a number of changes in here, all aimed towards refactoring Harper's core engine to be more faster, simpler, and more extensible.

The biggest change (so far) was to rework the `LintGroup` type to be runtime-extensible. This significantly reduces the number of types the Rust compiler must churn through and therefore significantly improves compile times.

Additionally, I've changed the naming conventions of the Harper lint setting keys. If you configure `harper-ls` through `lspconfig`, you may need change your settings to match the new schema (which will be outlined in our [Neovim configuration documentation](https://writewithharper.com/docs/integrations/neovim#Linters)). In VSCode or Obsidian, you may have to reconfigure your setup. See #667 for why we are doing this.

I will also be updating [the Rust version](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html) and removing a ton of boilerplate in `Pattern`s.

The goal is to do as many wide-reaching changes in one PR, in order to pay back tech debt in a way that avoids merge conflicts.


mcecode:
So are we settling on `harper-ls.camelCase.PascalCase` as the pattern for all of our configs now? If so, then the following also need to be updated for consistency's sake:

- `harper-ls.codeActions.forceStable` > `harper-ls.codeActions.ForceStable`
- `harper-ls.markdown.ignore_link_title` > `harper-ls.markdown.IgnoreLinkTitle`

---

As for the `precommit` workflow failing, it seems the order that diagnostics are sent has changed, maybe due to the changes in `harper-core`. So updating the second test from

```ts
	it('gives correct diagnostics', () => {
		compareActualVsExpectedDiagnostics(
			getActualDiagnostics(markdownUri),
			createExpectedDiagnostics(
				{
					message: 'Did you mean to repeat this word?',
					range: createRange(2, 39, 2, 48)
				},
				{
					message: 'Did you mean to spell “errorz” this way?',
					range: createRange(2, 26, 2, 32)
				}
			)
		);
	});
```

to

```ts
	it('gives correct diagnostics', () => {
		compareActualVsExpectedDiagnostics(
			getActualDiagnostics(markdownUri),
			createExpectedDiagnostics(
				{
					message: 'Did you mean to spell “errorz” this way?',
					range: createRange(2, 26, 2, 32)
				},
				{
					message: 'Did you mean to repeat this word?',
					range: createRange(2, 39, 2, 48)
				}
			)
		);
	});
```

should fix the issue.

---

Lastly, `packages/vscode-plugin/src/tests/fixtures/.vscode/settings.json` should also be updated to reflect the new changes. So it should be updated from

```json
{
	"files.associations": { "git-commit": "git-commit" },
	"harper-ls.linters.spell_check": true,
	"harper-ls.linters.repeated_words": true,
}
```

to

```json
{
	"files.associations": { "git-commit": "git-commit" },
	"harper-ls.linters.SpellCheck": true,
	"harper-ls.linters.RepeatedWords": true,
}
```

elijah-potter:
You're absolutely right @mcecode. I've updated the keys to match. Thanks for looking into why those tests were failing :).

Issue #730, opened 2025-02-20
bug(core): codebase treats pronouns as a type of noun


hippietrail:
Nouns and pronouns share some but not all properties.

In grammar and linguistics there is an umbrella term that covers both: **_nominal_**

At present for people who know some amount of grammar and linguistics terminology and want to contribute, the codebase is confusing and ambiguous.

Lint rules etc often refer to nouns where it means either nouns or pronouns.

When something specifically refers only to pronouns, it first checks if a word is noun and then checks for the pronoun metadata property.

Nouns and pronouns can both be the head of nominal phrases, and both have the concepts of number (singular vs plural) but only nouns have a "proper" property. Both nouns and pronouns have possessive but they are very different. I've already found odd problems because of this.

Nouns form possessives by the suffix -'s
Pronouns with the suffix -'s are contractions, and only 3rd person singular pronouns are amenable: he's, she's, it's
Many people confuse those contractions with possessives, and contributors to Harper are not immune.

Possessive nouns come before the possessed noun. But the equivalent of this with pronouns are possessive **_adjectives_**

_Earth's_ atmosphere
_man's_ intelligence
_my_ brother
_our_ idea
_its_ size

Possessive pronouns stand in for another noun, when it is the thing that's possessed. Nouns don't have an equivalent.

My car. That car is _mine_.
Is it _his_ or is it _hers_?

I highly recommend splitting up the current _noun_ functions and properties into _nominal_, _noun_, and _pronoun_.  I'm looking through the code now to see if I can do it on my own or not...

hippietrail:
This was fixed by #731

PR #731, opened 2025-02-20
refactor: nominal and determiner word types

hippietrail:
As per #730 this is the first step, moving the current "noun" types, which actually combine nouns and pronouns, into "nominal" types, that being the umbrella term in grammar/linguistics that encompasses both nouns and pronouns.

Because of the use of templates to derive type and function names from other type names, I had to rename a lot more identifiers than expected.

The next step will be to move pronouns out into their own type, with a "person" (1st/2nd/3rd) property but without a "proper" property.

Then noun can also be moved out, keeping things that apply to both, such as "nominal phrase" inside but things that don't such as "compound noun" outside the "nominal" type and inside the "noun" type.

This will make it easier to differentiate linters that work only on pronouns or only one nouns from those that work on both.


hippietrail:
Hmm runs fine but a test fails:
```
test lints_php_laravel_app_correctly ... FAILED

failures:

---- lints_php_laravel_app_correctly stdout ----
[harper-comments/tests/language_support.rs:55:1] & lints = [
    Lint {
        span: Span {
            start: 908,
            end: 911,
        },
        lint_kind: Spelling,
        suggestions: [
            ReplaceWith(
                [
                    'a',
                    'n',
                ],
            ),
            ReplaceWith(
                [
                    'a',
                    'n',
                    'd',
                ],
            ),
            ReplaceWith(
                [
                    'a',
                    'n',
                    'y',
                ],
            ),
        ],
        message: "Did you mean to spell “env” this way?",
        priority: 63,
    },
]
thread 'lints_php_laravel_app_correctly' panicked at harper-comments/tests/language_support.rs:55:1:
assertion `left == right` failed
  left: 1
 right: 2


failures:
    lints_php_laravel_app_correctly

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.00s
```

hippietrail:
Completes the split between nominals that group nouns and pronouns together vs places only nouns or only pronouns work.

Also renames _article_ to _determiner_ and groups the other words that work a lot like the articles, such as _this_, _that_, _these_, and _those_.

Adds some properties to other word types such as _comparative_ and _superlative_ to _adjective_, _subject_ and _object_ _case_ to _pronoun_, etc. Most need more logic to be coded and some refactoring since the same affix changes different properties on different word types.

Adds notes in comments about word types and properties for future implementation.

elijah-potter:
This is great work! I've gone ahead and made some minor fixes. I'm tempted to get this merged tomorrow, as long as you approve of my changes.

harper-core/src/linting/plural_conjugate.rs
```diff
@@ -23,7 +23,7 @@ impl Default for PluralConjugate {
             .then_whitespace()
             .then_exact_word("are");
 
-        let pat = EitherPattern::new(vec![Box::new(plural_case), Box::new(non_plural_case)]);
+        let pat = plural_case.or(Box::new(non_plural_case));
```
hippietrail, line 26: I wouldn't advise using `case` to refer to `singular` vs `plural` as `case` has a specific meaning in grammar and linguistics: I, we he, she, and they are `subject case`; me, us him, her, and them are `object case`; and you and it are both subject and object case. The correct technical term is `grammatical number` so `plural_number` and `singular_number`.

harper-core/src/linting/plural_conjugate.rs
```diff
@@ -101,14 +101,75 @@ mod tests {
...
-    // TODO currently there is no way to mark number on pronouns without making them also nouns
-    // #[test]
-    // fn pronoun_singular() {
-    //     assert_suggestion_result(
-    //         "If he are testing it.",
-    //         PluralConjugate::default(),
-    //         "If he is testing it.",
-    //     );
-    // }
+    #[test]
```
hippietrail, line 104: By the way, these tests are about what it termed `agreement` in grammar and linguistics. This means two (or more) words have a property in common, such as number, tense, case, person, etc. that must match. When two things don't match we don't necessarily know which of the two is wrong, the pronoun or the verb/copula. So a fully developed linter should suggest not only the verbs that match "he" but the pronouns that match "are". Since "he" is a third person singular subject pronoun we would suggest the third person plural subject pronoun "they". In contrived sentences it can seem very obvious which was intended but in real world sentences from non-native speakers it can go either way. Especially with "this" vs "these".

harper-core/src/word_metadata.rs
```diff
@@ -112,7 +112,7 @@ impl WordMetadata {
         adverb has
     );
 
-    /// Checks if the word is definitely a nominal.
+    /// Checks if the word is definitely a nominalpro.
```
hippietrail: My recommendation is to get familiar with the existing terminology in grammar and linguistics and try to use them except perhaps in cases where they might conflict with computer science terms. Especially if one of the goals is to attract people that know English grammar and linguistics to contribute linters.

elijah-potter: 🫡

Issue #678, opened 2025-02-14
"closed compound noun" false positive for catch all


jvoisin:
`they can't catch all the bugs.     ■ Did you mean the closed compound noun`

I'm pretty sure I do mean "can't catch all".

hippietrail:
> `they can't catch all the bugs. ■ Did you mean the closed compound noun`
> 
> I'm pretty sure I do mean "can't catch all".

What did it actually suggest? You seem to have cut that part off.

hippietrail:
Doesn't happen anymore in 0.37.0 

<img width="337" alt="Image" src="https://github.com/user-attachments/assets/533b8e84-9fd9-44c4-85aa-29016ec4c49b" />

PR #736, opened 2025-02-21
fix(core): restrict `GeneralCompoundNouns` to ignore adverb prefixes

### LINKED ISSUE #721: False positive with correct "any one" → wrong "anyone"
**Describe the bug**
"Any one" is a common phrase in English meaning any member of some set or group. "Anyone" is a pronoun meaning any non-specific person.

**To Reproduce**
1. Paste in the sentence "So if you adjust any one of these adjusters that can have a negative or a positive effect."
2. Notice that "any one" is flagged and the suggestion is "anyone"
3. Notice that the message mentions a noun though neither "any one" nor "anyone" is a noun.
4. Notice that following the suggestion makes the sentence ungrammatical and nonsensical.

**Expected behavior**
No error flagged.

**Screenshots**
Image
Image

**Platform**
VS Code, Windsurf

elijah-potter:
Resolves #721.

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -12252,7 +12252,7 @@ anvil/14MS
 anxiety/1SM
 anxious/5YP
 anxiousness/1M
-any/8~
+any/8~j5
 anybody/8SM
 anyhow/5
 anymore/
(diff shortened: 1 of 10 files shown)


hippietrail:
> Resolves #721.

"Any" is another case of a "determiner" like we were talking about: Any, every, each, some a, the, this that, these, those. The `-` annotation seems to work well promoted from "article" to "determiner". Just changed that in my PR.

elijah-potter:
Love it.

elijah-potter:
Also fixes #678.

PR #737, opened 2025-02-21
fix(core): edge case found in Harper's documentation

elijah-potter:
While working on documentation, I found a bug in the `PossessiveYour` rule. I've added a test case for it and modified the rule to work properly.

image

### DIFF
--- a/harper-core/src/linting/possessive_your.rs
+++ b/harper-core/src/linting/possessive_your.rs
@@ -15,7 +15,7 @@ impl Default for PossessiveYour {
             SequencePattern::aco("you")
                 .then_whitespace()
                 .then(|tok: &Token, _source: &[char]| {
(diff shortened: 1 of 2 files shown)


hippietrail:
> While working on documentation, I found a bug in the `PossessiveYour` rule. I've added a test case for it and modified the rule to work properly.

That "under-the-hood" is wrong too by the way. Phrases used as adjectives are normally hyphenated when they are used "attributively". Here it's used "predicatively" so should be written as three words.

[Compound modifier - Wikipedia](https://en.wikipedia.org/wiki/Compound_modifier#:~:text=Attributive%20compounds%E2%80%94modifiers%20within%20the,compounds%20attested%20as%20dictionary%20headwords.)

(I think but am not certain that there are exceptions that are always hyphenated.)

PR #738, opened 2025-02-21
test(core): added test with capital letters

### LINKED ISSUE #722: feat: Probably shouldn't suggest joining a word with a following capital letter
In a discussion of alphabets I found this sentence:

> C is gone, leaving S and K to do its job.

Harper wants to join "leaving" + "S" to "leavingS"

Image

There should probably be a heuristic for this. It would make sense if both were lower case or if both were in all caps.

----
And a more confusing example: 
Image

elijah-potter:
This may resolve #722. I would love your thoughts, @hippietrail.


elijah-potter:
I've looked further into this after the discussion over on #722. It was in a completely different part of the code than I was looking. I've gone ahead and fixed the issue. Would you mind taking one more look @hippietrail?

elijah-potter:
I'm running a release today, so I'm just going to merge it.

PR #747, opened 2025-02-22
feat: tests for sentences containing a decade

hippietrail:
# Issues 
I only [mentioned it in Discord](https://discord.com/channels/1335035237213671495/1335035237213671498/1342765690813157388) when nobody else was active.

On builds straight from master I was seeing the 's' after decades like '1990s' flagged, even though my decade lexer has been merged.

# Description
I checked the tests and we only had unit tests looking for edge cases. This adds a positive and a negative test against the sentence mentioned. Let me know if there's a better way to code them or if they belong in some other file. The other lexers in `harper-core/src/lexing/mod.rs` also didn't have any in-context tests.

# How Has This Been Tested?
`cargo test` passes all tests but `just test` gives strange new errors I've never seen before, including creating supposed screenshot png's but those are all blank. I think this is something unrelated that's been pushed to master in the same timeframe my decade lexer was pushed??

# Checklist

### DIFF
--- a/harper-core/src/lexing/mod.rs
+++ b/harper-core/src/lexing/mod.rs
@@ -470,4 +470,62 @@ mod tests {
         let source: Vec<_> = "'90s".chars().collect();
         assert!(lex_long_decade(&source).is_none());
     }
+
+    #[test]
+    fn accepts_sentence_with_decade() {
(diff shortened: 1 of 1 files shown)


elijah-potter:
Good stuff. For future reference: [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) has us use the `test:` verb rather than the `feat:` verb when we are only adding tests.

PR #754, opened 2025-02-23
feat(vscode-plugin): config updates

mcecode:
# Issues 

Related to #667 since that's what brought up talks about look and feel.

# Description

- Only send Harper specific settings to `harper-ls`. Prior to this change, VS Code sent up the whole user configuration by default. Aside from being the sane thing to do, this helps when debugging messages between VS Code and `harper-ls` since there's less cruft in the logs.
- Change setting key prefixes from `harper-ls` to `harper`. This change only applies to VS Code, the prefixes are changed to `harper-ls` before being sent to `harper-ls`. It's primarily for aesthetic purposes in the Settings UI.

# Demo

Before, with the awkward Harper-ls:

harper-pr-1

After, with just the title:

harper-pr-2

# How Has This Been Tested?

- Manually in the extension host
- `just test-vscode`

# Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -129,7 +129,7 @@ update-vscode-linters:
 
   linters=$(
     cargo run --bin harper-cli -- config |
-      jq 'with_entries(.key |= "harper-ls.linters." + . |
+      jq 'with_entries(.key |= "harper.linters." + . |
         .value |= {
           "scope": "resource",
           "type": "boolean",
@@ -144,7 +144,7 @@ update-vscode-linters:
   manifest_without_linters=$(
     jq 'walk(
       if type == "object" then
-        with_entries(select(.key | startswith("harper-ls.linters") | not))
+        with_entries(select(.key | startswith("harper.linters") | not))
       end
     )' package.json
   )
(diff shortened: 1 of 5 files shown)


mcecode:
On a side note, while I was making this PR I noticed that we pull the config every time we update a document:

https://github.com/Automattic/harper/blob/75d6fe40f4f52149dd8548e8961f6dd8c58828df/harper-ls/src/backend.rs#L140

I'm just wondering if that's really necessary since nowhere in `update_document` do we actually check and use any pulled updates. Additionally, any updates would already get picked up by `did_change_configuration`. It's not really important performance-wise since these exchanges take single-digit milliseconds, but it just seems like a waste to me if it's not really needed. Is it perhaps because there are editors that don't support `did_change_configuration` out of the box?

elijah-potter:
> Is it perhaps because there are editors that don't support did_change_configuration?

I recall having some weird behavior on Neovim without that line.

elijah-potter (approved):
Good stuff.

Issue #774, opened 2025-02-25
bug: "1s and 0s" flagged with "sand" suggested as fix


hippietrail:
## Description

> Claude Shannon saw the world in **1s and 0s**, Realizing they could represent words, sounds, images, even ideas.

Harper doesn't interpret "1s" and "0s" as nouns. It seems "1" and "0" as numbers despite their suffixes. It sees "s" and "and" as adjacent words.

We could add a lexer for suffixed numbers or just suffixed single digits that returns word type.

We'd have to take into account at least:
1. the `-s` suffix on numbers can have other meanings, such as "seconds".
2. people often use a `-'s` suffix for plurals for things like single letters, digits, acronyms, etc.

Actually can we add numbers to the curated dictionary? Wouldn't make sense for most numbers or even the other single digits. I'll check this... OK it makes no difference. No tests fail but behaviour doesn't change. I think this is due to the lexical analysis having higher priority than the dictionary lookup? The numbers get lexed as numbers rather than words so there is no chance for `1s` and `0s` to be looked up in the dictionary.

## Resources

If the problem is not obvious, some online resources clarifying the relevant grammar point can be useful.

## Screenshots

<img width="656" alt="Image" src="https://github.com/user-attachments/assets/692de0e1-75dc-4fbe-bd5c-de57d7ffde69" />
<img width="312" alt="Image" src="https://github.com/user-attachments/assets/49397586-efce-437e-baa4-fc4dd735e940" />

hippietrail:
I'm working on this one. Assign me if you read this before I submit the PR.

grantlemons:
Is this resolved now that your PR was merged?

hippietrail:
> Is this resolved now that your PR is merged?

This specific one is but similar problems still about wherever numbers can be suffixed with letters. A better solution that encompasses all of them is probably to include them all as words or as `word-like` or perhaps something similar to `word-like` since I haven't delved into exactly all the things that token type does.

hippietrail:
This specific issue solved by #775

PR #775, opened 2025-02-25
feat: lex 0s 1s 0's 1's etc as words

hippietrail:
# Issues 
This addresses #774

# Description
In sentences like "Computers work with 1s and 0s", `1` and `0` were lexed as numbers, cutting off the plural suffixes. This left `s and` to be a detected as a potential error for `sand`.

This change specifically lexes single digits followed by `s` or `'s` as words, leaving the rest up to the linters.

I've also added `0s` and `1s` to the curated dictionary. `0's` and `1's1 are omitted as those should be flagged and suggested to change to the non-apostrophe versions.

Other digits and longer numbers are omitted from the dictionary although they are still lexed as words. They can be added in if we find evidence of them causing similar problems.

One unexpected side-effect is that even with `0s` in the dictionary and `0's` flagged as an error, Harper does not manage to suggest `0s` as a correction for `0's`.

# Demo

## Before:  
Image
Image

## After, spelled correctly, without apostrophes:
Screenshot 2025-02-25 at 3 06 08 pm

## After, spelled incorrectly, with apostrophes:
Screenshot 2025-02-25 at 3 06 20 pm

# How Has This Been Tested?
I added new tests, including for edge cases I identified, such as more letters or digits following directly after the `s`.

# Checklist


elijah-potter:
> One unexpected side-effect is that even with 0s in the dictionary and 0's flagged as an error, Harper does not manage to suggest 0s as a correction for 0's.

This is a flaw in our spell-checker, which is my next target for improvement.

On the whole, this is great work. I'm a little hesitant to add such a special case, but I can see no downside. Onward!

hippietrail:
> > One unexpected side-effect is that even with 0s in the dictionary and 0's flagged as an error, Harper does not manage to suggest 0s as a correction for 0's.
> 
> This is a flaw in our spell-checker, which is my next target for improvement.
> 
> On the whole, this is great work. I'm a little hesitant to add such a special case, but I can see no downside. Onward!

The more dogfooding I do the more cases of numbers followed by various letter suffixes I find. I also noticed the "wordlikes" concept which I wasn't aware of when I made this or the "decades" stuff.

I think these all needed to be treated as words to prevent the suffixes being separated by the lexer and then taking part in various linters with what comes after the next space, which so far always feels wrong.

I think it's fine to have special-case lexers that catch either words or wordlikes that are hard to catch in monolithic lexers but I think I made the wrong choice with the decade lexer being its own new token type instead of either word or wordlike.

I'm not sure about actual words vs wordlikes yet. I'll dig into that soon I think.

PR #837, opened 2025-03-06
Dictionary curation 2025 03 06

hippietrail:
# Issues 
N/A

# Description
- Add words to dictionary
- Adjust affix annotations
- Move placenames into the new canonical case system
- Add "VS Code" to the canonical case system
- Sort parts of the canonical case JSON file that I modified

# Demo
N/A

# How Has This Been Tested?
Tested on place names that previously didn't work or needed special rules.

# Checklist


harper-core/proper_noun_rules.json
```diff
@@ -116,40 +116,43 @@
...
       "Abu Dhabi",
       "Addis Ababa",
+      "Andorra la Vella",
       "Bandar Seri Begawan",
+      "Belize City",
       "Buenos Aires",
+      "Cape Town",
+      "Dar es Salaam",
       "Diego Garcia",
+      "George Town",
```
elijah-potter, line 128: Would it be possible to move more `matcher.rs` rules here?

harper-core/src/linting/matcher.rs
```diff
@@ -139,9 +139,6 @@ impl Matcher {
 
         // countries and capitals with special casing or punctuation
         triggers.extend(pt! {
-           "andorra","la","vella" => "Andorra la Vella",
-           "Andorra","la","vella" => "Andorra la Vella",
-           "Andorra","La","Vella" => "Andorra la Vella",
            "guinea","bissau" => "Guinea-Bissau",
```
elijah-potter, line 142: These hyphenation rules can be made into `phrase_corrections` entries.

harper-core/proper_noun_rules.json
elijah-potter: Oh! I looks like you did!

harper-core/src/linting/matcher.rs
hippietrail: Cool. Will look into that today then.

elijah-potter: They can be made into separate rules. An entry for proper noun capitalization and an entry for hyphenation. In other words: just do the quick fix.

I may start looking into how we can collapse multiple `Lint`s into one soon, since we're starting to hit the critical mass you mentioned a while ago.

hippietrail: > They can be made into separate rules. An entry for proper noun capitalization and an entry for hyphenation. In other words: just do the quick fix.
> 
> I may start looking into how we can collapse multiple `Lint`s into one soon, since we're starting to hit the critical mass you mentioned a while ago.

I'm not sure how to proceed with this? Is it possible to at least merge in the `dictionary.dict` curation part as it's going to get more and more out-of-sync with later curation PRs.

hippietrail:
Oops I accidentally pushed [`3d0c3c3`](https://github.com/Automattic/harper/pull/837/commits/3d0c3c3573660ce5fa038856a77f14a1e4966e26) here instead of my new curation branch. I tried to revert it with `git reset --hard HEAD~1` and `git push -f origin dictionary-curation-2025-03-06` which I thought worked, but apparently it didn't. I don't want to fumble around and make it even worse \\-:

hippietrail:
> I'm a little confused to your meaning. Are you wanting to only submit the `dictionary.dict` changes?

I think one problem is I keep forgetting which file can fix which things because it's not in the name of the file, the name of the linter, or the comment at the top of the file.
By this I mean which file has lints that can add hyphens to phrases vs which has lints that can correct the case of hyphenated phrases.

I'm going to try to narrow this down, include tests, and include comments. But I might miss something. I'll push my best effort and then merge everything that looks right to you.

Maybe we could have a file somewhere just for testing place names and/or proper nouns that can include tests for cases, hyphenation, apostrophes, accents, etc, without having to know which source file is responsible.

harper-core/src/linting/phrase_corrections.rs
```diff
@@ -602,7 +602,34 @@ pub fn lint_group() -> LintGroup {
             ["day and age"],
             "Use `day and age` for referring to the present time.",
             "Corrects the eggcorn `day in age` to `day and age`, which properly means the current era or time period."
-        )
+        ),
+        "GuineaBissau" => (
+            // Note: this lint cannot correct wrong case, only the hyphenation
+            // Note: $input must already be the correct case
```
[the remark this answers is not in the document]
hippietrail: I believe it will _match_ either case and the result will stay in that case without fixing it so it will change `guinea bissau` to `guinea-bissau` instead of the fully correct `Guinea-Bissau` but let me double-check... yes:Screenshot 2025-03-20 at 11 27 31 am
That's why I also had fixes for these in `linting/matcher.rs` which can fix case and hyphenation at the same time but is going away. I'll clarify the comment.

harper-core/src/spell/fst_dictionary.rs
```diff
@@ -244,18 +244,23 @@ mod tests {
     fn fst_map_contains_all_in_full_dict() {
         let dict = FstDictionary::curated();
 
+        let mut is_first = true;
```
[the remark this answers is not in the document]
hippietrail: > What's the reasoning here?

That was my fix for the "blank entry in the dictionary" bug which was the flipside of the "no linefeed at the end of the dictionary" bug. I think it was a fleeting bug which appeared briefly and is now moot anyway since we now always end with a linefeed and specifically allow blank lines. I assumed it ended up last due to how the FST dictionary expanded itself.

Actually while merging the `dictionary.dict` changes I found there is indeed a blank line between "pounding" and "pour"!

I've removed this code.

elijah-potter (approved):

elijah-potter (approved):

PR #843, opened 2025-03-07
Rules page improvements

elijah-potter:
# Issues 

Missing information became apparent in discussion on Discord.

# Description

This PR does two things:

1. The rules page on the website now includes whether each rule in enabled by default.
2. The rules page is now formatted as a table, which makes it significantly easier to read.

# Demo

image


packages/web/src/routes/docs/rules/+page.svelte
```diff
@@ -1,23 +1,46 @@
...
+	<TableHead>
+		<TableHeadCell>Name</TableHeadCell>
+		<TableHeadCell>Enabled by Default</TableHeadCell>
+		<TableHeadCell>Description</TableHeadCell>
+	</TableHead>
+	<TableBody>
+		{#each Object.entries(descriptions) as [name, description]}
+			<TableBodyRow>
+				<TableBodyCell>{name}</TableBodyCell>
+				<TableBodyCell>{default_config[name] ? 'Enabled' : 'Disabled'}</TableBodyCell>
```
mcecode, line 41: I think it might be better to use Yes/No or ✔️/❌ as it might be difficult to differentiate at a glace between Enabled/Disabled in a large table like this.

elijah-potter: Good idea!

elijah-potter:
Updated screenshot: 

image

PR #847, opened 2025-03-07
feat(dict): added words to dictionary

ficcdaf:
# Issues 

# Description

Added some technical terms and other words to the dictionary.

# Demo

# How Has This Been Tested?

`just searchdictfor` to check the expansions.

# Checklist


hippietrail:
> # How Has This Been Tested?
> `just searchdictfor` to check the expansions.

Here's a tip in case you don't know: `just searchfor` greps the expanded dictionary so might return some coinicidental matches and might miss some expansions if you're not careful and the endings change.

You can also use `just getforms` with the exact line from `dictionary.dict`, or the line you're thinking of adding, and will return just the expanded forms of that entry.

Let me know if I should change its comment in the `justfile`, add it to the docs somewhere, or add missing functionality to it.

ficcdaf:
> You can also use just getforms with the exact line from dictionary.dict, or the line you're thinking of adding, and will return just the expanded forms of that entry.
> Let me know if I should change its comment in the justfile, add it to the docs somewhere, or add missing functionality to it.

I had no idea. I thought for example `just getforms cryptographic` should give me all the forms as defined in the dictionary so I was confused as to why it wasn't working. A justfile description and docs update would be stellar.

hippietrail:
> > You can also use just getforms with the exact line from dictionary.dict, or the line you're thinking of adding, and will return just the expanded forms of that entry.
> > Let me know if I should change its comment in the justfile, add it to the docs somewhere, or add missing functionality to it.
> 
> I had no idea. I thought for example `just getforms cryptographic` should give me all the forms as defined in the dictionary so I was confused as to why it wasn't working. A justfile description and docs update would be stellar.

It's funny. As I was writing my reply to you I was thinking it could do that - it's already maybe the longest recipe in the justfile but I think I'll improve it tomorrow. Thanks for the feedback! Is there a place in the docs it would belong?

PR #854, opened 2025-03-08
Avoid linting typst colors

alex-tdrn:
# Issues 
https://github.com/Automattic/harper/issues/849

# Description
Avoid linting typst calls to `color.rgb("...")`/`rgb("...")` since they represent hex codes. Accomplish this by ignoring any call to the `rgb` global function and any member access to the `color` global module. A typst user could *theoretically* shadow the `color` module and/or the `rgb` global function with their own implementation and add string parameters that represent text that should be checked for grammar/typos, but surely no reasonable user would do that. 

If this is seen as a good strategy, I think it could be used to tackle https://github.com/Automattic/harper/issues/846 as well as other such "do not check the string parameter of this specific function since it is not prose" cases.

I have also added 2 extra unrelated commits to make life a bit easier for contributors who:
- work on the typst files and have some kind of language server that automatically renders pdfs in the background and might commit them by accident
- work on windows and might commit CRLF files by accident

# How Has This Been Tested?
I have added a test file and I have built and tested harper with this change on personal typst projects that use the `rgb` function.

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests to cover my changes

PR #857, opened 2025-03-08
fix: Ignore hex codes inside rgb function calls

grantlemons:
- **fix: fall back to `grep` when `rg` is not available**
- **tests(typst): move integration tests to tests dir instead of lib**
- **fix(#854): mark inside of rgb calls as unlintable**


elijah-potter:
I went ahead an cherry-picked @alex-tdrn's general-purpose commits onto this branch, since discussion is already here.

harper-typst/src/typst_translator.rs
```diff
@@ -137,15 +137,23 @@ impl<'a> TypstTranslator<'a> {
...
+                self.parse_expr($inner, offset)
+            };
+            ($($inner:expr),+) => {
+                merge![
+                    $(recurse!($inner)),*
+                ]
+            };
+        }
+
+        macro_rules! get_text {
```
elijah-potter, line 150: Can this be a function rather than a macro?

grantlemons:
But we don't want the .editorconfig or .gitignore changes, right?

elijah-potter:
We do. Editors (until now) have been inferring this information. I'd like to tell them explicitly.

elijah-potter:
I want to run a release now, so I'm just going to go ahead and merge this.

Issue #825, opened 2025-03-04
Feat: ignore comments that begin with “spellchecker: ignore”


cylewaitforit:
I am using Harper in conjunction with cspell. When I do a comment ignore for cspell I then have to ignore the comment for Harper. It would be nice if Harper automatically ignored comments that began with ` spellchecker: ignore `.  Cspell also has other ways to add comment ignores like ` cspell: ignore ` but I think those could pollute this feature which is why I would suggest only ignoring comments with the more general `spellchecker:  ignore`.

hippietrail:
> I am using Harper in conjunction with cspell. When I do a comment ignore for cspell I then have to ignore the comment for Harper. It would be nice if Harper automatically ignored comments that began with `spellchecker: ignore`. Cspell also has other ways to add comment ignores like `cspell: ignore` but I think those could pollute this feature which is why I would suggest only ignoring comments with the more general `spellchecker: ignore`.

Are you saying this is already an existing cspell feature? Oh I found [this reference](https://joplinapp.org/help/dev/spellcheck/): `// cSpell:disable / // cSpell:enable`

This makes it seem like there's a set of related directives. A link to the official docs would be useful.

cylewaitforit:
Here is a link to cspell’s docs on using ignore.
https://cspell.org/configuration/document-settings/#ignore
In the docs they don’t call out using “spellchecker:” instead of “cspell:” for the ignore like they do just above it for the disable,  but it does work. I’ll see if I can work to get that clarified in their docs.

cylewaitforit:
The cspell docs now are updated to be more explicit about the ability to use “spellchecker: ignore” comments for ignoring words.

mcecode:
I think this was already closed in #861, feel free to ping me to reopen if it didn't and there's something I missed.

PR #861, opened 2025-03-09
feat: mask out comments beginning with spellchecker:ignore

grantlemons:
Still needs some tests. I'm not sure this is the cleanest way to accomplish this, but I based it on a comment @elijah-potter once made about something similar (I think shebangs).

### DIFF
--- a/harper-comments/src/comment_parser.rs
+++ b/harper-comments/src/comment_parser.rs
@@ -3,13 +3,13 @@ use std::path::Path;
 use comment_parsers::{Go, JavaDoc, JsDoc, Unit};
 use harper_core::parsers::{self, MarkdownOptions, Parser};
 use harper_core::{MutableDictionary, Token};
-use harper_tree_sitter::TreeSitterMasker;
 use tree_sitter::Node;
 
 use crate::comment_parsers;
+use crate::masker::CommentMasker;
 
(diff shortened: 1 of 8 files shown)


grantlemons:
Closes #825

grantlemons:
> Still needs some tests. I'm not sure this is the cleanest way to accomplish this, but I based it on a comment @elijah-potter once made about something similar (I think shebangs).

I was wrong, it was my own suggestion lol, I like elijah's better (making a new masker and wrapping the treesitter masker) so I'll mark this as draft for the time being.

grantlemons:
I made it  ignore on
```
spellchecker:ignore
spellchecker: ignore
spellcheck:ignore
spellchecker: ignore
cspell:ignore
cspell: ignore
harper:ignore
harper: ignore
```
These are all done with individual `.contains()` on the text of the comment, so I'm sure there's a better way to do this.

elijah-potter:
I like this implementation, but I'd like some more test-driven assurance that works as expected. Could you add a couple of integration tests (in `harper-comments/tests/language_support.rs`) that make sure it works in several contexts and languages?

They can be variations of the same fundamental information, but with different file extensions and with the ignore instruction in different places.

elijah-potter (approved):

PR #862, opened 2025-03-09
Just getforms improvements

hippietrail:
# Issues 
N/A

# Description

Previously `just getforms` (and `harper-cli forms`) would only expand the affix annotations provided on the commandline. Now it will first check if the word on the commandline is already in the dictionary, if it is annotated there, and if those annotations differ from those provided on the commandline. It will then print the old expansions, new expansions, or both.

See [ficcdaf's comments](https://github.com/Automattic/harper/pull/847#issuecomment-2708414662) in #847 

# Demo
Screenshot 2025-03-09 at 4 28 45 pm
Screenshot 2025-03-09 at 4 29 08 pm

# How Has This Been Tested?
I manually tried to test every combination of words and annotations already in the dictionary and not, etc.

# Checklist

I'm open to suggestions on adding unit tests.

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -235,10 +235,7 @@ fn main() -> anyhow::Result<()> {
                         "'{}' is already in the dictionary but not annotated.",
                         arg_bare_word
                     ),
-                    _ => format!(
-                        "'{}' is not in the dictionary yet.",
-                        arg_bare_word
-                    ),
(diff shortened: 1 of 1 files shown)


elijah-potter:
> I'm open to suggestions on adding unit tests.

It's up to you whether unit tests are necessary. I'm not going to require them, since this is a developer tool.

If this works for you, it's good enough for me.

PR #878, opened 2025-03-10
feat: Added Linux musl compilations

kiding:
# Issues 

**Branch created out of #877.**

# Description

This PR adds compilation support for Linux musl. **Depends on #877.**

`cross` supports g++ for musl targets, but the only way to make `actions-rust-cross` successfully compile Harper is to enable `force-use-cross` introduced with v1. After a few hours of headache, I spotted #877 by @elijah-potter which simplified my commit quite well.

# Demo

N/A

# How Has This Been Tested?

See https://github.com/kiding/harper/actions.

# Checklist

### DIFF
--- a/.github/workflows/build_harper_binaries.yml
+++ b/.github/workflows/build_harper_binaries.yml
@@ -48,6 +48,20 @@ jobs:
             bin: harper-ls
             name: harper-ls-aarch64-unknown-linux-gnu.tar.gz
             command: build
+          - release_for: Linux-x86_64-musl
+            os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+            project: harper-ls
+            bin: harper-ls
+            name: harper-ls-x86_64-unknown-linux-musl.tar.gz
+            command: build
+          - release_for: Linux-aarch64-musl
+            os: ubuntu-latest
+            target: aarch64-unknown-linux-musl
+            project: harper-ls
+            bin: harper-ls
+            name: harper-ls-aarch64-unknown-linux-musl.tar.gz
+            command: build
 
           - release_for: Windows-x86_64
             os: windows-latest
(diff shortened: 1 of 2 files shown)


elijah-potter:
I'm quite tempted to just merge this, since it feels "free" in the sense that it doesn't add any additional maintenance burden. Do you believe that to be the case?

kiding:
I think so. `actions-rust-cross` is simply using `cross` in the background, and I don't see any caveats in [supported targets of `cross` supported targets](https://github.com/cross-rs/cross#supported-targets). But I'm not an expert on musl nor cross.

For what it's worth, both binaries pass the tests using the script below. Looks like you can also add tests to the pipeline easily by changing `command: build`s to `command: both`.
```bash
# Running in a fresh Ubuntu 24.10 (amd64) virtual machine

# Install Docker Engine
curl -fsSL https://get.docker.com | sh
sudo chmod 666 /var/run/docker.sock

# Install rustup
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
export PATH="$HOME/.cargo/bin:$PATH"

# Install cross
curl https://github.com/cross-rs/cross/releases/download/v0.2.5/cross-x86_64-unknown-linux-gnu.tar.gz -L -o cross.tar.gz
tar xvfz cross.tar.gz

# Checkout
git clone https://github.com/kiding/harper.git

# Test
cd harper
../cross test --target x86_64-unknown-linux-musl --locked --release
../cross test --target aarch64-unknown-linux-musl --locked --release
```

elijah-potter:
Sorry, I didn't mean to close this. The other question is whether there is direct need for it right now. I see it as "nice to have" but are there use-cases for having a `musl` binary publicly distributed?

kiding:
Yes, it's certainly a good-to-have. I started digging into this because I more or less needed it. 

The company I work for write reports with Typst. Team members collaborate on GitHub repos, then on a new tag GitHub Actions would bake a PDF file which we deliver to a client. 

I wanted to use `harper-cli` in the process and/or pipeline. Some colleagues don't use any of the supported editors, and I didn't want to deviate from what we have with Alpine Linux thanks to its low footprint.

I think it's okay to close this PR or just simmer it for a while since I understand there would not be that many use cases. At least I know how to compile it anywhere, so that's good!

elijah-potter:
That's a good use-case, but be careful with using `harper-cli`. It isn't really meant for public consumption.

kiding:
Noted. Thank you.

PR #886, opened 2025-03-12
fix: two spelling mistakes based on homophones

hippietrail:
# Issues 
N/A

# Description
I spotted two homophone-spellos in the source:
- carrots are root vegetables
- carets are `^`
- unphased means not in phases
- unfazed means remaining calm in the face of adversity

# How Has This Been Tested?
`cargo test` runs with no failures.

# Checklist

### DIFF
--- a/harper-core/src/linting/sentence_capitalization.rs
+++ b/harper-core/src/linting/sentence_capitalization.rs
@@ -136,7 +136,7 @@ mod tests {
     }
 
     #[test]
-    fn unphased_unlintable() {
+    fn unfazed_unlintable() {
         assert_lint_count(
             "the linter should not be affected by `this` unlintable.",
             SentenceCapitalization,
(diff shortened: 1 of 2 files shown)


elijah-potter (approved):
Good eye! Maybe we'll need some rules based on these errors? I'm not sure how you'd implement those, though.

hippietrail:
> Good eye! Maybe we'll need some rules based on these errors? I'm not sure how you'd implement those, though.

I was thinking the same. There are basically two ways.

1. A file with curated sets of homophones
2. [Soundex](https://en.wikipedia.org/wiki/Soundex) or [Metaphone](https://en.wikipedia.org/wiki/Metaphone) algorithm as part of dictionary construction and on all words in document.

One caveat either way is that many words have multiple pronunciations, sometimes because they're homographs, sometimes regionally, so one pronunciation may have a soundalike but the other might not.
Eventually we can tie this to the dialect dictionary feature.

Another is that I think both Soundex and Metaphone are imperfect and will produce false positives. They're normally used for fuzzy searching. So not good for spotting mistakes but good for suggesting fixes - or to produce a factor to be used in calculating and sorting best suggestions. (I haven't looked at either for years now.)

I think starting with a curated JSON file should be easy. I'd have string => array. You can probably semiautomatically populated from Categories on the English Wiktionary.

Issue #498, opened 2025-01-25
feat: Flag Asian commas and "phone commas" with space on the wrong side or missing entirely.


hippietrail:
All these have become common in recent years:

These two seem to be due to how many phones enter commas and/or spaces automagically:
- this,or this
- this ,or this

I haven't seen a space on both sides yet, but I'll keep my eye open for it.
Don't accidentally flag thousands separators! Flagging Eurodecimals should be a separate lint.

These two, the ideographic comma `u+3001` and the full-width comma, are mostly due to people switching keyboard layouts on their computer or phone:
- this、or this
- this，or this

These might also occur with a space on one side or the other but I'm less certain. Because these characters are double width they look like they already contain a space. But I'll keep my eye open for those and other variants to be sure.

Examples found in the wild:
- ... but also for decency,justice,equality,fairness and freedom
- if open the TV ，that's normal
- criteria: prec、rec、F1、NDCG、oneCall、MRR、MAP、AUC
- TopK evaluation: precision、recall、 F1、 NDCG、 1-Call
- parser = argparse.ArgumentParser(description='Command line tools translated using Google、youdao and i18ns.com')
- \* is rotated to {b, c, d, e, f, g， a}. If k = 3 instead of 1, then 
- higher 2 bits of the number of nodes， whether abandoned or not decided by .index section
- map the original vertex id to a continuous id， key: original id, value: continuous id
- \* if the router is "/"，it means this router matches any http request
- GNSS system ID, 1(GPS)，2(GLONASS)，3(GALILEO)，4(QZSS)，5(BEIDOU)
- when cnt >= n + 1， move left

hippietrail:
Implemented in #891

Issue #20, opened 2024-01-30
feat: Include spaces after commas


elijah-potter:
Include a checker that ensures there is exactly one space after every comma. Quotes complicate things, since spaces should come __after__ the quote.

__Test cases:__

```
hello world,my friend
```

```
"Hello,"my friend said.
```

hippietrail:
Related: [feat: Flag Asian commas and "phone commas" with space on the wrong side or missing entirely. #498](https://github.com/Automattic/harper/issues/498)

hippietrail:
#891 solved the first case, but not the second: 

<img width="232" height="60" alt="Image" src="https://github.com/user-attachments/assets/03ea9b6b-1b01-4567-8e21-517dc77a1c3e" />

linkarzu:
Is #1573 related to this?
I pasted some examples there

hippietrail:
> Is [#1573](https://github.com/Automattic/harper/issues/1573) related to this? I pasted some examples there

The second part might be but I suspect it's not.

hulet:
Please add to the test cases a comma followed by a newline, e.g.

`Dear Sir or Madam,\n`

Formal letters, and many emails, will use a comma followed by a line break after the salutation. This should not be flagged as an error.

<img width="279" height="206" alt="Image" src="https://github.com/user-attachments/assets/14db4f29-f319-418a-a6a6-3990037f7baf" />

PR #891, opened 2025-03-12
feat: #20 : comma spacing and #498 : Asian commas

hippietrail:
# Issues 
#20 and #498 

# Description

Correct spaces around commas, only between two words.
Replace East Asian commas with English commas when the context is English.

# How Has This Been Tested?
- I added unit tests for every combination of space before, comma type, space after.
- I added some real-world examples from GitHub.
- I added a unit test to make sure it doesn't flag Asian commas in Asian text.

# Checklist

# Screenshot
image


harper-core/src/linting/lint_kind.rs
```diff
@@ -19,6 +19,7 @@ pub enum LintKind {
     WordChoice,
     #[default]
     Miscellaneous,
+    Punctuation,
```
elijah-potter, line 22: There are switch cases in the JS code that must be updated to reflect this change.

I'd fuzzy search for "Miscellaneous".

In the future I'd like to make sure the precommit tests fail when these other locations aren't updated.

harper-core/src/punctuation.rs
```diff
@@ -97,6 +97,8 @@ impl Punctuation {
             ':' => Punctuation::Colon,
             ';' => Punctuation::Semicolon,
             ',' => Punctuation::Comma,
+            '、' => Punctuation::Comma,
+            '，' => Punctuation::Comma,
```
elijah-potter, line 101: You may want to update `CharStringExt::normalized` to handle these cases too (although it doesn't matter much with commas).

elijah-potter (changes requested):
This looks good. Good test cases, too. Just a couple of consistency-related things.

harper-core/src/linting/lint_kind.rs
hippietrail: > Miscellaneous

Where can I test these colours? It seems `just precommit` and `just test` have once again stopped working on my system. I'll just commit it and see...

elijah-potter (approved):

PR #923, opened 2025-03-17
feat: linter for "each and everyone"

hippietrail:
# Issues 
N/A

# Description
Just the kind of thing phrase_corrections was intended for. Very common on GitHub:
- This community will help **each and everyone** who is facing issue in opensource and share job opportunities, resources, etc. 
- ... **each and everyone** of them is not a good idea and also just not feasible.
- ... for them to ship it same as they already do for vscode (so that not **each and everyone** has to ...
- Composing a state which can be scoped/pulledback to allow **each and everyone** of them to work off of the same piece of state can prove to be challenging.

# References
- [Is it each and every one or everyone?](https://quillbot.com/blog/frequently-asked-questions/is-it-each-and-every-one-or-everyone/)
- [Is "each and everyone" redundant? - English ...](https://english.stackexchange.com/questions/309813/is-each-and-everyone-redundant)
- [Everyone vs Every One | Difference & Examples - QuillBot](https://quillbot.com/blog/commonly-confused-words/everyone-vs-every-one/)
- [Everyone vs. Every One–What's the Difference](https://www.grammarly.com/blog/commonly-confused-words/everyone-vs-every-one/)
- [EACH AND EVERY ONE | English meaning](https://dictionary.cambridge.org/dictionary/english/each-and-every-one)

# How Has This Been Tested?
- I added an atomic unit test for just this phrase on its own.
- I added a real-world unit test for the phrase in context as found on GitHub.
- I ran `cargo test` which resulted in no fails

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_corrections.rs
+++ b/harper-core/src/linting/phrase_corrections.rs
@@ -751,6 +751,12 @@ pub fn lint_group() -> LintGroup {
             "Use `as far back as` for referring to a time in the past.",
             "Corrects nonstandard `as early back as` to `as far back as`."
         ),
+        "EachAndEveryOne" => (
+            ["each and everyone"],
+            ["each and every one"],
+            "Use `each and every one` for referring to a group of people or things.",
+            "Corrects `each and everyone` to `each and every one`."
+        ),
     });
 
     group.set_all_rules_to(Some(true));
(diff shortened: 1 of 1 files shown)


hippietrail:
I tried to add a lint for "everyone of" but doing so clashes with this lint. I've seen this before with another pair of related lints. I think they reveal a bug something like all lints being applied to the source in a loop rather than being applied to the result of the previous application?

elijah-potter:
> I tried to add a lint for "everyone of" but doing so clashes with this lint. I've seen this before with another pair of related lints. I think they reveal a bug something like all lints being applied to the source in a loop rather than being applied to the result of the previous application?

I believe #935 fixes this.

elijah-potter (approved):

PR #927, opened 2025-03-18
fix(vscode-plugin): ignore non-existent ".git" files, support untitled/unsaved files on VS Code

### LINKED ISSUE #382: VSCode: No diagnostics are provided for untitled/unsaved files
This was mentioned in #374. Filing as an issue so we can track it.

No diagnostics are provided when opening a new tab with an untitled/unsaved file. The logs show the following errors:

```text
[2m2025-01-15T04:01:32.370296Z[0m [31mERROR[0m [2mharper_ls::backend[0m[2m:[0m Unable to compute dictionary path for untitled:Untitled-1.
[2m2025-01-15T04:01:32.693595Z[0m [31mERROR[0m [2mharper_ls::backend[0m[2m:[0m Unable to compute dictionary path for untitled:Untitled-1.
[2m2025-01-15T04:01:33.717488Z[0m [31mERROR[0m [2mharper_ls::backend[0m[2m:[0m Unable to compute dictionary path for untitled:Untitled-1.
[2m2025-01-15T04:01:34.660662Z[0m [31mERROR[0m [2mharper_ls::backend[0m[2m:[0m Unable to compute dictionary path for untitled:Untitled-1.
[2m2025-01-15T04:02:10.866125Z[0m [31mERROR[0m [2mharper_ls::backend[0m[2m:[0m Unable to compute dictionary path for untitled:Untitled-1.
```

So it could be related to #251.

Diagnostics are shown after I saved the file, so this may be because `harper-ls` can't handle untitled/unsaved files since they're not saved to disk yet or VSCode doesn't give a correct path.

### LINKED ISSUE #487: VSCode: Showing problems for nonexistent `.git` files
This was mentioned in #441. Filing as an issue so we can track it.

VSCode is showing problems/diagnostics for `.git` files that don't exist. It happens when you open a file that's being tracked by Git, for example `extension.ts`. A corresponding `extension.ts.git` would show up in the "PROBLEMS" tab of the panel. The corresponding `.git` file seems to be treated as a plaintext file and that's why there's a lot of lints for it:

Image

kiding:
# Issues 

Fixes #487. Fixes #382.

# Description

This PR restricts `clientOptions.documentSelector` to the `file` and **`untitled` scheme, and changes `load_file_dictionary` to always succeed on the `untitled` scheme.**  This will prevent the non-existent `.git` files from showing up in Problems since they use `gitfs` scheme[^1], **and make Harper work on untitled/unsaved files. `HarperAddToFileDict` should fail as expected.**

VS Code's documentation[^2] suggests there may be other schemes like ~~`untitled`,~~ `http`, `ftp`, etc. Harper never worked with `untitled` with an error saying `ERROR harper_ls::backend: Unable to generate the file dictionary.` Harper's `file_dict_name` assumes `url` is a file path, so this shouldn't break any preexisting feature.

# Demo

N/A

# How Has This Been Tested?

To test for `untitled`: open a new document, type `Errorz`, check Output, see ~~no error.~~ **one error saying `Did you mean to spell “Errorz” this way?`.**
To test for `.git`: 
```bash
mkdir test && cd test
git init
echo 'Errorz' > test.txt
git add test.txt
code .
```
Check Problems, see no error for `test.txt.git`.

# Checklist

[^1]: https://github.com/microsoft/vscode/pull/86674
[^2]: https://code.visualstudio.com/api/references/document-selector

### DIFF
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -45,12 +45,18 @@ describe('Integration >', () => {
 		);
 	});
 
-	it('does nothing for untitled', async () => {
+	it('gives correct diagnostics for untitled', async () => {
 		const untitledUri = await openUntitled('Errorz');
+
+		// Wait for `harper-ls` to send diagnostics
+		await sleep(500);
 
 		compareActualVsExpectedDiagnostics(
 			getActualDiagnostics(untitledUri),
-			createExpectedDiagnostics()
+			createExpectedDiagnostics({
+				message: 'Did you mean to spell “Errorz” this way?',
+				range: createRange(0, 0, 0, 6)
+			})
 		);
 	});
 
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -9,6 +9,7 @@ import {
 	createRange,
 	getActualDiagnostics,
 	openFile,
+	openUntitled,
 	sleep
 } from './helper';
 
@@ -95,5 +96,14 @@ describe('Integration >', () => {
 		// Restore and reopen deleted file
 		await workspace.fs.writeFile(markdownUri, markdownContent);
 		await openFile('integration.md');
+	});
+
+	it('does nothing for untitled', async () => {
+		const untitledUri = await openUntitled('Errorz');
+
+		compareActualVsExpectedDiagnostics(
+			getActualDiagnostics(untitledUri),
+			createExpectedDiagnostics()
+		);
 	});
 });
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -29,7 +29,7 @@ describe('Integration >', () => {
 		expect(harper.isActive).toBe(true);
 	});
 
-	it('gives correct diagnostics', () => {
+	it('gives correct diagnostics for files', () => {
 		compareActualVsExpectedDiagnostics(
 			getActualDiagnostics(markdownUri),
 			createExpectedDiagnostics(
@@ -42,6 +42,15 @@ describe('Integration >', () => {
 					range: createRange(2, 26, 2, 32)
 				}
 			)
+		);
+	});
+
+	it('does nothing for untitled', async () => {
+		const untitledUri = await openUntitled('Errorz');
+
+		compareActualVsExpectedDiagnostics(
+			getActualDiagnostics(untitledUri),
+			createExpectedDiagnostics()
 		);
 	});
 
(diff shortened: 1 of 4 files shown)


mcecode (changes requested):
Thanks for addressing this issue! LGTM, just a few nitpicks.

packages/vscode-plugin/src/tests/suite/integration.test.ts
```diff
@@ -96,4 +97,13 @@ describe('Integration >', () => {
...
 	});
+
+	it('does nothing for untitled', async () => {
+		const untitledUri = await openUntitled('Errorz');
+
+		compareActualVsExpectedDiagnostics(
+			getActualDiagnostics(untitledUri),
+			createExpectedDiagnostics()
+		);
+	});
```
mcecode, line 108: Can you move this after the "gives correct diagnostics" test, just so they're better grouped together

packages/vscode-plugin/src/tests/suite/integration.test.ts
mcecode, line 1: Can you rename the "gives correct diagnostics" test to "gives correct diagnostics for files"? Just so it's clear that we only support files and not untitled.

mcecode:
Just to note, if this is merged, we may also want to close #382 as not planned or won't fix since this would stop the VS Code extension from sending untitled files to `harper-ls`, which would prevent "Unable to generate the file dictionary." errors from occurring when opening untitled files but also will stop us from being able to support them going forward.

kiding:
Done!

kiding:
@mcecode Decided to add the `untitled` support as well. `load_file_dictionary` returns an empty one when it encounters the `untitled` scheme; same behavior as there is no file dictionary for a given file. `HarperAddToFileDict` and `save_file_dictionary` would fail as expected.

elijah-potter:
@kiding, does that mean unsaved documents (i.e. blank buffers that aren't on-disk) will now work for text editors other than VS Code?

mcecode:
> does that mean unsaved documents (i.e. blank buffers that aren't on-disk) will now work for text editors other than VS Code?

Only if they support the same URI scheme that VS Code uses which is "untitled".

mcecode (approved):
Thanks @kiding!

kiding:
@elijah-potter Checked with the other supported editors, and it looks like it's VS Code-specific behavior.

* Obsidian: Works since it always saves a file.
* Neovim: Doesn't work due to a possible bug on Neovim's end.
  * For an unsaved buffer, the URL received by `harper-ls` is `file:///`.
    * It is actually `file://`, but the path probably defaults to `/`.
  * It seems Neovim fails to find which buffer is `file://`.
    * `:lua print(vim.uri_from_bufnr(1))` returns `file://` which is expected.
    * `:lua print(vim.uri_to_bufnr("file://"))` returns `2` which is wrong, then Neovim opens the directory listing for `/` at buffer 2.
* Helix: Doesn't work since it doesn't spawn `harper-ls` before saving to a file.
* Emacs: Doesn't work since it doesn't call `didOpen` and subsequent requests before saving to a file.
* Zed: Doesn't work since it doesn't start language servers before saving to and opening the file.

elijah-potter (approved):

PR #940, opened 2025-03-20
feat(core): added a bunch more common rules

elijah-potter:
# Description

I've gone ahead and added a bunch more phrase corrections.

# Checklist


harper-core/src/linting/for_noun.rs
```diff
@@ -0,0 +1,77 @@
...
+                "for",
+                problem_chars,
+            )],
+            message: "`For` is more common in this context.".to_owned(),
+            priority: 31,
+        })
+    }
+
+    fn description(&self) -> &'static str {
+        "Corrects the archaic or mistaken `fro` to `for` when followed by a noun."
```
ccoVeille, line 46:
```suggestion
        "Corrects the archaic or mistaken `fro` to `for`, or `from` when followed by a noun."
```

It could be a typo about `fro` used instead of `from`

> I received a letter fro Sarah
> I received a letter from Sarah

ccoVeille (changes requested):

harper-core/src/linting/for_noun.rs
hippietrail: Also, while `fro` is archaic, it's irrelevant since it's not an archaic form of `for`. It's an archaic form of `from` and is only used these days in `to and fro`.

I wonder if some logic for rare/archaic words could have a flag in the dictionary to allow them to get a warning and still undergo spellcheck. We'd want to find other instances similar to `fro` for that to be worthwhile. I know that back in the day spellcheck dictionaries purposefully left words out that were correct but rare and likely to cause confusion.

elijah-potter: Would you suggest we change the description, or remove this rule as well as the dictionary entry for `fro`?

hippietrail: > Would you suggest we change the description, or remove this rule as well as the dictionary entry for `fro`?

I would probably test if the spellcheck handles it. I've noticed it often doesn't make what I consider the 'obvious' suggestions for many mistakes. I haven't checked this one.

If the spellcheck suggests 'for' and 'from' we can probably do without the rule, but if it fails to suggest those then it's probably worth keeping the rule with a changed description.

At some point we probably want to add some code to 'score' suggestions in other ways than just pure edit distance to help get the best 3 into the suggestions.

elijah-potter: That's been around for a while now. See `harper-core/src/spell/mod.rs` `fn score_suggestion`. I've tweaked it a bit, but I'm not sure how to achieve the desired behavior.

elijah-potter:
I have no idea what is going here with all the merge commits.

hippietrail:
> I have no idea what is going here with all the merge commits.

I think about a month ago I tried to submit a PR to your PR... or something like that. And it seemed to work at the time... but it seems to have caused this awfulness in the longer run.

The worst case could be if my master somehow got in there because I turned off some of the GitHub CI stuff that didn't work in my fork and just kept causing recurring errors in my inbox.

If it's my fault I sincerely apologize!

PR #958, opened 2025-03-22
fix: 2 words in curated dict lack / before annotations

hippietrail:
# Issues 
N/A

# Description
While updating my [syntax highlighter VS Code extension for `dictionary.dict`](https://github.com/hippietrail/harper-dict-syntax) I noticed some lines in red, indicating a syntax error.
On closer inspection I found that had the new `<` annotation flag indicating "American dialect" but did not have a `/` separating the word from the annotations.

As far as I can tell only two entries were affected: `colored's` and `coloring's`

# Demo
Before:
Screenshot 2025-03-23 at 12 20 44 am
After:
Screenshot 2025-03-23 at 12 22 01 am

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -17597,14 +17597,14 @@ coloration/~1EM<
 coloratura/15MS<
 colorblind/5P<
 colorblindness/1M<
-colored's<
+colored's/<
 colored/~514U<
 coloreds/1<
 colorfast/5P<
 colorfastness/1M<
 colorful/~5PY<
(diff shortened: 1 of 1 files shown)


elijah-potter:
Good eye! A great use of a syntax highlighter. That must have happened during an erroneous call to a vim macro :shrug:.

elijah-potter (approved):

elijah-potter:
I know there's a plugin for Neovim that lets you query a buffer using tree-sitter.

PR #966, opened 2025-03-24
fix(harper-ls): handle language mode change and VS Code auto detect

kiding:
# Issues 

N/A

# Description

When someone changes the language mode of an opened document, Harper does not reflect the change, resulting in erroneous diagnostics. This has become apparent after #927 was introduced since `untitled`/unsaved documents default to auto detect or plain text, unless "Files: Default Language" (`files.defaultLanguage`) is configured.

The LSP specification states:[^1]
> If the language id of a document changes, the client needs to send a `textDocument/didClose` to the server followed by a `textDocument/didOpen` with the new language id if the server handles the new language id as well.

Harper so far has not implemented `did_close`. This PR implements `did_close` by removing the URL from `doc_state`. Sending an empty diagnostic is added since the `updates diagnostics when files are deleted` test would fail without it.

Might conflict with #960.

# Demo

For a `file`:
<video src="https://github.com/user-attachments/assets/86fdb4d4-2b19-470b-9657-55d54d558591">

For an `untitled` and auto detect:
<video src="https://github.com/user-attachments/assets/5230e0cc-5617-4920-84b4-9362276726cc">

# How Has This Been Tested?

Tested with VS Code manually, and using the newly added test case.
Indirectly tested with Helix, Zed, Emacs, Neovim; they don't call an extra `did_open` when the view is split. 

# Checklist

[^1]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didOpen

### DIFF
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -4,6 +4,7 @@ import { ConfigurationTarget, type Uri, commands, workspace } from 'vscode';
 
 import {
 	activateHarper,
+	closeAll,
 	compareActualVsExpectedDiagnostics,
 	createExpectedDiagnostics,
 	createRange,
@@ -18,6 +19,7 @@ describe('Integration >', () => {
 	let markdownUri: Uri;
 
 	beforeAll(async () => {
+		await closeAll();
 		harper = await activateHarper();
 		// Open test file so diagnostics can occur
 		markdownUri = await openFile('integration.md');
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -11,6 +11,7 @@ import {
 	getActualDiagnostics,
 	openFile,
 	openUntitled,
+	setTextDocumentLanguage,
 	sleep,
 } from './helper';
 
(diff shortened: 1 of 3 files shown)


packages/vscode-plugin/src/tests/suite/integration.test.ts
```diff
@@ -18,6 +20,7 @@ describe('Integration >', () => {
 	let markdownUri: Uri;
 
 	beforeAll(async () => {
+		await closeAll();
```
mcecode, line 23: Just curious, what's being closed here? Since this is the first test to run, shouldn't there be no tabs open at all?

mcecode:
If ever, this can be merged first, I'll just fix any conflict over at #960, if there are any.

packages/vscode-plugin/src/tests/suite/integration.test.ts
kiding: It may be a macOS thing where window states are saved.[^1] When I tested a bunch of times, the tabs from the previous test stayed, interfering with the new test.

[^1]: https://www.mothersruin.com/software/Archaeology/reverse/appstate.html

elijah-potter (approved):

packages/vscode-plugin/src/tests/suite/integration.test.ts
hippietrail: > It may be a macOS thing where window states are saved.[1](#user-content-fn-1-4e23ce261a044025ff34942684b8e0ec) When I tested a bunch of times, the tabs from the previous test stayed, interfering with the new test.

This was my guess too by the way, and I don't think it's Mac-only thought I can only test Mac at the moment.

PR #967, opened 2025-03-24
feat(core): flag "<adjective> of a"

hippietrail:
# Issues 
N/A

# Description
Flags phrases such as "How slow of a laptop are we talking here?" and suggests removing the "of".

It only flags adjectives in the positive degree (neither comparative or superlative).
It avoid a few hard-coded false positives where the adjective is much more common as a noun in this context.

This is so common, especially in American English, that many people might not know it's an error.
Some references:
- [WordReference](https://forum.wordreference.com/threads/not-that-adjective-of-a-noun.3260438/)
- [English StackExchange](https://english.stackexchange.com/questions/64605)
- [English StackExchange](https://english.stackexchange.com/questions/30011)
- [WordReference](https://forum.wordreference.com/threads/not-that-adjective-of-a-noun.2854358/)
- [r/grammar](https://www.reddit.com/r/grammar/comments/982svr/of_a/)
- [Language Log](http://itre.cis.upenn.edu/~myl/languagelog/archives/002283.html)
- [Language Log](https://languagelog.ldc.upenn.edu/nll/?p=25282)
- [Grammar Underground](https://www.grammarunderground.com/too-big-of-a-deal-or-too-big-a-deal.html)
- [Grammarphobia](https://www.grammarphobia.com/blog/2014/01/not-that-big-of-a-deal.html)
- [Pain in the English](https://painintheenglish.com/case/522)
- [Quora](https://www.quora.com/Which-of-the-following-is-correct-Its-too-big-a-problem-or-Its-too-big-of-a-problem)
- [Quora](https://www.quora.com/Are-both-of-these-phrases-grammatically-correct-it-s-not-that-big-of-a-deal-and-it-s-not-that-big-a-deal)
- [English StackExchange](https://english.stackexchange.com/questions/240248)
- [WordReference](https://forum.wordreference.com/threads/too-adjective-a-noun-too-expensive-a-desk.2634551/)

# Demo
image

# How Has This Been Tested?
- I added tests based on sentences found on GitHub.
- I added tests that ensure the false positives are not triggered.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -43980,7 +43980,7 @@ starveling/15MS
 stash/14MDSG
 stasis/~1
 stat/~514MS
-state/~145DRSMYGNLX
+state/~14DRSMYGNLX
 statecraft/~1M
 stated/~45U
 stateful/5Y
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -580,6 +580,7 @@ impl TokenStringExt for Document {
     create_fns_on_doc!(currency);
     create_fns_on_doc!(likely_homograph);
     create_fns_on_doc!(comma);
+    create_fns_on_doc!(adjective);
 
     fn first_sentence_word(&self) -> Option<&Token> {
         self.tokens.first_sentence_word()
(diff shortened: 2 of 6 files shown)


elijah-potter (approved):
This looks good! I don't have any comments. This is readable. I imagine a performance issue could crop up, but it would be easy to fix, and would not be worth investigating right now.

PR #975, opened 2025-03-26
test(core): confirm that #720 is no longer present

### LINKED ISSUE #720: False positives with then→than
**Describe the bug**
In some contexts a correct "then" is flagged and incorrect "than" is suggested.

**To Reproduce**
1. Paste in "And if just one of those is set incorrectly or it has the tiniest bit of dirt inside then that will wreak havoc with the engine's running ability."
2. Notice that this "then" connects a consequence of an earlier "if" with a result.
3. Hover over the flagged "then" and see it wants to replace it with "than", which won't make sense in this context.
4. Another example is "So let’s check it out then."

**Expected behavior**
No error flagged.

**Screenshots**
Image
Image

Image
Image

**Platform**
VS Code, Windsurf

elijah-potter:
# Issues 
Confirms that we previously fixed #720
# Description

Added unit tests from #720 to confirm that it has been truly resolved.

# Checklist

### DIFF
--- a/harper-core/src/linting/then_than.rs
+++ b/harper-core/src/linting/then_than.rs
@@ -206,4 +206,86 @@ mod tests {
             0,
         );
     }
+
+    #[test]
+    fn issue_720_school_but_then_his() {
+        assert_lint_count(
+            "She loved the atmosphere of the school but then his argument is that it lacks proper resources for students.",
+            ThenThan::default(),
+            0,
+        );
+        assert_lint_count(
+            "The teacher praised the efforts of the school but then his argument is that the curriculum needs to be updated.",
+            ThenThan::default(),
+            0,
+        );
+        assert_lint_count(
+            "They were excited about the new program at school but then his argument is that it won't be effective without proper training.",
+            ThenThan::default(),
+            0,
+        );
+        assert_lint_count(
+            "The community supported the school but then his argument is that funding is still a major issue.",
+            ThenThan::default(),
+            0,
+        );
+    }
+
+    #[test]
+    fn issue_720_so_then_these_resistors() {
+        assert_lint_count(
+            "So then these resistors are connected up in parallel to reduce the overall resistance.",
+            ThenThan::default(),
+            0,
+        );
+        assert_lint_count(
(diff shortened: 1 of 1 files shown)


hippietrail:
I still get #720 for the cases I posted in a comment on Feb 22: Yes so then sorry you didn't receive the memo about the meeting changes.
Yes so then sorry you had to wait so long for a response from our team.
Yes so then sorry you felt left out during the discussion; we value your input.
Yes so then sorry you missed the deadline; we can discuss an extension.
Yes so then sorry you encountered issues with the software; let me help you troubleshoot.
Screenshot 2025-03-27 at 3 09 31 pm

elijah-potter:
Hmmm... That's strange. The unit tests are coming up fine for your examples. They are also clean when I open a buffer.

hippietrail:
> Hmmm... That's strange. The unit tests are coming up fine for your examples. They are also clean when I open a buffer.

I had some doubts it was being checked by an out-of-date `harper-ls` process. But that doesn't seem to be the case. So I tried `just lint` but apparently that hasn't been updated to pass `--dialect` to `harper-cli lint`. So I tried `harper-cli lint` and it passed with no lints in different dialects. To make sure it was working I edited my test file with a random spello - now it flags all nine lines - and continues to even after removing the spello! same whether I run from `target/release` or `target/debug`

Hmm but does this maybe running these is not a fresh build and the fresh build `just` makes is hidden elsewhere?

But maybe I'm still doing something wrong trying to verify the test branch?

`cargo install --git https://github.com/automattic/harper --branch test-720 harper-cli --locked`

elijah-potter:
I ran that same command on my system, then ran `harper-cli lint` with your example text. No lints found.

hippietrail:
> I ran that same command on my system, then ran `harper-cli lint` with your example text. No lints found.
> 
> Maybe re-run `git clean -xf .`?

Argh! It was coming from a different rule - in `matcher-rs`
```rs
        // mixing up than/then in context
        triggers.extend(pt! {
            "then","her" => "than her",
            "then","hers" => "than hers",
            "then","him" => "than him",
            "then","his" => "than his",
            "then","last","week" => "than last week"
        });
```

PR #976, opened 2025-03-26
feat(harper-typst): ignore file path as arguments, regex, .display

kiding:
# Issues 

N/A

# Description

Some Typst functions accept file path, regular expression, display format string as an argument. Running spellcheck on them, `../icon.png` `(?i)` `[year repr:last_two]` for example, produces false positives.

This PR adds `parse_func_call` to handle such cases in a generalized way. It also handles the preexisting logic to ignore `#rgb("#000000")`. If Typst decides to add more arguments that are good to ignore, it should be fairly easy to add it to the entry.

One thing to note is that ignoring all `.display` function calls may affect other Typst Universe packages. I wasn't able to find any package that offers `.display` function. It makes sense since the name is quite generic, and the convention is that `.display` is a format function just like [`datetime`](https://typst.app/docs/reference/foundations/datetime/#format), whose argument is good to ignore anyways.

# Demo

image

# How Has This Been Tested?

Manually with VS Code, and with the newly added test case.

# Checklist


harper-typst/src/typst_translator.rs
```diff
@@ -203,6 +203,41 @@ impl<'a> TypstTranslator<'a> {
             )
         };
 
+        // Parse a function call
+        let parse_func_call = |func: FuncCall| {
+            let parse_args_without = |pos: bool, nameds: &[&str]| {
```
grantlemons, line 208: Can you clarify what `pos` represents?

It looks like every time you call this you pass true in for pos.

grantlemons (approved):

elijah-potter (approved):

Issue #982, opened 2025-03-27
feat: dialect indicator in editor plugins/extensions


hippietrail:
## Description

Add a little indicator, preferably next to the document format editor, that indicates Harper's selected dialect.

## Component

Which component of Harper would this feature affect? For example:
- Obsidian plugin
- VS Code extension
- Neovim plugin

## Use Cases

- Handy for users who aren't aware the dialect can be set.
- Handy for Harper contributors that want to work on dialect-specific problems.
- Handy for Harper contributors who want to test in multiple dialects in multiple editor windows without getting confused.

## Implementation Ideas

I'll look into the VS Code support myself.

## Screenshots/Examples
Add where either of the ❌ is marked in the status bar:
<img width="225" alt="Image" src="https://github.com/user-attachments/assets/47416d3c-96f1-4836-841d-dc7fe0c6f97d" />

elijah-potter:
I believe we can close this as it has been implemented in both Obsidian and VS Code. As for Neovim, it isn't possible to render such elements, since that is outside of the scope of language servers.

PR #985, opened 2025-03-28
Dialect indicator for VS Code extension

hippietrail:
# Issues 
#982 

# Description
Adds an indicator in the status bar for the English dialect set.

It reflects the **Workspace** setting, which seems to be the default. I don't know VS Code or its extension system deeply enough to know if this is the best solution.

On my setup I was able to confuse myself by changing the dialect in **User**.

Changing the Workspace setting instantly changes the flag, but does not reload the window. Some squiggles seem to be updated instantly, but not all of them.

It is possible to make the indicator clicking to change the setting but I haven't implemented this. It would be best to fully grok the Workspace vs User ramifications before attempting this.

It is possibly to add a Harper logo but not as a PNG or SVG. You need to do some font trickery: https://stackoverflow.com/a/76399730/527702

I tried to make the indicator appear as close to the document format indicator as possible. It doesn't seem to be possible to get them right next to each other.

# Demo
View from the settings:
image

View of position near file format indicator: 
image

# How Has This Been Tested?

Manually. I don't know if it would be possibly to add test code for this. Let me know.

# Checklist

### DIFF
--- a/packages/vscode-plugin/src/extension.ts
+++ b/packages/vscode-plugin/src/extension.ts
@@ -2,6 +2,7 @@ import type { ExtensionContext } from 'vscode';
 import type { Executable, LanguageClientOptions } from 'vscode-languageclient/node';
 
 import { Uri, commands, window, workspace } from 'vscode';
+// import { StatusBarAlignment, type StatusBarItem } from 'vscode';
 import { LanguageClient, ResponseError, TransportKind } from 'vscode-languageclient/node';
 
 // There's no publicly available extension manifest type except for the internal one from VS Code's
@@ -51,6 +52,8 @@ const clientOptions: LanguageClientOptions = {
 	},
 };
 
+// let dialectStatusBarItem: StatusBarItem | undefined;
+
 export async function activate(context: ExtensionContext): Promise<void> {
 	serverOptions.command = getExecutablePath(context);
 
@@ -99,6 +102,37 @@ export async function activate(context: ExtensionContext): Promise<void> {
 	);
 
 	await startLanguageServer();
+
+	// Add status bar item for dialect
+	// dialectStatusBarItem = window.createStatusBarItem(StatusBarAlignment.Left, 1000);
+	// dialectStatusBarItem.command = 'harper.setDialect';
+	// context.subscriptions.push(dialectStatusBarItem);
+
+	// // Add command to change dialect
+	// context.subscriptions.push(
+	// 	commands.registerCommand('harper.setDialect', async () => {
+	// 		const selected = await window.showQuickPick(['American', 'British'], {
+	// 			placeHolder: 'Select dialect',
+	// 		});
+	// 		if (selected) {
+	// 			await workspace.getConfiguration('harper').update('dialect', selected, true);
+	// 			updateDialectStatusBar();
+	// 		}
+	// 	}),
+	// );
+
+	// // Update status bar when dialect changes
+	// context.subscriptions.push(
+	// 	workspace.onDidChangeConfiguration(async (event) => {
+	// 		if (event.affectsConfiguration('harper.dialect')) {
+	// 			updateDialectStatusBar();
+	// 		}
+	// 		// ... existing code ...
+	// 	}),
+	// );
+
+	// // Initialize status bar
+	// updateDialectStatusBar();
 }
 
 function getExecutablePath(context: ExtensionContext): string {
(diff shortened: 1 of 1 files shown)


mcecode:
> It reflects the Workspace setting, which seems to be the default. I don't know VS Code or its extension system deeply enough to know if this is the best solution.

It's probably the right move, AFAIK, using `workspace.getConfiguration` like you did returns the setting for the current open workspace/folder. If that's not set, it returns the user level setting. If that's not set, it returns the default setting.

> Changing the Workspace setting instantly changes the flag, but does not reload the window. Some squiggles seem to be updated instantly, but not all of them.

The window doesn't need to reload on config change. Updating the diagnostics on config change is already handled here:

https://github.com/Automattic/harper/blob/77376c75538e13860c05cfeeb75535cf50d80d87/packages/vscode-plugin/src/extension.ts#L89

> I tried to make the indicator appear as close to the document format indicator as possible. It doesn't seem to be possible to get them right next to each other.

You can probably achieve this by setting `priority` to `1` or `-1`  like [Prettier](https://github.com/prettier/prettier-vscode/blob/main/src/StatusBar.ts) or [Biome](https://github.com/biomejs/biome-vscode/blob/main/src/status-bar.ts) do.

I have some nits, but for now, LGTM.

hippietrail:
> > It reflects the Workspace setting, which seems to be the default. I don't know VS Code or its extension system deeply enough to know if this is the best solution.
> 
> It's probably the right move, AFAIK, using `workspace.getConfiguration` like you did returns the setting for the current open workspace/folder. If that's not set, it returns the user level setting. If that's not set, it returns the default setting.

I know I got confused when coding it and then when testing it, but I'm not comfortable with the various levels of settings in VS Code. If it feels right to VS Code pros that's great!

> > Changing the Workspace setting instantly changes the flag, but does not reload the window. Some squiggles seem to be updated instantly, but not all of them.
> 
> The window doesn't need to reload on config change. Updating the diagnostics on config change is already handled here:
> https://github.com/Automattic/harper/blob/77376c75538e13860c05cfeeb75535cf50d80d87/packages/vscode-plugin/src/extension.ts#L89

Great! I got the feeling that some squigglies updated if they were hidden by another window and that others were not updated. But that may be a false impression. `.affectsConfiguration()` looks like the API I was hunting for and failed to find to force a window refresh/redraw/invalidation. Much of this documentation is written in too abstract a style IMO. Googling and asking LLMs didn't get me there either using more concrete terminology.

> > I tried to make the indicator appear as close to the document format indicator as possible. It doesn't seem to be possible to get them right next to each other.
> 
> You can probably achieve this by setting `priority` to `1` or `-1` like [Prettier](https://github.com/prettier/prettier-vscode/blob/main/src/StatusBar.ts) or [Biome](https://github.com/biomejs/biome-vscode/blob/main/src/status-bar.ts) do.

I did a manual bsearch over the priorities. `-1` behaves the same as any value <= 100, between Copilot and Notifications. If you feel that's a better spot I'm happy to change it. Here's what it looks like for comparison: 
image

> I have some nits, but for now, LGTM.

Please bring them on. I'm a lot less comfortable in this code than in the core Rust code.

mcecode:
> I know I got confused when coding it and then when testing it, but I'm not comfortable with the various levels of settings in VS Code.

I guess you can visualize it like CSS, the settings cascade. From less to more specific, it goes, default settings, user settings, then workspace settings. If the more specific setting is not set, then it falls back to less specific settings until you end up in the default setting. I don't know if that'll help, but that's how I imagine it.

> Much of this documentation is written in too abstract a style IMO. Googling and asking LLMs didn't get me there either using more concrete terminology.

I feel you on this. Docs and blogs outside of the official documentation are also sparse, that's probably why LLMs aren't very good at this, there's just not a lot of non-code data. If I need to do something specific or edge case, like in the case of middlewares, I usually have to hunt what API I need in `.d.ts` files and GitHub issues.

> If you feel that's a better spot I'm happy to change it.

I don't really have a preference, I just thought those values might put it closer to where you want it to be.

> Please bring them on. I'm a lot less comfortable in this code than in the core Rust code.

They're really just nits on the structure, nothing that should block the PR. Functionality-wise, this is good.

hippietrail:
Surprisingly, now that this has been merged and the extension was automatically updated on my main editor, Windsurf, which is a fork of VS Code, the indicator comes up in a totally different place! 
image

Even though Windsurf is my main editor, I had some trouble working on the Harper VS Code extension so did that work in VS Code itself.

Issue #384, opened 2025-01-15
feat: Flag nouns used instead of their related phrasal verb.


hippietrail:
A mistake that's been growing very common for some years now is joining the parts of a phrasal verb together into one orthographic word, but only when the result is also a noun in the dictionary/spellchecker.

This is a class of phrasal verbs consisting of two orthographic words, a verb in the present tense followed by an adverb or preposition:

- "I need to backup my hard drive" instead of "back up"
- "I will lookup the word in the dictionary" instead of "look up"
- "I workout every day" instead of "work out"

Note that it's only done when the result would be a valid noun: "He's not doing great now but he will comeback" but never "After this I will goback".

It's rare but not sometimes seen in conjugated forms such as "knockouted", but a spellchecker will catch those.

An incomplete list can be found on [notaverb.com](http://notaverb.com/)

By happy coincidence while cleaning up my full Google Photos I found this relevant photo I took a few years ago:
![breakup&workout](https://github.com/user-attachments/assets/96672db6-a98e-434e-8064-ee15849b090a)

ccoVeille:
In order to avoid other people to open duplicates, let me paste here the content of notaverb.com (an old site, not using https)

    backup
    carryout
    checkin
    checkout
    cleanup
    comeback
    cutover
    cutoff
    failover
    handout
    knockout
    lookup
    lockdown
    login
    logoff
    logon
    logout
    meetup
    meltdown
    pickup
    rollback
    rollout
    rollover
    setup
    showoff
    shutdown
    shutoff
    signin
    signoff
    signout
    signup
    slowdown
    splashdown
    startout
    startup
    takedown
    workaround
    workout 

    back up
    carry out
    check in
    check out
    clean up
    come back
    cut over
    cut off
    fail over
    hand out
    knock out
    look up
    lock down
    log in
    log off
    log on
    log out
    meet up
    melt down
    pick up
    roll back
    roll out
    roll over
    set up
    show off
    shut down
    shut off
    sign in
    sign off
    sign out
    sign up
    slow down
    splash down
    start out
    start up
    take down
    work around
    work out 

This way everything could be found via the search engine of GitHub

ccoVeille:
I edited my previous post to make it better, I also added examples I called "The Good, the Bad, the Ugly"

Take a look about their usage, it's frightening

[...]

PR #991, opened 2025-03-29
feat: Flag phrasal verbs spelled as compound nouns

hippietrail:
# Issues 
Implements #384

# Description
- Flags things such as "we breakup" and recommends "we break up".
- Related fixes to dictionary some annotations.
- Addad a phrasal verbs section to dictionary.
- Sorted token-related functions to make it easier to see what's there at a glance.
  `create_fns_on_doc`, `insert_struct_rule` / `insert_pattern_rule`, `create_decl_for`, `create_fns_for`
- Left in some commented-out debugging code that will help implement additional heuristics.

# How Has This Been Tested?
- Added unit tests for various cases.
- Added unit tests for false positives identified so far.

# Checklist


harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -0,0 +1,375 @@
...
+                (true, _) => Some("This word should be a phrasal verb, not a compound noun."),
+                (false, true) => {
+                    Some("This word might be a phrasal verb rather than a compound noun.")
+                }
+                _ => continue,
+            };
+
+            if let Some(prev_tok) = maybe_prev_tok {
+                if prev_tok.kind.is_adjective() || prev_tok.kind.is_determiner() {
+                    // eprintln!(
```
elijah-potter, line 153: Is there a reason you're keeping this stuff as a comment, rather than removing it entirely?

harper-core/src/linting/lint_group.rs
```diff
@@ -297,58 +297,58 @@ impl LintGroup {
 
         // Add all the more complex rules to the group.
         insert_struct_rule!(AdjectiveOfA, true);
```
elijah-potter, line 299: Whenever there's a place like this where you think it should remain sorted alphabetically, toss a comment there to remind other contributors.

hippietrail: > Whenever there's a place like this where you think it should remain sorted alphabetically, toss a comment there to remind other contributors.

Oh great idea - I didn't think of that! Just makes it easier to spot things without having to hunt.

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
hippietrail: > Is there a reason you're keeping this stuff as a comment, rather than removing it entirely?

In this case I'm pretty sure there's going to be more false positives/edge cases and these would have to be re-devised each time somebody wants to work on it.

elijah-potter:
This PR looks good to me. Would you mind resolving the conflicts so I can get it merged?

elijah-potter (approved):

Issue #993, opened 2025-03-30
Flag inflected verb forms after the infinitive particle "to"


hippietrail:
## Description

I just spotted this mistake in a comment in our source:
```
/// A location-agnostic structure that attempts to captures the context and content that a [`Lint`]
/// occurred.
```
This probably came from editing "that captures" to "that attempts to capture".
This is probably a common occurrence so worth our while to fix.

From the wide arrange of false positives listed below, it looks like the 3rd person singular present (the "-s" form) of the verb is the only relatively trivial one to identify.

## Examples

- A location-agnostic structure that attempts **to captures** the context and content that a Lint occurred.
- Whiteboard shape causes recording actor **to dies**
- Add linter **to finds** structures with uninitialized fields

## Potential False Positives

Many! "To" is a word with many uses.

- It's a preposition, which comes before nouns, and verbs with the `-ing` inflection are gerunds, which function as nouns
  - Is this a viable solution **to finding** all types of Flat Config?
- The `-ing` inflection is also the present participle, which functions as an adjective, which can occur between the preposition "to" and a noun:
  - Cannot write to **closing transport**
- The `-ed` inflection makes a verb a past participle, which functions as an adjective:
  - sending data **to closed** connection
- Adjectives, including past participles, can sometimes be used as "implied nouns":
  - If it exists how do I default it **to closed**, if not can it be implemented?
- Verbs in the third person singular present may have the same form as a plural noun:
  - Error due **to reads** >30000nt

hippietrail:
This is partly implemented but also has more false positives than expected.

PR #994, opened 2025-03-30
feat: implement #993

hippietrail:
# Issues 
#993 

# Description
Try to detect "to verb" where the verb is inflected.

There are many false positives so it errs on the side of not flagging some errors rather than flagging non-errors.

Due to Harper currently always treating the result of the `S` affix as a plural noun even when it's actually a 3rd person singular present verb (walks, eats, breathes, etc) the linter doesn't actually use the verb iterator it adds and uses the preposition iterator instead.

This can be redone when the affix system is improved to set different properties on `word/S` when `word` is a noun vs a verb (or both).

# How Has This Been Tested?
Added unit tests for positive and negative cases.

# Checklist


harper-core/src/linting/inflected_verb_after_to.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::{Dictionary, Document, FstDictionary, Span, TokenStringExt};
+
+use super::{Lint, LintKind, Linter, Suggestion};
+
+pub struct InflectedVerbAfterTo {
+    pub dictionary: Arc<FstDictionary>,
+}
+
+impl InflectedVerbAfterTo {
+    pub fn new() -> Self {
```
elijah-potter, line 12: Would you mind refactoring this to take a generic dictionary as an argument and passing it from the `LintGroup` constructor? There are other places in the code where we directly consume `FstDictionary::curated` from a linter, but it's a pattern I'd like to avoid moving forward.

harper-core/src/linting/lint_group.rs
```diff
@@ -343,6 +344,7 @@ impl LintGroup {
         insert_pattern_rule!(Hedging, true);
         insert_pattern_rule!(ExpandTimeShorthands, true);
         insert_pattern_rule!(ModalOf, true);
+        insert_struct_rule!(InflectedVerbAfterTo, true);
```
elijah-potter, line 347: You'll need to replace the macro call with something similar to how the `SpellCheck` rule is added.

harper-core/src/linting/inflected_verb_after_to.rs
hippietrail: Ah OK I'll look into that. So far I mostly see what other code does and copy the approaches I find.

hippietrail:
Let me know if I understood you correctly when implementing the changes you wanted.

While doing this I noticed some limitations in my linter and in the affix system as it's currently implemented, so I've added more tests that work now plus other tests that will work once those limitations are removed.

elijah-potter (approved):

Issue #978, opened 2025-03-27
Add a feature to flag deprecated names or spellings of places etc.


hippietrail:
Sometimes official names or spellings of countries, cities; and presumably products, trademarks, and people can change:

- Ayers Rock → Uluru
- Bangalore → Bengaluru
- Bombay → Mumbai
- Burma → Myanmar
- Calcutta → Kolkata
- Czech Republic → Czechia
- Ivory Coast → Côte d'Ivoire
- Madras → Chennai
- The Olgas → Kata Tjuta
- Peking → Beijing
- Saigon → Ho Chi Minh City
- Turkey → Türkiye
- Ulan Bator → Ulaanbaatar
- Upper Volta → Burkina Faso

(I left out some that are so old that they are unlikely to be used in ignorance: Formosa/Taiwan, Persia/Iran, Siam/Thailand, Sri Lanka/Ceylon)

This would make a nice little feature and probably an easy first project for a new contributor!

hippietrail:
This was implemented in #1013

PR #1013, opened 2025-04-03
Update place names 978

hippietrail:
# Issues 
#978 

# Description
Many place names have changed over the years. This lint looks for uses of known deprecated names and offers the new name.

It doesn't attempt to include all place names. My rule of thumb was to include ones where I recognized at least the old name or the new name, and isn't too "ancient".

It can:

- If known, it includes the year the new name was accepted in the message.
- It can handle multiple old names.

It cannot yet:

- Handle old place names with hyphens.
- Offer multiple new names, such as with and without accents.
- Recognize old names that don't use the correct case.
- Recognize old names with variations in accents.
- Replicate the exact whitespace between words if both the old and new names are multi-word.

# How Has This Been Tested?
- I included unit tests for all the straightforward cases and edge cases I could think of.
- `cargo test` passes with no problems.

# Checklist


harper-core/src/linting/update_place_names.rs
```diff
@@ -0,0 +1,301 @@
...
+/// Detect deprecated place names and offer to update them.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct UpdatePlaceNames;
+
+type PlaceNameEntryStr<'a> = ((i16, &'a str), &'a [&'a str]);
+type PlaceNameEntryVecChar = ((i16, Vec<char>), Vec<Vec<Vec<char>>>);
+
+// (year, new name), [old names]
+const RAW_PLACE_NAME_UPDATES: &[PlaceNameEntryStr] = &[
+    // Africa
```
elijah-potter, line 14: This looks very expensive to iterate through. Could you refactor it into a `PatternLinter` to take advantage of caching?

elijah-potter: Doing it once at startup doesn't negate the fact that you're iterating through each element for every word in the document. Run `cargo bench` and see what happens.

harper-core/src/linting/lint_group.rs
```diff
@@ -494,6 +496,7 @@ impl LintGroup {
         insert_struct_rule!(ThrowRubbish, true);
         insert_expr_rule!(Touristic, true);
         insert_struct_rule!(UnclosedQuotes, true);
+        insert_struct_rule!(UpdatePlaceNames, true);
```
elijah-potter: Change this to `insert_expr_rule` so we can take advantage of caching and I'll merge it.

hippietrail: > Change this to `insert_expr_rule` so we can take advantage of caching and I'll merge it.

Damn! How'd that get in there?

harper-core/src/linting/update_place_names.rs
hippietrail: Resolving this old thread, since it dealt with the old implementation.

PR #1019, opened 2025-04-04
feat: use the old code if `parallel` unavailable

hippietrail:
# Issues 
N/A

# Description
Shortly after my system regained its ability to run `just test` it broke again. This time the cause was easy to track down. A new optimization in the doc generation script uses [`parallel`](https://www.gnu.org/software/parallel/) but vanilla Mac systems don't have it.

# How Has This Been Tested?
`just test` works again for me

# Checklist


hippietrail:
Hmm this works fine on my machine. `cargo test`, `just test`, `just precommit`.

I thought I found and fixed a problem in a perl command which I cut and pasted over from the version before `parallel` was brought in.

But it seems to still be failing on GitHub.

elijah-potter:
Does this still work properly on your machine @hippietrail?

elijah-potter (approved):

PR #1030, opened 2025-04-05
feat: Expand "cuz" to "because"

hippietrail:
# Description

When dogfooding on YouTube transcripts they're always for of the word "cuz" for because. I know it's common not just on YouTube.

So I added an expansion to it similar to the ones for "stdout" and "w/o" but note that "cuz" is informal, which is a bit different from the others.

# How Has This Been Tested?

I added a unit test with a typical sentence from a YouTube transcript.

# Checklist


elijah-potter (approved):
Good stuff. I imagine `on face value` is quite a common mistake.

PR #1042, opened 2025-04-07
fix: update vscode deps

hippietrail:
# Issues 
N/A

# Description

Without this `just test-vscode` crashes at the end after all tests pass anyway.
```
30 specs, 0 failures
Finished in 6.822 seconds
[main 2025-04-07T08:27:08.325Z] Extension host with pid 52461 exited with code: 0, signal: unknown.
(node:52439) [DEP0168] DeprecationWarning: Uncaught N-API callback exception detected, please run node with option --force-node-api-uncaught-exceptions-policy=true to handle those exceptions properly.
Exit code:   0
```

Just a draft because I'm not totally sure about this. Needs domain expert.

# Demo

# How Has This Been Tested?

I couldn't debug the VS Code extension after the new VS Code update. With this test I can.

# Checklist


mcecode:
I'm not sure why this is happening either, but if updating the VS Code engine fixes the issue, then I think that's an okay fix.

I searched the VS Code repo for similar issues and stumbled upon microsoft/vscode#238600. Not sure if it's related, but they have the same exact error message.

P.S. The precommit failing is just a fluke.

hippietrail:
> I'm not sure why this is happening either, but if updating the VS Code engine fixes the issue, then I think that's an okay fix.

I still feel uncertain about it but I've taken this out of draft and ready for review.

> I searched the VS Code repo for similar issues and stumbled upon [microsoft/vscode#238600](https://github.com/microsoft/vscode/issues/238600). Not sure if it's related, but they have the same exact error message.

Will check it out.

> P.S. The precommit failing is just a fluke.

Very helpful - thanks!

elijah-potter (approved):

PR #1044, opened 2025-04-07
feat: trail and error→trial and error

hippietrail:
# Issues 
N/A

# Description

While looking into adding an "and" suggestion to the "a/an" linter I saw the typo "trail and error". This turns out to be common, so here's a lint.

# How Has This Been Tested?

Added the comment from the "a/an" linter as a unit test.
`cargo test`

# Checklist


elijah-potter (dismissed):
I love it when we get new rule ideas from blunders in the documentation.

Issue #544, opened 2025-01-30
feat: Should check text in string literals as well as comments


hippietrail:
I actually assumed it already checked both, but:

<img width="609" alt="Image" src="https://github.com/user-attachments/assets/9c3d4db5-7486-447d-ab05-9cbaa656bc3f" />

PR #1045, opened 2025-04-07
feature: highly kept (secret)→well-kept

hippietrail:
# Issues 
N/A

# Description

I heard somebody mention a "highly kept secret" in a YouTube video and it sounded strange to keep things in a high manner.
But it seems to be pretty common since the 2000s. (I think I've heard other odd uses of "highly" recently too.)

# How Has This Been Tested?

I added two test based on sentences I found in the wild: one with a hyphen, which is more correct, and one without, wich is more common nowadays.

# Checklist


harper-core/src/linting/phrase_corrections.rs
```diff
@@ -923,6 +923,13 @@ pub fn lint_group() -> LintGroup {
             "Use `without` instead of `w/o`",
             "Expands the abbreviation `w/o` to the full word `without` for clarity."
         ),
+        "WellKept" => (
+            ["highly-kept", "highly kept"],
+            // There may be other good alternatives such as closely-guarded or tightly-held
+            ["well-kept"],
+            "`Highly-kept` is not stanard. To describe secrets, `well-kept` is the most used phrase.",
```
elijah-potter, line 930: Did you mean `standard` here?

hippietrail: Indeed. After 4 years my keyboard has started to double some keypresses and miss others and I don't always catch it. Being able to lint inside quoted text would help a lot! #544

elijah-potter (approved):

elijah-potter (approved):

PR #1046, opened 2025-04-08
fix(comments): ignore comments CSpell compatibility

mcecode:
# Issues 

Related to #825.

# Description

While I was updating our docs in #1000, I noticed that there's a problem with our CSpell ignore comment compatibility based on [their docs](https://cspell.org/docs/Configuration/document-settings#ignore). It's supposed to be `cSpell` not `cspell` and they also support `spell-checker`, so I updated our implementation.

# How Has This Been Tested?

There are already tests in place for this that shouldn't break. I could add more for the other ignore comments, but I think that'll be too redundant.

# Checklist


cylewaitforit:
I agree that the goal should just be _some_ compatibility. I think keeping the more generic “spellchecker: ignore” and its permutations. Possibly dropping the support for the “cspell: ignore” permutations would suffice and be more straightforward to document/maintain. Since what would remain would be the general, as well as  harper specific, ignores.

mcecode:
I agree with @cylewaitforit's points since that's what #825 asked for too. I've removed `cspell:` ignore comments and have updated the docs accordingly.

elijah-potter (approved):

Issue #1051, opened 2025-04-10
`correct_number_suffix` matches non-number suffixes


hippietrail:
## Description

The `correct_number_suffix` lint seems to be about wrong "[ordinal suffixes](https://en.wikipedia.org/wiki/English_numerals#Ordinal_numbers)", as in non-native English speakers writing 3th instead of 3rd etc.

But in practice it is triggered by any text appended directly to a number. Another reason for this to happen is simply a missing space or hyphen as in `33story`.

It further assumes without checking that the appendage is 2 letters and only checks/replaces those. In my example is puts squigglies under `ry` and changes it to `33stord` instead of `33rd`.

## Screenshots

<img width="659" alt="Image" src="https://github.com/user-attachments/assets/13811764-053b-41b7-8f6c-b52b2141fc69" /><img width="636" alt="Image" src="https://github.com/user-attachments/assets/33f62ce5-bf96-49cb-a177-5f35fb8b2bb3" />

<img width="184" alt="Image" src="https://github.com/user-attachments/assets/23453cc2-e224-412d-85af-fa594503e343" />

hippietrail:
I hadn't noticed the function `condense_number_suffixes` before now. Its documentation isn't very specific. Is it intended specifically for ordinal suffixes (1st, 2nd, 3rd, 4th) or any suffixed numbers (0s and 1s, 100km, 50hrs, 1980s, 1024kb, etc)?

When I added my lexers for decades and 0s and 1s I didn't spot this function so those lexers create tokens that are not `Number` so are not subject to the previously mentioned function.

My decade lexer returns a new type of token `Decade` while my plural digit lexer treats `0s` and `1s` as words. Mainly because I wasn't sure how either would pan out but was pretty sure that would emerge later as more examples of numbers with suffixes revealed themselves.

I'm not sure if the correct way forward is to have a unified concept of suffixed numbers and merge the decade and plural digit stuff there and correct ordinals along with that.

Or whether to rename `_number_suffix` stuff to `_ordinal_suffix` and have those only deal with `st`, `nd`, `rd`, and `th`.

Does the `Number` token cover suffixed numbers as well as plain numbers? Should it? Should it only deal with long numbers.

There's also a concept of `word_like` so if need be, things like suffixed numbers, decades, plural digits, etc could more specifically work with that concept... if they don't already.

I could come up with a quick fix for the `33story` false positive, but I think more thought of the architecture is the proper way to proceed.

PR #1054, opened 2025-04-10
fix(#1051): Ignore potential ordinal suffixes of length greater than 2

grantlemons:
# Issues 
#1051

# How Has This Been Tested?
Added a unit test for the function and wrote an example sentence:
```
This 33 story building is located in the heart of the financial district.
```

# Checklist


elijah-potter:
What do we need to do to get this ready for review?

hippietrail:
> What do we need to do to get this ready for review?

I believe it fixes the immediate problem just fine. At the time I started to dig deeper and wondered if it was kind of a symptom of a deeper problem with text concatenated onto numbers that could maybe be handled in a systematic way?

grantlemons:
I hadn't marked it ready for review because @hippietrail had the above concern, but I figure I might as well mark it ready since it does fix the immediate issue. @hippietrail you might want to open an issue about your concern if you haven't already.

hippietrail:
> I hadn't marked it ready for review because @hippietrail had the above concern, but I figure I might as well mark it ready since it does fix the immediate issue. @hippietrail you might want to open an issue about your concern if you haven't already.

I was hoping to get more feedback on it but it's not all in my head any more. I think we should just go ahead and merge this and in the meantime I'll collect samples of other kinds of suffixed number problems for a future issue.

PR #1058, opened 2025-04-12
added words to dictionary

ficd0:
# Issues 

# Description
Added some words to the dictionary

# Demo

# How Has This Been Tested?

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -51617,7 +51617,7 @@ worksheet/14MS
 workshop/~14MS
 workshy/5
 worksite/1S
-workspace/1
+workspace/1S
 workstation/~1MS
 worktable/1MS
 worktop/1S
@@ -52362,6 +52362,28 @@ WD40/1M
 whitespace/~1SM     # dictionaries prefer: white space
 wikilink/~1SM       # !! please check and comment !! elsewhere we have Wikilink
 
+ChaCha20/M2 # Cryptography
+NSERC/M2 # Natural Sciences and Engineering Research Council of Canada
+Ed25519/M2 # Cryptography
+ECDH/M2 # Cryptography
+Poly1305/M2 # Cryptography
+Waybar/M2 # Linux software
+CSE/M2 # Communication Security Establishment
+aerc/M2 # Linux software
+Niri/M2 # Linux Software
+DSA/M2 # Cryptography
+X25519/M2 # Cryptography
+TOTP/M2 # Cryptography
+SpeechBrain/M2 # Machine learning
+tmux/M2 # Linux Software
+OpenAI/M2 # Machine Learning
+codec/M1S # Technical term
+LLM/M1S # Machine learning
+HackerRank/M2 # Platform
+Maildir/M1S # Linux term
+MITM/1MS # Cybersecurity
+Diffie/M2 # Cryptography (Diffie-Hellman)
+
 # Experimental multi-word entries
 # Harper may not recognize these as words in documents
 # But it will be able to include them in suggestions
@@ -52674,3 +52696,10 @@ write-off/1MS
 # And they will be reviewed
 # Word added using the `just addnoun` command will be added here
 
+Liu/M2 # Surname
+SMTP/M2
+monorepository/M1S
+inherence/M1S
+remap/ABGD
+neurodivergent/PY
+exfiltration/NS


elijah-potter (dismissed):

harper-core/dictionary.dict
```diff
@@ -52391,3 +52391,34 @@ write-off/1MS
 # And they will be reviewed
 # Word added using the `just addnoun` command will be added here
 
+ChaCha20/M2
```
hippietrail, line 52394: Not sure the best way to review this, seems a bit awkward so let me know if there's a better way.
I'd put ChaCha20, NSERC, Ed25519, ECDH, Poly1305 in the proprietary/specialized section and I think they're all obscure enough to warrant a very short comment each.
`tmux` seems to be like `npm` in that it's a proper noun (or trademark?) spelled lowercase, so it should get `/M2`
Conversely, `LLM` is an abbreviation for a common noun, so should get `/1MS`
maildir looks like it should be Maildir going by Wikipedia
MITIM should have `/1MS` like LLM
I can't find what Waybar is - that one definitely needs at least a comment and maybe a justification?
CSE has many expansions. for ones like this I include the most relevant one or two in a comment. Probably should be `/1MS`
I can't find "aerc". Wiki has two for "AERC" - needs comment and maybe justification?

(I know the wrong `/M1S` and `/M2` are probably due to `just addnoun` - it only makes simplistic guesses about common vs proper noun based on capitalization. I probably should change it for all-caps words.)

ficcdaf: I agree, I didn't realize we could leave comments on the words now. I'm a bit iffy about affix rules but how about this I'll comment explanations for all of them, which could be helpful in figuring out the correct affix rules.

harper-core/dictionary.dict
```diff
@@ -51617,7 +51617,7 @@ worksheet/14MS
 workshop/~14MS
 workshy/5
 worksite/1S
-workspace/1
+workspace/1S
```
hippietrail, line 51620: I always include `M` with all nouns because it means `'s` and you can always have things like `my workspace's setup` etc. It could probably be done internally to every noun without an annotation but since `'s` is also used in contractions of `is` and `has` that might be the reason to make it explicit.

harper-core/dictionary.dict
```diff
@@ -52674,3 +52696,10 @@ write-off/1MS
 # And they will be reviewed
 # Word added using the `just addnoun` command will be added here
 
+Liu/M2 # Surname
+SMTP/M2
+monorepository/M1S
+inherence/M1S
+remap/ABGD
```
hippietrail, line 52703: Don't forget how we discussed this before. (Apologies if it's just me doing the review wrong, I'm a noob at GitHub stuff)

`remap` needs `p` to be doubled before endings/suffixes that start with `e` or `i` and `A` contributes `re-` which is already there. So these annotations all lead to nonwords:
- reremap
- remapable
- remaping
- remaped

These suggest a base word "mape", which is not a word. This is because the affix system isn't smart enough to know about when to double the last letter. Because of this we enter all the forms in longhand:
```
- remap/4S
- remappable/5
- remapped/4
- remapping/4
```
`remap` is a verb `4` - at this point we don't yet have different annotations for root form vs tense forms of a verb. We  should probably think about that at some point. So `remaps`, `remapped`, and `remapping` are also verbs. `remappable` is an adjective `5`

harper-core/dictionary.dict
```diff
@@ -52674,3 +52696,10 @@ write-off/1MS
...
+neurodivergent/PY
+exfiltration/NS
```
hippietrail, line 52705: The first is an adjective so should have the `5` annotation. `P` and `Y` look right for -ness and -ly.
The second is a noun so should have the `1` annotation. `N` adds a nominalization suffix but that's already supplied so would end up doubled like the `re-` prefix before. Again as a noun `M` is always applicable. That would give us:
```
neurodivergent/5PY
exfiltration/1MS
```
Let's use `just getforms` to compare:
```
neurodivergent/5PY
 - neurodivergent
 - neurodivergently
 - neurodivergentness
```
```
exfiltration/NS
 - exfiltration
 - exfiltrations
 - exfiltrationen
```
vs
```
exfiltration/1MS
 - exfiltration
 - exfiltration's
 - exfiltrations
 ```

hippietrail (changes requested):

elijah-potter:
@ficcdaf, I'd love to get this PR merged. Is there anything you need from us get it polished up?

hippietrail:
> @ficcdaf, I'd love to get this PR merged. Is there anything you need from us get it polished up?

I'm new to GitHub reviewing so feel free to critique my attempts at doing so.

elijah-potter:
No rush. But I'd love to get this merged, @ficcdaf.

elijah-potter:
Closing because this PR is stale.

PR #1077, opened 2025-04-16
fix: #1075 package logo font for vscode extension

### LINKED ISSUE #1075: Harper logo in status bar dialect indicator renders as a box
**Describe the bug**

When VS Code updated Harper to 0.29.0 it tries to draw the Harper logo in the dialect indicator in the status bar but only draws a box character.

Running the VS Code extension from it's sub-project in the Harper repo draws it correctly.

**To Reproduce**
1. Make sure Harper is updated to 0.29.0
2. Look for the dialect indicator in the status bar. Does it have a Harper logo or a box?

**Expected behavior**

It should look like this: 

Image

And when running the extension from it's subproject in the harper repo it does.

**Screenshots**

When just using VS Code normally after the Harper 0.29.0 update it actually looks like this:

Image

**Platform**
Apple Silicon macOS, VS Code Version: 1.99.2

**Additional context**

I'm assuming this is some missing file or missing bit of config from when I added the logo but everything seems to be in the repo. I'm not very well-versed in VS Code extension dev though.

hippietrail:
# Issues 

Attempts to fix #1075 

# Description

When building and testing the Harper extension locally the new Harper logo appears next to the dialect indicator in VS Code, but when installing the Harper VS Code extension the logo renders as a box.

This seems to be due to the `harper.woff` single-character custom logo font not getting packaged into the `.vsix` file.

This is quite far outside what I feel I have competence in but I asked a couple of LLMs and tried a bunch of things, and I think this works. It includes tests but 

# How Has This Been Tested?

I added, with the help of an llm, two new tests:
- Test that the `media` directory exists and that the `harper.woff` file is in it.
- Extract the `.vsix` file that's been built and make sure the `harper.woff` file is in that too.

THIS DOES NOT MEAN IT ACTUALLY WORKS. In fact when I tried manually installing the resulting `.vsix` file I still saw a box rendered.

I am assuming the tests are testing hard-coded paths and files that it's assuming are right but might be different from those actually used in some places?

# Checklist


packages/vscode-plugin/.vscodeignore
```diff
@@ -5,3 +5,12 @@
...
 !icon.png
+
+# Ignore build files
+build/
+
+# Ignore test files
+tests/
+
+# Include media directory
+!media/
```
mcecode, line 16: I think this line is the only thing that's needed to actually fix the issue.

As an explanation, the first line of this file is `**/**` which tells `@vscode/vsce` to ignore everything when packaging and all the subsequent lines starting with `!` tell it what is an exception to that. So, when packaging the extension without this line this is the output:

pr-1077-1

But with this line included, this is the output:


pr-1077-2

packages/vscode-plugin/src/tests/suite/font.test.ts
mcecode, line 1: I don't think this test is very useful because we already know that the font file is there, we just need to know if it's being included when the extension is being packaged. Additionally, this will always pass whether the fix in the `.vscodeignore` file is there or not because when the extension is being tested using `just test-vscode` it's testing with `packages/vscode-plugin` as its root, not the packaged extension, and therefore `packages/vscode-plugin/media/harper.woff` will always be there because it's in the repo.

packages/vscode-plugin/src/tests/suite/vsix.test.ts
mcecode, line 1: I don't think this test is very useful either, but for the opposite reason. This test will always fail. That's because we're not packaging the extension when testing it, so `packages/vscode-plugin/harper-x.x.x.vsix` won't exist when this runs, thus, the `ENOENT` error in the precommit CI.

packages/vscode-plugin/.vscodeignore
mcecode: You can check the latest VS Code extension release action and see that this is indeed the case: https://github.com/Automattic/harper/actions/runs/14476652533/job/40603810278#step:7:87

packages/vscode-plugin/src/tests/suite/vsix.test.ts
hippietrail: Thanks so much. Awesome to have some clarity from somebody that understands this world! Do you think there's any value in having any tests at all or should I just ditch everything completely other than the `!media` line in the `.vscodeignore` file?

Also is `media` the best name? The LLM insisted it is but from trying to find similar things to look at I found a `files` field seemed to be used in a similar way and, unlike my `media` field the LSP for this format did give hover info for `files`. The LLM insisted they were different but it could've just been doubling down on a hallucination. **WAIT!** I see I'm confusing myself again with `media` used as a directory name to put the font in and `media` used as a field name in `package.json`, which may be a valid thing or may be where the LLM got its hallucination from.

mcecode: > Thanks so much. Awesome to have some clarity from somebody that understands this world! Do you think there's any value in having any tests at all or should I just ditch everything completely other than the `!media` line in the `.vscodeignore` file?

I think just adding `!media` to the `.vscodeignore` file would suffice.

> Also is `media` the best name? The LLM insisted it is but from trying to find similar things to look at I found a `files` field seemed to be used in a similar way and, unlike my `media` field the LSP for this format did give hover info for `files`. The LLM insisted they were different but it could've just been doubling down on a hallucination. **WAIT!** I see I'm confusing myself again with `media` used as a directory name to put the font in and `media` used as a field name in `package.json`, which may be a valid thing or may be where the LLM got its hallucination from.

AFAIK, the `files` field is the opposite of the `.vscodeignore` file, it tells what to include instead of what to ignore, but it might only be respected by npm and other JS package managers, it's been a while since I've looked into it. I'm not sure if the `media` field is a thing though, it's neither in the [VS Code docs](https://code.visualstudio.com/api/references/extension-manifest) nor in the [npm docs](https://docs.npmjs.com/cli/v11/configuring-npm/package-json), so I'm guessing it's an LLM hallucination. `media` as the name of the directory that contains `harper.woff` is fine.

packages/vscode-plugin/package.json
```diff
@@ -1843,5 +1843,6 @@
 	},
 	"dependencies": {
 		"vscode-languageclient": "^9.0.1"
-	}
+	},
```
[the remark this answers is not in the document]
mcecode: See my comment above for more info, but I think the `media` field can be removed as I don't think it's really needed.

elijah-potter:
@hippietrail, @mcecode, is this ready for merging?

elijah-potter (approved):

Issue #1089, opened 2025-04-20
"The light of a star" falsely reported by "Adjective Of A" lint


RunDevelopment:
## Description

The following sentence gets the error "The word `of` is not needed here." for "light of a".

```
The light of a star.
```

The suggested fix is "Replace with: light a", which results in the sentence: "The light a star."

I turned off some lints and found that the "Adjective Of A" lint is responsible. So it seems that harper thinks that "light" is an adjective here. While "light" can be an adjective, it's a noun here.

More examples that also get reported:

- The **red of a** drop of blood.
- The **top of a** hill.
- The image needs to be a **multiple of a** certain size.

## Resources

n.a.

## Screenshots

![Image](https://github.com/user-attachments/assets/e63f4ef1-be1f-4520-a853-b27c9a4d2c5e)

hippietrail:
Thanks for these! I constantly add more false positives this linter. I'll add these three to PR #1090 and let me know of any more you find!

RunDevelopment:
Maybe I'm misunderstanding the PR, but did you add these words to something like a whitelist? If so, then this seems like the wrong approach to me. All the examples I gave (except the "multiple of" example) are just sentences I made up to illustrate a pattern of behavior. There's nothing special about them. Especially for colors, it's easy to make up more examples.

Wouldn't it be a lot better to simple *not* report all sequences of the form `<a/an/the> <adjective> of <a/an>`? If the supposed adjective is preceded by an article, we know that it's not actually an adjective but a noun. Of course, even this approach is easy to defeat (e.g. with actual adjectives: "the deep red of a drop of blood", "the faint light of a distant star"), but it seems a lot more scalable than an ever-growing list of exceptions, which will eventually also cause false negatives.

---

This is unrelated, but does Harper not create some kind of AST to store semantic information about tokens? I initially expected this to an issue in how the AST is parsed, so I was really surprised to see that you seemingly just added those words to an ignore list.

hippietrail:
> Maybe I'm misunderstanding the PR, but did you add these words to something like a whitelist? If so, then this seems like the wrong approach to me. All the examples I gave (except the "multiple of" example) are just sentences I made up to illustrate a pattern of behavior. There's nothing special about them. Especially for colors, it's easy to make up more examples.
> 
[...]

PR #1098, opened 2025-04-21
Updated `justfile` to check tests and benches

RunDevelopment:
This PR fixes the compiler error [I mentioned here](https://github.com/Automattic/harper/issues/1089#issuecomment-2817334955) and changes the `justfile` to make sure something like this never happens again.

Changes:

- Update `justfile` to change `cargo build` to `cargo build --all-targets`. [The `--all-targets` flag](https://doc.rust-lang.org/cargo/commands/cargo-check.html#target-selection) will instruct cargo to build not just the main libraries and binaries, but also tests and benchmarks. This allows the command to catch compiler errors in rarely-used files.
- Fixed the compiler error in `harper-core\benches\parse_demo.rs` .

This PR comes in 2 commits. The first only updates the `justfile`. CI will fail for this commit, because CI will now detect the compiler error in `parse_demo.rs`. The second commit will fix the compiler error.


elijah-potter (approved):
Great stuff! Thanks for taking a look at this.

PR #1099, opened 2025-04-21
feat: `harper cli annotate` and `just annotate`

hippietrail:
# Issues 

Doesn't solve #1089 but it was discussed there.

# Description

Pass a text in to have the affix system apply the dictionary metadata to each word.
Due to `just`'s nature you need to use two different quotes to pass in a sentence:
`'"The cat sat on the mat"'`

Every word can have multiple parts-of-speech. The major ones are given their common abbreviations and each is given a different ANSI colour.

In addition, words which are not in the dictionary at all are underlined. Words which are only in the dictionary after normalizing to all lowercase are italicised.

# Demo
image

# How Has This Been Tested?

No unit tests, just trying random sentences.

# Checklist

# Help

The code actually treats the input string as markdown because I'm not sure how to implement it as plain text. If anyone who can show me how to modify the code to do so I would be grateful.

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -76,6 +76,8 @@ enum Args {
         /// The document to mine words from.
         file: PathBuf,
     },
+    /// Annotate a provided string.
+    Annotate { text: String },
 }
 
 fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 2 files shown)


RunDevelopment:
Thanks for working on this @hippietrail!

This is pretty good, but not really what I need. The most important properties that I care about are:

1. Diffable. I want to be able to use my IDE to see the differences. This means that lines shouldn't be much longer than 80 characters.
2. Readable. It should be immediately obvious what's the text and what are the annotations. Annotations should also be terse.
3. Information rich. The annotations should encode as much information as possible.

You did 2 and 3 perfectly, but I really need diffs. This also unfortunately also means that your nice ANSI colors will disappear when I save the command output to disk. 

I also want to make tests with this. Basically, have a few documents and save their annotated text as a snapshot. When I make the changes to fix #1089, I want the reviewers to see that my changes are doing exactly what I say they do directly in the diff of the PR.

I would propose the following format:

```md
# This      is  a         demonstration  of      the       
  prep.det  v   prep.det  n              v.prep  prep.det  
# new      feature  to    annotate  text.
  n.v.adj  n.v      prep  v         n.v
```

The format works as follows:
1. Break the text into lines at most 40 characters wide. This will ensure that the annotated text isn't too wide.
2. Each line of text will be represented with 2 lines: the text line and the annotation line. The 2 lines are aligned such the text and metadata of each token start at the same column.
3. 2 spaces between tokens for easier readability.
4. The text line starts with `# `. This allows us to (ab)use Markdown for some light syntax highlighting in IDE.

Alternatively, we could also use your proposed syntax with newlines between token. This is syntactically valid JS, it can be (ab)used for highlighting:

```js
'This'.prep.det
'is'.v
'a'.prep.det
'demonstration'.n
'of'.v.prep
'the'.prep.det
'new'.n.v.adj
'feature'.n.v
'to'.prep
'annotate'.v
'text'.n.v
```

 Thoughts?

elijah-potter:
Superseded by #1119. That will be the canonical way to do annotations like this.

PR #1101, opened 2025-04-22
feat: examples for using `harper.js` from `justfile`

### LINKED ISSUE #1096: Add an example of how to use `harper.js` from the `justfile`
## Description

I want to add a new commandline tool. Since I haven't used `harper.js` yet I thought it would be a useful experiment to try to use it for my tool rather than implement it in `harper-cli` in Rust.

But once again I'm finding that "difficult" Rust is trivially easy compared to "easy" JavaScript. Maybe just for me.

## Component

Which component of Harper would this feature affect? For example:
- `harper.js`

## Use Cases

- Shows how to use `harper.js` in another environment.
- Shows how to add `justfile` recipes without having to use Rust.

## Implementation Ideas

I wish I knew! Nothing I tried worked and the ideas the LLMs were giving me were looking more and more convoluted.

hippietrail:
# Issues 

Fixes #1096

# Description

Inlcudes bun and node examples

Node seems to require an IIFE, but I may be missing a trick. Bun is cleaner and can use TypeScript directly in the `justfile`

All I needed to get it to work was the `just` special which I'd used before but forgotten:
```js
require('{{justfile_directory()}}/packages/harper.js/')
```

# Demo
image

# Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -474,7 +474,7 @@ newest-dict-changes *numCommits:
     });
   });
 
-# Test using harper.js from packages/harper.js in `justfile`
+# Example of using harper.js with Node.js in `justfile`
 harperjsnode test_string:
   #! /usr/bin/env node
   (async () => {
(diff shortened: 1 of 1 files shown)


elijah-potter:
Is there utility for this that I'm not seeing. If it's just an example, might it be better shown in the docs, rather than in a part of our build system?

hippietrail:
> Is there utility for this that I'm not seeing? If it's just an example, might it be better shown in the docs, rather than in a part of our build system?

Yep you're right. I'll close it. I might add a real `just` command this way at some point now that I know how to do it when I have the need for one.

PR #1102, opened 2025-04-22
feat: implement dialect indicator for Obsidian plugin

hippietrail:
# Issues 
Part 2 of fixing #982

# Description

Next to the Harper logo/button in the status bar the country code and flag for the selected dialect are now displayed.
These update when the dialect setting is changed.

# Demo
image

# How Has This Been Tested?

Manual testing.

# Checklist

### DIFF
--- a/packages/obsidian-plugin/src/index.ts
+++ b/packages/obsidian-plugin/src/index.ts
@@ -39,7 +39,7 @@ export default class HarperPlugin extends Plugin {
 	private harper: Linter;
 	private editorExtensions: Extension[];
 	private delay: number;
-	private statusBarItem: HTMLElement | null = null;
+	private dialectSpan: HTMLSpanElement | null = null;
 
 	constructor(app: App, manifest: PluginManifest) {
 		super(app, manifest);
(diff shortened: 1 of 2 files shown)


packages/obsidian-plugin/src/index.ts
```diff
@@ -269,6 +299,12 @@ export default class HarperPlugin extends Plugin {
 			},
 		);
 	}
+
+	updateStatusBar(dialect: Dialect) {
```
elijah-potter, line 303: Could we use `updateStatusBar` for _both_ its initial creation _and_ further updates?

hippietrail: > Could we use `updateStatusBar` for _both_ its initial creation _and_ further updates?

I'll have a look. I had it kind of that way in the first commit but it was a bit heavy handed. There might be a middle way though...

hippietrail: It's just a one-line change and adds a tiny bit of indirection. Maybe it's clearer. Will commit...

elijah-potter (approved):

PR #1108, opened 2025-04-22
feat(core): wrote rule for do not wan

elijah-potter:
# Description

This PR adds several rules from the backlog. It would better for you to read the code, rather than read me list them out here. 

# How Has This Been Tested?

Unit tests.

# Checklist


harper-core/src/linting/it_is.rs
```diff
@@ -0,0 +1,251 @@
...
+        let span = its_token.span;
+        let text = span.get_content(source);
+        Some(Lint {
+            span,
+            lint_kind: LintKind::Miscellaneous,
+            suggestions: vec![Suggestion::replace_with_match_case(
+                "it's".chars().collect(),
+                text,
+            )],
+            message: "Did you mean `it's` (short for “it is”) instead of `its`?".to_string(),
```
hippietrail, line 79: To me this is asking the user if they intend to replace "its" with "it's".

I was just thinking about doing a feature request for "instead of" vs "rather than". The former is about replacement whereas the latter is about preference.

hippietrail: > I'm a little confused by your phrasing. What do you think it should say?

It's definitely hard to explain. "instead of" means "in the place of" so it means "Did you mean to replace its with it's" which implies they know it's is right but decided to replace it with its for some reason. Whereas usually they're using its or it's intentionally in the way they believe is right but is not, or they use them haphazardly because they never really understand which is which.

We want a wording that means "Rather than using the word that means 'it is' didn't you mean the word that means 'of it'?"

But that seems a bit wordy. Let me see what an LLM can suggest...

- "Did you mean 'it's' (it is) rather than 'its' (possessive)?"
- "Consider using 'it's' (it is) instead of 'its' (possessive form)."
- "You might want to use 'it's' (it is) for clarity, as 'its' indicates possession."
- "Please check if 'it's' (it is) is the correct choice for your sentence."
- "Did you intend to write 'it's' (it is) in this context?"

Notice that it uses "instead of" in a different way, now asking the user to replace one.

It really is subtle. I hope I'm not making it even more confusing.

elijah-potter: Nope! That makes sense.

harper-core/src/linting/phrase_corrections.rs
```diff
@@ -1105,6 +1105,12 @@ pub fn lint_group() -> LintGroup {
             "Use `best regards` to convey sincere well wishes in a closing.",
             "In valedictions, `best` expresses your highest regard—avoid the typo `beat regards`."
         ),
+        "Freezing" => (
```
hippietrail, line 1120: You could also have "Starving" for "very hungry". And also an opposite rule for detecting hyperbole?
Which reminds me I started looking into how to implement settings beyond just on-off today but I didn't get very far.

elijah-potter: Done.

PR #1110, opened 2025-04-23
Fix Lua syntax in neovim example config

nikosavola:
# Issues 

# Description

Fix lua syntax for the default config in the docs.

# Demo

# How Has This Been Tested?

# Checklist


elijah-potter (approved):
Good catch. Thanks!

PR #1112, opened 2025-04-23
Flags the definite article used together with a possessive

hippietrail:
# Issues 
N/A

# Description

People often write "the my" or "my the" accidentally. This is usually the result of editing but probably sometimes comes from non-native English speakers. I just noticed this is a very old post of mine.

All personal possessive adjectives are handled, but not "its" since that's more often a misspelling or typo for "it's".

I added the `Clone` trait to `AnyCapitalization` because I think that made a slight optimization or avoided a redundancy, but let me know if I was wrong.

Two edge cases are avoided:

1) If the pronoun follows "the" and is capitalized, it's usually part of a proper noun etc such as "the My Documents folder"

2) "her" is both a possessive adjective and an objec-case pronoun so "her the" is not flagged as in "give her the book"

# How Has This Been Tested?

I added atomic unit tests for both orders, capitalized, and accepting either the article or the possessive as the correction.

I mined GitHub for examples of as many combinations as possible, including false positives.

# Checklist


elijah-potter:
I see that your using `WordSet`s rather than using metadata. Is there a reason for that? What would we need to do for metadata to work just as well?

hippietrail:
> I see that your using `WordSet`s rather than using metadata. Is there a reason for that? What would we need to do for metadata to work just as well?

I put comments in the metadata code for the majority of obvious missing flags. I believe I added annotations for some of them in the affix json and a few in `dictionary.dict`

For pronouns and possessives it gets very subtle because some things which in common parlance are referred to as possessive pronouns are possessive adjectives, and English also has possessive pronouns, but the syntax is different between them and needs to be unambiguous to be able to build linting rules on top of.

For instance people call "my" a possessive pronoun but it can't take the place of a noun so is not a pronoun. "Mine" is a possessive pronoun and can take the place of a noun. The two cannot be interchanged and if they are that's the kind of grammatical mistake we want to be able to detect.

Butted up against this is the possessive `M` on nouns which is itself ambiguous since it's probably inherited from Hunspell where it just meant `'s` but `'s` has three meanings in English: possessive on nouns and both 'is' and 'has' on various kinds of words I think.

So I've been mulling this all over slowly in my head before diving in and doing anything overly rash to the affix system. Each little improvement has a flow-on effect.

For instance if we're going to fix `M` we probably also want to fix `S` which for words themselves means "add (e)s on the end" but in the metadata it's treated as plural, ignoring that it also adds (e)s onto verbs to create the 3rd person singular present verb form: writes vs write, something we definitely want to lint!

For `S` the two uses don't conflict so we could keep one annotation for both meanings and fix the logic to first check/apply the base word property and deciding based on that what derived property to give.

But for `M` they do conflict: "the car's tyres are flat" vs "the car's got flat tyres" and "the car's in the garage".

Many/most of the affix system properties do not get propagated properly when the cross-product and related steps are done. I first noticed this when I wanted to lint inflected forms of swear words and when I investigated I found that a lot were missing like verb tenses no longer marked as verbs, prefixes and suffixes either don't change the part-of-speech of the derived form based on the pos of the base form, or keep the same one when that is correct.

I think a lot of the 'generated' forms in the dictionary are missing most of their metadata.

I haven't gone down all of these rabbitholes completely so some of what I said is likely wrong but the overall gist is.

Importantly, it all works as it is now pretty well so I didn't want to break it trying to improve it (-:

PR #1117, opened 2025-04-23
feat: `Document` QoL functions for tokens & words

hippietrail:
# Issues 
N/A

# Description

- `get_token_offset()` allows you to look in either direction from a document index iterator. Since `get_token()` takes a `usize` you had to first check if you were going to look before index 0, whereas looking past the end would just return None.
  - With `get_token_offset()` you'll get a Token or None in either direction.
  - Having a base and an offset lets you incrementally step further bit by bit in either direction from an index to get more context for disambiguation heuristics.
- `get_next_word_from_offset()` takes this a bit further since often such heuristics require sequences of words separated by whitespace. Now you can pass the offset from your iteration index to the whitespace and this will check both that whitespace then continue to the next word, returning `None` if either the whitespace or the word wasn't found.
  - Works in either direction based on the sign of the offset. So you can use a loop or nested logic looking at the next or previous word as needed for disambiguation.

# How Has This Been Tested?

I added a bunch of unit tests to cover most use cases:
- Positive offset
- Negative offset
- Returning a token
- Returning None when out of range
- Returning None when expected whitespace token not found or expected word token not found.

# Checklist

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -249,6 +249,14 @@ impl Document {
         self.tokens.get(index)
     }
 
+    /// Get a token at a signed offset from a base index, or None if out of bounds.
+    pub fn get_token_offset(&self, base: usize, offset: isize) -> Option<&Token> {
+        match base.checked_add_signed(offset) {
+            None => None,
+            Some(idx) => self.get_token(idx),
+        }
+    }
+
     /// Get an iterator over all the tokens contained in the document.
     pub fn tokens(&self) -> impl Iterator<Item = &Token> + '_ {
         self.tokens.iter()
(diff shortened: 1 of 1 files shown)


harper-core/src/document.rs
```diff
@@ -259,6 +267,27 @@ impl Document {
         self.tokens().map(|token| token.to_fat(&self.source))
     }
 
+    /// Get the next or previous word token relative to a base index, iff separated by whitespace.
+    /// Returns None if the next/previous token is not a word or does not exist.
+    pub fn get_next_word_from_offset(&self, base: usize, offset: isize) -> Option<&Token> {
+        // Look for whitespace at the expected offset
+        if !self.get_token_offset(base, offset)?.kind.is_whitespace() {
+            eprintln!(
```
elijah-potter, line 275: Take a second look at this function. We don't want any kind of stdio in `Document` methods. It also looks like you misspelled `iff`.

elijah-potter (approved):

harper-core/src/document.rs
```diff
@@ -267,7 +267,7 @@ impl Document {
         self.tokens().map(|token| token.to_fat(&self.source))
     }
 
-    /// Get the next or previous word token relative to a base index, iff separated by whitespace.
+    /// Get the next or previous word token relative to a base index, if separated by whitespace.
```
hippietrail, line 270: [Not a typo by the way (-:](https://en.wikipedia.org/wiki/If_and_only_if)

PR #1119, opened 2025-04-23
Add snapshot tests for token metadata (aka part-of-speech tags)

RunDevelopment:
# Description
This PR adds a new test to `harper-core` that creates snapshots for the token metadata of documents. I also exported the `Degree` enum (for adjective degrees), so that tests and users can actually do use the adjective degree in `WordMetadata`.

In my opinion, one of the best way to test large complex system is to just throw a bunch of data at them and see the result. While not exhaustive or scientific, this approach does tend to catch a lot of bugs if the dataset is large enough.

So to do this kind of testing, I added a new test which just parses a few documents and saves a snapshot of the parsed document's tokens + metadata. Metadata is encoded using a custom terse tag format inspired by the Penn Treebank tagset. The meaning of all tags is documented in the docs of the test, but to give a short summary:

- `N` = noun
- `I` = pronoun
- `V` = verb
- `J` = adjective
- `R` = adverb
- `C` = conjunction
- `D` = determiner
- `P` = preposition
- `.` = punctuation
- `#` = numbers
- Tags can have addition info appended to them. E.g. `NPrSg` mean noun, proper, singular.
- Words can have multiple tags. E.g. `N/V` means that a word is a noun and/or a verb.

I added 5 text documents:
- Alice's Adventures in Wonderland and The Great Gatsby.
- The US Constitution.
- Wikipedia articles about computer science and part-of-speech tagging.

All of these works are either public domain or permissively licensed.

Related to #1099.

# Demo

```md
> In          my younger and more        vulnerable years my father  gave me        some  advice that    I’ve
# NPrSg/V/J/P D  J       V/C NPrSg/I/V/J J          NPl   D  NPrSg/V V    NPrSg/ISg I/J/R NSg/V  N/I/C/D W?
> been  turning over      in          my mind  ever since .
# NSg/V NSg/V   NSg/V/J/P NPrSg/V/J/P D  NSg/V J    C/P   .
```

# How Has This Been Tested?
This *is* a test.

# Checklist

- [x] I have added tests to cover my changes (technically)

# Bugs this uncovered

1. `Document::articles_imply_nouns` doesn't work. This method is supposed to fix word metadata, but word metadata is only added AFTER `articles_imply_nouns` is called. So it can't do anything, since there isn't any metadata yet.

    This is a simple fix, but I'd recommend not fixing it yet. `articles_imply_nouns` in is current for has **many** false positives.

2. The tokenizer or the part that handles contractions can't deal with "’" (U+2019). E.g. "I’m", "I’ve", "She’s" aren't recognized as contractions.

3. A lot of words are assigned *empty* metadata (so roughly `Some(WordMetadata::default())`). E.g. "applied", "inclined", "also", "accused", "suggested", "there", "heightened", "attracted", "only".

   Use the regex `^#.*W\?` and you'll find thousands of them.

# Weird things and rough edges

4. Why is "an" a conjunction? "The", "A", and "An" are all prepositions and determiners, but "an" is also a conjunction. Why?
5. Why are conjunctions so many things are once? "And" is tagged as both verb and conjunction. "Or" is tagged as a proper singular noun, a preposition, and a conjunction. "But" is tagged as a proper singular noun, a verb, a preposition, and a conjunction. How can "and" be a verb?
6. Why are number words tagged as nouns? "One", "two", "three", etc. are tagged as nouns. They don't seem very nouny to me.
7. Adjective tags should probably be revised. E.g. "inevitable" is tagged as both a noun and an adjective. "Inevitable" is obviously an adjective, so my guess is that sentences like "You cannot stop the inevitable" caused problems. This should probably be fixed using a rule that turns adjectives into nouns depending on context.
8. A lot of states in `WordMetadata` seem to be unused. E.g. `NounData::is_proper: Option<bool>`, but I have not seen any nouns with `is_proper == Some(false)`. They all use `None` instead. Same for `is_possessive`.


hippietrail:
> # Weird things and rough edges
> 4. Why is "an" a conjunction? "The", "A", and "An" are all prepositions and determiners, but "an" is also a conjunction. Why?

I look all these up one by one as I run into them. I first check Wiktionary if need to I do a google of with `site:` for all of: Cambridge, Collins, Longman, OED, Oxford, and Webster which are the main professional dictionaries online. Often these oddities are obsolete, archaic, or dialectal. Sometimes they are total mistakes. Some are actually right in special circumstances. They surely came from importing from an over-inclusive source.

When I remove weird annotation flags for some of these reasons I put a comment in the dictionary in case somebody put it there or wants to add it back.

- "an" being a conjunction is archaic: https://en.wiktionary.org/wiki/an#Conjunction
- "an" is a preposition in "twenty miles an hour: etc.: https://en.wiktionary.org/wiki/an#Preposition
- "the" is a preposition like "an" but looks archaic to me: valued at half a pound the bushel; paying seven dollars the year interest https://en.wiktionary.org/wiki/the#Preposition

So `an.conj` and `the.prep` can definitely go but `an.prep` needs some thought.

> 5. Why are conjunctions so many things are once? "And" is tagged as both verb and conjunction. "Or" is tagged as a proper singular noun, a preposition, and a conjunction. "But" is tagged as a proper singular noun, a verb, a preposition, and a conjunction. How can "and" be a verb?

- "and" and "or" are verbs due to programming logic: anding a bitmask etc. but "and" is actually in Wiktionary not for that reason but because it has a dialectal sense: https://en.wiktionary.org/wiki/and#Verb
- "Or" is a surname. It should only get tagged as proper noun when capitalized, if it doesn't that's a bug and there are definitely bugs I've seen in the affix system where cases seem to be conflated at some level https://en.wiktionary.org/wiki/Or#Proper_noun
- "or" as a preposition is archaic: https://en.wiktionary.org/wiki/or#Preposition
- "But" is a surname but should only get the tag when capitalized https://en.wiktionary.org/wiki/But#Proper_noun
- "but" as a verb is archaic: https://en.wiktionary.org/wiki/but#verb
- "but" is a preposition in "I like everything but that" etc.: https://en.wiktionary.org/wiki/but#Preposition

> 6. Why are number words tagged as nouns? "One", "two", "three", etc. are tagged as nouns. They don't seem very nouny to me.

- "one" as a noun has half a dozen senses, computing is all ones and zeroes: https://en.wiktionary.org/wiki/one#Noun_2
- Similar for other digits/numerals

> 7. Adjective tags should probably be revised. E.g. "inevitable" is tagged as both a noun and an adjective. "Inevitable" is obviously an adjective, so my guess is that sentences like "You cannot stop the inevitable" caused problems. This should probably be fixed using a rule that turns adjectives into nouns depending on context.

I haven't run into this in testing yet but I have thought about it. I would explain this as something like such words being lexically adjectives that can become syntactically nouns. But note that you can't do it with arbitrary adjectives... I think. If it only applies to some adjectives it can get a new annotation flag at some point.

> 8. A lot of states in `WordMetadata` seem to be unused. E.g. `NounData::is_proper: Option<bool>`, but I have not seen any nouns with `is_proper == Some(false)`. They all use `None` instead. Same for `is_possessive`.

I think this is just an artefact of how the options are handled via JSON objects for practical reasons. I'm not sure if it needs to change or not.

Now I don't know where the dictionary data was imported from. When I started playing with this kind of stuff 20 years ago people used to import from a century-old public domain Websters. These days there are probably a bunch of scraped resources around the place. My go-to is Wiktionary but it's a good idea to get familiar with its flaws and quirks and treat everything as questionable but good faith. Its data isn't primarily intended for direct use in software but it's been slowly getting easier to parse over the years.

I think it was a wise move to use a "curated dictionary" and name it such. That means we can iron out these quirks and problems as we go. You have to start somewhere and some data sources will be missing too much and some will include too much stuff you have to weed out later.

I've been doing "dictionary curation PRs" for a few weeks now, mostly adding missing words, but also fixing weird stuff in the annotations, always after looking into each one first. I'll add all the easy ones you mention here to my next one.

RunDevelopment:
> "Or" is a surname. It should only get tagged as proper noun when capitalized

I think it might be good idea to do that for a lot of common words. Looking through the snapshots, `NPrSg` is **very** common.

> I haven't run into this in testing yet but I have thought about it. I would explain this as something like such words being lexically adjectives that can become syntactically nouns. But note that you can't do it with arbitrary adjectives... I think. If it only applies to some adjectives it can get a new annotation flag at some point.

Are there adjectives that don't work? I mean, I'm sure it's uncommon for some adjectives to be used as nouns, but it might not be wrong depending on context. Even words like "funny" are used as [nouns](https://en.wiktionary.org/wiki/funny#Noun) nowadays. E.g. "Streamer, do the funny."

I also asked ChatGPT and it gave some examples for adjectives that can't be used as nouns: happy, sad, tall, nice, kind, ineffective. But wiktionary lists their usage as nouns [for](https://en.wiktionary.org/wiki/happy#Noun) [the](https://en.wiktionary.org/wiki/sad#Noun) [first](https://en.wiktionary.org/wiki/tall#Noun) [four](https://en.wiktionary.org/wiki/nice#Noun), and the last can be used as nouns too. E.g. "Ben is a good person. He's one of the kind." and "My company only hires motivated, skilled, and fast-working people, not the lazy and inefficient." 
Common ChatGPT L.

Comparative and superlative also work. E.g. "When the rich become the richer and the poor become the poorest, something needs to change."



> > 8\. A lot of states in `WordMetadata` seem to be unused. E.g. `NounData::is_proper: Option<bool>`, but I have not seen any nouns with `is_proper == Some(false)`. They all use `None` instead. Same for `is_possessive`.
> 
> I think this is just an artefact of how the options are handled via JSON objects for practical reasons. I'm not sure if it needs to change or not.

Having simple bools would make the API simpler, and would allow for more optimizations.

Most `Pattern`s just check token kind and word metadata. So by compressing some of this information into bitsets, a lot of patterns can be replaced with a few bit-wise operations. This is huge, because it means that you don't need to call into the `Box<dyn Pattern>`, saving multiple indirections and the function call.

The more information from `WordMetadata` can be crammed into small bitsets, the more efficient pattern matching can become. And since 1 `bool` needs 1 bit and 1 `Option<bool>` needs 2 bits, I want to avoid `Option<bool>`s.

elijah-potter:
Both @hippietrail and @RunDevelopment have done some truly great work here. I'm a big fan of the approach for annotations in this PR and prefer it over #1099, particularly for the diffing ability.

> I think it was a wise move to use a "curated dictionary" and name it such. That means we can iron out these quirks and problems as we go. You have to start somewhere and some data sources will be missing too much and some will include too much stuff you have to weed out later.

This was precisely my reasoning. When the original version of the dictionary was imported, there were very few rules the used it. As more rules were added to support the functionality, curation became more important. 

> Most Patterns just check token kind and word metadata. So by compressing some of this information into bitsets, a lot of patterns can be replaced with a few bit-wise operations. This is huge, because it means that you don't need to call into the Box<dyn Pattern>, saving multiple indirections and the function call.
> The more information from WordMetadata can be crammed into small bitsets, the more efficient pattern matching can become. And since 1 bool needs 1 bit and 1 Option<bool> needs 2 bits, I want to avoid Option<bool>s.

This is something I've considered. The original versions of these actually implemented some of the ideas you're discussing here. 

There's just one problem: actual profiling of Harper shows that most of the runtime is dedicated to other things (particularly HashTable lookups). We're not at a point yet where bitset optimization is the most effective use of our time. There are still a good many linters that aren't using metadata exclusively.

> I've been doing "dictionary curation PRs" for a few weeks now, mostly adding missing words, but also fixing weird stuff in the annotations, always after looking into each one first. I'll add all the easy ones you mention here to my next one.

You've been doing great work here. I'd love to find a way to automate the process a bit more, but haven't found a method that's reliable enough.

hippietrail:
@RunDevelopment You may be interested in my tool I'm playing with to process Wiktionary dumps: [wikters](https://github.com/hippietrail/wikters) - I intend it more to be something people can hack on to extract info they need out of Wiktionary than a fully useable command-line tool. I've been processing Wiktionary dumps for well over a decade in various ways and using Wiktionary for over two. But I don't "trust" it (-;

RunDevelopment:
> There's just one problem: actual profiling of Harper shows that most of the runtime is dedicated to other things (particularly HashTable lookups).

Yeah... I did a quick and dirty test of this idea and I couldn't measure any difference. Maybe I just didn't take it far enough, but it does suggest that you're right and that other areas should be considered for optimization first. The root of all evil struck again.

> You may be interested in my tool I'm playing with to process Wiktionary dumps: [wikters](https://github.com/hippietrail/wikters)

Noted. I don't need it right now, but now I know it exists for later.

RunDevelopment:
@elijah-potter Is this ready to merge, or would you like me to change anything?

hippietrail:
> This looks good. I just wanted to make sure there wasn't any more discussion from @hippietrail. Merging now...

I'm excited to see it!

elijah-potter (approved):

PR #1120, opened 2025-04-23
No invalid `private = true` fields in `Cargo.toml`

RunDevelopment:
# Description
When running `cargo test`, I got warnings about these invalid fields in `Cargo.toml`. So I removed them.

If they are used somewhere, please tell me, and I'll add a comment to these files explaining why they contain invalid fields instead.

# How Has This Been Tested?
N/A

# Checklist

- [x] I have added tests to cover my changes (N/A)


elijah-potter:
These fields are not in the official `cargo` spec, but we use [`cargo workspaces`](https://github.com/pksunkara/cargo-workspaces). Setting `private = true` stops `cargo workspaces` from publishing those crates to `crates.io`.

elijah-potter (approved):

PR #1124, opened 2025-04-24
Use `Option<NonZeroUsize>` as the return type of `Pattern::matches`

RunDevelopment:
# Description
I changed the return type of `Pattern::matches` from `usize` to `Option<NonZeroUsize>`.

Previously, a return value of 0 meant "no match," but this was not obvious from the signature of the function. So I changed it to `Option<NonZeroUsize>`. This makes it very obvious that (1) a match always contains at least one token and (2) that matching can fail.

As a nice bonus, using `Option` enables the use of `?` and Rust usual APIs around `Option`. This makes the implementations of most patterns simpler.

## Other minor changes:

- I added documentation to `Pattern::matches`.
- I used `cfg_attr` on the `Pattern` trait to avoid copying the trait definition.

# How Has This Been Tested?
This only changes an existing API without adding or removing functionality. As such, I didn't add any new test since the existing tests already cover the expected behavior.

# Checklist

- [x] I have added tests to cover my changes (Since I only modified an existing API, the existing tests suffice.)


RunDevelopment:
> It looks like you're going to have to run `just format`.

On it. I was just occupied by something else.

> Would you also mind following the commit conventions discussed on [this page](https://writewithharper.com/docs/contributors/committing)?

Sorry, I didn't see that. I'll follow it.

On that note. Any reason why you aren't squashing commits? In my repos, I always squash commits, because I want a clean commit history on `main` with one PR being one feature.

elijah-potter:
> On that note. Any reason why you aren't squashing commits? In my repos, I always squash commits, because I want a clean commit history on main with one PR being one feature.

I've personally never seen the utility, so it's not something I have an opinion on. If we squashed moving forward, would that make things easier for you?

elijah-potter (approved):
This all looks good to me! Great idea.

RunDevelopment:
> I've personally never seen the utility, so it's not something I have an opinion on. If we squashed moving forward, would that make things easier for you?

From my experience, squash & merge reduces friction. It gives you complete control over the (single) commit made to master, so PR authors don't have to worry about their commit history (it just needs to be easy to review). It also keeps the history of master clean. During review, seeing all commits is useful, but nobody needs to know that it took me 3 tries to make CI pass after the PR is merged.

At the end of the day, this is a workflow question, so do what works best for you. I found that squash and merge works best for me (when in the role of maintainer), but you might feel otherwise.

Either way, formatting my commit messages a certain way is not a big deal, so I'm going to play by your rules. (Or at least, I want to. Sorry again...)

elijah-potter:
> From my experience, squash & merge reduces friction. It gives you complete control over the (single) commit made to master, so PR authors don't have to worry about their commit history (it just needs to be easy to review). It also keeps the history of master clean. During review, seeing all commits is useful, but nobody needs to know that it took me 3 tries to make CI pass after the PR is merged.

You've convinced me. I've changed the merge queue to squash from now on.

> Either way, formatting my commit messages a certain way is not a big deal, so I'm going to play by your rules. (Or at least, I want to. Sorry again...)

Absolutely no worries. In case it isn't clear, this is not something I'm super worried about.

PR #1126, opened 2025-04-24
Simplified `Pattern::find_all_matches`

RunDevelopment:
# Description
This PR simplifies the implementation of `Pattern::find_all_matches` by removing the code to remove overlaps. It's a lot easier to just avoid overlaps to begin with, instead of removing them after the fact.

This also fixes a small bug in the removed overlap code. If the pattern found overlapping patterns like this:

```
[ m1 ]
   [ m2 ]
      [ m3 ]
```
(m1 overlaps m2 overlaps m3, but m1 and m3 don't overlap)

The old code removed m2 and m3, despite m1 and m3 not overlapping.

If this behavior was intentional, please tell me and I'll restore this behavior without allocation a `VecDeque`.

# How Has This Been Tested?
Existing tests verify the behavior.

# Checklist


elijah-potter (dismissed):
Good stuff! This is way more efficient. Thanks for tackling it.

Issue #408, opened 2025-01-17
Flag "how" used together with "look like"


hippietrail:
This is very common among non-native English speakers whose first language is a European language:

- How does it look like? → What does it look like? / How does it look?
- This is how it looks like. → This is what it looks like. / This is how it looks.

hippietrail:
Implemented in #1146 for simple but common cases involving the pronouns it's most often seen with.

PR #1146, opened 2025-04-26
feat: how it looks like→how it looks/what it looks like

hippietrail:
# Issues 

Partial solution of #408.

# Description

It only handles the cases with pronouns, not the full range of possibilities which can include arbitrary nominal phrases.

As a bonus it also fixes cases where the verb looks doesn't agree with a plural pronoun. I doubt anyone says "how he look like"...

Actually I checked and people do use "how it look like"... so let me include that too.

It doesn't handle other cases of two or more mistakes together that I noticed:
- here is screenshot of how it's look like
- this is how it look's like with the fresh downloaded

# How Has This Been Tested?

I added a unit test for each case using a sentence found on GitHub, and a 2nd unit test for most of them using the second suggestion.

# Checklist


hippietrail:
> Would it be possible to refactor this into a `PatternLinter` to match against all pronouns?

I put a fair bit of thought and research into that, did lots of searching for examples, and other pronouns seem to be far down the list, and I was worried edge cases and false positives might be introduced.

Much more common that the other pronouns are various nouns and then nominal phrases, but those will need a fair bit of effort and debugging because they will definitely have lots of edge cases etc. My thought was to get an initial implementation out that can be iterated upon that takes care of the low-hanging fruit that are both really common and not likely to produce too many false positives.

elijah-potter (dismissed):

elijah-potter (approved):

PR #1149, opened 2025-04-26
refactor: Rename `ACO` to `Word`

RunDevelopment:
# Description
I renamed `AnyCapitalization` to `Word` to be consistent with `WordSet`. I also changed the API to be more similar to `WordSet`. There is now a nice symmetry between `Word` and `WordSet`.

I also added `Word::new_exact` for exact matches. This now powers the implementation of `SequencePattern::then_exact_word`.

# How Has This Been Tested?
Tested by existing rules and new tests.

# Checklist


elijah-potter (approved):
Great stuff. Thanks for polishing that up!

Issue #1138, opened 2025-04-25
New API for patterns


RunDevelopment:
@elijah-potter

I'm thinking about making a rather large change to the `Pattern` API, so I wanted to discuss this beforehand. Please tell me what you think.

## Motivation

I have worked on a [regex library](https://github.com/RunDevelopment/refa), a [regex linter](https://github.com/ota-meshi/eslint-plugin-regexp), and a [type system](https://github.com/chaiNNer-org/Navi) before. Because of that, I have a bit of experience with ASTs, state machines, and representing regex-like patterns in code.

Harper `Pattern`s very similar to that. There are terminals (e.g. `WordSet`, `Fn(&Token, &[char])`) and combiners (e.g. `SequencePattern`, `EitherPattern`, `Invert`).

I don't like how patterns are currently represented in code. 

1. Combiners are very verbose.
2. There are multiple ways to do one thing.
3. Some functionality is only available on `SequencePattern`
4. In general, everything is very centered around `SequencePattern`

I would like to change that.

## Description

What I want is a terse syntax without any god-classes like `SequencePattern`.

Here are a few examples of the syntax I want (all "old" code is currently used):

```rs
// old 
let pattern = SequencePattern::aco("here")
    .then_whitespace()
    .t_aco("by")
    .then_whitespace()
    .then_verb();

// new
let pattern = Sequence::new(("here", WS, "by", WS, Verb));
```

<details>
<summary>More examples</summary>

```rs
// old
let pattern = SequencePattern::default()
    .then_number()
    .then_whitespace()
    .t_aco("day")
    .then(EitherPattern::new(vec![
        Box::new(
            SequencePattern::default()
                .then_whitespace()
                .then(NominalPhrase),
        ),
        Box::new(
            SequencePattern::default()
                .then_hyphen()
                .then_adjective()
                .then_whitespace()
                .then(NominalPhrase),
        ),
    ]));
// new
let pattern = Sequence::new((
    Number,
    WS,
    "day",
    Choice::new((
        Sequence::new((WS, NominalPhrase)),
        Sequence::new((Hyphen, Adj, WS, NominalPhrase)),
    )),
));

// old
let the_how = SequencePattern::default()
    .t_aco("the")
    .then_whitespace()
    .t_aco("how")
    .then(Invert::new(
        SequencePattern::default().then_whitespace().t_aco("to"),
    ));

let the_who = SequencePattern::default()
    .t_aco("the")
    .then_whitespace()
    .t_aco("who")
[...]

PR #1150, opened 2025-04-26
feat: new pattern syntax

RunDevelopment:
# Issues 
#1138

# Description
This PR adds the basics of the basics for the new pattern API I proposed in #1138. This includes the necessary traits, `Sequence`, `Choice`, and a few helpers.

I also updated 2 existing lints to use the new pattern syntax to show it off.

This is still is a work in progress.

# How Has This Been Tested?
Hasn't been yet.

# Checklist

### DIFF
--- a/harper-core/src/patterns/new_syntax_experiment.rs
+++ b/harper-core/src/patterns/new_syntax_experiment.rs
@@ -1,0 +1,170 @@
+use std::num::NonZeroUsize;
+
+use crate::{CharString, Token};
+
+use super::{AnyCapitalization, AnyPattern, Pattern, WhitespacePattern};
+
+pub trait IntoPattern {
+    type Output: Pattern + 'static;
+    fn into_pattern(self) -> Self::Output;
+}
+impl<T: Pattern + 'static> IntoPattern for T {
+    type Output = T;
+    fn into_pattern(self) -> Self::Output {
+        self
+    }
+}
+impl IntoPattern for &str {
+    type Output = AnyCapitalization;
+    fn into_pattern(self) -> Self::Output {
+        AnyCapitalization::of(self)
+    }
+}
+
+pub trait IntoPatternList {
+    fn into_pattern_list(self) -> Vec<Box<dyn Pattern>>;
+}
+impl IntoPatternList for Vec<Box<dyn Pattern>> {
+    #[inline]
+    fn into_pattern_list(self) -> Vec<Box<dyn Pattern>> {
+        self
+    }
+}
+impl IntoPatternList for () {
+    #[inline]
+    fn into_pattern_list(self) -> Vec<Box<dyn Pattern>> {
+        Vec::new()
+    }
+}
+impl<P: IntoPattern> IntoPatternList for P {
+    #[inline]
+    fn into_pattern_list(self) -> Vec<Box<dyn Pattern>> {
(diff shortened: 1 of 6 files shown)


harper-core/src/patterns/new_syntax_experiment.rs
```diff
@@ -133,5 +179,38 @@ pub mod prelude {
...
+    /// An assertion that matches anything but the given sequence of patterns.
+    /// No tokens are consumed.
+    ///
+    /// ## Examples
+    ///
+    /// ```rust
+    /// use crate::patterns::new_syntax_experiment::preluse::*;
+    /// let love_no_ego = seq!["I", WS, "love", not_ahead![WS, "myself"]];
+    /// ```
+    macro_rules! not_ahead {
```
hippietrail, line 205: This might not be the best name as it suggests any word up ahead rather than the very next word.

RunDevelopment: Sorry, but do you mean by "any word up ahead rather than the very next word"? Are you trying to say that people will read `not_ahead![A, B]` as `not_ahead![choice![A, B]]` instead of `not_ahead![seq![A, B]]`?

hippietrail: I mean it could be misinterpreted as meaning "if the rest of the sentence/phrase/chunk does not contain <word>".

RunDevelopment: Ah, so you mean that people could think that there could be any number of tokens between `"love"` and `WS, "myself"` in the above example, right?

If so, then I guess. They are called lookahead in regex-land, so I named them after that. If you have suggestions for a different name, please share them. I'm open for bikeshedding. I couldn't think of anything better that is also:

1. Short.
2. Possible to negate (lookaheads and negated lookaheads).
3. Has an opposite we can later use for lookbehinds and negated lookbehinds.

That's why I went with "ahead" and (later) "behind".

hippietrail: > Ah, so you mean that people could think that there could be any number of tokens between `"love"` and `WS, "myself"` in the above example, right?
> 
> If so, then I guess. They are called lookahead in regex-land, so I named them after that. If you have suggestions for a different name, please share them. I'm open for bikeshedding. I couldn't think of anything better that is also:

Yes I realized that, but often I do wish I could look further ahead or back, especially for trying to disambiguate a word. But nothing of that nature exists yet and would likely be at a different level... But I think we'll attract users who don't know regex terminology.

To me the obvious would be `not_next` so I assumed you ruled that out for a reason.

> 1. Short.
> 2. Possible to negate (lookaheads and negated lookaheads).
> 3. Has an opposite we can later use for lookbehinds and negated lookbehinds.
> 
> That's why I went with "ahead" and (later) "behind".

They're good reasons of course!

RunDevelopment: I didn't think of `next` and `prev`. That could work!

```rs
let love_no_ego = seq!["I", WS, "love", not_next![WS, "myself"]];
let might_no_noun = seq![not_prev![Det | Adj, WS], "might"];
```

elijah-potter:
Any updates on this? I'd love to get something like this across the finish line. If not, I totally understand. Thanks for everything you've done!

PR #1152, opened 2025-04-27
Spanish skeleton for feedback

hippietrail:
# Issues 

# Description

This PR adds multiple human language support, starting with Spanish.

English remains built-in, other languages are attempted to load from disk.

`dictionary.dict` and `affixes.json` are for the built-in English. The new Spanish skeletons are `dictionary-es.dict` and `affices-es.json`

I've chosen the same clunky name for the human language argument/parameter/member/getter through to avoid any possibility of ambiguity with "language" where we mean programming language/file format/markup language etc. That name is currently `langiso639`, easy to grep for.

I tried to avoid defaulting anything to `en` except what is clearly English-specific. This got tricky in places due to "language" being used to refer to programming languages etc, "English" being used to check document language against dictionary language, and edges that was too unfamiliar with.

I started at the dictionary loading and worked my way out to the edges. By the time I got to the interfaces with the LSP and harper.js there was a lot more I wasn't familiar with and more things that demanded defaults.

In a few places where a dictionary or some other item needed to be instantiated that was unclear I used different fake language codes with the word `fake` in them so they'll be easy to identify and find.

In a couple of places that are clearly not using a language I pass a fake language tag `__` (double underscore). I pass `en` where comments or typenames or string literals contain English, include "English" as part of the identifier, or where a Dialect field was already hard-coded to "American".

Test and lints are intended to be in parallel files or directories mimicking the English ones but with `-es` or `_es` or `Es` appended. I renamed one type by appending `En`.

I currently create no new linters for Spanish and only one file with a test or two mimicking low-level English ones.

`cargo test` runs with no errors. `cargo clippy` makes no complaints.

I haven't tried `just test` or `just precommit` yet. I expect failures with integrations since I'm very unfamiliar with those parts of the codebase.

Unfortunately I had to modify many lines in many files to propagate the language.

# How Has This Been Tested?

`cargo test` and `clippy` and one new Spanish-specific test file with one or two simple tests.

# Checklist

# Feedback

Please give me feedback! Especially about the design decisions and the areas where I use "fake" language codes or default to "en" and if I've missed some ways to avoid having to explicitly pass language codes around as much.


mu22le:
Hi @hippietrail , what happened to this PR?

hippietrail:
> Hi @hippietrail , what happened to this PR?

It was just for feedback on my approach but nobody offered any in five months. It was never going to be *the* solution. @elijah-potter knows the codebase much better than me and has said it wouldn't be hard to implement when the time is right.

PR #1153, opened 2025-04-27
feat: zero-width patterns

RunDevelopment:
# Description
This PR adds support for zero-width patterns. I need this to implement lookaheads, lookbehinds, and optional patterns (think regex `?`) for #1150. 

Implementation-wise, I just changed the return type of `Pattern:matches` from `Option<NonZeroUsize>` to `Option<usize>`.

# How Has This Been Tested?
Existing lints serve as tests

# Checklist


elijah-potter (approved):
This is a pretty straightforward but likely tedious change. Thanks for doing it!

RunDevelopment:
I resolved all conflicts.

harper-core/affixes.json
```diff
@@ -417,7 +417,9 @@
 				}
 			],
 			"target_metadata": {
-				"verb": {},
+				"verb": {
+					"tense": "Present"
```
hippietrail: "Tense" is a syntactic property once words are in sentences and the sentences have been parsed. Words like "run", "walk", "eat" can be future if they come after "will", "shall", "going to", "gonna" or infinitive if they come after "to", etc.

The dictionary and affix system are lexical rather than syntactic. Terms used for this form of verbs are "dictionary form", "citation form", "lemma" - these three apply to other inflected parts of speech too. For verbs specifically "infinitive", literally "no time", "not inflected for tense".

RunDevelopment: How did you make this comment here? Did GitHub mess up?

harper-core/dictionary.dict
```diff
@@ -52867,8 +52867,62 @@ write-off/1MS
 write-up/1MS
 year-over-year/5
 
+# Phrasal verbs. These help correct the error where people write "to login" instead of "to log in"
+# For this reason the most important ones to include are the ones that have a noun counterpart.
+# Remember not to add any affixes since only the verb part at the start undergoes inflection.
+
+back up/4
```
hippietrail: I've been thinking about having a special annotation flag for these. It would already be useful in linters. The one I worked on weeks ago dealing with the "backup"/"back up" problem I think finally got merged today and it would benefit already.

But I wasn't sure if the multi-word part of the dictionary would be accepted. I guess it has been. But that's why I'd put them in a separate part of the dictionary. To make best use we would really want some kind of multi-word awareness in parsing and spell checking so I hadn't pushed it any further yet.

elijah-potter (approved):

PR #1154, opened 2025-04-27
feat:one in the same→one and the same

hippietrail:
# Issues 
N/A

# Description

This is a common mistake I thought I'd already done a Linter for.
"One in the same", which was in a YouTube video a couple of hours ago, should be "One and the same".

# How Has This Been Tested?

I made atomic unit tests for the basic cases and edge cases to avoid.
I made unit tests from real sentences from GitHub for both flagging actual misuses and not flagging when it's not a mistake.

# Checklist


harper-core/src/linting/one_and_the_same.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn corrects_before_as_github() {
+        assert_suggestion_result(
+            "In our case the slicedState is one in the same as the featureState",
+            OneAndTheSame::default(),
+            "In our case the slicedState is one and the same as the featureState",
+        );
+    }
+
+    // TODO needs zero-width end-of-chunk pattern akin to regex `$`
+    // #[test]
```
elijah-potter, line 137: Would you disable these tests instead of commenting them out?

hippietrail: > Would you disable these tests instead of commenting them out?

Oh I had grepped for all of those when I found out about `#[ignore]` but I must've missed some...

PR #1155, opened 2025-04-27
feat: iterator API for pattern matches

RunDevelopment:
# Description
I added an `iter_matches` method to patterns which returns an `impl Iterator<Item = Span>`. It works exactly as `find_all_matches`, but without allocating a `Vec<Span>`.

I did this, because I noticed that 3 linters re-implemented the logic from `find_all_matches`. Using the iterator API, everything can now use the same implementation.

# How Has This Been Tested?
It's tested by all existing tests.

# Checklist


elijah-potter (approved):
Good eye! Thanks for working to make our code a bit cleaner. Keep 'em coming.

PR #1158, opened 2025-04-28
Fix false positive in let/lets→let's

### LINKED ISSUE #1157: "Let us" incorrectly flagged when followed by an adverb
## Description
`harper-cli 0.1.0`
`harper-ls 0.31.0`

The unexpanded form "Let us" is incorrectly flagged as missing the 'us' when the two-word phrase is immediately followed by an adverb like 'then' or 'now', as in the example sentence

"Let us then have a good time."

"Let" is flagged, suggesting a correction "Let us", even though the 'us' is indeed present.
The following screenshot shows a few similar sentences. The red marks demonstrate this incorrect flagging.

## Screenshots

Image

hippietrail:
# Issues 

Fixes #1157

# Description

It turned out not to be due to being followed by an adverb.

Rather it's a symptom of `us` getting tagged as a noun as well as a pronoun due to something in the harper dictionary/affix system conflating `us` and `US`, which is a noun.

I added a special case to the part that checks if the word after "let" is a noun.

# How Has This Been Tested?

I added a new unit test which failed the previous version.

# Checklist

### DIFF
--- a/harper-core/src/linting/lets_confusion/no_contraction_with_verb.rs
+++ b/harper-core/src/linting/lets_confusion/no_contraction_with_verb.rs
@@ -17,11 +17,13 @@ pub struct NoContractionWithVerb {
 
 impl Default for NoContractionWithVerb {
     fn default() -> Self {
+        // Only tests "let".
         let let_ws = SequencePattern::default()
             .then(WordSet::new(&["lets", "let"]))
             .then_whitespace();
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
I'll be tackling a more permanent solution to the WordId problem once the Chrome Extension is online.

PR #1159, opened 2025-04-28
feat: minimal, minutely documented example of using Harper from C

hippietrail:
# Issues 
No issues but the idea was briefly discussed on Discord.

# Description

Provides a copiously commented but very minimal example of how to write a C program using Harper's Rust library.

Includes:

- `ffi.rs` file showing Rust code needed to provide C-style access to the Rust functions and types.
- `harper.h` showing how to make a C header file for the `ffi.rs` Rust file.
- `florb.c` showing how to use some of the simpler Harper APIs. Named after the user in Discord who inspired this.
- A minimal `Cargo.toml` file

The shell commands needed to build:
```
cargo build --release
cc src/florb.c -L target/release -lharper_c -o florb
./florb
```

# Demo

image

# How Has This Been Tested?

Only manual testing in the command-line.

# Checklist

# Feedback

This is an area of Rust I haven't touched before. I may have got things wrong, such as:
- Files not minimally needed
- Stuff in any file that's not needed
- Wrong or misleading comments in any file
- Poor directory structure

Please let me know!


HobbitJack:
In support of this, I have implemented using this interface a small C program to do CLI grammar checking:
https://github.com/HobbitJack/grammar/

elijah-potter:
I am not an expert on this type of project _at all_, so my input would be misleading at best. 

The bigger issue at hand is what purpose this patch holds. Are you trying to build a C library for consumption from other languages, or are you trying to build an example. If it's the latter, I think it would be better placed in its own repository, rather than here.

PR #1160, opened 2025-04-28
ci: add `cargo hack`

RunDevelopment:
# Description
Just like how I forget to update a `cfg`-disabled block, which I fixed in #1140, I forgot this again in #1153. And again, the CI did not catch this. So in this PR, I'm adding [`cargo hack`](https://github.com/taiki-e/cargo-hack). This will check that all crates compile with all features enabled and disabled. So silly mistakes like #1140 won't ever make it into `master` again.

# How Has This Been Tested?
This pretty much is a test.

# Checklist


elijah-potter:
Would it be possible to add this as part of `just precommit`?

RunDevelopment:
> Would it be possible to add this as part of `just precommit`?

If you can tell me [how you want it to be installed](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#installation), yes.

> It should be as easy as moving your stuff from the workflow to the `justfile`.

I intentionally made this a separate workflow, because I think the current way CI is done via `precommit` is pretty horrible. I literally have to wait half an hour for the CI to tell me that one test didn't pass. Adding this `recommit` would probably add another minute. Usually, I consider CI too slow, when it takes more than two minutes. Running the justfile locally is out of the question for me, because I'm on Windows.

So I made a separate workflow, so people making PRs can quickly see whether this check passed or not.

hippietrail:
> the justfile locally is out of the question for me, because I'm on Windows.

Does `just` totally not support Windows? You can't use `wsl`? I've been carrying my old Intel MacBook Air around for ages intending to install Windows on it just because I'm passionate about cross-platform.

RunDevelopment:
> Does `just` totally not support Windows?

Just does support windows, but the scripts inside this repo's justfile mostly require bash and some use unix-specific commands.

> You can't use `wsl`?

Maybe. Haven't used it in ages.

That said, me not using `just` is somewhat orthogonal to the reasons I didn't include `cargo hack` into `justfile`.

elijah-potter:
We can keep this as a separate workflow, but I would like to have it be part of the `precommit` step as well. 

> If you can tell me [how you want it to be installed](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file&rgh-link-date=2025-04-29T08%3A32%3A45Z#installation), yes.

It would be sufficient for you to add it to the [environment set up](https://writewithharper.com/docs/contributors/environment) documentation.

It's in `packages/web/src/routes/docs/contributors/environment/+page.md`.

RunDevelopment:
Alrighty. Should I still keep as a separate workflow or just as a part of `precommit`?

PR #1162, opened 2025-04-28
fix(core): don't lex the last period as part of a number

### LINKED ISSUE #1010: Paragraph detected as Single Long Sentence
## Description

Harper detects a paragraph that is made of multiple sentence as one long sentence. 

From the SS we can see the sentences are not 50 words long and the the paragraph consists of multiple sentences.

I took a look at multiple notes in my Vault but this paragraph seems to be the only one that has this problem.

It is shown as a Readability issue.

## Screenshots

Image

Image

elijah-potter:
# Issues 

Fixes #1010

# Description

I don't think it makes sense in pretty much any context to parse a terminating dot as part of a number.

For example, the sentence: `The I like the Switch 3.` 

# How Has This Been Tested?

Additional unit tests.

# Checklist

### DIFF
--- a/harper-core/src/lexing/mod.rs
+++ b/harper-core/src/lexing/mod.rs
@@ -84,15 +84,17 @@ pub fn lex_number(source: &[char]) -> Option<FoundToken> {
         if let Ok(n) = s.parse::<f64>() {
             let precision = s.chars().rev().position(|c| c == '.').unwrap_or_default();
 
-            return Some(FoundToken {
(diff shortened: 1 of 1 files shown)


hippietrail:
> I don't think it makes sense in pretty much any context to parse a terminating dot as part of a number.
> 
> For example, the sentence: `The I like the Switch 3.`

Yeah the only place I can think of seeing it is in shader code, and even then it would have to be commented out.

PR #1131, opened 2025-04-25
feat: abanonedware→abandonware

hippietrail:
# Issues 
N/A

# Description

Corrects abanonedware→abandonware
Spellcheck does not suggest the correct word.

# Demo

# How Has This Been Tested?

I made a unit test based on the first sentence I found using it in GitHub.

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests to cover my changes

PR #1166, opened 2025-04-28
fix(core): allow erroneous `shift` usage in spellcheck

elijah-potter:
# Issues 

For some reason this allows the spellcheck to properly handle #1131 without a special case. 

# Description

We give preference to spelling suggestions that start with the same letter. When this was originally implemented, it didn't handle uppercase versions of the same letter. This PR instructs Harper to give preference regardless of case.

# How Has This Been Tested?

# Checklist


hippietrail:
That's great! I'd really like to include a unit test though to ensure we keep suggesting it. Is there a recommended place for such 'random' tests? I'll have a quick look and edit an answer in if the answer is easy...

PR #1168, opened 2025-04-29
Add `core_version` to `harper-core`

HobbitJack:
# Description
This commit adds a method `core_version()` to `harper-core`, and also provides a `core-version` command in `harper-cli`.
This can then be used to get the current version of the core library while debugging, and is also useful for e.g. the C library, to be able to compare against the version of the core library (and, for example, abort if it's too old, or if there is some breaking change down the line to the core, to be able to handle that).

I will note this is my first _ever_ time working with Rust, so it is likely the way I did this isn't exactly idiomatic.
However, the tests seem happy, and it seems to work, so might as well start somewhere, right?

# Demo
image

# How Has This Been Tested?
`core-version` has been added to `harper-cli`. C.f. above, and it can be verified this is producing the same version as in `Cargo.toml`. 

# Checklist


HobbitJack:
looks like I forgot to squash a bunch of merge commits... whoops...

elijah-potter (approved):
This is a simple enough change, and I from what I've seen, it is idiomatic. Good work for your first time!

PR #1178, opened 2025-04-29
feat(harper-core/linting): add expand pull request phrase correction

uncenter:
# Issues 
n/a

# Description

Adds a phrase correction rule that is similar to the other "expand" rules present in the file (such as for expanding dep/deps) to correct "pr" or "PR" to "pull request" for clarity.

# Demo
n/a
# How Has This Been Tested?

Unit test (currently failing). I'm not sure how I can, with the current setup, have the correction *not* preserve the casing of the original text. The test fails because "PR" becomes "PULL REQUEST" instead of "pull request". Would appreciate feedback/guidance on how to make that happen.

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_corrections.rs
+++ b/harper-core/src/linting/phrase_corrections.rs
@@ -1177,6 +1177,12 @@ pub fn lint_group() -> LintGroup {
             "The correct term is `as well` with a space.",
             "Corrects `aswell`, which should be written as two words."
         ),
+        "ExpandPullRequest" => (
+            ["pr"],
+            ["pull request"],
+            "Use `pull request` instead of `pr`",
+            "Expands the abbreviation `pr` to `pull request` for clarity."
+        ),
     });
 
     group.set_all_rules_to(Some(true));
--- a/harper-core/src/linting/phrase_corrections.rs
+++ b/harper-core/src/linting/phrase_corrections.rs
@@ -2506,4 +2506,13 @@ mod tests {
             "format Cargo.toml as well #5893 - rust-lang/rustfmt",
         );
     }
+
+    #[test]
+    fn expand_pull_request() {
+        assert_suggestion_result(
+            "I'm implementing the feature in a PR.",
+            lint_group(),
+            "I'm implementing the feature in a pull request.",
+        );
+    }
 }


elijah-potter:
This is a rule that might be better implemented as a `PatternLinter`. If I were you, I'd make a copy of `expand_time_shorthands` and modify it.

hippietrail:
PR is also "public relations".

uncenter:
> PR is also "public relations".

Good point. I don't see a way to differentiate on context, so this doesn't seem feasible. Sorry for the bother 😅

hippietrail:
> > PR is also "public relations".
> 
> Good point. I don't see a way to differentiate on context, so this doesn't seem feasible. Sorry for the bother 😅

No no - phrase corrections accepts lists both for matching and suggestions! You can include both.

uncenter:
What should I call the lint rule to make it more generic than `ExpandPullRuquest` - `ExpandPR`? Would it make sense to make a new `acronym_expansions.rs` lint category maybe, which could solve what I mentioned earlier with the casing?

hippietrail:
> What should I call the lint rule to make it more generic than `ExpandPullRuquest` - `ExpandPR`? Would it make sense to make a new `acronym_expansions.rs` lint category maybe, which could solve what I mentioned earlier with the casing?

I would name it `ExpandPR`.

I was thinking about a general abbreviation and acronym expansion feature. I'm just a bit worried it could end up getting jammed with everything. Look at the Wikipedia pages listing them for instance. I would recommend to file a discussion or feature request to get feedback on the idea.

uncenter:
Gotcha. How do you think I can have it not preserve the original casing? For acronyms in general you'd want the expanded version to be in either sentence/title case and not all uppercase like it seems to do now.

hippietrail:
> Gotcha. How do you think I can have it not preserve the original casing? For acronyms in general you'd want the expanded version to be in either sentence/title case and not all uppercase like it seems to do now.

Oh wow yeah I didn't even notice that! Separate bug report (-:

uncenter:
Would you be willing to file it? I'm not sure I understand the internals of lint rules enough to understand it well 😅

hippietrail:
> Would you be willing to file it? I'm not sure I understand the internals of lint rules enough to understand it well 😅

I don't understand them perfectly either yet (-: Just report what happened vs what you expected to happen. I haven't struck this problem where you seem to be saying that when it offers to expand an abbreviation if the abbreviation is all caps then the expansion is all caps? If I understand correctly I tried and can't reproduce it. So you need to list the steps to reproduce, which is basically a bug report anyway.

uncenter:
See the failing test for this pull request? That is reproducible on both my machine and the CI. That's the example I'm referring to.

hippietrail:
> See the failing test for this pull request? That is reproducible on both my machine and the CI. That's the example I'm referring to.

Sorry I'm so bad at git and github - I'll try...

uncenter:
Ah, I think I found where this comes in:

https://github.com/Automattic/harper/blob/47aa8495f0949661e1fd2c5cde2ff9c1e9c1f404/harper-core/src/linting/map_phrase_linter.rs#L101

We make a `MapPhraseLinter` for each entry in `phrase_corrections.rs`, and the `PatternLint::match_to_lint` method for `MapPhraseLinter` uses [`Suggestion::replace_with_match_case`](https://github.com/Automattic/harper/blob/47aa8495f0949661e1fd2c5cde2ff9c1e9c1f404/harper-core/src/linting/suggestion.rs#L30) to match and presumably apply.

hippietrail:
> Ah, I think I found where this comes in:
> 
> https://github.com/Automattic/harper/blob/47aa8495f0949661e1fd2c5cde2ff9c1e9c1f404/harper-core/src/linting/map_phrase_linter.rs#L101
> 
> We make a `MapPhraseLinter` for each entry in `phrase_corrections.rs`, and the `PatternLint::match_to_lint` method for `MapPhraseLinter` uses [`Suggestion::replace_with_match_case`](https://github.com/Automattic/harper/blob/47aa8495f0949661e1fd2c5cde2ff9c1e9c1f404/harper-core/src/linting/suggestion.rs#L30) to match and presumably apply.

Yes `replace_with_match_case` was there before but it was buggy so I think somebody fixed it.
I think this shows that `phrase_corrections` is not suitable for acronym expansions in the same way it is for abbreviation expansions for this reason. It's also not flexible. Elijah made the right call back up here: https://github.com/Automattic/harper/pull/1178#issuecomment-2840217135

elijah-potter:
Since we haven't seen any activity in a while, I'm going to close this.

PR #1182, opened 2025-04-30
feat(core): avoid providing multiple lints for a single long hyphen sequence in `Dashes`

86xsk:
# Description

This PR modifies the `Dashes` linter to prevent it from suggesting multiple lints for a single long sequence of hyphens. Previously, if you had a line of 60 hyphens, you would get 20 lints suggesting you replace each triple hyphen with an em dash. With these changes, only a single lint will be provided (encompassing the whole sequence).

While working on these changes, I noticed it can be very difficult if not impossible to visually discern between en and em dashes in some contexts. To reduce the potential for confusion, I extracted the dash literals into module-level `const`s. This did require making some of the tests more verbose with `format!`s though, so there is a tradeoff. I'm certainly open to undoing or modifying this "const extraction" if it's deemed to be detrimental.

# How Has This Been Tested?
- `cargo test`
- Briefly testing the modified build of `harper-ls` in a code editor.

# Checklist

### DIFF
--- a/harper-core/src/linting/dashes.rs
+++ b/harper-core/src/linting/dashes.rs
@@ -5,6 +5,9 @@ use crate::{
 
 use super::{Lint, LintKind, PatternLinter, Suggestion};
 
+const EN_DASH: char = '–';
+const EM_DASH: char = '—';
+
 pub struct Dashes {
     pattern: Box<dyn Pattern>,
 }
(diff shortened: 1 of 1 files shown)


hippietrail (commented):
This looks good to me.
I would suggest though that two or three hyphens in a row are not even a mistake but a convention followed by many tools since en dash and em dash are not always available on keyboard. And as you note, look alike in source code monospaced fonts.
Long strings of hyphens are not usually used to represent long strings of en or em dashes but some kind of ASCII/ANSI format such as an underline, horizontal rule, or part of a table diagram.

I think it makes sense to just not even flag long sequences of hyphens.

86xsk:
> Long strings of hyphens are not usually used to represent long strings of en or em dashes but some kind of ASCII/ANSI format such as an underline, horizontal rule, or part of a table diagram.
>
> I think it makes sense to just not even flag long sequences of hyphens.

Agreed. I sometimes use such long strings of hyphens to separate sections in plain text files, which is what let me notice this issue in the first place. I was hesitant to make such a change at first since I wasn't sure if handling such cases is the responsibility of the linter, or if it was on the parser to filter them out.

I pushed a commit to avoid providing any `Dashes` lint for hyphen sequences longer than 3 hyphens.

elijah-potter (approved):
Great stuff. Thanks!

PR #1186, opened 2025-04-30
fix: heuristics to fix "comparison to expected results"

### LINKED ISSUE #1185: Comparison to Expected Results -> Comparison to Expect Results
## Description
Harper thinks that the "base form of the verb is needed here" and suggests "to expect" instead of "to expected"

Using harper-ls in helix in a typst document, on the Australian dialect

## Screenshots

Image

Image

hippietrail:
# Issues 

Fixes #1185

# Description

Added heuristics for -ed form of a verb after "to"

Refactored to allow for adding similar heuristics for other verb forms.
Added heuristics to keep the old tests passing with the new heuristics in place.

# How Has This Been Tested?

I added the phrase from the issue as a new unit test.

# Checklist

### DIFF
--- a/harper-core/src/linting/inflected_verb_after_to.rs
+++ b/harper-core/src/linting/inflected_verb_after_to.rs
@@ -44,36 +44,82 @@ impl<T: Dictionary> Linter for InflectedVerbAfterTo<T> {
                 continue;
             }
 
-            let mut check_stem = |stem: &[char]| {
+            let check_stem = |stem: &[char]| {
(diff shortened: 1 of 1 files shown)


hippietrail:
> Great stuff. Thanks for addressing the problem so promptly!
> 
> This code is started to get a little hairy. It's fine for now, but next time you're in this module, would you mind refactoring or commenting it a bit more?

I agree. I did refactor it a bit but I'm not that satisfied with the structure of it.

PR #1187, opened 2025-04-30
feat: disambiguate prepositions followed by determiners

RunDevelopment:
# Description
Preposition-like words followed by a determiner are typically prepositions. This disambiguates prepositions in a lot of cases.

# How Has This Been Tested?
Via the PoS snapshot tests. I also added a new document with hundreds of new example sentences.

# Checklist


hippietrail:
Will this affect phrases like "I woke up the dog"?

RunDevelopment:
No. I only enabled this for a few prepositions, which didn't cause many (if any) false positives.

harper-core/src/document.rs
```diff
@@ -176,6 +178,26 @@ impl Document {
         }
     }
 
+    /// A proposition-like word followed by a determiner or number is typically
+    /// really a preposition.
+    fn known_preposition(&mut self) {
+        let pattern = SequencePattern::default()
```
elijah-potter, line 184: You may want to cache this per-thread somewhere (see `ARTICLE_PATTERN`). The memory allocation of the pattern can be costly.

RunDevelopment: More costly than the Vec we need to allocate for `pattern.find_all_matches_in_doc`?

elijah-potter (approved):

PR #1188, opened 2025-04-30
feat: articles implies no verb

RunDevelopment:
# Description
This is a weaker, but much more reliable version `articles_implies_noun`. Basically, the first word after a determiner typically can't be a verb. Unfortunately, this doesn't work for all determines, since the same word may have multiple functions. E.g. "This point ..." "This points (to) ...". So I curated a small list of determiners that produce no false positives that I could find or think of.

# How Has This Been Tested?
PoS snapshot tests.

# Checklist


hippietrail:
Gerunds and to-infinitives are both nouns and verbs, they keep their verbal properties while acting syntactically like nouns:

Gerunds:
- I like **eating**. ← looks like a verb that became a noun
- I like **eating** cake. ← but actually can act like both

Infinitives:
- My goal is **to finish** the project.

But then I'm pretty sure there are definite cases of verbs being fully used as nouns:
- After my sleep I'm going for a run.

We probably want at least gerunds to have noun and verb properties, infinitives almost definitely not, and words like "sleep" and "run" are normally listed in the dictionary with both parts of speech already.

(Related: We should have an "ing" property or similar for inflected verbs. We could have "gerund" but as far as I know every verb in gerund form is also a present participle, which is both a verb and an adjective. So I probably wouldn't call such a property "gerund".)

RunDevelopment:
The point of part-of-speech tagging is to annotate the syntactic use of words in a sentence [*given the surrounding context*](https://en.wikipedia.org/wiki/Part-of-speech_tagging). If an "-ing"-word is cannot be a verb based on context, it shouldn't be tagged as a verb. If you want to have access to the information that the same word can be used as a verb *in a different context*, then you should use a different system. E.g. you could do another lookup into the dictionary.

Not disambiguating will just make it harder to impossible to implement certain rules correctly. E.g. we already have the AdjectiveOfA rule, which needs to use separate heuristics to figure out whether a word tagged as an adjective, is actually an adjective is in this context.

I understand that the function of word can be ambiguous (see your examples), but there are also cases where it's not. This PR handles such cases.

---

I'm also not sure why you brought up to-infinitives as examples of syntactic nouns. While infinitive *phrases* can act as nouns, the verb in the phrase is still a verb. This can be most easily seen with phrasal verbs.

- To err is human. (err is a verb.)
- To wake up is human. (wake up in a phrasal verb)

In the last example, just like "wake" isn't a noun, "up" isn't a noun either. Yet, "to wake up" acts as a noun.

hippietrail:
I'm not aware of how POS tagger actually deal with these issues. I know last time I looked at some a long time ago they tended to all do it differently. That may have settled into some consensus since then.

I don't know how many other contributors are aware of issues like these so they're always worth bringing up.

In an AST nodes could have POS as well as the individual tokens. Without an AST each node just has its word properties so they don't really have an inner and outer POS.

I recall reading/hearing several times that modern POS taggers use something on the order of 100+ word types. I'm not sure if the Penn is just older or aimed at familiarity. I'm just an interested observer, not an expert.

Anyway we'll learn as we go as we get deeper into lints that need whole noun phrases and verb phrases.

RunDevelopment:
> I'm not aware of how POS tagger actually deal with these issues.

Inconsistently :(

The [Penn Treebank P.O.S. Tags](https://www.ling.upenn.edu/courses/Fall_2003/ling001/penn_treebank_pos.html) defines a VBG	(Verb, gerund or present participle) but taggers that implement this tag set (e.g.[Standford](https://nlp.stanford.edu/software/tagger.shtml)) use either VBG or NN (noun). E.g. "talking" in "The talking stopped" gets NN, but "waking" in "The waking stopped" gets VBG. (You can use [this website](https://parts-of-speech.info/) to see for yourself.)


> I'm not sure if the Penn is just older or aimed at familiarity.

Older. It was released in the 90s. Not sure if it was updated since. Wikipedia says it's probably ["the most popular tag set."](https://en.wikipedia.org/wiki/Part-of-speech_tagging#:~:text=The%20most%20popular%20%22tag%20set%22%20for%20POS%20tagging%20for%20American%20English%20is%20probably%20the%20Penn%20tag%20set%2C%20developed%20in%20the%20Penn%20Treebank%20project.) If I had to guess, I'd say that Penn is old and was "good enough" that researchers and devs kept using it. (Probably because comparing the performance of taggers that use different tag sets isn't easy.)

> I recall reading/hearing several times that modern POS taggers use something on the order of 100+ word types.

We probably do too. We just hide the true number by grouping tags together and differentiating them with boolean attributes. E.g. `NounData` has 3 bool attributes, which makes 2^3=8 tags (technically 3^3=27, because `Option<bool>`, but whatever).

> I'm just an interested observer, not an expert.

I'm neither of those... I just want a good grammar checker. The things I tell you about come from what I gathered researching the topic and the NLP course I took at uni.

> Anyway we'll learn as we go as we get deeper into lints that need whole noun phrases and verb phrases.

I believe we are already getting there. I mean, the `NomialPhrase` pattern is quite similar to that.

---

Anyway, let's solve this. What do you need the verb information for, when the word is used as a noun?

There are also cases, when "-ing"-verbs are used as adjectives. E.g. "A shining light", "your dazzling beauty." This PR removes the verb tag there too. What about those?

And how do you feel about the plural forms of gerund nouns and obvious nouns that happen to use gerund form? E.g. "I'm feeling okay." "I have a feeling." "The feelings I have for you."

For nouns, we could add a `NounData::is_gerund: bool` property if you need this information. However, this opens the question, whether adjectives should also have it. (And what about adverbs? E.g. "Interestingly")

The most important thing right now is what you intend to use this information for.

hippietrail:
> > I'm not aware of how POS tagger actually deal with these issues.
> 
> Inconsistently :(

Well that's not always bad. It helps avoid getting bogged down in a good-enough local minimum. Though that kinda happened anyway and got blown out of the water by transformers.

> The [Penn Treebank P.O.S. Tags](https://www.ling.upenn.edu/courses/Fall_2003/ling001/penn_treebank_pos.html) defines a VBG (Verb, gerund or present participle) but taggers that implement this tag set (e.g.[Standford](https://nlp.stanford.edu/software/tagger.shtml)) use either VBG or NN (noun). E.g. "talking" in "The talking stopped" gets NN, but "waking" in "The waking stopped" gets VBG. (You can use [this website](https://parts-of-speech.info/) to see for yourself.)

Ah great stuff. It's definitely the only one I could probably name.

> > I'm not sure if the Penn is just older or aimed at familiarity.
> 
> Older. It was released in the 90s. Not sure if it was updated since. Wikipedia says it's probably ["the most popular tag set."](https://en.wikipedia.org/wiki/Part-of-speech_tagging#:~:text=The%20most%20popular%20%22tag%20set%22%20for%20POS%20tagging%20for%20American%20English%20is%20probably%20the%20Penn%20tag%20set%2C%20developed%20in%20the%20Penn%20Treebank%20project.) If I had to guess, I'd say that Penn is old and was "good enough" that researchers and devs kept using it. (Probably because comparing the performance of taggers that use different tag sets isn't easy.)

I think it was good enough for study but industry needed more. Before transformers took over everything used a clunky pipeline where the next stage couldn't look at the source if the POS tagging wasn't detailed enough etc. Same for phonetic transcriptions in TTS and all that NLP stuff. So you lost stuff at each interface.

> > I recall reading/hearing several times that modern POS taggers use something on the order of 100+ word types.
> 
> We probably do too. We just hide the true number by grouping tags together and differentiating them with boolean attributes. E.g. `NounData` has 3 bool attributes, which makes 2^3=8 tags (technically 3^3=27, because `Option<bool>`, but whatever).

I think it's qualitatively different. We have a lot because it's combinatorial. They have a lot because they want fine-grained. Adverbs alone are just a junk drawer POS. But there's tons of subdivisions you can make on most categories.

> > I'm just an interested observer, not an expert.
> 
> I'm neither of those... I just want a good grammar checker. The things I tell you about come from what I gathered researching the topic and the NLP course I took at uni.

That's great!

> > Anyway we'll learn as we go as we get deeper into lints that need whole noun phrases and verb phrases.
> 
> I believe we are already getting there. I mean, the `NomialPhrase` pattern is quite similar to that.
> 
> Anyway, let's solve this. What do you need the verb information for, when the word is used as a noun?

Well we'll see. I'm trying to be pragmatic. We don't know what we need until we need it by making more and more clever linters.

> There are also cases, when "-ing"-verbs are used as adjectives. E.g. "A shining light", "your dazzling beauty." This PR removes the verb tag there too. What about those?

I would've kept them but we'll see if that info is needed or not.

> And how do you feel about the plural forms of gerund nouns and obvious nouns that happen to use gerund form? E.g. "I'm feeling okay." "I have a feeling." "The feelings I have for you."

Those are usually lexical. Plurals of gerunds are not standard/productive, which is why "trainings" sounds wrong and "campings" and "parkings" sound very wrong. "Feelings" on the other hand will be specifically covered in dictionaries as a separate headword or a separate sense. In fact all 3 of those.

> For nouns, we could add a `NounData::is_gerund: bool` property if you need this information. However, this opens the question, whether adjectives should also have it. (And what about adverbs? E.g. "Interestingly")

I see know reason why any such properties can't be added to the "affix" system as needed. Adverbs of manner are about as regular a category as English has. We'll definitely want different properties to differentiate types of adverbs at some point. The ones that are not derived from adjectives in -ly.

> The most important thing right now is what you intend to use this information for.

It will be an iterative process.

harper-core/src/document.rs
```diff
@@ -176,6 +182,42 @@ impl Document {
...
+                        // "I like her", "Something about her struck me as odd."
+                        "my", "your", "thy", "thine", "his", /*"her",*/ "its", "our", "their",
+                        "whose", // "no" is also a determiner
+                        "no",
+                    ]))
+                    .then_whitespace()
+                    .then_verb(),
+            )
+        }
+        thread_local! {static PATTERN: Lrc<SequencePattern> = create_pattern()}
```
elijah-potter, line 204: I like this positioning.

elijah-potter (approved):
Looks good to me. You're quickly becoming an expert in this.

PR #1192, opened 2025-04-30
feat: Correct "infront" and other phrases written as a single word

hippietrail:
# Issues 
N/A

# Description

Corrects "infront" which these days is one of a set of phrases increasingly written as though it's a single word.

Now all of:
- "a lot"
- "a while"
- "as well"
- "at least"
- "each other"
- "in case"
- "in front" 

In looking for examples I found a few cases are very common in GitHub that should not be corrected:
- It's used as identifiers and company/product names or trademarks as `inFront` and `InFront`
- It's also commonly used as part of hyphenated terms which look like things such as HTML class names.

Even in prose those are pretty common, not just in commented-out code.

When I generalized this code I moved over phrases that were already in `phrase_corrections.rs` and their tests too.

# How Has This Been Tested?

I added a unit test for each edge case I found while looking for examples.
I added unite tests for the problems I had when trying first to implement this with an `All` pattern and an exception sequence before giving up and moving all the logic into `match_to_lint()`

# Checklist


elijah-potter (dismissed):
Looks good to me. Keep an eye on issues after I run the release.

elijah-potter (approved):

Issue #1169, opened 2025-04-29
Sort dictionary.txt before write


averms:
## Description

It would be nice if harper sorted `~/.config/harper-ls/dictionary.txt` before writing the words. I keep it under version control and every time I run the "add word to dictionary" code-action, the order of words in the dictionary seems to be randomized. This causes large git diffs.

hippietrail:
> ## Description
> It would be nice if harper sorted `~/.config/harper-ls/dictionary.txt` before writing the words. I keep it under version control and every time I run the "add word to dictionary" code-action, the order of words in the dictionary seems to be randomized. This causes large git diffs.

I agree. I don't actually use user dictionary at all myself yet but I'm a huge fan of keeping any kind of list sortable. I've sorted a few places in the source code already with lists of modules etc. It makes it easier to spot things with a quick scan, and your problem with diffs is even more important.

averms:
Looks to be fixed as of harper 0.32.1. Thanks @86xsk!

PR #1195, opened 2025-05-01
feat(ls): write sorted dictionaries

86xsk:
# Issues 
#1169 

# Description

This PR makes it so we sort the words iterator when saving the dictionary to disk. Thereby writing a sorted dictionary.

This uses the [default ordering for strings](https://doc.rust-lang.org/std/primitive.str.html#impl-Ord-for-str), which means the resulting sort won't necessarily be alphabetic.

# How Has This Been Tested?

Tested by using the code action to add to the global dictionary from Neovim. Running `sort -c` showed the dictionary file was properly sorted after adding a word. After an initial sort, adding a word no longer produced unreasonably large diffs in the dictionary file.

# Checklist

### DIFF
--- a/harper-ls/src/dictionary_io.rs
+++ b/harper-ls/src/dictionary_io.rs
@@ -1,4 +1,5 @@
 use anyhow::anyhow;
+use itertools::Itertools;
 use std::path::{Component, Path, PathBuf};
 
 use harper_core::{Dictionary, MutableDictionary, WordMetadata};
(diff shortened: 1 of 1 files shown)


elijah-potter (commented):
> This uses the [default ordering for strings](https://doc.rust-lang.org/std/primitive.str.html#impl-Ord-for-str), which means the resulting sort won't necessarily be alphabetic.

It'll be close-enough. This is a simple-enough change. Would it be possible to add a test or two to ensure the behavior doesn't change in the future?

elijah-potter (approved):
> This uses the [default ordering for strings](https://doc.rust-lang.org/std/primitive.str.html#impl-Ord-for-str), which means the resulting sort won't necessarily be alphabetic.

It'll be close-enough. This is a simple-enough change. Would it be possible to add a test or two to ensure the behavior doesn't change in the future?

PR #1197, opened 2025-05-01
feat: dictionary curation and dialect tests

hippietrail:
# Issues 
N/A

# Description

I merged, sorted, and commented new dictionary contributions.

I noticed that `afterwards` was being flagged even for Australian English where it's the preferred spelling.

- Fixed annotations for both `afterward` and `afterwards`
- Added tests for American vs Commonwealth dialects

I had trouble getting some tests to work with Australian English despite this. It turns out that the `dialect` field of the metadata is a string and can only hold one dialect. Changed to simpler tests, some of which are set to ignore with notes.

# How Has This Been Tested?

Added +ive and -ive unit tests for `afterward` and `afterward` for each dialect. Two are disabled due to the above limitation.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -11769,7 +11769,7 @@ aftershave/1SM
 aftershock/1SM
 aftertaste/1SM
 afterthought/14SM
-afterward/~S<
+afterward/j~<
 afterwards/j!@_
 afterword/1MS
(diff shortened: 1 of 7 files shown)


elijah-potter (approved):
I see. I'll get to that dialect issue soon.

PR #1199, opened 2025-05-01
chore: convert commented out tests to use `#[ignore]`

hippietrail:
# Issues 
N/A

# Description

I just discovered that Rust tests allow a `#[ignore]` attribute that has an optional 'reason' field. I wasn't aware of so this before so now I've gone ahead and converted all commented-out tests to use it.

# How Has This Been Tested?

`cargo test` is still happy.

# Checklist

### DIFF
--- a/harper-core/src/lexing/mod.rs
+++ b/harper-core/src/lexing/mod.rs
@@ -358,29 +358,29 @@ mod tests {
         ));
     }
 
-    // #[test]
-    // fn lexes_negative_1() {
-    //     let source: Vec<_> = "-1".chars().collect();
-    //     assert!(matches!(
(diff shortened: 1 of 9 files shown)


elijah-potter (approved):

elijah-potter:
I like this system much more. It improves visibility and search-ability.

hippietrail:
> I like this system much more. It improves visibility and search-ability.

It gives us feedback during testing about ignored tests too during `cargo test`!
I'll actually feel confident adding tests that I know fail now but should be handled ideally.

I meant to check if there's an option that will run the ignore tests and report if some of them have started to pass, but I didn't get around to it yet.

PR #1200, opened 2025-05-01
fix: exception for "how did you" in how-to linter

hippietrail:
# Issues 
N/A

# Description

While dogfooding I noticed sentences like "How did you do that" were flagged by the `HowTo` linter so I added a special exception.

# How Has This Been Tested?

I added a unit test with the sentence that revealed the problem.
All tests pass.

# Checklist

### DIFF
--- a/harper-core/src/linting/how_to.rs
+++ b/harper-core/src/linting/how_to.rs
@@ -22,10 +22,12 @@ impl Default for HowTo {
             .then_anything()
             .then_anything()
             .then(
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
I hope there aren't too many edge cases here.

hippietrail:
> I hope there aren't too many edge cases here.

I expect most of the linters are going to have numerous edge cases. We'll start to see overarching patterns though, but there will also be a smaller number of edge cases after we get a good feeling for it.

PR #1201, opened 2025-05-01
fix: add exception to possessive_your linter

hippietrail:
# Issues 
N/A

# Description

Prevent flagging "showing you what's available".
I'm not generalizing yet until we have more data, just treating "what's" as special for now.

I also tweaked the description wording a little.

While I was there I change another commented-out test to use `#[ignore]`

# How Has This Been Tested?

I added a unit test using the sentence that revealed the false positive.

# Checklist


elijah-potter (approved):
Good stuff.

Issue #1202, opened 2025-05-01
'Let go' not recognized as an expression


Coccocoahelper:
## Description

The idiom 'let go' is not recognized by Harper, even though it is correct.

2 spurious fixes are proposed, 'let's go' and 'let us go'.

## Resources

Merriam-Webster is reliable, so here's a link for its entry of 'let go of':

https://www.merriam-webster.com/dictionary/let%20go%20of

## Screenshots

Here's a screenshot of the issue (if requested, I may send the full text):

![Image](https://github.com/user-attachments/assets/ab3530ad-b414-440a-bb1d-6476e5b2c0d2)

Coccocoahelper:
NOTE: This does not seem to happen when I use 'let go of W'.

hippietrail:
Thanks for the report. I'll look at this now.

Coccocoahelper:
> Thanks for the report. I'll look at this now.

Make sure to check my comment, it has important notes.

hippietrail:
> NOTE: This does not seem to happen when I use 'let go of W', or if I add a comma (as in 'let go, (…)') — maybe this is an issue with the correction mechanism itself not giving the right suggestions?

Thanks. At present commas delimit "chunks" in Harper's lint system, along with sentence-final punctuation (. ? !) and paragraph breaks. To figure out if something is a mistake or not it often has to look at what comes before and after, and as it's quite young there's edge cases to discover.

elijah-potter:
Has this issue been resolved?

hippietrail:
Yes fixed in #1203

PR #1203, opened 2025-05-01
fix: heuristic for no-contraction-with-verb for "let go"

hippietrail:
# Issues 

Solves issue #1202

# Description

The phrasing in the bug report is inherently ambiguous even looking forward more and more tokens.
However, `lets go` is a very common mistake for `let's go` so worth flagging even though in a context like `he lets go of the ball` would an error. On the other hand `let go` is more likely to be part of a context like `he let go of the ball` that a mistake for `let's go`.
So this boils down to a frequence heuristic.

Which means false positives and negatives are still possible but less likely now.

I've also applied some refactoring I've learned to clean up the other code a little.

# How Has This Been Tested?

I made a unit test from the sentence in the screenshot in the issue.

# Checklist

### DIFF
--- a/harper-core/src/linting/lets_confusion/no_contraction_with_verb.rs
+++ b/harper-core/src/linting/lets_confusion/no_contraction_with_verb.rs
@@ -33,20 +33,15 @@ impl Default for NoContractionWithVerb {
                 return false;
             }
 
-            let tok_chars = tok.span.get_content(source);
+            let lowercase = tok.span.get_content_string(source).to_lowercase();
 
             // If it ends with 'ing' and is at least 5 chars long, it could be a gerund or past participle
             // TODO: replace with metadata check when affix system supports verb forms
-            if tok_chars.len() < 5 {
+            if lowercase.len() < 5 {
                 return true;
             }
 
-            let is_ing_form = tok_chars
-                .iter()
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/lets_confusion/no_contraction_with_verb.rs
```diff
@@ -210,4 +215,13 @@ mod tests {
     fn dont_flag_let_us() {
         assert_lint_count("Let us do this.", NoContractionWithVerb::default(), 0);
     }
+
+    #[test]
+    fn dont_flag_let_go_1202() {
```
elijah-potter, line 220: Moving forward, let's name all tests that come from an issue with the schema `*_issue_<NUMBER>`. 

In this case, that would be `dont_flag_let_go_issue_1202`.

elijah-potter (approved):
I have one tiny nitpick. Otherwise, this is ready to go.

harper-core/src/linting/lets_confusion/no_contraction_with_verb.rs
elijah-potter: It is not very important, so I'm not going to let it slow down merging this PR.

PR #1205, opened 2025-05-01
test(ls): ensure we write sorted dictionaries

86xsk:
# Description

Adds a test to ensure we write stably sorted dictionaries with the current sort method. Though the limited word count might not cover every edge case, it should ensure we don't accidentally do something like write unsorted dictionaries or reverse the sort order in the future.

(Also adds some `const`s and a test utility function to hopefully make it easier to write `dictionary_io` tests in the future.)

Related PR: #1195 

# How Has This Been Tested?
`cargo test`, making modifications until the test was passing successfully.

# Checklist

### DIFF
--- a/harper-ls/src/dictionary_io.rs
+++ b/harper-ls/src/dictionary_io.rs
@@ -82,3 +82,57 @@ pub fn file_dict_name(url: &Url) -> anyhow::Result<PathBuf> {
 
     Ok(rewritten.into())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Cursor;
+
+    const TEST_UNSORTED_WORDS: [&str; 10] = [
+        "peafowl",
+        "housebroken",
+        "blackjack",
+        "Žižek",
+        "BMX",
+        "icebox",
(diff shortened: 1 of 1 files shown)


grantlemons:
Wait, I'm confused, iirc dictionaries don't need to be sorted?

`dictionary_io` is only responsible for writing to file and user dictionaries, and both are represented as hashmaps under the hood, so sort order doesn't matter.

Storing in lexicographical order may speed up the construction of an `FstDictionary`, but that is only relevant for the curated dictionary.

86xsk:
> Wait, I'm confused, iirc dictionaries don't need to be sorted?
> 
> `dictionary_io` is only responsible for writing to file and user dictionaries, and both are represented as hashmaps under the hood, so sort order doesn't matter.

PR #1195 makes it so we write sorted dictionaries (per user request). This test just helps ensure we write the words in a lexicographical order when saving file/user dictionaries.

elijah-potter (approved):
I like this. I was going to ask you to add these on the other PR, but deemed it of lower concern. Thanks for taking initiative!

PR #1206, opened 2025-05-01
fix: Remove verb property from "of"

RunDevelopment:
# Description
According to wikionary, "of" can have the meaning of "have" in [Eye dialect](https://en.wiktionary.org/wiki/Appendix:Glossary#eye_dialect), a dialect of deliberate misspelling. Given how common the word "of" is in English, I believe it is more important to unambiguously tag "of" as a preposition than to support a deliberate misspelling. This will help prevent false positives in rules dealing with verbs.

# How Has This Been Tested?
See snapshots.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -35967,7 +35967,7 @@ oestrogen/1MS!_
 oestrous/5!_
 oestrus/1MS!_
 oeuvre/~1MS
-of/~+4
+of/~+
 off/~5+41SZGDRJ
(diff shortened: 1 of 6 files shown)


hippietrail:
There's quite a few of these anomalies from dictionary data import. I've mostly found obsolete, archaic, or dialectal stuff. For those I put a note in a comment when I remove them. I don't think eye dialect warrants that (-:

elijah-potter (approved):

PR #1208, opened 2025-05-02
rename harper-cli to harper-debug

grantlemons:
One of the most common misunderstandings I see is that the `harper-cli` tool is intended for users to grammar check files in the terminal.
While there may be ways to use it this way, it is intended as a development aide, as is reflected in its unpublished status.

This PR changes the name to `harper-debug`, to hopefully clear up this confusion.


hippietrail:
Another good name might be "harper-tool"

elijah-potter:
It's a point of confusion, but I don't want to make the change since it would cause a _huge_ downstream headache from the various package managers that expect a `harper-cli` binary.

I don't see it as worth it.

PR #1209, opened 2025-05-02
feat: add single token pattern trait

RunDevelopment:
# Description

I intend to change the `Pattern` trait in the future, so I added a simpler version of it called `SingleTokenPattern`. This trait is less likely to change over time, meaning that less code has to be updated as the `Pattern` trait evolves.

# How Has This Been Tested?
It doesn't change functionality

# Checklist


elijah-potter (approved):
This is a good call.

PR #1210, opened 2025-05-02
fix: prevent crash in PhrasalVerbAsCompoundNoun

RunDevelopment:
# Issues 
I noticed that Harper crashed with the following error:

```
thread 'main' panicked at harper-core\src\linting\phrasal_verb_as_compound_noun.rs:54:17:
internal error: entered unreachable code
```

The VSC language server then repeatedly restarts and crashes until the following message pops up:

image

# Description
The crash was caused by an `unreachable!` statement being very reachable. The dictionary contains some multi-word entries, which contain hyphens and spaces. The tokenizer seems to use this information to parse words like "add-on" as one token. This triggered the following code path:

```rs
            // * Can't contain a hyphen or space.
            if nountok_charsl.contains(&'-') || nountok_charsl.contains(&' ') {
                // This means Harper's tokenizer's concept of a word is different to what we expect
                // or has changed since this code was written.
                unreachable!();
            }
```

The fix is to *skip* tokens that contain these characters instead of crashing.

# How Has This Been Tested?
With a new test.

# Checklist

### DIFF
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -47,27 +47,25 @@ impl Linter for PhrasalVerbAsCompoundNoun {
                 continue;
             }
             let nountok_charsl = document.get_span_content(&token.span);
-            // * Can't contain a hyphen or space.
-            if nountok_charsl.contains(&'-') || nountok_charsl.contains(&' ') {
-                // This means Harper's tokenizer's concept of a word is different to what we expect
-                // or has changed since this code was written.
-                unreachable!();
-            }
-            // * Can't contain an apostrophe
-            if nountok_charsl.contains(&'\'') || nountok_charsl.contains(&'’') {
+            // * Can't contain space, hyphen or apostrophe
+            if nountok_charsl.contains(&' ')
+                || nountok_charsl.contains(&'-')
+                || nountok_charsl.contains(&'\'')
+                || nountok_charsl.contains(&'’')
+            {
                 continue;
             }
             // * Must end with the same letters as one of the particles used in phrasal verbs.
             let particle_endings: &[&[char]] = &[
-                &['a', 'r', 'o', 'u', 'n', 'd'][..],
-                &['b', 'a', 'c', 'k'][..],
-                &['d', 'o', 'w', 'n'][..],
-                &['i', 'n'][..],
-                &['o', 'n'][..],
-                &['o', 'f', 'f'][..],
-                &['o', 'u', 't'][..],
-                &['o', 'v', 'e', 'r'][..],
-                &['u', 'p'][..],
+                &['a', 'r', 'o', 'u', 'n', 'd'],
+                &['b', 'a', 'c', 'k'],
+                &['d', 'o', 'w', 'n'],
+                &['i', 'n'],
+                &['o', 'n'],
+                &['o', 'f', 'f'],
+                &['o', 'u', 't'],
+                &['o', 'v', 'e', 'r'],
+                &['u', 'p'],
             ];
 
             // * Must not be in the set of known false positives.
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -369,4 +369,13 @@ mod tests {
             0,
         );
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


hippietrail:
This is odd. The dictionary doesn't contain "tokens". Tokens come from the tokenizer, not from the dictionary. Code in linters often needs to compare tokens to the dictionary.

For this to happen the tokenizer would've had to change, which is why crashing is OK because it brings attention to run-on effects. Do you have some example text which causes this?

RunDevelopment:
While word tokens typically don't contain hyphens and other symbols, there's nothing in the code that guarantees this. 

The problem isn't that the definition for a word token changed, but that this lint assumed a particular definition in the first place. The lint has no control over tokens, so it should just deal with what it has been given.

> which is why crashing is OK because it brings attention to run-on effects

Crashing is absolutely NOT okay in user space. I didn't encounter this crash when working on Harper as a dev, but when using it as a regular user. I only noticed this bug, because Harper repeatedly crashed for me the last couple of days, which made it neigh unusable. Crashing in tests is okay, but crashing for users is not.

If you want to catch run-on effects, test more. I'd recommend snapshot tests again. Just like I did for PoS tagging, we could run all lints on a few documents and create snapshots of the results. [Kinda like this](https://github.com/RunDevelopment/eslint-snapshot-rule-tester).

> Do you have some example text which causes this?

I used "add-on" in the test I added, but any word with a hyphen [in the dictionary](https://github.com/Automattic/harper/blob/a2e0da7a841fba9436c2fa1bbfbabf4bee4e314d/harper-core/dictionary.dict#L52540) works.

hippietrail:
> > which is why crashing is OK because it brings attention to run-on effects
> 
> Crashing is absolutely NOT okay in user space. I didn't encounter this crash when working on Harper as a dev, but when using it as a regular user. I only noticed this bug, because Harper repeatedly crashed for me the last couple of days, which made it neigh unusable. Crashing in tests is okay, but crashing for users is not.

I don't know what industry standard is currently in Rust or in general these days when it comes to unexpected input that reveals something wrong in another part of the code. Sorry I'm still learning.
I suppose somewhere a test against spaces and hyphens inside tokens. I'm not sure where/how to do that test. I'll have to think about it.

I'm still puzzled how this was reached when you were using Harper as a regular user?

> > Do you have some example text which causes this?
> 
> I used "add-on" in the test I added, but any word with a hyphen [in the dictionary](https://github.com/Automattic/harper/blob/a2e0da7a841fba9436c2fa1bbfbabf4bee4e314d/harper-core/dictionary.dict#L52540) works.

I still can't reproduce it. I must still be misunderstanding something.

RunDevelopment:
> I'm still puzzled how this was reached when you were using Harper as a regular user?

I use Harper via the VSC plugin. Then I just wrote a comment in Rust code that contained the sentence "I like this add-on!". Boom. Crash. See above for the exact error.



> I don't know what industry standard is currently in Rust or in general these days when it comes to unexpected input that reveals something wrong in another part of the code. Sorry I'm still learning.

No need to apologize. There aren't any standards for grammar checkers that I'm aware of. Just think about this from the perspective of a user. If the grammar checker has a bug that flags too much, the user can ignore it. If the grammar checker can't be used because it constantly crashes, the user looks for another grammar checker. 

No one will use software that constantly crashes. This is why crashes can kill projects. I am so vocal about this, because I want Harper to live a little longer. It's currently the only VSC grammar checker that works properly for large Markdown documents. LT and co all become unusable when the document is too large.

hippietrail:
Ah OK I can finally reproduce this. I can't get it to happen in a test yet, only in a comment in a rust document. I'm going to hunt the root cause.

RunDevelopment:
> Ah OK I can finally reproduce this. I can't get it to happen in a test yet, only in a comment in a rust document. I'm going to hunt the root cause.

Wdym? I added a test that paniced before. Just copy it.

hippietrail:
The root cause is `is_case_separator` used by `collapse_identifiers.rs`, which treats snake_case and kebab-case as words.

It doesn't trigger in Markdown and in testing in code comments I couldn't find any other trigger that worked besides `add-on`

So I missed that and your fix is correct but it does seem to have revealed something strange going on that I don't fully understand yet.

elijah-potter:
> Crashing in tests is okay, but crashing for users is not.

Preach.

elijah-potter (approved):
Good eye! I'll run a release today so this fix can get sent out.

PR #1214, opened 2025-05-02
feat: Improved metadata applied by affix system

### LINKED ISSUE #1212: How does harper's dictionary handle verb forms?
I noticed that "point" is tagged as noun singular or verb (which is correct), but "points" is only tagged as noun plural. This is obviously incorrect, since "points" is the 3rd person singular form of "point". E.g. "Julia points at Eric."

While it would be easy to fix this particular case, it made me wonder whether Harper's dictionary has any way to say "this is a regular verb" which implies that "\<verb>-ed" and "\<verb>-s" are also verbs?

---

Similarly, if the dictionary contains a noun singular, is there something in place to check that the plural version is also in the dictionary?

RunDevelopment:
# Issues 

Resolves #1212

# Description

I changed how the `target_metadata` of affixes is applied. Instead of applying the same metadata to all words, it is now possible to specify a condition to only apply certain metadata if the base word's metadata meets certain conditions. 

This means that e.g. the "-s" suffix can now apply the metadata "noun plural" or "verb present tense" depending on whether the base word is a noun or verb. So e.g. "algorithms" is always a noun, while "words" is a noun or verb depending on context. (e.g. "He has his way with words. He words it so well").

Besides the "-s" suffix, I also fixed the metadata applied by the "-(i)(e)d" suffix. This one was intended to make the derived words adjectives, but this was declared incorrectly. I fixed this, meaning that all derived "-(i)(e)d" words are now tagged as adjectives. I also added that the derived "-(i)(e)d" words are verbs if they were derived from verbs. E.g. see the word "derived" in my last sentence.

I also added a few words to the dictionary that weren't tagged correctly.

# How Has This Been Tested?
See the snapshots. So many incorrect word tags were fixed.

# Checklist

### DIFF
--- a/harper-core/tests/text/linters/this and that.snap.yml
+++ b/harper-core/tests/text/linters/this and that.snap.yml
@@ -1,7 +1,7 @@
 Lint:    Spelling (63 priority)
 Message: |
        2 | As such, disambiguating them is necessary.
-         |          ^~~~~~~~~~~~~~ Did you mean “disambiguation”?
+         |          ^~~~~~~~~~~~~~ Did you mean `disambiguation`?
 Suggest:
   - Replace with: “disambiguation”
 


harper-core/src/linting/possessive_your.rs
```diff
@@ -65,6 +65,7 @@ mod tests {
     use super::PossessiveYour;
 
     #[test]
+    #[should_panic] // currently fails because comments is a homographs (verb or noun)
```
hippietrail, line 68: Ooh handy! Didn't know about this. I'll remember to use it in other tests I wanted to include for things that don't work but should.

elijah-potter (approved):
Love this kind of stuff. I don't see any possible issues with this code, so I'm going to push it through. It might cause some merge conflicts with other PRs, though.

PR #1223, opened 2025-05-03
feat(cli): fallback to `PlainEnglish` parser

86xsk:
# Description

Makes `harper-cli` fallback to the `PlainEnglish` parser when encountering an unimplemented file extension. When this happens, a message will be printed to notify the user of this occurring.

This change allows `harper-cli` to be used on a wider selection of files, like `.txt` files and those without an extension.

A potential concern is that, by printing the fallback warning message, we are changing the output of some `harper-cli` commands. If something relies on `harper-cli`'s output being in a particular format, it could break.

# How Has This Been Tested?
- `cargo test`
- Compiled and manually tested. I tested `harper-cli` with `.cs`, `.md`, and `.txt` files, and also those without an extension. The expected behavior was observed and no issues were noticed.

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -7,13 +7,12 @@ use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 use std::{fs, process};
 
-use anyhow::format_err;
 use ariadne::{Color, Label, Report, ReportKind, Source};
 use clap::Parser;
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
I don't see the output changing to be an issue. This is a debug program that we only support for internal use. If someone raises a stink, we can always add a flag to retain the previous behavior.

PR #1225, opened 2025-05-04
refactor: use `LSend` trait to simplify trait definitions

RunDevelopment:
# Description
I found the `LSend` trait today and used it to simplify the trait definitions of `Pattern` and `Parser`

# How Has This Been Tested?
I just made sure that everything compiles. I also checked that tests runs.

# Checklist


elijah-potter (approved):
Good stuff. I've been meaning to do this, but never got around it. Thanks for getting this in!

PR #1232, opened 2025-05-05
fix: `Y` flag for `-ly` suffix adverb POS fixes

hippietrail:
# Issues 
N/A

# Description

`Y` adds the `-ly` suffix. This suffix changes normal adjectives to adverbs of manner.
Coincidentally a couple of adjectives happen to also end in `-ly`, such as `friendly`.

The `affixes.json` was seting the correct `adverb` POS but also the incorrect `adjective` POS.
And it didn't specify `adjective` for the base POS.

I had to correct the `pear` and `state` dictionary entries which were using the `Y` flag incorrectly, probably due to Hunspell legacy affix compression. These prevented tests from passing in `pharse_correction.rs` and `lint_group.rs`

I'm also added adverb support to `then_than.rs` for a test that began to fail.

# How Has This Been Tested?

Manual test and tweaking to get `cargo test` to succeed.

# Checklist


harper-core/tests/text/tagged/Part-of-speech tagging.md
```diff
@@ -242,8 +242,8 @@
 # NSg    . V/C NSg/V  D/P NSg/V V/P D   NPl           V/P I/J     NPl       . C/P
 > example , once  you've seen  an  article such  as    ' the ' , perhaps the next    word  is a
 # NSg/V   . NSg/C W?     NSg/V D/P NSg/V   NSg/I NSg/R . D   . . NSg     D   NSg/J/P NSg/V VL D/P
-> noun  40 % of  the time  , an  adjective 40 % , and a   number  20 % . Knowing   this , a
-# NSg/V #  . V/P D   NSg/V . D/P NSg/V/J   #  . . V/C D/P NSg/V/J #  . . NSg/V/J/P I/D  . D/P
+> noun  40 % of  the time    , an  adjective 40 % , and a   number  20 % . Knowing   this , a
+# NSg/V #  . V/P D   NSg/V/J . D/P NSg/V/J   #  . . V/C D/P NSg/V/J #  . . NSg/V/J/P I/D  . D/P
```
RunDevelopment, line 246: [Time](https://en.wiktionary.org/wiki/time#English) isn't an adjective. Similarly, the following words are also incorrectly tagged as adjectives now: year, hour, night, love, death, sight, earth, father, daughter, manner, purpose, swimming.

The adjective property assigned via `base_metadata` should probably be removed.

hippietrail: Actually that's because many of those `-ly` words are not adverbs but adjectives so need to be fixed in the dictionary.

All from yearly to fatherly are adjectives. Swimmingly is an adverb though and swimming is a present participle and present participles are adjectives. The others I've never seen before.

Correction: They are adjectives but some are both adjective+adverb. The dictionary entries need to be fixed either way.

I also think the base POS property shouldn't be assigned for any 'affix' but it's all through the system for the other affixes. It is going to be needed for fixing the system though, which should check the base POS before adding the affix and assigning the derived POS.

I'll fix all the POSes in a new commit after cross-checking with dictionaries...

RunDevelopment: > All from yearly to fatherly are adjectives.

That's not what I was saying. I was talking about year to father. No -ly. See the line of code I commented on. "time" is incorrectly tagged as an adjective.

hippietrail: > > All from yearly to fatherly are adjectives.
> 
> That's not what I was saying. I was talking about year to father. No -ly. See the line of code I commented on. "time" is incorrectly tagged as an adjective.

Yes it was wrong for them to use `Y` so I've corrected them in the dictionary. I was only explaining why they had `Y` for Hunspell. We've gotta clean that kinda stuff up for Harper.

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-	"typescript.tsdk": "node_modules/typescript/lib"
-}
+  "typescript.tsdk": "node_modules/typescript/lib",
```
elijah-potter, line 2: Was this intentional?

elijah-potter (approved):

PR #1228, opened 2025-05-04
test: add snapshots for linters

RunDevelopment:
# Motivation

Why I added snapshots:

1. They make it easy to see the impact of changes to lints (if the text corpus is large enough).
2. Changes in the tokenizer, dictionary, or PoS tagger that affect lints will also (likely) become visible.
3. It becomes easy to see when the span of a report is suboptimal/wrong.

# Description
Similar to the PoS tag snapshots, I added snapshots for the reports of all lints. Just like before, snapshots are automatically created and updated just by running the snapshot test.

Snapshots are formatted in a YAML-like syntax, and reports have a Rust-error-messages like syntax. Example:

```yml
Lint:    Spelling (63 priority)
Message: |
      46 | calculator in 1623. In 1673, Gottfried Leibniz demonstrated a digital mechanical
         |                              ^~~~~~~~~ Did you mean to spell “Gottfried” this way?
Suggest:
  - Replace with: “Lotteries”
  - Replace with: “Notified”
  - Replace with: “Pottered”
```

# How Has This Been Tested?
This is a test.

# Checklist

- [x] I have performed a self-review of my own code
- [ ] I have added tests to cover my changes

# Notes

1. If you look through the snapshots, you'll see that long-sentence reports will sometimes start at the space or punctuation before the first word of the sentence. This is a bug in the lint, not the snapshot formatter.
2. The snapshots reveal several bugs and false positives. I recommend to look through them.

PR #1237, opened 2025-05-06
feat(core): use Rayon in snapshot tests

elijah-potter:
# Issues 

Follow up to #1228 

# Description

I saw an easy candidate to parallelize snapshot tests using Rayon. This PR reduces total test time on `harper-core` from 33 seconds to 24 seconds. Linter tests alone go from 22 to 16 seconds. I have a 16 core laptop. Your mileage may vary.

Test speed is important to iteration time.

# How Has This Been Tested?

Manually.

# Checklist


RunDevelopment:
It might be more effective to add `opt-level = 1` instead. On my machine, this reduces the time for `tests/linters.rs` from 33 sec to 4.5 sec. From my experience, `opt-level = 1` doesn't take much longer to compile than regular debug mode, so this should be a win overall. The main problem is that break points stop working, so you need to manually comment out `opt-level = 1` when debugging.

(I do the same for [`dds`](https://github.com/image-rs/image-dds/tree/main), where it reduces the time to run tests [from 1 min to 5 sec.](https://github.com/image-rs/image-dds/blob/main/CONTRIBUTING.md#debugging-tests))

I don't want to say that we shouldn't parallelize. I just want to point out that tests can be even faster.

---

This might be unrelated, but are there plans to parallelize the linting of harper itself?

elijah-potter:
> This might be unrelated, but are there plans to parallelize the linting of harper itself?

It's not necessary yet. When it becomes so, it should be quite straightforward to do it at the rule level using Rayon as well.

> It might be more effective to add opt-level = 1 instead. On my machine, this reduces the time for tests/linters.rs from 33 sec to 4.5 sec. From my experience, opt-level = 1 doesn't take much longer to compile than regular debug mode, so this should be a win overall. The main problem is that break points stop working with optimizations, so you need to manually comment out opt-level = 1 when debugging.

I'll look into it right now.

Cargo.toml
```diff
@@ -2,6 +2,9 @@
 members = [ "harper-cli", "harper-core", "harper-ls", "harper-comments", "harper-wasm", "harper-tree-sitter", "harper-html", "harper-literate-haskell", "harper-typst" , "harper-stats"]
 resolver = "2"
 
+[profile.test]
+opt-level = 1
```
RunDevelopment, line 6: I really think the issues with debug breakpoints in tests not working should be documented. The first time, I hit this issue, it took me a while to figure out.

RunDevelopment (approved):

PR #1239, opened 2025-05-06
fix(core): don't misplace `ParagraphBreak` tokens

### LINKED ISSUE #1181: Harper-ls crashes after they are a certain number of words that do not end with a period after a sentence
**Describe the bug**
If you have a line that ends with a sentence/period, then has a lot of words which do not end with a period, harper-ls crashes. I tested a bit, and the numbers of words seems to be inconsistent. But from what I could tell, it cannot just be characters. 

Below is some helix logs:

[helix.log](https://github.com/user-attachments/files/19969156/helix.log)

**To Reproduce**
1. Open a text file and to an empty line, paste the following: ". one two three four five six seven eight nine ten eleven twelve thirteen fourteen fifteen sixteen seventeen eighteen nineteen twenty twenty-one twenty-two twenty-three twenty-four twenty-five twenty-six twenty-seven twenty-eight twenty-nine thirty thirty-one"
2. Sometimes you need to write the file in order for harper-ls to crash

**Expected behavior**
There wouldn't be a crash. 

**Platform**
Helix on Ultramarine Linux 41 (Fedora fork)

86xsk:
# Issues 
(Hopefully) fixes #1181.

# Description

This ensures we don't incorrectly place a `ParagraphBreak` token at the start of a sentence/sequence in the Markdown parser. Currently, a `ParagraphBreak` token which is intended to be placed at the end of a sequence will sometimes (or always?) get placed at its beginning instead.

Incidentally, this also seems to fix the `LongSentence` lint sometimes not spanning the whole sentence. I first noticed this when testing in `harper-cli`, but this is also visible in the adjusted snapshots.

I'm not 100% confident this is the ideal fix, because I can't say I fully understand the inner workings of the Markdown parser yet. Hopefully someone more confident in this area of the code can double-check this.

# How Has This Been Tested?
`cargo test`
Manual testing with `harper-cli` and `harper-ls`. With these changes, I could no longer replicate the crash mentioned in #1181.

# Checklist


RunDevelopment:
> Hopefully someone more confident in this area of the code can double-check this.

I did some quick debugging, and it seems that the root cause of the problem is the way the MD parser assigns spans to event and the way. For the input `"Something cool\n"`, the following is emitted by the MD parser:

- `Event::Start(Tag::Paragraph)` from `0..15`
- `Event::Text` from `0..14`
- `Event::End(Tag::Paragraph)` from `0..15`

These spans are a problem, because tokens get their span assigned from the `start` of the MD spans.

(All of this is a little obfuscated, because the MD parser uses byte spans while Harper uses char spans. `traversed_chars` is used for the conversion.)

This leaves us with 2 choices:

1. Derives the token span from the `end` of the MD spans. In other words, `Span::new_with_len(range.end - 1, 0)` (ignoring byte→char index translations).
2. Use the `end` span of the last token (implemented in this PR).

I believe that what is implemented in this PR is the better choice. The reason this PR exists is that the ranges for start/end events don't fit our use case, so I think that not using them and relying on known good data (=other tokens) is more robust.

That said, I think a comment explaining this should be added to the code. After all, it's not obvious why the old `Span::new_with_len(traversed_chars, 0)` is incorrect.

elijah-potter (approved):

PR #1241, opened 2025-05-07
feat(core): add several new rules

elijah-potter:
# Description

This PR introduces several new rules. I don't think there's any better description of what they are then the code itself.

# How Has This Been Tested?

Unit tests.

# Checklist


harper-core/src/linting/ask_no_preposition.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        let verb = toks[0].span.get_content_string(src).to_lowercase();
+        let span = Span::new(toks[2].span.start, toks[3].span.end);
+
+        Some(Lint {
+            span,
+            lint_kind: LintKind::WordChoice,
+            suggestions: vec![Suggestion::ReplaceWith(Vec::new())],
+            message: format!(
+                "The verb `to {verb} someone` should not be preceded by the preposition “to”."
```
hippietrail, line 50: The verb itself can be preceded by "to" but the object of the verb, in this case the pronoun, shouldn't be preceded by "to". Or you could say the verb shouldn't be followed by "to".

harper-core/src/linting/else_possessive.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn default() -> Self {
+        let pronouns = WordSet::new(&[
+            "somebody",
+            "someone",
+            "anybody",
+            "anyone",
+            "everybody",
+            "everyone",
+            "nobody",
+            "noone",
```
hippietrail, line 21: "no one" is <b>also</b> legit. In fact I just checked and "noone" is actually a mistake. You should be able to use an `EitherPattern` with the `WordSet` on one side and `SequencePattern` of `no` and `one` on the other.

harper-core/src/linting/nominal_wants.rs
```diff
@@ -0,0 +1,81 @@
...
+    patterns::{Pattern, SequencePattern, WordSet},
+};
+
+pub struct NominalWants {
+    pattern: Box<dyn Pattern>,
+}
+
+impl Default for NominalWants {
+    fn default() -> Self {
+        let miss = WordSet::new(&["wont", "wonts"]);
```
hippietrail, line 13: You could include "wan'ts" which I've seen around and just checked that it's on GitHub too.

harper-core/src/linting/nominal_wants.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.pattern.as_ref()
+    }
+
+    fn match_to_lint(&self, toks: &[Token], _src: &[char]) -> Option<Lint> {
+        let offender = toks.last()?;
+        Some(Lint {
+            span: offender.span,
+            lint_kind: LintKind::Miscellaneous,
+            suggestions: vec![Suggestion::ReplaceWith("wants".chars().collect())],
+            message: "Did you mean `wants`?".to_owned(),
```
hippietrail, line 35: `then_nominal` should be matching on singular and plural if plurality is propagating properly. This means either `wants` or `want` will be correct and the wrong one will introduce a different error.

harper-core/src/linting/save_to_safe.rs
```diff
@@ -0,0 +1,77 @@
...
+impl Default for SaveToSafe {
+    fn default() -> Self {
+        let pattern = SequencePattern::default()
+            .then(InflectionOfBe::new().or(Box::new(Word::new("it"))))
+            .then_whitespace()
+            .t_aco("save")
+            .then_whitespace()
+            .t_aco("to")
+            .then_whitespace()
+            .then_verb();
```
hippietrail, line 20: You might want to check if the verb is also an adjective or noun, because those could be false positive. You really only want infinitive forms of verbs too, the only form that can come after "to". Past and continuous forms can act as adjectives and nouns. I can't think of examples off the top of my head right now though.

harper-core/src/linting/win_prize.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl Default for WinPrize {
+    fn default() -> Self {
+        let verbs = WordSet::new(&["win", "wins", "won", "winning"]);
+        let miss = WordSet::new(&["price", "prices", "prise", "prises"]);
+
+        let pattern = SequencePattern::default()
+            .then(verbs)
+            .then_whitespace()
+            .then_determiner()
```
hippietrail, line 19: You could also catch "win prizes" etc without determiner, any verb form, but "prizes" must be plural.

harper-core/src/linting/nominal_wants.rs
elijah-potter: That gets caught by spellcheck.

harper-core/src/linting/else_possessive.rs
```diff
@@ -18,8 +18,10 @@ impl Default for ElsePossessive {
             "everybody",
             "everyone",
             "nobody",
-            "noone",
-        ]);
+        ])
+        .or(Box::new(
+            SequencePattern::aco("no").then_whitespace().t_aco("one"),
```
hippietrail, line 23: At the time I forgot there's also `ExactPhrase`, unless that's also case-sensitive?

PR #1242, opened 2025-05-08
refactor: don't allocate when linting chunks

RunDevelopment:
# Description
I didn't like that `run_on_chunks` allocates a new `Vec` for every chunk with a lint in it. So I made it return an iterator instead.

This doesn't change the behavior of Harper in any way.

# How Has This Been Tested?
Hasn't been.

# Checklist


elijah-potter (approved):
I have nothing to say. This is just raw improvement. Thank you.

PR #1245, opened 2025-05-08
feat: improve spell check suggestions by accounting for common misspellings

RunDevelopment:
# Description
The fix for some dialect-based misspellings is very obvious. E.g. color ↔ colour and recognize ↔ recognise. So I changed the spell checker to **only** suggest the version of the word in the current dialect, if it detects an "ou↔o" or "s↔z" misspelling.

I also refactored and slightly optimized `SpellCheck::cached_suggest_correct_spelling`. It now only caches the words that will be suggested (without any excess) and generally works a bit more efficiently.

# Demo
See tests.

# How Has This Been Tested?
Using existing tests and a new document.

# Checklist

### DIFF
--- a/harper-core/src/linting/spell_check.rs
+++ b/harper-core/src/linting/spell_check.rs
@@ -6,7 +6,10 @@ use smallvec::ToSmallVec;
 use super::Suggestion;
 use super::{Lint, LintKind, Linter};
 use crate::document::Document;
-use crate::spell::{is_ou_misspelling, is_sz_misspelling, suggest_correct_spelling};
+use crate::spell::{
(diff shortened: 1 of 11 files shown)


hippietrail:
> > And many words end with `-er` in US spelling but `-re` in Commonwealth spelling, such as `metre`, `litre`, `theatre`.
> 
> I always wondered what the difference between meter and metre is. Good to know!

Commonwealth English uses "meter" for devices that measure things and "metre" for the unit of length.

> Might as well fix this and `ll` too.

Nice!

elijah-potter (approved):
This is some good work. I appreciate the improvements to readability you've also introduced.

Issue #1220, opened 2025-05-03
Persistent lint ignores with LSP


kuukunen:
## Description

For context, I'm using harper-ls in nvim through lspconfig/mason.

I would like to have persistency for lint ignores. For example, I have a .txt file (I configured 'text' to be included in the filetypes. I'm almost tempted to request having that as default, but I guess it'd be more of an lspconfig issue.)

Let's say in a file I have a way too long sentence that I think is ok, heck the conventions. I mark it as ignored through the LSP code actions system. Alas, when I reopen the file, the same lint issue reappears. There are no words in the extended English language to describe how much this annoys me.

I jest, but I think it would be nice to have some sort of out-of-the-box default way to persist the ignored lints, like how dictionary additions are saved. It could just automatically serialize and deserialize the IgnoredLints.context_hashes and store them with the file path as key to some file(s)/sqlite and, say, have configurable/hard coded 1000 file LRU cache for it.

I tried searching for issues and the code. I saw there was some typescript stuff related to exporting the ignored lints, but I felt I was getting a bit too deep without finding anything relevant. I apologize if this has been discussed before or even solved.

## Component

- The `harper-ls` language server

## Implementation Ideas

- Simple file/sqlite LRU cache storing IgnoredLints with file path as key.

## Concerns

Theoretically there could be a usability issue of "But what if I ignored something by accident and I want it back? Now I can't just reopen the file." First, I'm not entirely sure how the current system works. If I rewrite the same part, it seems like it will rediscover the lint, so it should be mostly fine. I'm not sure if this would ever be a realistic problem. Technically there could be some LSP custom request for resetting all ignored lints for the current file, but I'm not sure if that would be even worth implementing and documenting. In the worst case scenario the documentation could just mention that it is possible to delete the LRU cache file from the user config dir.

elijah-potter:
Thanks for the wonderful write up! I'm working on this right now, so you should see a PR for it later today.

My initial thoughts are to implement this in a similar manner to the file dictionaries—in fact, I may be able to reuse some of the code.

elijah-potter:
This should be implemented as of the latest version. Mind giving it a whirl?

kuukunen:
Seems to be working for me so far. Thank you so much! It feels great.
From my perspective this can be closed.

PR #1250, opened 2025-05-08
feat(ls): persist lint ignores

elijah-potter:
# Issues 

#1220

# Description

This PR allows `harper-ls` users to persist lint ignores across sessions. This works by saving the relevant context hashes to the data directory on the system—similar to how we handle file dictionaries.

I've also taken the opportunity to refactor some of the now-shared logic between ignored lints and file dictionaries.

# How Has This Been Tested?

This has been tested manually.

# Checklist

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -220,7 +220,7 @@ impl Backend {
         );
 
         let mut doc_lock = self.doc_state.lock().await;
-        let ignored_lints = self.load_ignored_lints(url).await?;
+        let ignored_lints = self.load_ignored_lints(url).await.unwrap_or_default();
 
         let doc_state = doc_lock.entry(url.clone()).or_insert_with(|| {
             info!("Constructing new LintGroup for new document.");
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -10,7 +10,8 @@ use harper_comments::CommentParser;
 use harper_core::linting::{LintGroup, LintGroupConfig};
 use harper_core::parsers::{CollapseIdentifiers, IsolateEnglish, Markdown, Parser, PlainEnglish};
 use harper_core::{
-    Dialect, Dictionary, Document, FstDictionary, MergedDictionary, MutableDictionary, WordMetadata,
+    Dialect, Dictionary, Document, FstDictionary, IgnoredLints, MergedDictionary,
+    MutableDictionary, WordMetadata,
 };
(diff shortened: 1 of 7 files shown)


RunDevelopment:
What's the default location for the file this is saved to?

kuukunen:
> What's the default location for the file this is saved to?

Based on the code, looks like in the data folder, the same as the per-file dictionaries.

So, for example: `~/.local/share/harper-ls/ignored_lints/file%path.ext%`

RunDevelopment:
The problem with that is that on Windows with the VSC extension, per-file dicts create files like `%Users%micha%Git%repo%posts%my files.md%` at the root of my repos. These files just spam the repo and aren't usable for anyone else, so per-file dicts aren't usable for me.

So if ignore-lints create files in the same way, they won't be usable either.

kuukunen:
> The problem with that is that on Windows with the VSC extension, per-file dicts create files like `%Users%micha%Git%repo%posts%my files.md%` at the root of my repos. These files just spam the repo and aren't usable for anyone else, so per-file dicts aren't usable for me.
> 
> So if ignore-lints create files in the same way, they won't be usable either.

That sounds like an issue in the VSC extension, though. And by root, you mean all the files in the actual root? Even if it resolves the data dir as the repo root for some reason, I would imagine that it should at least put them in `file_dictionaries` directory, unless I misunderstood the code. In which case you could put that directory in .gitignore. Or if you don't want to push that into the repo, .git/info/exclude or core.excludesFile. You could even set the ignore globally for all repos.

mcecode:
> The problem with that is that on Windows with the VSC extension, per-file dicts create files like `%Users%micha%Git%repo%posts%my files.md%` at the root of my repos.

That shouldn't be the case. Currently, the implementation uses [dirs](https://crates.io/crates/dirs) to find the local data dir to save the file dictionaries there. For Windows, it's whatever `FOLDERID_LocalAppData` is set to. Maybe that variable isn't set on your system?

> That sounds like an issue in the VSC extension, though.

It shouldn't be, if there's an issue, it would be in `harper-ls` since that's what creates the files, not the extension.

elijah-potter:
> The problem with that is that on Windows with the VSC extension, per-file dicts create files like %Users%micha%Git%repo%posts%my files.md% at the root of my repos. These files just spam the repo and aren't usable for anyone else, so per-file dices aren't usable for me.

This is huge problem. If there isn't an issue open already for this, would you make one?

Issue #1216, opened 2025-05-02
Review and rethink affixes


RunDevelopment:
After making #1215, I look at all other affixes as well to understand how they are used. What I found is listed below.

I'm making this issue, because I want to talk about the usage of affixes. I believe they are currently both overused (re-fine) AND underused (event-ful). Because some affixes also assign metadata, this causes problems for the PoS tagger (e.g. "her" being an adjective). Something needs to change.

Specifically, I think we consider the following changes:

1. Remove `X` (-ions) and `J` (-ings). These plural forms are unnecessary, because the `-ion`/`-ing` word can just use `S` affix.
2. Clean up the silly/incorrect usages of affixes.
3. Either use or remove rarely-used affixes.

## Prefixes

- `C`: `de-` prefix. Used 247 times, almost always sensibly.

  The weirdest usages are: def, dept, dean, deed, deer, debit, decay, delay

- `A`: `re-` prefix. Used 589 times, almost always sensibly.

  Funny examples: rev, reed, react, rebut, reeve, relay, reply, resin, reactor, refine, reflex, resent, resign, retire, return, revert

- `K`: `pro-` prefix. Used 53 times, almost exclusively in silly ways.

  Funny examples: prob, product, profile, profit, program, prolapse, prolong, prom, promote, propane, propose, proposition, protest, proverb

- `E`: `dis-` prefix. Used 218 times, rarely in silly ways.

- `U`: `un-` prefix. Used 49 times. I couldn't find any silly usages.

- `F`: `con-` prefix. Used 75 times. Used mostly sensibly.

  Funny examples: cone, conch, conduct, confine, confuse, consort, contact, content, contest, context, contour, confront, conserve

- `I`: `in-` prefix. Used 335 times. Used mostly sensibly.

  Funny examples: ink, inn, into, inch, input, invest, incite

## Suffixes

- `L`: `-ment` suffix. Used 223 times, almost always sensiblly.

  The weirdest usages are: comment, garment (clothing), vestment (clothing), abutment, basement, casement (window), statement

  The weirdest usages are: disaster, disc, disco, discard, discount, discourse, discuss, disease, disgust, dismiss, display, [disport](https://en.wiktionary.org/wiki/disport), distill, distress

- `Y`: `-ly` suffix. Used 2,466 times, almost always sensibly.

  The weirdest usages are: rely, Lily, holy (holy is a separate entry too), Polly, Sally, stably, Bradly, Philly

[...]

PR #1251, opened 2025-05-08
feat: split affixes and properties and support propagating properties

### LINKED ISSUE #969: The annotation `6` for swears/curses does not propagate
**Describe the bug**

In `dictionary.dict` the annotation flag `6` indicates bad words.
Other flags are for affixes for derivation and inflection, such as `S`, `D`, `G` for -(e)s, -ed, -ing.
But when using `6` along with those flags, the 'bad word' property does not propagate.

**To Reproduce**
1. Type your favourite bad word in the uninflected dictionary form: `fart`
2. Notice it's flagged with "Try to avoid offensive language".
3. Now type inflected forms: `farts`, `farted`, `farting`.
4. Notice none of these are flagged.

**Expected behavior**
Inflected forms of offensive words should be just as offensive as the singular noun form or present indicative form.

**Screenshots**

Image

**Platform**
Windsurf (VS Code form)

RunDevelopment:
# Issues 
Fixes #969
Related to #1216

# Description
This PR implements 3 main changes:

1. I split `affixes.json` into affixes (prefixes/suffixes) and properties. This makes `affixes.json` a lot cleaner, since affixes are no longer (mis)used to assign metadata. This also allows for affixes and properties to carry different data.
2. Add support for propagating properties. We want certain properties (such as being a swear) to propagate to derived words. So I added a `propagate: bool` field for properties, which implements exactly this functionality.
3. Use `"propagate": true` for swears and dialects. This fixes #969 and the problem that harper didn't know that derived words keep their dialect (e.g. colours).

    To that end, I also had to make minor changes to the dictionary. "ax" and "axe" were assigned a dialect incorrectly. It's not that "axe" is British, but that "ax" is American ([wiktionary](https://en.wiktionary.org/wiki/axe#English)). I rather unorthodoxly fixed the derived words of "programme".

I also added a new tag for swears to the PoS tagger.

# Demo
See snapshots

# How Has This Been Tested?
With snapshots

# Checklist


hippietrail:
I'm in the middle of a big road trip right now so can't dig in and check but I seem to recall that even some of the POS properties might not have been propagating. At least I'm pretty sure `just getmetadata` on inflected words was returning no or almost no metadata surprisingly often.

elijah-potter (approved):
Accidentally hit "approve". Don't mind this.

elijah-potter (changes requested):
This is fantastic work. This has been sorely needed, so thank you for pushing it through.

One note though: would you mind updating `packages/web/src/routes/docs/contributors/dictionary/+page.md` to include details from this PR? Nothing complex, just a short blurb on the diff between properties and affixes. Thanks!

PR #1256, opened 2025-05-09
feat: adieu→ado

hippietrail:
# Issues 
N/A

# Description

"Adieu" is a German word meaning farewell but it's become common for people to use it in place of the English word "ado" which means "fuss". I usually only occurs in a couple of contexts, so we can use `phrase_corrections`.

- further ado
- much ado

# How Has This Been Tested?

I added a unit test for each variant using a sentence found on GitHub.

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_corrections.rs
+++ b/harper-core/src/linting/phrase_corrections.rs
@@ -1212,7 +1212,19 @@ pub fn lint_group() -> LintGroup {
             ["definite articles"],
             "The correct term for `the` is `definite article`.",
             "The name of the word `the` is `definite article`."
-        )
+        ),
+        "FurtherAdo" => (
+            ["further adieu"],
+            ["further ado"],
+            "Don't confuse the German `adieu`, meaning `farewell`, with the English `ado`, meaning `fuss`.",
+            "Corrects `adieu` to `ado`."
+        ),
+        "MuchAdo" => (
+            ["much adieu"],
+            ["much ado"],
+            "Don't confuse the German `adieu`, meaning `farewell`, with the English `ado`, meaning `fuss`.",
+            "Corrects `adieu` to `ado`."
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/phrase_corrections.rs
```diff
@@ -1212,7 +1212,19 @@ pub fn lint_group() -> LintGroup {
             ["definite articles"],
             "The correct term for `the` is `definite article`.",
             "The name of the word `the` is `definite article`."
-        )
+        ),
+        "FurtherAdo" => (
+            ["further adieu"],
+            ["further ado"],
+            "Don't confuse the German `adieu`, meaning `farewell`, with the English `ado`, meaning `fuss`.",
```
ccoVeille, line 1219: I would say Adieu is French.

French here.

Adieu is a old French word made from the contraction of "à Dieu", which is literatly "to God". Which is a way to say good bye, whish wishing someone to be protected by God.

I'm not saying Adieu is not used in German, I'm just raising the attention. Adieu comes from French.

https://de.m.wiktionary.org/wiki/adieu

> um 1600 von französisch adieu entlehnt

Another way to consider it is that is comes from `ad Deum`, Latin


https://fr.m.wiktionary.org/wiki/adieu

hippietrail: Thanks! That was my first thought and then I couldn't remember using it in French but knew I used it in German-speaking Switzerland. I think I'll change the wording to `French/German` or `French and German`.

ccoVeille (approved):

harper-core/src/linting/phrase_corrections.rs
ccoVeille: You're welcome

elijah-potter (approved):

Issue #403, opened 2025-01-17
feat: Flag adjective "everyday" used in place of adverbial phrase "every day"


hippietrail:
This has become very common among both native and non-native English speakers.
Surprisingly, the browser only detects adverb used in place of the adjective, which is less common.

This is something I do everyday. → every day (compare every night, every year, etc)
It's nothing special, just an every day thing. → everyday (meaning mundane, typical, not special)

<img width="357" alt="Image" src="https://github.com/user-attachments/assets/a98f6311-5e3c-4978-9084-cd23c6b192f2" />

hippietrail:
This was implemented in #1265 - open new issues for false positives and/or negatives.

PR #1265, opened 2025-05-13
feat: Every day vs Everyday linter

hippietrail:
# Issues 
Implements #403

# Description

Attempts to fix the very common error of using the adjective "everyday" (normal, mundane, not special) in place of the adverb "every day" (repeated daily).

Also attempts to fix the less common error of using the adverb "every day" when it should be the adjective "everyday".

Some common cases can't be covered yet until we can match end-of-chunk since attributive-only adjectives like "everyday" must come before a noun and cannot be the last word of a sentence.

Added a new shorthand `t_any()` to `SequencePattern` for `then_anything()` just like `t_aco` and `t_ws`.

# Demo

# How Has This Been Tested?

I've added a bunch of unit tests using sentences I found on GitHub, including many tests against potential false positives.

I don't yet have enough test cases so some errors are surely not getting flagged. Please submit them!

# Checklist


harper-core/src/linting/everyday.rs
```diff
@@ -0,0 +1,318 @@
...
+            Box::new(
+                SequencePattern::default()
+                    .then_any_word()
+                    .t_ws()
+                    .then(every_day.clone()),
+            ),
+            Box::new(|tok: &Token, src: &[char]| is_article(tok, src)),
+        ]);
+
+        // Can we detect all mistakes with just one token before or after?
```
elijah-potter, line 72: I'm a little confused what purpose these comments serve.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -473,15 +463,6 @@ Message: |
 
 
 
-Lint:    WordChoice (31 priority)
-Message: |
```
elijah-potter, line 477: I like to see these go away.

elijah-potter (dismissed):
Some tweaks, but otherwise close to being ready.

harper-core/src/linting/everyday.rs
hippietrail: It expresses that I'm not confident that this implementation will be able to catch all mistakes yet and that perhaps more pre or post context will be needed. But more real-world text would probably bring deeper false positives or misses to light as I haven't seen examples yet and didn't think any up.

harper-core/src/linting/everyday.rs
```diff
@@ -0,0 +1,318 @@
...
+    fn correct_adjective_at_end_of_chunk() {
+        assert_suggestion_result(
+            "This is something I do everyday.",
+            Everyday::default(),
+            "This is something I do every day.",
+        );
+    }
+
+    #[test]
+    fn correct_adverb_after_article_before_noun() {
```
elijah-potter, line 176: I'd love to see some more cases that test for false-negatives.

hippietrail: I've messed up my sleeping pattern and it's almost 4am. Can you give me an example so I'm sure I know what you mean? Test for something the linter doesn't flag but actually is an error?
Would that involve
```rs
#[should_panic]
```

elijah-potter: Sorry, I should have explained a bit better.

You have a lot of cases that make sure the linter does not throw any false-positives. You have added some that test for false-negatives (which would be the linter doesn't do anything) by running `assert_suggestion_result` with some example text.

Would you add a few more tests with `assert_suggestion_result` that make sure the linter catches the cases we expect it to?

hippietrail: Oh so "true positives"? Sure I'll take a look soon.

elijah-potter (approved):
Fantastic work. I wonder if we can extract anything from here into a more general module. Do any of the heuristics here apply to other open/closed compounds?

hippietrail:
> Fantastic work. I wonder if we can extract anything from here into a more general module. Do any of the heuristics here apply to other open/closed compounds?

I keep wondering about that kind of thing too. For now I think the best is more data from real-world mistakes and more false positives and keep looking for patterns.

It comes down to which parts-of-speech the right and wrong versions are and which environments each can occur in. And resolving ambiguous parts-of-speech of words needed for heuristics to work.

Most compounds are compound nouns but here the open compound is an adverb and the closed compound is an adjective. If there are other pairs like that we could use this. I'm not sure if there are?

Issue #1269, opened 2025-05-13
Various corrections


uncenter:
## Description

Words:
- [x] `toolchain`
- [x] `deserialization` (opposite of "serialization", which doesn't trigger an error)
- [x] `parseable`/`parsable` (something that can be parsed - debatable if this is a real word? https://en.wiktionary.org/wiki/parsable)
- [x] `roadmap`
- [x] `timeframe`
- [x] `symlink`
- [x] `vendored` ("this folder is for vendored code", as in including/copying third-party code locally - not technically a word in most dictionaries, pretty common in dev spaces though?)
- [x] `encodings` (plural of `encoding`, which can be plural; "documents can be written in various encodings")

Acronyms (e.g. https://github.com/Automattic/harper/pull/1178):
- [x] `UX` (user experience)
- [x] `DX` (developer experience)
- `PR` ("pull request" or "public relations", see my earlier draft PR linked above)
- [x] `WIP` (work-in-progress)
- [x] `JS` (JavaScript)

Bug?
- ~~Surrounding text in inline Markdown syntax prevents context from working: Joe's (`Joe's`) is fine, but _Joe_'s (`_Joe_'s`) is not because I guess Harper doesn't see that the `'s` is connected to the word `Joe`? Stripping out inline syntax like that seems like an easy fix but I'm not sure if it would break other things.~~ https://github.com/Automattic/harper/issues/385/https://github.com/Automattic/harper/issues/187
- "This test is ignored" has an odd word choice suggestion for using "testis" instead of "test is"?

Proper nouns:
- ~~"StackOverflow" should be corrected to "Stack Overflow"~~ https://github.com/Automattic/harper/issues/1066
- ~~"VSCode" should be "VS Code" or "Visual Studio Code".~~ https://github.com/Automattic/harper/issues/643

Other:
- How are "units" handled? E.g. "gb" for gigabytes after a number in "the computer has 512gb of storage" is flagged. "512GB" is fine, with the unit being capitalized, fwiw, but that isn't the suggested correction for "512gb".
- [x] "UTF-8", "UTF-16" are (falsely) flagged.
- ~~The "(s)" method around if something could be both singular and plural; "Your configuration file(s) have been successfully migrated."~~ https://github.com/Automattic/harper/issues/476

hippietrail:
Related to #643

hippietrail:
I've added all the words to my next Dictionary Curation PR which I'll link here.
[...]

PR #1270, opened 2025-05-14
chore: new words, annotation corrections, adj-of-a exceptions

hippietrail:
# Issues 

#1269

# Description

Added new words
Corrected some affix annotation
Added two more exceptions to the adjective-of-a linter for words which are technically adjectives too but much more commonly nouns so don't get used in the way the linter corrects.

# How Has This Been Tested?

I added a unit test for the adjective-of-a exception.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -52939,8 +52939,8 @@ uh-oh
 un-American/5
 user-friendly/5
 UTF-16/1M
-UTF-8/1M
 UTF-32/1M
+UTF-8/1M
 voice-over/1MS
 von Neumann/2
 walk through/4
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -13485,6 +13485,7 @@ avert/~4GDS
 avg/~1
 avian/~51
 aviary/~1SM
+aviate/41DSG
 aviation/~1M
 aviator/~1MS
 aviatrices/19
@@ -24581,6 +24582,7 @@ fisher/~1M
 fisherman/~1M
 fishermen/~9
 fishery/~1SM
+fisheye/1SM
 fishhook/14SM
 fishily/j
 fishiness/1M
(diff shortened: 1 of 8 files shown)


harper-core/dictionary.dict
```diff
@@ -18256,6 +18260,7 @@ companionably/j
 companionship/~1M
 companionway/1MS
 company/~14SM
+companywide/5
```
uncenter, line 18263: Is this a word? I always thought it was supposed to be "company-wide", using the [-wide suffix](https://www.collinsdictionary.com/dictionary/english/wide#wide__2).

harper-core/dictionary.dict
```diff
@@ -52827,6 +52856,7 @@ red wolf/1M
 red wolves/9
 right-hand/5
 right-wing/5
+road map/1MS
```
uncenter, line 52859: I've only seen it written as one word together in technological contexts - https://github.com/search?q=%22roadmap%22&type=code (3.5m) is used 24x more often than https://github.com/search?q=%22road+map%22&type=code (145k). "road map" seems to refer to a literal map of roads whereas roadmap is more distinctly a plan of action? https://en.wikipedia.org/wiki/Technology_roadmap

harper-core/dictionary.dict
hippietrail: I was surprised too but I try to check every word against the English Wiktionary. Wiktionary isn't perfect though and I don't always check "proper" dictionaries and didn't this time.

harper-core/dictionary.dict
hippietrail: I'm never sure of these any more. I'm pretty good at spelling but don't trust my intuition for compounds and always check dictionaries. When I don't trust Wiktionary I usually check: Chambers, Collins, OED, Oxford (same company but their dictionaries can differ), and Websters.

elijah-potter: I suspect both are valid. The closed compound means something different from the open compound.

elijah-potter (approved):

harper-core/dictionary.dict
hippietrail: "roadmap" overtakes "road map" in popularity around 2000 on Google Ngrams so that will be the modern sense we use in the tech world. I added it in the techy section with a comment.

PR #1271, opened 2025-05-14
feat(comments): add Kotlin support

elijah-potter:
# Issues

Per #79

# Description

I did not add Kotlin to the VS Code extension, since it does not appear that VS Code supports the language out of the box. I'd love to be proved wrong, @mcecode.

It's been added to `harper-ls`, though, which should pave the way for Neovim, Helix, Emacs, etc.

# How Has This Been Tested?

Unit and integration tests.

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -842,6 +842,7 @@ dependencies = [
  "tree-sitter-haskell",
  "tree-sitter-java",
  "tree-sitter-javascript",
+ "tree-sitter-kotlin",
  "tree-sitter-lua",
  "tree-sitter-nix",
  "tree-sitter-php",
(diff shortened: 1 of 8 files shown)


mcecode:
> it does not appear that VS Code supports the language out of the box

You are correct, it does not. However, I do think you can still add it to the VS Code extension because even if out of the box VS Code would just ignore the language ID, once a user installs an extension that gives Kotlin support, VS Code would then recognize that language ID, enabling the Harper extension to pass Kotlin files to `harper-ls`.

elijah-potter:
'Kay. I've skipped the integration test, since installing the plugin may complicate our pipeline.

Does this look good?

mcecode (approved):
Yup, LGTM.

PR #1279, opened 2025-05-15
fix: add special case for "how come"

hippietrail:
# Issues 
N/A

# Description

The `HowTo` linter had a false positive. It was flagging "how come" (which means "why") to add "to" after "how".

# How Has This Been Tested?

I added a unit test to let "how come" go unflagged.

I also removed the `ignore` from the unit test for "how did you" since it works now.

# Checklist


harper-core/src/linting/how_to.rs
```diff
@@ -23,11 +23,15 @@ impl Default for HowTo {
...
-                    tok.kind.is_auxiliary_verb()
+                    if tok.kind.is_auxiliary_verb()
                         || tok.kind.is_adjective()
                         || tok.kind.is_present_tense_verb()
-                        // Special case for "did" as in "how did you do that?"
-                        || tok.span.get_content_string(src).eq_ignore_ascii_case("did")
+                    {
+                        true
+                    } else {
+                        let normed = tok.span.get_content_string(src).to_ascii_lowercase();
```
elijah-potter, line 32: This is not a cheap comparison to do. Is the metadata for `come` and `did` correct?

elijah-potter (approved):

harper-core/src/linting/how_to.rs
hippietrail: It needs to be a text comparison because "how come" is a term meaning "why" which doesn't just depend on the metadata. I always do the string fetching last when all other tests fail, and if they're one or two characters you can do `.get_content()` and do `2^len` comparisons in a match.

I would really love to write some string ↔ [char] comparison functions though. I can't believe they don't already exist.

PR #1282, opened 2025-05-16
refactor: reduce false positive blacklist with new logic

hippietrail:
# Issues 
N/A

# Description

As pointed out by @elijah-potter the list of adjectives that caused false positives was getting unwieldy and gave the impression it would grow forever.

Instead we'll check for a small set of adjectives known to commonly occur in this construction.

Additionally I've identified more specific contexts it's very often used in that makes false positives much less common, but there's still a couple we blacklist.

The contexts are:
- as adj of a
- how adj of a
- that adj of a
- this adj of a
- too adj of a

Of these, "this" looks more likely to produce false positives, so we'll have to keep an eye and in the worst case may have to remove it from the list.

I added a bunch of new tests, most using sentences from GitHub.
I removed on questionable test which was failing but was probably ungrammatical in other ways.

This removes some false positives I hadn't realized were in the snapshots too!

# How Has This Been Tested?

New unit tests and ensuring all the old ones (minus that one bad one) still pass.

# Checklist

### DIFF
--- a/harper-core/src/linting/adjective_of_a.rs
+++ b/harper-core/src/linting/adjective_of_a.rs
@@ -5,64 +5,60 @@ use crate::{Document, Span, TokenStringExt};
 #[derive(Debug, Clone, Copy, Default)]
 pub struct AdjectiveOfA;
 
-const FALSE_POSITIVES: &[&str] = &[
-    // Different valid constructions.
-    "all",
-    "emblematic",
-    "equivalent",
-    "full",
-    "fun",
-    "illustrative",
-    "inside",
-    "more",
-    "much",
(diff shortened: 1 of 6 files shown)


elijah-potter (approved):
Thanks for the revision. It looks _way_ sturdier now. Look at the decimated false positives in the snapshot tests!

hippietrail:
> Thanks for the revision. It looks _way_ sturdier now. Look at the decimated false positives in the snapshot tests!

It's kind of what I expected when the blacklist started to grow. Sometimes the patterns are hard to identify until you've handled a number of cases, and some take more cases than others.

Issue #468, opened 2025-01-22
Flag mixing up "apart" and "a part"


hippietrail:
This one is particularly egregious because they're pretty much opposites.

I'll add some examples as I spot them in the wild.

hippietrail:
Implemented in #1284

PR #1284, opened 2025-05-17
feat: correct mixing up "apart" and "a part"

hippietrail:
# Issues 

#468

# Description

Tries to fix "apart" used when "a part" is correct and "a part" when "apart" is correct.
Some cases are too difficult to disambiguate correct vs incorrect so far, those are still flagged but the message also advises when to leave it without "correcting".

# How Has This Been Tested?

I added a bunch of tests, some made up for simple cases, some mined from GitHub for real-word examples.

I also added comments with other sentences from GitHub where I couldn't even figure out what they were trying to say.

# Checklist

### DIFF
--- a/harper-core/src/linting/a_part.rs
+++ b/harper-core/src/linting/a_part.rs
@@ -1,5 +1,5 @@
 use crate::{
-    Lrc, Token, TokenStringExt,
+    Token, TokenStringExt,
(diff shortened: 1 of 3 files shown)


elijah-potter (approved):
Stellar stuff.

PR #1292, opened 2025-05-19
fix: 'a very' no longer flagged to change to 'Avery'

hippietrail:
# Issues 
N/A

# Description

I had noticed for some time that `a very` was sometimes flagged and the suggested fix was `Avery`.
I finally investigated and the context was after `that's`, which the `implied_ownership_compound_nouns` linter treated as a possessive when it's actually a contraction.

While checking things out I realized the special case for `let's` only worked with the straight typewriter apostrophe but not the curved typography apostrophe. I added apostrophe and case normalization before checking the special case exceptions.

This fixed a couple of lints in the snapshots too.

# Demo

Before: 
Screenshot 2025-05-19 at 3 38 01 pm

After: 
Screenshot 2025-05-19 at 5 59 21 pm

# How Has This Been Tested?

I added a unit test for `that's` and a second one for `let's` that used the curved apostrophe: `let’s`.

# Checklist

### DIFF
--- a/harper-core/tests/text/linters/The Great Gatsby.snap.yml
+++ b/harper-core/tests/text/linters/The Great Gatsby.snap.yml
@@ -3404,15 +3404,6 @@ Suggest:
 
 
 
-Lint:    WordChoice (63 priority)
-Message: |
-    2212 | “Well!” I inspected them. “That’s a very interesting idea.”
-         |                                   ^~~~~~ The possessive noun implies ownership of the closed compound noun “Avery”.
-Suggest:
-  - Replace with: “Avery”
-
-
-
 Lint:    Spelling (63 priority)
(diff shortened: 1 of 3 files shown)


harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -3404,15 +3404,6 @@ Suggest:
 
 
 
-Lint:    WordChoice (63 priority)
-Message: |
-    2212 | “Well!” I inspected them. “That’s a very interesting idea.”
```
elijah-potter, line 3409: I'm glad these pass now.

elijah-potter (approved):
Good stuff.

PR #1296, opened 2025-05-19
chore: sort and dedup lint group rules

hippietrail:
# Issues 
N/A

# Description

While trying to learn how to start using `lint_group` I noticed that the rules were partly but not fully alphabetized. And then I noticed the `TheHowWhy` rule was added twice! Once as a struct and once as a pattern. I'm surprised that didn't upset clippy or anything else.

# How Has This Been Tested?

`cargo test` still has no failures.

# Checklist


elijah-potter (approved):
Good tweak.

PR #1318, opened 2025-05-23
feat: add "except if" and "except when" for "unless if"

hippietrail:
# Issues 
N/A

# Description

I was going to implement a fix for the apparently common mistake `unless if`...
But it's already implemented (-:
However, it was only suggesting one of the three possible corrections: "unless"
This little PR adds "except if" and "except when" suggestions.

# How Has This Been Tested?

I added two new unit tests using sentences from GitHub.

# Checklist


harper-core/src/linting/phrase_corrections.rs
```diff
@@ -1243,6 +1243,12 @@ pub fn lint_group() -> LintGroup {
             "The more standard, less colloquial form is `take it personally`.",
             "Corrects `take it personal` to `take it personally`."
         ),
+        "Unless" => (
```
elijah-potter: Since these two rules have the same name, I believe the second one will override the first. Maybe remove the first one to clean up the code and keep the same behavior?

elijah-potter (approved):

PR #1320, opened 2025-05-24
feat:amount for→account for/amount to

hippietrail:
# Issues 
N/A

# Description

I'd never seen this before yesterday but it occurs all over the place when you look for it.
"amounts for" / "amount for" seem to be used for both "accounts for" and "amounts to".

# How Has This Been Tested?

I made a bunch of unit tests with sentences sourced from around the web.

# Checklist


elijah-potter (approved):
Testing on my end looks good!

Issue #1001, opened 2025-04-01
Precommit failing intermittently


elijah-potter:
**Describe the bug**

Precommit will sometimes fail on the `untitled` buffer VS Code integration test.

**To Reproduce**

It's hard to reproduce, since it doesn't happen deterministically. My best advice: run `just precommit` over and over again until it fails. 

**Screenshots**

![Image](https://github.com/user-attachments/assets/858acad5-a8b1-49b9-b14c-fd414f4b9f06)

**Platform**

The VS Code integration tests.

**Additional context**

I think this started with #927. Any ideas @kiding?

kiding:
I think I know what the problem is. Originally in #927, it slept for 500ms after `await openUntitled`. #960 changed it to `waitForUpdatesFromOpenedFile` which only sleeps for 75ms. I remember in my testing it reliably failed when the sleep time was that short. I'll make a PR.

mcecode:
I set out in #960 to standardize the waits we use for different things, that's why I changed it. How much we need to wait for diagnostics to change really depends on the machine, what's running on it, etc. I didn't want to use long sleep times every time diagnostics change so that the VS Code tests won't take forever to finish since the startup with the downloads already takes a while, but if it's really become flaky and non-deterministic, should we just set 500ms as our base time every time we need to wait for diagnostics? That should be enough time so we don't get false positives.

kiding:
I’m okay with the idea to fix all delays to 500ms (precommit takes a LONG time anyways), but I think it’d be better to keep this for now and find a better solution. I’m hoping there has to be a decent way to make use of VS Code APIs to sense when the process is finished / the diagnostics are updated / harper-ls responded / etc.

mcecode:
> I’m hoping there has to be a decent way to make use of VS Code APIs to sense when the process is finished / the diagnostics are updated / harper-ls responded / etc.

I'm open to ideas, as far as I can remember, I couldn't really find an alternative to adding delays when I started writing the integration tests.

elijah-potter:
Would you be able to send a command to `harper-ls` and use the response as an indication that it's ready?

kiding:
Not entirely sure. I'll try to investigate more when I get a chance.

PR #1322, opened 2025-05-24
fix(vscode-plugin): make test reliable with `waitForDiagnosticChange`

kiding:
# Issues 
#1001

# Description
It has been a constant headache testing `vscode-plugin` since it's not quite easy to detect when VS Code actually finished updating diagnostics. #200 simply added a sleep function, but #1004 show it becomes problematic as the project gets complicated.

After some dabbling, it was clear to me that VS Code doesn't provide any API to detect it. The APIs don't provide any access to extension logs or intercept LSP response, which is understandable. It does provide the `languages. onDidChangeDiagnostics` listener, but it wasn't particularly useful since 1) it gets fired strangely too many times, 2) still not possible to detect when it stops, 3) sometimes doesn't get triggered at all. 

In this PR, `waitForDiagnosticChange` simply compares diagnostics before an action and after. It simply remembers the diagnostics before doing anything, and repeatedly check if the current diagnostics look different. Since it's possible they may be initially changed to `[]` then to an actual value, the function tries multiple times after a change is detected.

# Demo
N/A

# How Has This Been Tested?
Manually, https://github.com/kiding/harper/pull/4, and hopefully in this PR.

# Checklist

### DIFF
--- a/packages/vscode-plugin/src/tests/suite/helper.ts
+++ b/packages/vscode-plugin/src/tests/suite/helper.ts
@@ -27,10 +27,12 @@ export async function activateHarper(): Promise<Extension<void>> {
 	return harper;
 }
 
-export async function openFile(...pathSegments: string[]): Promise<Uri> {
-	const uri = Uri.joinPath(Uri.file(workspace.workspaceFolders![0].uri.path), ...pathSegments);
+export function getUri(...pathSegments: string[]): Uri {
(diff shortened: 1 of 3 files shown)


elijah-potter (approved):
Thanks for drafting a fix for this! I see no better way to test this than to just start using it for our workflows. Is it cool if I ping you if any problems arise?

PR #1335, opened 2025-05-27
feat(ls): configurable max length

### LINKED ISSUE #1310: allow linting large documents
**Describe the bug**
Filing an issue for this because it just took me way too long to find the undocumented hardcoded character limit of 120000 (in `harper-ls/src/backend.rs:345`) in the hopes someone else can find it easier or it will eventually make it's way into the configuration (and thus documentation).

**To Reproduce**
1. Create any document with more than 120000 characters
2. `harper-cli lint` works just fine (but might take a while)
3. try using an editor with `harper-ls`
4. doesn't work

**Platform**
`harper-ls`

86xsk:
# Issues 

Closes #1310

# Description

This PR makes the max document length configurable. Currently, it is set to a hardcoded limit of 120,000 bytes.

I have tried to update the documentation to reflect this change in all the relevant areas, however, it is possible that I missed some places. In addition, I'm not 100% confident that the updates I made to the documentation are technically correct, as I am not familiar with the configuration syntax of all our supported clients. I am also not sure if the changes I've made in `packages/vscode-plugin/package.json` are correct.

# How Has This Been Tested?

- `cargo test`
- `just precommit`

Brief manual testing with VS Code and Neovim indicated that the feature works as expected. `harper-ls` would stop providing lints when the file was larger than the max length, and would resume when the file was shrunk again.  

~~However, there is a notable issue in that the existing lints do not seem to go away when `harper-ls` stops linting the file. Repeatedly going over and under the `maxFileLength` limit causes many outdated and incorrectly positioned lints to pile up.~~ Never mind, further testing seems to indicate that I misunderstood the problem. The lints do not continue to pile up, but they do remain after `harper-ls` stops linting the file, and can become incorrectly positioned if additional edits to the file are made afterwards. If the file is shrunk enough that `harper-ls` resumes linting it, the out of sync lints disappear. I'll see if I can find a way to clear existing lints before we stop linting the file. (Done/fixed in 2f9cab4fd8e7c61b2c8680479021c399686f9ec8)

# Checklist

### DIFF
--- a/packages/vscode-plugin/package.json
+++ b/packages/vscode-plugin/package.json
@@ -122,7 +122,7 @@
 					"scope": "resource",
 					"type": "number",
 					"default": 120000,
-					"description": "Maximum length of file to be linted. If a file has more characters than this, it will not be linted."
+					"description": "Maximum length of file to be linted (in bytes). If a file is larger than this, it will not be linted."
 				},
 				"harper.userDictPath": {
 					"scope": "resource",
--- a/packages/vscode-plugin/package.json
+++ b/packages/vscode-plugin/package.json
@@ -118,6 +118,12 @@
 					"default": false,
 					"description": "Skip linting link titles."
 				},
+				"harper.maxFileLength": {
+					"scope": "resource",
+					"type": "number",
+					"default": 120000,
+					"description": "Maximum length of file to be linted. If a file has more characters than this, it will not be linted."
+				},
 				"harper.userDictPath": {
 					"scope": "resource",
 					"type": "string",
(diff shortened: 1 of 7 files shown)


packages/vscode-plugin/package.json
```diff
@@ -118,6 +118,12 @@
 					"default": false,
 					"description": "Skip linting link titles."
 				},
+				"harper.maxFileLength": {
```
mcecode, line 121: LGTM. The docs for the different editors also look good. Thanks for tackling this!

PR #1341, opened 2025-05-28
Implement server_info

jvoisin:
Implement a proper answer for ServerInfo

This PR supersedes #1326


harper-ls/src/backend.rs
```diff
@@ -41,6 +41,11 @@ use crate::ignored_lints_io::{load_ignored_lints, save_ignored_lints};
 use crate::io_utils::fileify_path;
 use harper_stats::{Record, Stats};
 
+/// Return harper-ls version
+pub fn ls_version() -> &'static str {
```
elijah-potter, line 45: I believe this function is a duplicate of what was added in #1338.

jvoisin: It is, but I moved it from main.rs to backend.rs

elijah-potter: Right you are! I've been caught not paying attention.

elijah-potter (approved):

PR #1344, opened 2025-05-29
feat(brill): train and use Brill tagger

elijah-potter:
# Description

This PR significantly improves the POS tagging system by building a framework to train and use a Brill tagger. This adds some overhead to the parse stage, but I've deemed an acceptable cost for better tagging (plus, there are plenty of opportunities for optimization).

# How Has This Been Tested?

The tagger itself (as of writing this comment) has been trained for two hours to achieve a 95% accuracy on the UD_English_GUM dataset. I plan to train further overnight or over the weekend.

Many existing tests fail, but I expect them to succeed once sufficient training has been completed.

# Checklist


hippietrail:
Should be interesting!

harper-pos-utils/src/brill_tagger.rs
```diff
@@ -145,7 +145,7 @@ impl BrillTagger {
                 let form = token.form.clone();
                 if let Some(last) = toks.last_mut() {
                     match form.as_str() {
-                        "'snt" | "'ll" | "'ve" | "'re" | "'d" | "'m" => {
+                        "'snt" | "'ll" | "'ve" | "'re" | "'d" | "'m" | "'s" => {
```
hippietrail: I'm pretty sure `"'snt"` here should be `"sn't"` or maybe just `"n't"`?

elijah-potter: Good eye. It should be fixed now.

harper-brill/src/lib.rs
```diff
@@ -1,18 +1,18 @@
 use lazy_static::lazy_static;
 use std::sync::Arc;
 
-pub use harper_pos_utils::{BrillTagger, UPOS};
+pub use harper_pos_utils::{BrillTagger, FreqDict, Tagger, UPOS};
```
hippietrail: A frequency dictionary could be handy in linters possibly. Such as when words match all the POS rules but are marginal, as in technically correct/grammatical but much more likely to be a typo for a more common word. I'll have to check it out.

elijah-potter: That struct is kind of a misnomer. It actually exists as a baseline tagger which just assigns each word its most common POS tag. Inside, it's just a <WORD> -> <UPOS> map.

elijah-potter:
Right now there are just two things left to do: train the models significantly more and set up `NominalPhrase` to use the new `np_member` tags.

PR #1345, opened 2025-05-30
feat: touristic → tourist / tourism / touristic

hippietrail:
# Issues 
N/A

# Description

"Touristic" is a legit English adjective in dictionaries and every native speaker understands it. But we almost never use it. On the other hand it's very commonly used in Euroenglish and by non-native speakers worldwide.

Depending on the context it can be replaced with either "tourist" or "tourism" used attributively or with "touristy".

This linter does its best to take into account all the contextual factors to choose which suggestions to offer.

I also added two new linter suggestion assertions:

- `assert_not_in_suggestion_result` ensures a certain result is not among the suggestions.
- `assert_good_and_bad_suggestions` takes both a whitelist and a blacklist and checks that all the results in the whitelist and none of those in the blacklist were suggested.

# How Has This Been Tested?

I added a bunch of unit tests using sentences found on GitHub.

# Checklist


harper-core/src/linting/mod.rs
```diff
@@ -284,10 +286,111 @@ mod tests {
...
+                        Expected to not find suggestion: \"{}\"\n\
+                        But found: \"{}\"",
+                        i, j, bad_suggestion, suggestion_text
+                    );
+                }
+            }
+        }
+    }
+
+    pub fn assert_good_and_bad_suggestions(
```
elijah-potter, line 337: Make sure you `#[track_caller]`.

elijah-potter: Also, leave a little description of what this assertion actually does. It looks quite useful and is likely to be used often. Let's make what it does clear to new contributors.

harper-core/src/linting/mod.rs
```diff
@@ -284,10 +286,111 @@ mod tests {
...
             // I think it's not possible for more than one suggestion to be correct
-            (true, true, false) => unreachable!(),
-            (true, false, true) => unreachable!(),
-            (false, true, true) => unreachable!(),
-            (true, true, true) => unreachable!(),
+            _ => {}
+        }
+    }
+
+    /// Asserts that none of the suggestions from the linter match the given text.
```
elijah-potter, line 306: Good.

harper-core/src/linting/mod.rs
hippietrail: > Make sure you `#[track_caller]`.

I've seen that in some of the asserts but haven't had an understanding of what it actually is.

harper-core/src/linting/mod.rs
hippietrail: I thought I commented it (-: I'm already forgetting some of the details since I finished it.

harper-core/src/linting/mod.rs
elijah-potter: It essentially makes any panics that occur within the function get reported as if they were fired at the location the function was called.

elijah-potter (approved):

PR #1347, opened 2025-05-30
fix: don't close compound if open is in dictionary

hippietrail:
# Issues 
N/A

# Description

Only suggest joining adjacent words into a closed compound noun if there isn't a two-word entry in the dictionary.

I also took the liberty of renaming `SplitCompoundWord` to `MergeableWords` as the old name is a bit ambiguous and makes it sounds like its job is to split compounds.

# How Has This Been Tested?

I added new unit tests to cover the three cases I could think of:
- Two words that can be compounded.
- Two words that can't be compounded due to the existing rules.
- Two words that can't be closed due to being in the dictionary as an open compound.

# Checklist


harper-core/src/patterns/mergeable_words.rs
```diff
@@ -0,0 +1,133 @@
...
+    ) -> Option<CharString> {
+        let a_chars: CharString = word_a.span.get_content(source).into();
+        let b_chars: CharString = word_b.span.get_content(source).into();
+
+        // First check if the two words with a space exist in the dictionary
+        let mut two_word_lexeme = a_chars.clone();
+        two_word_lexeme.push(' ');
+        two_word_lexeme.extend_from_slice(&b_chars);
+
+        if self.dict.get_word_metadata(&two_word_lexeme).is_some() {
```
elijah-potter, line 46: I feel that performing this check _in_ the pattern might be confusing to those reading the source for the compound noun rules. Could we move it to the predicate?

elijah-potter (approved):
> Only suggest joining adjacent words into a closed compound noun if there isn't a two-word entry in the dictionary.

Solid.

> I also took the liberty of renaming SplitCompoundWord to MergeableWords as the old name is a bit ambiguous and makes it sounds like its job is to split compounds.

That's a wildly better name.

PR #1348, opened 2025-05-30
feat: `harper-cli` command to rename affix annotation flags

hippietrail:
# Issues 

N/A

# Description

`harper-cli rename-flag` `<old>` `<new>` `<path-to-dictionary-dict-and-affixes.json>

Will change all occurrences of an affix annotation flag throughout the dictionary and affix files, withouth changing the order, format, spacing, etc.

It will check that the flag is valid, exists, and the new one is not already in use.
It will then verify that the `affixes.json` file is still valid JSON.

This makes it very easy to move an old flag to a new letter to make way for a new flag, move flags to more mnemonic letters, etc.

Backups of the old `dictionary.dict` and `affixes.json` are kept with the `.bak` file extension added.

A few tags are used in `just addnoun` so if you change those, don't forget to also change the `justfile`.

# Demo

# How Has This Been Tested?

Lots of manual testing.

# Checklist


elijah-potter (approved):
LGTM

PR #1361, opened 2025-06-02
test: `exact_phrase` and `exact_word` differ in case sensitivity

hippietrail:
# Issues 
N/A though perhaps this PR should also be considered an issue?

# Description

I added a new rule to one to verify that it's not case-sensitive. I renamed an existing rule in the other to verify that it is case-sensitive.

Perhaps one or both patterns/methods names need to be bikeshedded?

Are there other places where `exact` is used inconsistently?

# How Has This Been Tested?

I added one new unit test as mentioned above, and renamed an old one so the tests' names align.

# Checklist


elijah-potter:
I think you're right. When I initially made `ExactPhrase`, I was describing it in contrast to another pattern (I may have been called `ApproxPhrase` or something?). 

Maybe just `Phrase` would work?

hippietrail:
> I think you're right. When I initially made `ExactPhrase`, I was describing it in contrast to another pattern (I may have been called `ApproxPhrase` or something?).
> 
> Maybe just `Phrase` would work?

In some places such as `phrase_corrections` we deal in single words as well as phrases but I think in this case it's phrases only really. Could be something like `StaticPhrase` to make clear it has no changing parts such as tense or plurals. Probably not `SetPhrase` though as that has a specific meaning pretty close to Idiom so would be misleading. I think there are other words than `Static` that would work that aren't popping into my head right now. But yeah `Phrase` also works.

elijah-potter:
Alright. Feel free to bikeshed here. Whatever you decide, go ahead and make the change as part of this PR.

hippietrail:
> Alright. Feel free to bikeshed here. Whatever you decide, go ahead and make the change as part of this PR.

I think the term I was searching for was "fixed phrase", so I've gone with that. I got the LLM to generate comments in the style of the other `Pattern` implementations and to me it looks right but you might want to double check.

harper-core/src/patterns/fixed_phrase.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl FixedPhrase {
+    /// Creates a [FixedPhrase] from a plain text string.
+    /// Uses plain English tokenization rules.
+    pub fn from_phrase(text: &str) -> Self {
+        let document = Document::new_plain_english_curated(text);
+        Self::from_document(&document)
+    }
+
+    /// Creates a [FixedPhrase] from a pre-tokenized document.
```
elijah-potter, line 30: In the future, it's more idiomatic to put backticks right inside the square brackets.
For example: ``[`FixedPhrase`]``.

It's super pedantic, so I won't let it get in the way of this PR.

elijah-potter (approved):

Issue #402, opened 2025-01-17
feat: Flag verbs and nouns with similar endings interchanged


hippietrail:
I think these mistakes are mostly done by non-native English speakers:

This is what I intent to do next. → intend
Was that really your intend? → intent

I advice you to stop that. → advise
Do you have any advise for me? → advice

There are other word paris in this category. I'll add them when I spot them in the wild.

Finding some:
- It is not the specific software, but a **prove** of principal → It is not the specific software, but a **proof** of principal

My browser catches both nouns used in place of the verbs but only one of the verbs used in place of the noun, which in this case is actually far less common so not a bad choice.

<img width="345" alt="Image" src="https://github.com/user-attachments/assets/7f380707-acb4-493f-81a9-1c688406dda6" />

hippietrail:
Add emphasis vs emphasise / emphasize

hippietrail:
I'm just going to close this due to [#1565](https://github.com/Automattic/harper/pull/1565) and [#1703](https://github.com/Automattic/harper/pull/1703)

PR #1362, opened 2025-06-02
feat: fix noun that should be related verb

hippietrail:
# Issues 

Partial implementation of #402

# Description

It only corrects nouns to verbs, and only in a few contexts in which the noun is ungrammatical and the verb is grammatical.

The currently supported word pairs are: advice→advise, belief→believe, breath→breathe, effect→affect, and intent→intend

Proof→prove didn't seem to work well in this implementation.

# How Has This Been Tested?

I made lots of unit tests using sentences mined from GitHub. A couple of unexpected edge cases are `#ignore`d but might make good projects for anybody interested.

# Checklist


hippietrail:
> Tested on my machine, with some brief fuzzing. Looks to work as advertised 👍

There are probably more pairs I didn't think of and some probably have edge cases I didn't hit. But they will reveal themselves.

PR #1369, opened 2025-06-04
feat: Added ORGMODE parser

eloycoto:
I heavilly use ORGMODE[0] for my notes, and I also love harper. This PR add the parser for ORGMODE, I also add support to haper-ls and harper-cli.

I know that maybe it's not complete at 100%, but I'm going to work over needed changes.

0: https://orgmode.org/

How Has This Been Tested?
```
cargo run --bin harper-cli -- lint /tmp/example.org
```

Checklist


eloycoto:
@elijah-potter could you tell me if this is something interesting for you and if it is ok to merge after review?

harper-ls/src/backend.rs
```diff
@@ -337,6 +339,7 @@ impl Backend {
             "html" => Some(Box::new(HtmlParser::default())),
             "mail" | "plaintext" | "text" => Some(Box::new(PlainEnglish)),
             "typst" => Some(Box::new(Typst)),
+            "org" => Some(Box::new(OrgMode)),
```
elijah-potter, line 342: Is there a file extension for org-mode documents? If so, make sure you add it in the related function below.

harper-core/src/parsers/org_mode.rs
```diff
@@ -0,0 +1,300 @@
+use super::{Parser, PlainEnglish};
+use crate::{Span, Token, TokenKind};
+
+/// A parser that wraps the [`PlainEnglish`] parser that allows one to parse
```
elijah-potter: An initial look appears solid to me. I appreciate the test coverage you have so far, but would love to see even more end-to-end tests. Would it be possible to toss some in `harper-core/tests/test_sources/`? You might need to modify the macro in `harper-core/tests/run_tests.rs`.

elijah-potter (changes requested):
This is great work. I'm glad to see Org mode finally coming to Harper! 

I would love to include this in the next release, but want a bit more assurance to its stability. Could we get a few more tests?

harper-ls/src/backend.rs
eloycoto: I think so, at least it's the default one for sure.

harper-core/src/parsers/org_mode.rs
eloycoto: make sense, I would create a file, I will use this one as default:
https://github.com/nvim-orgmode/orgmode/blob/32ef9e95f43a6e951fb931b438372546a4f0c524/docs/index.org

And I'll update the macro (Looks ok ish)

harper-ls/src/backend.rs
elijah-potter: I was a bit confused. We're good on this.

harper-core/src/parsers/org_mode.rs
```diff
@@ -0,0 +1,300 @@
...
+#[derive(Default, Clone, Debug, Copy)]
+pub struct OrgMode;
+
+impl OrgMode {
+    // Check if a line starts with a header (starts with one or more '*')
+    fn is_header_line(chars: &[char], start: usize) -> bool {
+        if start >= chars.len() {
+            return false;
+        }
+        chars[start] == '*'
```
grantlemons:
```suggestion
        chars.get(start).is_some_and(|c| *c == '*')
```
I think something like this would be a bit more idiomatic

harper-core/src/parsers/org_mode.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    // Check if a line starts with a source block begin/end
+    fn is_source_block_marker(chars: &[char], start: usize) -> Option<bool> {
+        let line = Self::get_line_from_start(chars, start);
+        let line_str: String = line.iter().collect();
+        let line_str = line_str.trim();
+
+        if line_str.starts_with("#+BEGIN_SRC") || line_str.starts_with("#+begin_src") {
```
grantlemons: worth mentioning, using an enum might be clearer here

harper-core/src/parsers/org_mode.rs
```diff
@@ -0,0 +1,300 @@
...
+            // If we're in a source block, make everything unlintable
+            if in_source_block {
+                let line_end = Self::find_line_end(source, line_start);
+                tokens.push(Token {
+                    span: Span::new(line_start, line_end),
+                    kind: TokenKind::Unlintable,
+                });
+                cursor = line_end;
+                continue;
+            }
```
grantlemons: isn't this redundant?

harper-core/src/parsers/org_mode.rs
```diff
@@ -0,0 +1,300 @@
...
+/// Org-mode files.
+///
+/// Will ignore code blocks, source blocks, and other org-mode specific elements
+/// that should not be linted for prose.
+#[derive(Default, Clone, Debug, Copy)]
+pub struct OrgMode;
+
+impl OrgMode {
+    // Check if a line starts with a header (starts with one or more '*')
+    fn is_header_line(chars: &[char], start: usize) -> bool {
```
grantlemons: fyi this, and the other `is_xyz` functions don't really need to be in this `impl` block

you just end up with a lot more `Self::`s

harper-core/src/parsers/org_mode.rs
```diff
@@ -0,0 +1,300 @@
...
+        if matches!(
+            tokens.last(),
+            Some(Token {
+                kind: TokenKind::Newline(_) | TokenKind::ParagraphBreak,
+                ..
+            })
+        ) && source.last() != Some(&'\n')
+        {
+            tokens.pop();
+        }
```
grantlemons, line 172: Not entirely sure what this is for just by looking at it. Could you add a comment to explain?

grantlemons (changes requested):
It seems like your code all operates around lines, but you are always just passing around the entire source. You might want to rework your functions to accept lines, and then loop through the lines (split by newline) in your parse function.

elijah-potter:
I have opened it up on my editor. From my limited understanding of orgmode, it looks good. If there are further issues, can we ping you as a resident expert on orgmode?

eloycoto:
> I have opened it up on my editor. From my limited understanding of orgmode, it looks good. If there are further issues, can we ping you as a resident expert on orgmode?

If I'm still alive by them, yes! I'm happy to check issues if happens :-)

PR #1372, opened 2025-06-06
fix(core): handle words with multiple dialects

### LINKED ISSUE #1337: Internal word metadata only has a single dialect field but `dictionary.dict` correctly supports multiple dialects per entry
**Describe the bug**

If one word/variant/spelling is used on multiple dialects, it will be marked as an error in all but the first.

For example, "aluminium" and "aluminum" are currently both errors in American English. The same happens for some of the other English dialects.

This is the only word you'll see it with the dialect set to American as it's the only one that has the Canadian `@` before the American `<` dialect attribute flag. But with dialect set to Canadian or Australian it's extremely common and with it set to British it's somewhere in-between.

**To Reproduce**
1. If using the American dialect, just type "aluminum".
2. If using Canadian dialect, type any word that shares the American spelling and differs from the British spelling.
3. If using Australian dialect, type any word that shares the British spelling and differs from the American spelling.

**Expected behavior**
All correct spellings for the chosen dialect should be allowed without flagging as errors.

**Screenshots**
Image

**Platform**
All, including the Chrome extension as shown in the screenshot.

86xsk:
# Issues 
Closes #1337

# Description

This PR adds proper handling for words that correspond to multiple dialects. This fixes the issue reported in #1337, as the dialects defined for a word no longer conflict with each other.

This adds the `bitflags` dependency to store the activated dialects as a bitfield in `WordMetadata`. As an alternative, it should not be too difficult to switch to using something like a boolean array instead if preferable.

# How Has This Been Tested?
- `cargo test`
- `just precommit`

I did some brief manual testing with the modified builds of `harper-ls` and `harper-cli` with seemingly succesful results. Switching dialects in VS Code showed that "aluminum" would be linted as a spelling error if the dialect was set to British or Australian, whereas "Aluminium" would be linted as such if the dialect was set to Canadian or American. The snapshots look promising as well. Still, this does feel like a pretty wide-reaching change, so I can't be 100% certain this doesn't break something.

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,9 +179,12 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.8.0"
+version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8f68f53c83ab957f72c32642f3868eec03eb974d1fb82e453128456482613d36"
+checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "blanket"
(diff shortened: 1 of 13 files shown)


hippietrail:
Great stuff! I'm on the road right now so might not be able to try it out for a couple of days. But if you grep the repo for `ignore` I think there are at least a couple of tests that are ignored with the reason that this bug was blocking them... OK there's 3 in `spell_check.rs`

Issue #1365, opened 2025-06-03
url has no authority component


Fydon:
**Describe the bug**
Harper crashes on start. This is repeatable occurring in one of my workspaces that I use for work so can't share it, but isn't occurring in others.

From output window
```
[Info  - 9:36:02 AM] Server initialized!

thread 'main' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\tower-lsp-server-0.21.1\src\uri_ext.rs:79:46:
url has no authority component
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[Error - 9:36:03 AM] Server process exited with code 101.
[Info  - 9:36:03 AM] Connection to server got closed. Server will restart.
true
[Info  - 9:36:03 AM] Server initialized!

thread 'main' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\tower-lsp-server-0.21.1\src\uri_ext.rs:79:46:
url has no authority component
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[Error - 9:36:04 AM] Server process exited with code 101.
[Info  - 9:36:04 AM] Connection to server got closed. Server will restart.
true
[Info  - 9:36:04 AM] Server initialized!
[Error - 9:36:05 AM] Client Harper: connection to server is erroring.
write EPIPE
Shutting down server.

thread 'main' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\tower-lsp-server-0.21.1\src\uri_ext.rs:79:46:
url has no authority component
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[Error - 9:36:05 AM] Server process exited with code 101.
[Error - 9:36:05 AM] Client Harper: connection to server is erroring.
Cannot call write after a stream was destroyed
[Error - 9:36:05 AM] Request textDocument/codeAction failed.
  Message: Cannot call write after a stream was destroyed
  Code: -32099 
[Error - 9:36:05 AM] Connection to server got closed. Server will not be restarted.
[Error - 9:36:05 AM] Stopping server failed
  Message: Cannot call write after a stream was destroyed
  Code: -32099 
[Error - 9:36:05 AM] Stopping server failed
  Message: Cannot call write after a stream was destroyed
  Code: -32099 

```

[...]

PR #1385, opened 2025-06-10
feat: update untitled handling

mcecode:
# Issues 

#1365

# Description

Currently, the crash happens when `harper-ls` starts up because somewhere down the line `tower-lsp-server`'s `UriExt::to_file_path` gets called and panics on Windows machines when passed an untitled URI, as they don't have an authority.

I traced the flow and put checks so that `UriExt::to_file_path` doesn't get called with an untitled URI. This affects the ignore lints feature we currently have and changes it's flow a bit:

- When running `HarperIgnoreLint`, like for `HarperAddToUserDict` and `HarperAddToFileDict`, the extension now prompts the user to save the file before the command can be run.
- When loading ignored lints, if the URI passed is an untitled URI, it just returns an empty `IgnoredLints` instance.

Though tower-lsp-community/tower-lsp-server#53 should fix `tower-lsp-server`'s issue of panicking when passed a URI without an authority, I think this new handling of ignoring lints for untitled files is better and should be kept for the following reasons:

- Untitled files are meant to be ephemeral, so anything persistent that points to it would probably go stale fast.
- VS Code doesn't open untitled files with unique names, so when saving ignored lints, they get saved like so: `path/to/harper-ls/ignored_lints/Untitled-1%`, which means the same ignore file could point to multiple files since the first untitled file opened in any workspace is named `Untitled-1` and the number just gets incremented with every new untitled file opened.

There's still one instance where `harper-ls` could crash because of this bug, and that's on `did_change_configuration`. However, that should be resolved when tower-lsp-community/tower-lsp-server#53 gets merged. At the very least, `harper-ls` won't just crash on start up and changing settings/configs is rare enough that the churn probably won't be as bad while we wait for the fix.

# How Has This Been Tested?

Manually. Unfortunately, even if the logic should hold up, I don't have a Windows machine to test if this truly prevents the crash, so I'd appreciate it if anyone with a Windows machine can check.

# Checklist

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -105,6 +105,14 @@ impl Backend {
     }
 
     async fn load_ignored_lints(&self, uri: &Uri) -> Result<IgnoredLints> {
+        // VS Code's unsaved documents have "untitled" scheme
+        if uri
+            .scheme()
+            .is_some_and(|scheme| scheme.eq_lowercase("untitled"))
+        {
+            return Ok(IgnoredLints::new());
+        }
+
         Ok(load_ignored_lints(
             self.get_ignored_lints_path(uri)
                 .await
--- a/packages/vscode-plugin/src/extension.ts
+++ b/packages/vscode-plugin/src/extension.ts
@@ -30,15 +30,11 @@ const clientOptions: LanguageClientOptions = {
 		},
 		executeCommand(command, args, next) {
 			if (
-				['HarperAddToUserDict', 'HarperAddToFileDict'].includes(command) &&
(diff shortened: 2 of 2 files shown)


elijah-potter:
I don't have a Windows computer either. It seems like a relatively simple change, so if you're confident, I think we can push it out.

mcecode:
> I don't have a Windows computer either. It seems like a relatively simple change, so if you're confident, I think we can push it out.

I think we can try pushing it out to see if does fix #1365. At the very least, it's only a breaking change of ignore files for untitled files which should only affect a small percentage of users, if any.

PR #1391, opened 2025-06-12
feat:Shoot oneself into a leg → oneself in the foot

hippietrail:
# Issues 
N/A

# Description
I read an online comment where somebody wrote "shooting oneself **into** the foot". It turns out to be quite common.
I then checked other common mistakes and found "in the **leg**" and "in **a** foot" are also common.
In the case of "in the **feet**" it sounds like an error if the agent is singular but if the agent is plural then "**foot**" and "**feet**" both sound idiomatic:
- You guys shot yourselves in the foot.
- Prevents our users from shooting themselves in the feet.

How Has This Been Tested?
I made a bunch of unit tests from sentences mined from GitHub. Both correct and incorrect.

# Checklist


harper-core/src/linting/shoot_oneself_in_the_foot.rs
```diff
@@ -0,0 +1,264 @@
...
+            .iter()
+            .rev()
+            .take(5)
+            .rev()
+            .zip("elves".chars())
+            .all(|(c, s)| c.to_ascii_lowercase() == s);
+        let plural_foot = toks.get(8)?.kind.is_plural_noun();
+
+        // case-insensitive comparison without allocating a String
+        let is_in = prep.len() == 2
```
elijah-potter: Would it be possible to refactor this into a method on `CharStringExt`?

harper-core/src/linting/shoot_oneself_in_the_foot.rs
```diff
@@ -0,0 +1,264 @@
...
+
+pub struct ShootOneselfInTheFoot {
+    pattern: Box<dyn Pattern>,
+}
+
+impl Default for ShootOneselfInTheFoot {
+    fn default() -> Self {
+        let verb_forms = WordSet::new(&["shoot", "shooting", "shoots", "shot", "shooted"]);
+
+        let reflexive_pronouns = WordSet::new(&[
```
elijah-potter: I'd love to refactor this into a `SingleTokenPattern` or a member of `WordMetadata`. Is that feasible?

harper-core/src/linting/shoot_oneself_in_the_foot.rs
```diff
@@ -0,0 +1,264 @@
...
+    fn fix_a_foot() {
+        assert_suggestion_result(
+            "Shot ourselves in a foot, \"Wrong X-Request-Key\" error #589.",
+            ShootOneselfInTheFoot::default(),
+            "Shot ourselves in the foot, \"Wrong X-Request-Key\" error #589.",
+        );
+    }
+
+    #[test]
+    fn ignore_shoots_himself() {
```
elijah-potter, line 175: Great coverage.

elijah-potter (changes requested):
Looks good. There are just a few edits I'd love to be made to go the extra mile.

harper-core/src/linting/shoot_oneself_in_the_foot.rs
hippietrail: Yes! Exactly what I was asking you about the other day (-:

harper-core/src/linting/shoot_oneself_in_the_foot.rs
hippietrail: I think there's no way to correlate the different inflected forms of a word if that's what you're getting at. That's probably something we'll want though. Care would need to be taken to work with the common variations of attaching the endings that's already in the affix code but also for irregular and defective verbs, which would probably need a JSON file.
Same for nouns with irregular plurals. Might not be needed for irregular adjectives though since there's so few: good/better/best, bad/worse/worst, less/least, and more/most.

harper-core/src/linting/shoot_oneself_in_the_foot.rs
hippietrail: This part is now done if you can check it over.

harper-core/src/linting/shoot_oneself_in_the_foot.rs
hippietrail: I'll await further feedback before trying anything on this one since I'm not sure I get it yet.

elijah-potter: A list of reflexive pronouns would be pretty useful for some other linters. If you don't mind, refactoring it into it's own reusable pattern would make that sort of thing easier.

hippietrail: It is done!

harper-core/src/expr/reflexive_pronoun.rs
```diff
@@ -0,0 +1,99 @@
+use crate::{
+    Span, Token,
+    expr::{Expr, LongestMatchOf},
+    patterns::WordSet,
+};
+
+// These are considered ungrammatical, or are at least not in `dictionary.dict` but are commonly used anyway.
+// The tests below check if this changes so we can update this `Expr`
+const BAD_REFLEXIVE_PRONOUNS: &[&str] = &[
+    "hisself",
```
elijah-potter, line 10: Should we have a linter that checks for usage of "bad" reflexive pronouns?

hippietrail: > Should we have a linter that checks for usage of "bad" reflexive pronouns?

Quite possibly. Maybe more than one though. Some are bad to everybody and some are bad to some people, such as "themself". Some might be caught well enough by the spellchecker?

Feel free to file a feature request.

PR #1393, opened 2025-06-12
refactor(core): `Pattern` -> `Expr`

elijah-potter:
# Description

This PR represents a serious refactor of how the pattern system works, to enable more complex behavior (i.e. anchors and capture groups) and codify a more flexible system that can be iterated upon more quickly.

## How it Works

Up until now, almost all rules in Harper have been implemented as `PatternLinters`.
These worked by defining a `Pattern`:

```rust
pub trait Pattern: LSend {
    /// Check if the pattern matches at the start of the given token slice.
    ///
    /// Returns the length of the match if successful, or `None` if not.
    fn matches(&self, tokens: &[Token], source: &[char]) -> Option<usize>;
}
```

The linter would then run this pattern over a given sentence by passing in a sliding window of tokens.
We could string multiple patterns together (using a `SequencePattern`) by simply requiring that multiple patterns must pass, one next to the other.
Simple enough.

There are now two new traits: `Expr` and `Step`:

```rust
/// A common problem in Harper is that we need to identify tokens that fulfil a certain criterion.
/// An `Expr` is a way to express whether a certain set of tokens fulfil that criteria.
/// When supplied a specific position in a token stream, the job of an `Expr` is to determine the window of tokens (including the cursor itself) that fulfils whatever criteria the author desires.
/// It is then the job of another system to identify portions of documents that fulfil this criteria.
pub trait Expr: LSend {
    fn run(&self, cursor: usize, tokens: &[Token], source: &[char]) -> Option<Span>;
}


/// An atomic step within a larger expression.
///
/// Its principle job is to identify (if any) the next position of the cursor.
/// When cursor is moved, all tokens between the current cursor and the target position will be
/// added to the match group.
pub trait Step: LSend {
    fn step(&self, tokens: &[Token], cursor: usize, source: &[char]) -> Option<isize>;
}
```

There are a few observations to be had here.

First: A `Pattern` is simply a special case of a `Step`. A `Pattern` is a `Step` that can only move in the positive direction and does not have access to positional information.
In fact, through a blanket implementation, we can make every `Pattern` fulfill the `Step` trait automatically.
This means we can reuse all of our existing `Pattern`s.

```rust
impl<P> Step for P
where
    P: Pattern,
{
    fn step(&self, tokens: &[Token], cursor: usize, source: &[char]) -> Option<isize> {
        self.matches(&tokens[cursor..], source).map(|i| i as isize)
    }
}
```

Second: notice that the `Step` trait now returns a signed value.
This means we can match text _behind_ the current cursor. Cool, no?

Finally, did you notice that `Expr`s return `Span`s instead of indices?
That means we can create expressions that start in the middle of a pattern and expand outward in either direction.

There are plenty more juicy technical details, but I think they're better represented via code. 
If you're interested, read it.

## What This Means

As I said before, all existing `Pattern`s are also valid `Step`s which are also valid `Expr`s.
From now on, instead of chaining `Pattern`s in a `SequencePattern`, you'll be doing the same thing in a `SequenceExpr`.
Because of the greater flexibility you get with expressions, you'll also have access to the new `AnchorStart` and `AnchorEnd` expressions, which behave similarly to `$` and `^` in RegEx.

# How Has This Been Tested?

All existing tests pass, plus some new ones.

# Checklist


harper-core/src/expr/anchor_end.rs
```diff
@@ -0,0 +1,49 @@
+use crate::Token;
+
+use super::Step;
+
+/// A [`Step`] which will match only if the cursor is over the last non-whitespace character in stream.
+/// It will return that token.
```
hippietrail, line 6: I haven't tried to use any of this yet so I'm just thinking out loud. In regexes anchors are zero-width. So you'd match say `a`, `b`, `c`, `d`, and then the anchor. It's looking like this `Expr` would mean that to express the same you would ask to match `a`, `b`, `c`, and then d would be a both part of that and also the anchor?

As a potentially bad made up example you might want to match a verb that's both transitive and intransitive: (I [ws] eat (([ws] noun) <or> [anchor-end])

Another way to describe it might be that it's external context rather than part of the match. Or perhaps that because it's zero-width it makes no difference to the matched text returned whether it's part of the match or following it.

harper-core/src/expr/anchor_start.rs
```diff
@@ -0,0 +1,41 @@
+use crate::{Token, TokenStringExt};
+
+use super::Step;
+
+/// A [`Step`] which will match only if the cursor is over the first word-like of a token stream.
+/// It will return that token.
```
hippietrail, line 6: Same as for my comment above on `anchor_end.rs`. This one I've ran into and been blocked by more often. A common case is that a `Pattern` matches what we're looking for if it's not preceded by some context. But if we make a pattern like (before-context <ws> our interesing pattern) then it only matches if the pattern comes after a word and whitespace, whereas if our pattern is at the start of the chunk that means it is not preceded by 'some context', which means we want it. But our pattern couldn't do it.
If on the other hand we make an `EitherPattern` with (before-context <ws> our interesting pattern) on one branch and just (our interesting pattern) in the other, then it will match twice: First because it had the context, and then because the second branch is at the start of the sliding window.

But if the anchor has to do double-duty as the word "our" in (our interesting pattern) then this might be too restrictive.

harper-core/src/expr/repeating.rs
```diff
@@ -0,0 +1,84 @@
+use super::Expr;
+use crate::{Span, Token};
+
+/// An expression that will match one or more repetitions of the same expression.
+///
+/// Somewhat reminiscent of the `.*` operator in Regex.
```
hippietrail: `.*` is two things: `.` matches any atom. `*` matches **zero or more** of the previous atom. (`.` may or may not match linebreaks depending on various things).

The regex thing that matches **one or more** is `+` so `.+` would match one or more atoms.

Also both `*` and `+` are "greedy", meaning they match the longest possible match. To change this you follow either with the `?` operator to make them "lazy".

harper-core/src/expr/step.rs
```diff
@@ -0,0 +1,19 @@
+use crate::{LSend, Token, patterns::Pattern};
+
+/// An atomic step within a larger expression.
+///
+/// Its priciple job is to identify (if any) the next position of the cursor.
```
hippietrail: Typo: `priciple` should be `principle`.

harper-core/src/expr/word_expr_group.rs
```diff
@@ -0,0 +1,56 @@
+use hashbrown::HashMap;
+
+use super::first_match_of::FirstMatchOf;
+use super::{Expr, SequenceExpr};
+use crate::{CharString, Span, Token};
+
+/// A expression collection to look for expressions that start with a specific
```
hippietrail: Typo: `A expression` should be `An expression`.

harper-core/src/expr/anchor_end.rs
elijah-potter: This anchor _should_ behave as zero-width as well. 

For example, if you built `SequenceExpr::default().t_aco("word").then(AnchorEnd)` and ran it on `This is a word`, the resulting `Span` would only cover the final `word` token. 

I might be misunderstanding what you're trying to say here, so bear with me.

harper-core/src/expr/repeating.rs
elijah-potter: Yup. You're right. That doc-comment is pretty old (which means it's been wrong for a long time).

harper-core/src/expr/step.rs
elijah-potter: Yup. I accidentally had Harper off while writing much of this code.

harper-core/src/expr/anchor_end.rs
hippietrail: Oh OK I think I may have misunderstood the code comments then?

Issue #1401, opened 2025-06-16
Words missing from the dictionary/not recognized


insalt-glitch:
## Description

The following words are not known to Harper (I'm using the British version at the moment) (unknown -> spurious fix):
- gray -> gay/grab/grad
- discretization -> discrimination
- ansatz -> ersatz
- Plasmas -> Plasma's/Plasma
- unmagnetized -> demagnetized/magnetized
- collisionless -> collision's collision
- isothermal -> isotherm/isotherm's/isotherms
- wavefronts -> wavefront/waterfronts/lakefronts
- bitangent -> tangent
- magnetohydrodynamics -> No suggestion
- collisional -> collision/collision's/collisions
- symplectic -> simplistic/apoplectic
- inhomogeneous -> homogeneous
- extremum -> extreme
- dimensionality -> dimension
- separatrix -> separate
- wavenumber -> renumber
- isotropy -> isotope
- rotationally -> rotational
- gyrofrequencies ->frequencies

More uncommon words and thus forgivable errors:
- heliosphere -> hemisphere
- magnetosheath -> magnetosphere
- magnetopause -> magneto's

Finally, this might be a grammar error? Not sure how to categorize it:
- This leads to a sum over all elements. -> This leads to a sum overall elements. (over all -> overall)

If you need anything else from me, let me know. I haven't looked into how to fix these things on my own. So if anyone else has the time to add the words to the dictionary/dictionaries, that would be fine by me :). Thank you for making this software open-source and available ❤

hippietrail:
"gray" is spelled "grey" in UK English unless there's some specific context perhaps?

The last one with "over all" being combined to "overall" is a false positive bug in a grammar linter.

The words all look legit to me. I can include them in my next dictionary curation of somebody else can make up a PR.

PR #1403, opened 2025-06-17
fix: ensure grey is suggested when gray is wrong

hippietrail:
# Issues 

Fixes part of #1401

# Description

Tweak the suggestion ordering logic for cross-dialect word pairs containing `gray` or `grey`.

I added a function similar to RunDevelopment's functions that check for other cross-dialect spelling variant patterns. In this case it checks for `ay` in one vs `ey` in the other.

I also had to tweak the constant subtracted from the `score` from `5` to `6`

# Demo

image

# How Has This Been Tested?

All tests still pass.

# Checklist

### DIFF
--- a/packages/vscode-plugin/src/tests/suite/integration.test.ts
+++ b/packages/vscode-plugin/src/tests/suite/integration.test.ts
@@ -108,7 +108,7 @@ describe('Integration >', () => {
 					range: createRange(2, 26, 2, 32),
 				},
 				{
-					message: 'Did you mean `realize`?',
+					message: 'Did you mean to spell `realise` this way?',
 					range: createRange(4, 26, 4, 33),
 				},
 			),


hippietrail:
I've fallen into a rabbit hole wrt Precommit.
If I change the logic to bump the score to `-6` instead of `-5` and add an `ay` vs `ey` spelling check, then the number of suggestions changes, and the number of suggestions when it's 1 creates a different linter message than when it's plural, and this makes the javascript tests fail.

There's two ways to address this:
1. In the Javascript tests. This is not my wheelhouse and I tried this way and got nowhere. But I also doubted it would be the right fix.
2. I hunted through the spelling and suggestion code for why the number of suggestions changes but I can't seem to locate it.

Perhaps @elijah-potter would see what's up immediately with his thorough understanding of the codebase.

elijah-potter:
I suspect your problem has something to do with this section in `harper-core/src/linting/spell_check.rs`:

```
let mut possibilities = self.suggest_correct_spelling(word_chars);
if let Some(most_likely) = possibilities.first() {
    // If the most likely suggestion is a common misspelling, ignore all others.
    if is_ou_misspelling(most_likely, word_chars)
        || is_cksz_misspelling(most_likely, word_chars)
        || is_er_misspelling(most_likely, word_chars)
        || is_ll_misspelling(most_likely, word_chars)
    {
        possibilities.truncate(1);
    }
}

```

hippietrail:
It seems I still need help with this. Some of the integration tests relied on the test of the lint message, and that changed depending on whether one or multiple suggestions were offered, and that changed due to the number of suggestions being truncated when a dialect spelling variation was found. I removed the truncation and tried to fix the tests.

But I *think* there's a second problem that the integration test code assumes the dialect is set to American at the start where my system was set to British. But even after changing my settings and trying to add logic to set the dialect to American before running the tests I'm still getting test failures.

The integration system and/or code seems to be too gnarly for me to grok well enough to fix this deterministically.

hippietrail:
Well this makes no sense to me because it fails on my machine but out of frustration I had a hunch it might work on CI anyway. Lo and behold it does.

elijah-potter (approved):

PR #1423, opened 2025-06-21
feat: patterns for `have`, `need`, and `want` based on the one for `be`

hippietrail:
# Issues 
N/A

# Description

Adds patterns for inflections of three more key verbs based on `inflection_of_be.rs`

These three are picked because they are all used in similar grammatical structures known as `semi modals`: `have to`, `need to`, `want to`

They may in the future be used in bigger patterns for when any semi-modal or when any modal or semi-modal is wanted.

They could in theory be replaced by an `inflection_of`. `Have` is like `be` in that they are both irregular verbs, though `have` is less odd than `be`. `Need` and `want` are both totally regular so could be replaced by a `regular_inflection_of` pattern.

I will probably add a ctor to optionally include common mistakes like `need's` and `want's`

# Demo

# How Has This Been Tested?

No tests yet but `cargo test` still has no fails.

# Checklist


elijah-potter:
I'm somewhat hesistant to merge this, just because I don't personally see the value yet. Could you propose (or better yet, implement) a rule that uses these new patterns?

hippietrail:
> I'm somewhat hesistant to merge this, just because I don't personally see the value yet. Could you propose (or better yet, implement) a rule that uses these new patterns?

In the PR where you asked me if I could convert anything to use `UPOS` I mentioned a special class of verbs that can take a "to verb" as an object instead of an NP. These are the major verbs in this set. So would be useful for making a `Pattern` or `Expr` for such verbs. "want to", "need to", and "have to" are similar to the "modal verbs" (can, should, must, etc.) in this regard so often where you need modal verbs you ideally also want these. But unlike the modals, these ones inflect.

But it also makes sense to be hesitant because these are not the only verbs that take "to + verb" as an object. "Like to" is the best example. And it's not really like a modal. So maybe it will pan out that there are multiple groups of words.

Also "need" and "want" are totally regular so could be handled by some "any inflection of regular verb", whereas "have" is an irregular verb like "be".

Let's wait and see what concepts gel...

PR #1433, opened 2025-06-22
fix(core): auto detect dialect for snapshots

86xsk:
# Description

Adds dialect detection functionality to `Dialect` and `DialectFlags`, and uses it to automatically detect the dialect used in the source texts used for snapshots. Previously, the dialect was always considered to be American, and this caused some erroneous spelling lints to be included in some snapshots.

# How Has This Been Tested?
- `cargo test`
- `just precommit`

# Checklist


hippietrail:
Nice!

I actually found the wrong-dialect side-effect kind of useful since it tested the cross-dialect spelling suggestion logic as a free extra.

Maybe we could run each test in both native dialect and "other" dialect? Would there be value in that?

86xsk:
> Maybe we could run each test in both native dialect and "other" dialect? Would there be value in that?

Do you mean that in the sense of generating two snapshots, with one preserving the old behavior? It's certainly possible and I could throw that together. I do feel it would add some complexity though. (For instance, duplication of some snapshot changes in diffs.) Perhaps it would be better to create a specific snapshot for cross-dialect spelling?

One way to implement this could be to prepend the dialect in some way or another to the file extension for the text source(s) (e.g. `Spell.US.md`), and then try to parse that as a dialect override in the code that runs the snapshots. That might be a little hacky though, so I'm very much open to other ideas for the implementation.

elijah-potter (approved):
Great stuff. We're making great progress on reducing the false-positives in these snapshots. Thanks!

There's nothing in this PR I can see to improve. Keep it up.

PR #1442, opened 2025-06-23
feat(core): add rule for pronouns followed by an inflection of `be`

elijah-potter:
# Description

This PR implements a linter for agreement between a pronoun and an accompanying inflection of `be`. This has been a common complain among first-time users, so I figured it could be implemented as a special rule rather than a general one.

@hippietrail, thoughts? I'd like to get this pushed out in a release ASAP.

# How Has This Been Tested?

Unit tests.

# Checklist


hippietrail:
I've actually been working on improvements to the pronoun system in the dictionary / `affixes.json` / `TokenKind` and `WordMetadata` and running into unexpected problems. But I finally filed the PR tonight: https://github.com/Automattic/harper/pull/1445

So my advice is to make sure you add a test for each individual inflected form. I see you have many tests.

Also, does this only cover present tense? If so you might want to clarify in the name.

If you want to add past tense:
I was
We were
You were
He / she / it was
They were

So unlike present where 3rd person singular is the odd one out, here 1st person and 3rd person singular are the odd ones out.

elijah-potter:
@hippietrail. I've gone ahead and covered more cases, as well as added a tool to make inline expressions optional. Thoughts?

harper-core/src/word_metadata.rs
```diff
@@ -311,6 +311,28 @@ impl WordMetadata {
...
+            })
+        )
+    }
+
+    pub fn is_first_person_singular_pronoun(&self) -> bool {
+        matches!(
+            self.pronoun,
+            Some(PronounData {
+                person: Some(Person::First),
+                is_plural: Some(false),
```
hippietrail: Have a look at my big Pronoun metadata PR: `is_plural` being `false` no longer implies singular, since "you" is both singular and plural. You should now check `is_singular: Some(true)` - that is if you think my PR is up to snuff.

harper-core/tests/text/linters/Difficult sentences.snap.yml
```diff
@@ -316,6 +316,15 @@ Suggest:
 
 
 
+Lint:    Miscellaneous (30 priority)
+Message: |
+     399 | Precisely to get away from you was why I did what I did.
+         |                                ^~~ Make the verb agree with its subject.
```
hippietrail: You should check the case of the pronoun. Support for this is also in my new PR. You'll want something like `is_subject_pronoun && !is_object_pronoun` because "you" is both subject and object. Same applies to "it".

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -7103,6 +7103,16 @@ Suggest:
 
 
 
+Lint:    Miscellaneous (30 priority)
+Message: |
+    5575 | “He come out to see me two years ago and bought me the house I live in now. Of
+    5576 | course we was broke up when he run off from home, but I see now there was a
+         |           ^~~ Make the verb agree with its subject.
+Suggest:
+  - Replace with: “were”
```
hippietrail, line 7112: Nice! 👌

harper-core/src/word_metadata.rs
elijah-potter: Fixed.

Issue #1439, opened 2025-06-23
Dart tree-sitter crate is unmaintained


beeb:
I am trying to add support for the Solidity language in harper, and I've managed to update all tree-sitter crates to their latest versions: https://github.com/beeb/harper/blob/add-solidity-support/harper-comments/Cargo.toml

However, `tree-sitter-dart` is archived on GitHub and is not compatible with the latest tree-sitter version (doesn't expose a `LanguageFn`):

https://github.com/ast-grep/tree-sitter-dart

Should support for dart be dropped until a newer tree-sitter crate is made available? This is currently blocking me from adding Solidity support.

elijah-potter:
I agree that we should not let an unmaintained Dart grammar hold us back, but I'm hesistant to remove it altogether.Go ahead and remove Dart support on your PR. I'll find another solution in the meantime.

beeb:
Looks like this can be closed!

Btw here's the grammar that helix uses. It's not published on crates.io but could serve as inspiration or to double-check your fork: https://github.com/UserNobody14/tree-sitter-dart

PR #1443, opened 2025-06-23
feat: add solidity support

beeb:
# Issues 
https://github.com/Automattic/harper/issues/1439

# Description
This PR adds Solidity support to `harper-comments` by way of the widely used [tree-sitter-solidity](https://github.com/JoranHonig/tree-sitter-solidity) crate (used by the Helix editor amongst others).

The vscode-plugin has also been updated to register for Solidity files.

To be able to use the latest version of the Solidity grammar, I took the opportunity to update all tree-sitter dependencies.

# How Has This Been Tested?
Tests were added based on the model from various other languages.

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -849,6 +849,7 @@ dependencies = [
  "harper-core",
  "harper-html",
  "harper-tree-sitter",
+ "harper-tree-sitter-dart",
  "itertools 0.14.0",
  "paste",
  "tree-sitter",
@@ -993,6 +994,17 @@ version = "0.44.0"
 dependencies = [
  "harper-core",
  "tree-sitter",
+]
+
+[[package]]
+name = "harper-tree-sitter-dart"
+version = "0.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2cffcbd6eedacbf7df10ddb03de6e5e2fa52c0c183f6f01f7bfec9d0f6b38fb7"
+dependencies = [
+ "cc",
+ "tree-sitter",
+ "tree-sitter-language",
 ]
 
 [[package]]
--- a/harper-comments/Cargo.toml
+++ b/harper-comments/Cargo.toml
@@ -34,6 +34,7 @@ tree-sitter-solidity = "1.2.11"
 tree-sitter-swift = "0.7.1"
 tree-sitter-toml-ng = "0.7.0"
 tree-sitter-typescript = "0.23.2"
+harper-tree-sitter-dart = "0.0.5"
 
 [dev-dependencies]
 paste = "1.0.15"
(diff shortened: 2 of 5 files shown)


harper-tree-sitter/src/lib.rs
```diff
@@ -115,29 +115,29 @@ impl Masker for TreeSitterMasker {
...
-    let cloned = byte_spans.clone();
-
-    let mut i: usize = 0;
-    byte_spans.retain(|cur| {
-        i += 1;
-        if let Some(prev) = cloned.get(i.wrapping_sub(2)) {
-            !cur.overlaps_with(*prev)
-        } else {
-            true
+/// If any spans overlap, it will merge them.
```
beeb, line 121: This seemed more correct than ignoring the subsequent spans and didn't break any tests.

Original implementation was not working with the newer version of some grammars which have nested items for each line. Starting with 2 lines, the older logic was not sound (because 3 or more spans were overlapping) so I re-implemented it.

beeb:
@elijah-potter almost done, but I'm not sure why, the VSCode extension checks the whole file instead of just comments. Isn't it supposed to infer the parser to use from the language identifier?

Added some debug logs in the extension tests:

```
1) Languages > gives correct diagnostics for Solidity files
  Message:
    Error: Expected 1 diagnostics, got 4. [{"severity":"Information","message":"There are 4 spaces where there should be only one.","range":[{"line":5,"character":0},{"line":5,"character":4}],"source":"Harper"},{"severity":"Information","message":"Did you mean to spell `SPDX` this way?","range":[{"line":0,"character":3},{"line":0,"character":7}],"source":"Harper"},{"severity":"Information","message":"Did you mean to spell `Errorz` this way?","range":[{"line":3,"character":4},{"line":3,"character":10}],"source":"Harper"},{"severity":"Information","message":"Did you mean to spell `uint256` this way?","range":[{"line":5,"character":4},{"line":5,"character":11}],"source":"Harper"}]
```

The parser itself behaves properly in the rust tests.

mcecode:
> the VSCode extension checks the whole file instead of just comments

Hi, just to chime in on this. We haven't been able to set up the VS Code tests to support languages like Solidity that VS Code doesn't support out of the box. It doesn't understand that what `.sol` files are so it passes it to `harper-ls` as plaintext and that's why the whole file gets linted. You can see CMake, Haskell, and other language tests that are commented out in `packages/vscode-plugin/src/tests/suite/languages.test.ts` for the same reason. That said, as long as `harper-ls` understands it and a user installs a Solidity extension, it should work just fine for users.

beeb:
Oh thanks! That makes sense, I'll update the PR accordingly and mark is as ready now.

justfile
```diff
@@ -10,7 +10,7 @@ build-wasm:
 
 # Build `harper.js` with all size optimizations available.
 build-harperjs: build-wasm 
-  #! /bin/bash
+  #!/usr/bin/env bash
```
beeb, line 13: Had to fix those for nixOS

harper-tree-sitter/src/lib.rs
elijah-potter: Fantastic. Thank you.

elijah-potter:
I've gone ahead and returned Dart to the supported list using a fork of the original `tree-sitter-dart` dependency.

PR #1445, opened 2025-06-24
feat: improvements to pronoun metadata

hippietrail:
# Issues 
N/A

# Description

Improves the pronoun metadata and to a lesser extent to the noun metadata:

- Pronouns are no longer either singular or plural. "You" can be marked as both singular and plural.
- Nouns are no longer either singular or plural. Words like "aircraft", "biceps", "fish", "sheep", etc. can now be marked both singular and plural.
- To save having to add a `singular` flag to thousands of nouns in `dictionary.dict`, common nouns marked neither singular nor plural are treated as singular.
- `a` and `.` flags added to `affixes.json` for `personal pronoun` and `singular pronoun` respectively. Since the `plural pronoun` flag is `:` it makes a handy mnemonic pair.
- The `first person pronoun` flag was corrected from `f` to `1`.
- The `w` `first person plural` flag is now redundant and has been removed.
- The new pronoun flags have been applied to all the main personal pronouns in `dictionary.dict`
- A few noun and other entries have had wrong flags fixed.
- Renamed `not_plural_pronoun` and `not_possessive_pronoun` to `non_..._pronoun` for clarity to match the semantics since they match pronouns without those properties set, not any token that is not a pronoun or is a pronoun without those properties set.
- Changed `then_not_plural_nominal` in `hop_hope/to_hope.rs` to `then_nominal` since one of the tests used "we" which is plural.
- Replaced the `case` property of `PronounData` with `is_subject` and `is_object` since both "you" and "it" are both subject and object pronouns.
- Added a complete matrix of tests for all of the English pronouns covering all of the properties currently implemented.
- Updated the `pos_tags` code to be able to deal with nouns and pronouns that are both singular and plural.

I *did not* touch much of the new `UPOS` code since I'm not familiar with it. @elijah-potter may want to check if there's any mismatch etc between this new code and the current UPOS code.
But I did add pronouns to `is_nominal`, which previously only considered nouns and proper nouns to be nominals.

There are more changes in the snapshots than I would like, but most of them reveal quirks in the dictionary and affix system. It looks like "proper noun" and/or "possessive noun" properties might not be propagating to word derived by affixation. Many others are due to lettercase quirks due to the dictionary system conflating upper and lower case resulting in words like `in` being treated as noun, etc.

# How Has This Been Tested?

Many unit tests covering all supported properties of each personal pronoun. All old tests still pass.

# Checklist


elijah-potter:
> I did not touch much of the new UPOS code since I'm not familiar with it.

That's fine. The dictionary-derived metadata and the UPOS data serve separate functions, so they shouldn't overlap.

harper-pos-utils/src/upos.rs
```diff
@@ -63,6 +63,6 @@ impl UPOS {
     }
 
     pub fn is_nominal(&self) -> bool {
-        matches!(self, Self::NOUN | Self::PROPN)
+        matches!(self, Self::NOUN | Self::PROPN | Self::PRON)
```
elijah-potter, line 66: Good catch!

Issue #535, opened 2025-01-29
Flag mixing up "thing" and "think".


hippietrail:
This is always going to just be a typo, but the spell checker won't pick it up. I just made it myself in Discord:
- some people **thing** it's about plurals → some people **think** it's about plurals

hippietrail:
Examples found on GitHub:
- Whad do you **thing** about tinygo?
- bcz i **thing** hugging face embeddings and models are very complex
- one should always **thing** of whether the efforts are better targeted to the improvement
- terraform will always **thing** it has changed
- As I always **thing** in 'sets of devices of the same kind'
- which information we **thing** to be missing
- So far we really like it featurewise and we **thing** it is really ...
- On second thought, I **thing** my ROIs are off
- One thing that I sometimes **thing** would be nice is if I could make different instances
- But when I **think** about composition, I often thing about Bartok. And when I think about performance, I often **thing** about Coltrane.
- When working with workflows on many forms I often **thing** I need to do the same over and over
- Sub builders don't always **thing** about ...
- otherwise without it I **thing** I would have to pass it as parameter
- I **thing** documentation is not enough and I have to do a lot of things with Superset.
- The exe file dosen't work allways, because antivirus can **thing** it is a virus.
- I **thing** lowering fog density and lowed detail spread somewhat mitigates the issue
- I **thing** something going wrong with permission.
- ... they **thing** something is a good idea.
- I **thing** it's faster to just type the date
- I **thing** something like “Controversial topics” would be much ...
- Only thing i can **thing** of is 'Number' or 'Number of Bits', but that doesn't look right
- Trying to **thing** about it a bit more, its the same issue with all the initial props.
- Here I'm trying to **thing** about the following questions:
- Fix comparison of config.xml strings that would sometimes **thing** XMLs were different in cases when they were not.
- I **thing** this is not the correct thread to this discussion but to have a library that supports your request ...
- I **thing** it could be related with the way stardist ...
- Reason I **thing** this is useful is when I tried scramble on my vapor hosted site, it would fail.
- When I **thing** about it again, after writing this, Yii3 does not feel like Yii anymore
- I **thing** Addons are the solution, with addons you can include
- I **thing** idea of these panels is nonsense anyway.
- I always **thing** that less is more
- It's good practice to always **thing** in the context of the data
- I **thing** you are doing same way
- I **thing** my issue is the combination of several technologies:
- I **thing** I found a way to achieve it, not the best one, but it should work.
[...]

PR #1451, opened 2025-06-24
feat: thing→think

hippietrail:
# Issues 
#535 

# Description

Implements #535 

Also introduces a new `LintKind`, `Typo` since `Spelling` is reserved for the spellchecker and typos are a specific unintentional type of spelling mistake.

# How Has This Been Tested?

Dozens of unit tests with sentences found on GitHub.

# Checklist


harper-core/src/linting/thing_think.rs
```diff
@@ -0,0 +1,403 @@
...
+
+        let verb_to = SequenceExpr::default()
+            .then(WordSet::new(&[
+                "have", "had", "has", "having", "need", "needed", "needs", "needing", "want",
+                "wanted", "wants", "wanting", "try", "tried", "tries", "trying",
+            ]))
+            .t_ws()
+            .t_aco("to");
+
+        let modal = WordSet::new(&[
```
elijah-potter, line 39: Would it be possible to replace some of these `WordSet`s with queries against POS tags? If you want to match an adjective, for example, replace your `WordSet` with a `UPOSSet::new(&[UPOS::ADJ])`.

harper-core/src/linting/lint_kind.rs
```diff
@@ -20,6 +20,7 @@ pub enum LintKind {
     /// This should only be used by linters doing spellcheck on individual words.
     Spelling,
     Style,
+    Typo,
```
elijah-potter, line 24: I like this, but there are several other places (in integrations) which need to be updated to reflect this new item. I would just run a search for `LintKind` and you'll likely find all the relevant places.

harper-core/src/linting/thing_think.rs
hippietrail: I can't see any places where the `WordSet` is general enough. I see `UPOS` covers pronouns but not subgroups by property:

- `subject_pronouns` and `indefinite_pronouns`. If we used just any kind of pronoun we would get false positives for ones like "this", "that", and personal pronouns with object case.
- `verb_to` is a special subset of verbs that can be followed by `to` and then another verb. (That one should be generalized into a `Pattern`, `Expr`, or `WordMetadata`.)
- `modal` is a subset of verbs which don't inflect, don't have person agreement, plus most of them have negative forms which are contractions with "not".
- `adverb_of_frequency` is a special subtype of adverb. My hunch is just using arbitrary adverbs would result in many false positives.

We really need some kind of corpus of real-word mistake-ridden text to test new linters on. So far we use bug reports, false positive reports, and the results of Google searches for ideas we already had, and we can't Google by POS (-: StackOverflow used to provide downloadable dumps, but I believe those were stopped when AIs started to mine them and they started going broke.

elijah-potter: Understood. In that case, we will leave them as they are.

elijah-potter (approved):

elijah-potter (approved):

PR #1454, opened 2025-06-25
feat: in the cards ⇔ on the cards

hippietrail:
# Issues 
N/A

# Description

Corrects "in the cards" or "on the cards" when the wrong variant is used for the currently set dialect.

In British English, "on the cards is used". In American English "in the cards" is used. Australian and Canadian English seems to follow American English based on some quick research.

There would be quite a few false positives if not depending on the phrase following an inflection of "be", the word "not", or a contraction ending in a form of "be" or "not": it's, wasn't, etc.

It also works if the apostrophe is missing from the contractions.

Surprisingly, it didn't work when "smart" typographic apostrophes are used. I had to handle those explicitly.

# How Has This Been Tested?

Many example sentences from GitHub were used to craft unit tests for most variants and combinations.

# Checklist


harper-core/src/linting/lint_kind.rs
```diff
@@ -16,6 +16,7 @@ pub enum LintKind {
     Punctuation,
     Readability,
     Redundancy,
+    Regionalism,
```
elijah-potter, line 19: I believe I looked at this on another PR. You need to update the other references of `LintKind` in the TypeScript code. Just do a search for `LintKind` in the codebase.

elijah-potter (changes requested):
This is very close to being ready. I made some changes to fix the typographic problems. You need to update the remaining references to `LintKind`, then we can merge.

harper-core/src/linting/lint_kind.rs
hippietrail: > I believe I looked at this on another PR. You need to update the other references of `LintKind` in the TypeScript code. Just do a search for `LintKind` in the codebase.

Thanks! I knew there was something somewhere.

I just checked and some of the colours are reused multiple times: `540D6E` is capitalization and formatting, `0EAD69` is enhancement, readability, and word choice, and `3BCEAC` is miscellaneous and repetition.

Is this intentional? In other words should I make `Redundancy` and `Regionalism` reuse colours that are used for semantically similar lint kinds, or should I redo the colours to remove redundancy?

I've gone ahead and replaced the duplicate colours with new distinct colours. Let me know if I should change it back.

Hey I wonder if it would be possible to get the `LintKind` colours into the playwright test report for a nice visualization?

elijah-potter: > Is this intentional? In other words should I make Redundancy and Regionalism reuse colours that are used for semantically similar lint kinds, or should I redo the colours to remove redundancy?

You can reuse stuff. That's fine. We want to maintain the color pallet.

hippietrail: > > Is this intentional? In other words should I make Redundancy and Regionalism reuse colours that are used for semantically similar lint kinds, or should I redo the colours to remove redundancy?
> 
> You can reuse stuff. That's fine. We want to maintain the color pallet.

Thanks for the clarification!

elijah-potter (approved):
Thanks for making those changes. Looks good to me.

Homework: Do some thinking about what existing rules could be better described with a different `LintKind`.

PR #1458, opened 2025-06-26
feat: no match against/of/to → no match for

hippietrail:
# Issues 
N/A

# Description

I noticed somebody use "no match to" when they clearly meant the idiom "no match for". A bit of research led to finding that "no match against" and "no match of" are also pretty common.

This linter will match against any preposition. I noticed there would be many false positives if I only matched this exact phrase, so matching only after an inflection or contraction of "be" such as "is" or "it's" rules out most false positives.

# How Has This Been Tested?

I made unit tests for many variations, preferably from GitHub but from other sites too.

# Checklist


harper-core/src/linting/no_match_for.rs
```diff
@@ -0,0 +1,156 @@
...
+        let prep_tok = toks.last()?;
+        let prep_chars = prep_tok.span.get_content(src);
+        if prep_chars.eq_ignore_ascii_case_chars(&['f', 'o', 'r']) {
+            return None;
+        }
+
+        let phrase_toks = &toks[2..];
+        let phrase_span = phrase_toks.span()?;
+
+        let suggestion = Suggestion::replace_with_match_case(
```
elijah-potter: You should be able to use `replace_with_match_case_str` here.

hippietrail: > You should be able to use `replace_with_match_case_str` here.

Ah I've been intentionally trying to avoid it since I learned more about `&[char]` vs `&str` etc. I thought `replace_with_match_case_str` did a conversion and called `replace_with_match_case` internally that I was avoiding for literals. Is this wrong or just a pointless optimization that makes reading the code a bit harder?

elijah-potter: In this case, I don't believe there is any benefit to create the `Vec` directly rather than let `replace_with_match_case_str` do the conversion. Even if it did, it wouldn't be worth optimizing anything outside the hot loop (which is the pattern itself).

hippietrail: Right. I'll switch to preferring `replace_with_match_case_str()`!

PR #1466, opened 2025-06-27
Added an HarperAPI with rust

Eeman1113:
Description
This pull request significantly enhances the harperAPI linting service by introducing two major features: advanced overlap filtering and paragraph-relative coordinates. These changes provide more accurate, useful, and context-aware linting suggestions.

Key Changes:

Advanced Overlap Filtering:

Previously, the API could return multiple, overlapping lints for the same text span (e.g., flagging "geat" for both Spelling and WordChoice).

A new priority system (LintKindPriority) has been implemented to resolve these conflicts. Lints are now prioritized in the following order: WordChoice > Repetition > Style > Spelling.

The API now filters overlapping suggestions, ensuring only the highest-priority and most relevant lint is returned for any given span of text.

Paragraph-Relative Coordinates & Context:

The JSON response has been updated to make lint coordinates (start, end, length) relative to the parent paragraph, not the entire document. This makes it much easier for a client application to highlight and process suggestions.

Each lint object in the response now includes the full text of the paragraph it belongs to, providing immediate context.

The final list of lints is now sorted by paragraph number and then by start position, ensuring a logical and predictable order.

How Has This Been Tested?
The changes have been tested manually by running the server and sending requests with text designed to trigger overlapping lints and multiple paragraphs.

To test:

Ensure the harper-core crate is located at the correct relative path.

Navigate to the harperAPI directory and run the server with cargo run.

Use a tool like curl or Postman to send a POST request to http://127.0.0.1:8000/lint.

Sample Request Body:

{
    "text": "Your going to have a geat time.\n\nThis is a big event, and I want it to be perfect."
}

Verification Steps:

Check for Overlap Resolution: In the example, "geat" could be a misspelling of "great" (Spelling) or a poor word choice (WordChoice). Verify that the API returns only the higher-priority WordChoice lint for this word.

Check for Relative Coordinates: Confirm that the start and end positions for each lint are calculated from the beginning of their respective paragraphs, not from the beginning of the entire text block.

Check for Paragraph Context: Ensure the paragraph field in the response correctly contains the text of the paragraph where the lint was found.

Checklist
[x] I have performed a self-review of my own code

[x] I have added tests to cover my changes


hippietrail:
Wow this looks really interesting! I wonder if you noticed that each lint has a priority field? So far I've got the impression that they don't do a lot so rarely pay much attention to what value they have in the linters I write. But if it's going to make a noticeable difference I'll definitely start trying to set them as meaningfully as possible.

harper-core/src/linting/suggestion.rs
```diff
@@ -88,10 +88,10 @@ impl Display for Suggestion {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Suggestion::ReplaceWith(with) => {
-                write!(f, "Replace with: “{}”", with.iter().collect::<String>())
+                write!(f, "{}", with.iter().collect::<String>())
```
elijah-potter: Why? This will make suggestions on `harper-ls` more difficult to understand for newcomers.

harperAPI/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "harperAPI"
```
elijah-potter: I love the idea. Could we get a slightly different name for this crate? Ideally, it would be something that follows the existing naming convention, like `harper-api` or `harper-serve`.

harperAPI/backup.txt
```diff
@@ -0,0 +1,175 @@
+// use actix_web::{web, HttpResponse, Responder};
+// use harper_core::{
+//     linting::{Lint, LintGroup, LintKind, Linter},
+//     Dialect, Document, FstDictionary,
```
elijah-potter: What does this file do? We used to have an API server for debugging purposes. Did you dig that up or is this new?

harperAPI/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+        LintKind::Capitalization => LintKindPriority::Capitalization,
+        _ => LintKindPriority::Miscellaneous,
+    }
+}
+
+pub async fn lint_text(request: web::Json<LintRequest>) -> impl Responder {
+    let document = Document::new_plain_english_curated(&request.text);
+    let curated_text = document.get_source();
+    let dictionary = FstDictionary::curated();
+    let mut linter = LintGroup::new_curated(dictionary, Dialect::American);
```
elijah-potter: The `LintGroup` should be instantiated once at startup and persisted throughout runtime. It contains lots of caching, so persistence is necessary in order to take full advantage.

harperAPI/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+enum LintKindPriority {
+    Miscellaneous,
+    Capitalization,
+    Spelling,
+    Style,
+    Repetition,
+    WordChoice,
+}
+
+fn get_priority(kind: &LintKind) -> LintKindPriority {
```
elijah-potter: This is not a bad idea.

elijah-potter (commented):
This is a pretty interesting PR. Are you trying to mimic the LanguageTool spec?

Eeman1113:
> Wow this looks really interesting! I wonder if you noticed that each lint has a priority field? So far I've got the impression that they don't do a lot so rarely pay much attention to what value they have in the linters I write. But if it's going to make a noticeable difference I'll definitely start trying to set them as meaningfully as possible.

Yes I actually intentionally did that because there were cases when I was having issues such as 

"sho ul" -> "should"
"ul" -> "la" "lm"

so for same word it was gicing multiple suggestions to handle that I added priority in my code

Eeman1113:
> This is a pretty interesting PR. Are you trying to mimic the LanguageTool spec?

Exactly my intention.
Also I have read through all other of your points will make the changes and will PR again.

PR #1467, opened 2025-06-27
fix(core): we cannot respect `spellcheck="off"`

elijah-potter:
# Issues 

#1459 

# Description

We cannot ignore fields with `spellcheck=off` without significantly impairing Harper's ability to help.

# Checklist


hippietrail:
Would it make sense to leave grammar checking on but just turn of spellcheck when encountering this?

PR #1469, opened 2025-06-27
fix(core): support progressive/perfect forms after "its" + 1

### LINKED ISSUE #1463: False positive re "its" vs "its's" in `lint_group.rs`
## Description

The comment `/// Get map from each contained linter's name to its associated description.` gets the `its` wrongly flagged as an error:

Image

kiding:
# Issues 

Fixes #1463. (@hippietrail)

# Description

This PR fixes `harper-core` to fix false positives when progressive/perfect forms appear after "its"; e.g., its existing properties, its vanished trees, etc. The bug was probably introduced with f0bcaa0.

Commit-by-commit explanations:
- 044941d: `NominalPhrase` is modified to support an input where the last word can be both adjective/determiner and also nominal(noun/pronoun); e.g., a way. 
  - Now that I think of it, this is not *technically required* for this PR, but I'm hoping it's good to have.
- ed4cbd9: `NominalPhrase` is again modified to support progressive/perfect forms; e.g., a winning way, a failed way. Since perfect forms are always adjectives, we have that for free. 
- 1ca65d3: Finally `ItsContraction` is modified to fix the logic for ignoring a nominal phrase after "its."
  - As a `ExprLinter`, sending `&toks[2..]` to `NominalPhrase.matches` isn't right since `toks` is always `["its", " ", VERB or AUX or DET]` thanks to `positive`. `ItsContraction` is now changed to `Linter` so that `toks` can contain more tokens; e.g., `["its", " ", "existing", " ", "properties", "."]`. 
  - The `NominalPhrase.matches` check is invoked only when the next word is a `VERB`, rather than `not AUX`. If not, "Its a nice day" passes the test since "a nice day" is a nominal phrase. Excluding `DET` out of the three makes sense in that any nominal phrase that start with a determiner can't be used after "its."

**A side note:** Currently progressive forms are not parsed as adjectives unlike perfect forms. Do you think this is correct?

# Demo

image

# How Has This Been Tested?

Manually and with unit tests.

# Checklist


elijah-potter (approved):
These are entirely reasonable changes. Thanks for lowering the false-positive rate!

elijah-potter:
> A side note: Currently progressive forms are not parsed as adjectives unlike perfect forms. Do you think this is correct?

That's more of a question for @hippietrail. Are you referring to the UPOS tags or the standard WordMetadata?

hippietrail:
> > A side note: Currently progressive forms are not parsed as adjectives unlike perfect forms. Do you think this is correct?
> 
> That's more of a question for @hippietrail. Are you referring to the UPOS tags or the standard WordMetadata?

I wanted to flesh out more of the verb morphology system since the previous stage of it was merged over the last week.
When @RunDevelopment was contributing he was adamant that past and present participles shouldn't be marked with adjective and/or noun POSes where I think they should be. Both views are valid.

Basically the -ing form has several functions but I chose the most recognizable and verbish name for it in the code. When it acts like an adjective it's called the `present participle` and when it acts like a noun it's called the `gerund`. Similar for the `-ed` form. I believe I just called it `past` because that's the most recognizable name, even though there are two pasts in English morphology. For regular nouns the past tense form is the same as the past participle but for irregular verbs there's an extra form. In both cases the past participle can act as an adjective.

For this reason my preference would be that in the affix expansion code the past participle should be given the adjective property and the present progressive should be given the adjective and noun properties. This way linters won't have to check for adjective *and* progressive *and* past. But it's not the only solution.

kiding:
> Are you referring to the UPOS tags or the standard WordMetadata?

I was looking at `harper-cli parse` result. For the word "presenting" it is (excluding nulls):
```json
{
  ...
  "kind": {
    "kind": "Word",
    "value": {
      "verb": {
        "verb_form": "ProgressiveForm"
      },
      "dialects": "",
      "preposition": false,
      "common": false,
      "derived_from": {
        "hash": 16732856634846968943
      },
      "np_member": false,
      "pos_tag": "VERB"
    }
  }
}
```
and for the word "presented":
```json
{
  ...
  "kind": {
    "kind": "Word",
    "value": {
      "verb": {
      },
      "adjective": {
      },
      "dialects": "",
      "preposition": false,
      "common": false,
      "derived_from": {
        "hash": 16732856634846968943
      },
      "np_member": false,
      "pos_tag": "VERB"
    }
  }
}
```

Additionally, some/all? progressive forms in `dictionary.dict` lack "J"; e.g., slurring/VN. 

> For this reason my preference would be that in the affix expansion code the past participle should be given the adjective property and the present progressive should be given the adjective and noun properties. 

This makes sense to me... although I'm not too familiar with the code base and also English in general for that matter.

PR #1475, opened 2025-06-28
feat: replace terms from dialects with local synonyms

hippietrail:
# Issues 
N/A

# Description

Flags words from other English dialects than the one set.
Suggests synonyms that are used in the set dialect.

Does not flag words used in the current dialect that have a different meaning in the other dialect to avoid false positives.
Thus `soccer` will be flagged to be replaced by `football` when the dialect is set to British.
But `football` will not be flagged to be replaced by `soccer` when the dialect is set to American since `football` is a legit word in both dialects though it refers to different sports.

# Demo

image

# How Has This Been Tested?

I added several unit tests to cover terms that should and should not replaced in a mix of dialects and for various reasons.

# Checklist


harper-core/src/linting/regionalisms.rs
```diff
@@ -0,0 +1,610 @@
...
+
+#[derive(PartialEq)]
+enum CanFlag {
+    Flag,
+    DontFlag,
+}
+
+use CanFlag::*;
+
+#[derive(PartialEq)]
```
elijah-potter, line 25: It's not entirely clear why this `enum` exists. Could we get away without it?

hippietrail: > It's not entirely clear why this `enum` exists. Could we get away without it?

We need the logic to prevent many false positives, but it can be a `bool`. I just made it a newtype because it made the logic easier to reason about while I was working out how to implement it. Change to bool then?

elijah-potter: Sorry, I was referring to the `Concept` enum. I think the `CanFlag` enum is just fine as-is.

hippietrail: > Sorry, I was referring to the `Concept` enum. I think the `CanFlag` enum is just fine as-is.

I added a comment for `Concept`. Doing it this way we can cover the same word that means different things in different dialects (like "chips"), one concept in one dialect mapping to multiple concepts with different words in other dialects (like "RV"), etc.

It's basically more flexible than the naïve 1:1 expectation.

PR #1476, opened 2025-06-28
feat(cli): allow direct text (non-file) input

86xsk:
# Description

Allows source/input text to be passed directly as a command line argument for some `harper-cli` commands.
This includes: 
- `lint`
- `parse`
- `spans`

This should hopefully make it easier to debug short snippets of text, as it would no longer be necessary to first place them into a file. For instance, it makes it possible to run `lint` like so: `harper-cli lint 'This is an example sentence'`.

This also upgrades `harper-cli` to Rust 2024 (from 2021).

# How Has This Been Tested?
Manual testing of the modified commands.

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -137,12 +137,14 @@ fn main() -> anyhow::Result<()> {
             let mut merged_dict = MergedDictionary::new();
             merged_dict.add_dictionary(dictionary);
 
+            // Attempt to load user dictionary.
             match load_dict(&user_dict_path) {
                 Ok(user_dict) => merged_dict.add_dictionary(Arc::new(user_dict)),
                 Err(err) => println!("{}: {}", user_dict_path.display(), err),
             }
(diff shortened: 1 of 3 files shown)


kiding:
I like this, but would it make more sense to make use of stdin, as in: `echo 'An apple' | harper-cli lint`? It's more of a preference though.

harper-cli/src/main.rs
```diff
@@ -695,3 +712,64 @@ fn file_dict_name(path: &Path) -> PathBuf {
 
     rewritten.into()
 }
+
+/// Represents an input/source passed via the command line. For example, this can be a file, or
```
elijah-potter: This might be worth separating into its own module. I enjoy the idea of having a dedicated struct to contain logic. Good stuff.

elijah-potter (approved):
Thanks for making that change! This is good work and it's ready to merge.

PR #1487, opened 2025-07-01
refactor(core): rename `insert_pattern_rule!`

86xsk:
# Description

Rename `insert_pattern_rule!` -> `insert_expr_rule!` to match documentation.

The last code block in the [author-a-rule#Register-Your-Rule](https://writewithharper.com/docs/contributors/author-a-rule#Register-Your-Rule) section references `insert_expr_rule!`. However, the mentioned code still seems to use the obsolete name of `insert_pattern_rule!` at the moment.

Related to PR #1393.

# How Has This Been Tested?
- `cargo test`

# Checklist


hippietrail:
My understanding is that there are both `Expr`s and `Pattern`s and that all `Pattern`s are `Expr`s, so I guess this makes sense...? @elijah-potter

elijah-potter:
Makes sense. I remember making this change with the documentation... I suppose that didn't actually happen.

PR #1488, opened 2025-07-01
chore: augment and refactor metadata

hippietrail:
Work-in-progress

# Description

## Part 1

Add `0` flag to `affixes.json` for singular nouns. Only needed for nouns which are both singular and plural. Nouns marked neither are singular by default.

Mark a few dozen words in the dictionary which are both singular and plural.

Ensure the proper hierarchy of metadata properties and methods is respected:

`TokenKind` is the highest level with `WordMetadata` being a lower level for tokens representing words.

`token_kind.rs` implements `is_xxx()` methods on `TokenKind`.

They can call each other (`nominal` might combine `noun` and `pronoun` methods).

They can call `is_yyy()` methods on `WordMetadata`.

They must not use the low-level properties of `WordMetaData` directly when there are methods for them.

This ensures properties with special behaviour such as defaults that are not `false` only have to be implemented in one place.

`word_metadata.rs` implements `is_yyy()` methods on `WordMetadata`.

They can call each other (`nominal` might combine `noun` and `pronoun` methods).

They can use the low-level properties of `WordMetadata` directly.

They can implement special logic such as properties that do not default to false. In particular the new `singular` noun flag `0` does not have to be applied to every word in `dictionary.dict` and nouns will be treated as singular if they either have `singular` set to `Some(true)` or if they have both `singular` and `plural` set to either `None` or `Some(false)`.

(In the next stage, countable nouns will be implemented this same way.)

Tests in `word_metadata.rs` must use the `WordMetadata` `is_yyy` methods and must not use the low-level properties directly.

All `is_xxx()` and `is_yyy()` methods in both files have been standardized as above.

Un-`ignore`d all the tests on `singular` nouns and they all now pass.

Comments have been added to `token_kind.rs` in the `impl TokenKind` blocks to label the `is_xxx()` vs other methods, the word `is_yyy()` methods vs other token types `is_yyy()` methods, and the parts of speech of the various `is_yyy()` methods on words. In a following step I will group them and sort them. Until then this commit will be much easier to compare against the old code.

## Part 2

- `m` and `w` added to `affixes.json` for mass nouns and countable+mass nouns respectively. Nouns with neither flag in their annotation default to countable only.

- The auxiliary verb property flag was moved from `m` to `A`.

- A dozen or so words in `dictionary.dict` annotated for countability.

- `is_countable_noun` and `is_mass_noun` methods added to `TokenKind`.

- Also added the missing `is_singular_xxx` methods for `_nominal`, `_noun`, and `_pronoun`

- A first couple of tests added to `token_kind.rs`

- `is_countable` and `is_mass` properties added to `WordMetadata`

- `is_countable_noun` and `is_non_countable_noun` methods added to `WordMetaData`

- countable and mass noun unit tests added to `word_metadata.rs`

This will help make possible linting for errors like `source codes` and `an advice` etc.

I did not add countable and mass information to the snapshots. Let me know if it's needed.

I also removed a few bogus entries I spotted in the dictionary.

# How Has This Been Tested?

There are unit tests for all the word and part of speech properties.

## Part 3a

Replace all the straightforward `is_xxx` functions with a `delegate_to_metadata` macro is in the spirit of the one in `word_metadata.rs` but simpler.

Everything is still in the same order.

## Part 3b

Rearrange the methods into some kind of sensible order and grouping.

# Checklist


harper-core/src/token_kind.rs
```diff
@@ -31,6 +41,91 @@ pub enum TokenKind {
 }
 
 impl TokenKind {
+    // Word metadata delegation methods grouped by part of speech
+    delegate_to_metadata! {
```
elijah-potter, line 45: Gorgeous.

elijah-potter (approved):
This is fantastic PR. You've just paid back so much technical debt. I have no comments. Thanks!

PR #1489, opened 2025-07-01
fix(core): use tagger to improve certainty in `HowTo` lint

elijah-potter:
# Description

Tweaks the `HowTo` rule to be more accurate by using the Brill Tagger's output.

# How Has This Been Tested?

Existing and additional unit test.

# Checklist


hippietrail:
Another way to tackle this is to only work with the lemma form of the verb using `.is_verb_lemma()` on the verb token. The lemma is the dictionary form often referred to as the present tense informally. All the examples in the tests seem to expect that and `has` is the third-person singular present form.

But I just checked and `has` doesn't have the `*` annotation flag needed to mark it as 3p-sing-pres so you'd want to add that too.

Issue #386, opened 2025-01-15
Flag indefinite article used before mass nouns (uncountable nouns)


hippietrail:
Non-native English speakers from all cultures have trouble with mass nouns / uncountable nouns / non-count nouns in English. A little of this may have started to spill over into younger native speakers:

- An evidence
- An info/an information
- An infrastructure
- A luggage

Erroneous pluralizations of these seem to all already be flagged by the spellchecker.
Suggestions could include:
- Simply removing the article
- Change a/an into "some"
- Add a quantifier, usually "item of" or "piece of", but some words have one or more idiomatic ones.

hippietrail:
A couple of these have become more acceptable in the programming context, so I suppose they should be optional. They sound wrong to people my age and older who are native speakers, but sound fine to non-native speakers and younger native speakers. I'm not sure if it crossed over in or after the Millennial generation:

- I wrote a code → I wrote some code
- This is a software → This is a software package

There are probably word-specific replacements. I'll start a list for each:

code / source code
- an app
- a piece of code
- a program
- a snippet
- a snippet of code
- some code
- code

software
- a piece of software
- an app
- a program
- a software tool
- a software package
- some software
- software

"Hardware" may fall into the same category. I'll add it if I find concrete examples in the wild. "Firmware" seems to be different in some way and at least has a plural though "I installed a firmware." still sounds wrong where "They're running different firmwares." sounds more acceptable, possibly not 100% acceptable though.

hippietrail:
I'm going to close this as fixed by #1498 - feel free to reopen if false negatives are found, or open a new focused issue.

Issue #950, opened 2025-03-21
feat(core): Add an `assert_any_suggestion_result` test function


hippietrail:
It would be nice to have a testing assertion along with `assert_suggestion_result()` and `assert_nth_suggestion_result()` that passes of the test no matter which suggestion appears in the result.

I looked at implementing this briefly a week or two ago and decided I didn't know enough about the architecture to do it efficiently.

hippietrail:
Reopening because it seems we do need a proper `assert_any_suggestion_result()` after the `top3` one wasn't enough for one of my linters and I made a `top5` one. But I'm not 100% happy with how those are coded. And nor am I happy with my code for the `good_and_bad` assertion, even though it works for the linters it's currently used on, at least to pass all the tests.

hippietrail:
This is still important. I run into it every couple of days.

hippietrail:
I should be submitting a PR for this tomorrow. Just need to do some more testing.

PR #1498, opened 2025-07-02
feat: mass nouns after countable determiners/quantifiers

hippietrail:
# Issues 
#386 

# Description

Looks for mass nouns being treated as singular or plural nouns.

- group and sort `sequence_expr` `gen_then_from_is` etc.
- Added a new `agreement` `TokenKind` and added it to two other linters.
- Adding singular/plural & countable/mass flags to dictionary.

Does not yet handle adjectives between the determiner/quantifier and the noun.

<s>Does not yet handle the case where the mass noun is part of a compound noun and therefore not an error.</s>

<s>Does not yet offer specific synonym such as "a tip" for "an advice".</s>

Does not yet offer specific units for certain nouns, such as `item` for `furniture` and `clothing`.

<s>Does not yet support not suggesting the default unit `piece` for nouns it sounds wrong with: `piece of traffic`.</s>

# How Has This Been Tested?

Unit tests have been included sourced from sentences on GitHub.

# Checklist


elijah-potter:
@hippietrail, I see you've marked it as ready for review, but it still says "Work in Progress". Which is right?

hippietrail:
> @hippietrail, I see you've marked it as ready for review, but it still says "Work in Progress". Which is right?

Oh good point. There's features it's missing but it's ready to go now. I'll edit that out.

harper-core/src/linting/mod.rs
```diff
@@ -337,6 +339,48 @@ pub mod tests {
         }
     }
 
+    pub fn assert_top5_suggestion_result(
+        text: &str,
```
elijah-potter: Maybe we should just create a more abstract `assert_top_n_suggestion_result`? This is getting verbose.

harper-core/src/linting/lint_kind.rs
```diff
@@ -28,6 +29,7 @@ pub enum LintKind {
 impl LintKind {
     pub fn new_from_str(s: &str) -> Option<Self> {
         Some(match s {
+            "Agreement" => LintKind::Agreement,
```
elijah-potter, line 32: Good.

harper-core/src/linting/have_pronoun.rs
```diff
@@ -36,7 +36,7 @@ impl ExprLinter for HavePronoun {
 
         Some(Lint {
             span,
-            lint_kind: LintKind::Miscellaneous,
```
elijah-potter, line 39: Good.

elijah-potter (changes requested):
I have some minor tweaks, but this is close.

harper-core/src/linting/mod.rs
hippietrail: > Maybe we should just create a more abstract `assert_top_n_suggestion_result`? This is getting verbose.

No not really. I don't like the way they're implemented but needed them. What we really want is `assert_any_suggestion_result` implemented in a different way. I tried but kept getting confused about iterating through both linters and lints. I mentioned this at the time and think I filed a feature request...

**Ah I misremembered.** I did file a [feature request](https://github.com/Automattic/harper/issues/950) but I also implemented the `_top3_` one as a solution that closed the issue! https://github.com/Automattic/harper/pull/1033

I'm still not that confident about that assert. I think it may be missing something subtle in comparison to the original `assert_suggestion_result` which had nested loops over linters and lints but also a loop to iterate (up to?) 100 times that I never fully grokked.

I've got a couple of things on the go right now but then I want to look at those...

elijah-potter: > but also a loop to iterate (up to?) 100 times that I never fully grokked.

That loop was to make sure that the lint did not create an infinite loop of linting. That is, if applying that lint would result in a new error that could be detected by that same linter.

Maybe we should reopen that feature request?

hippietrail: > > but also a loop to iterate (up to?) 100 times that I never fully grokked.
> 
> That loop was to make sure that the lint did not create an infinite loop of linting. That is, if applying that lint would result in a new error that could be detected by that same linter.

Yes that part I do understand now at least since I've had a few linters in progress that got caught looping. I don't think I got it at the time or got confused trying because of the other loops and maybe some identifier names at the time?

> Maybe we should reopen that feature request?

That's a good idea. Have a look at the good_and_bad assert I wrote. It worked on the linters I used it on but I wrote it with the help of an LLM and it passed all the tests for the linters I made it for, but I always felt like there could be something subtle missing because it looked so different to the code for the basic suggestion linter. It might not have that 100 loop part?

hippietrail: For now I've removed `pub fn assert_top5_suggestion_result` and the feature request for a better more general solution has also been reopened.

Issue #1495, opened 2025-07-02
False positive for the word "plugin"


chrisgrieser:
follow-up to #1280, there seem to still be a few cases where the problem occurs.

## Description

Sample text:
```
Color schemes are available by using the Style Settings plugin.
```

"plugin" is flagged with `This word might be a phrasal verb rather than a compound noun`

## Screenshots

![Image](https://github.com/user-attachments/assets/762b1ab6-fc45-4619-84a6-10c8f345a211)

elijah-potter:
I believe @hippietrail wrote this rule. Mind taking look?

hippietrail:
> I believe [@hippietrail](https://github.com/hippietrail) wrote this rule. Mind taking look?

Wow this is more subtle than I first expected. To guess what POS an NV (maybe noun maybe verb) actually is:
- If it follows a noun, it could be part of a compound noun (dictionary lookup) or a verb (couples breakup).
- To decide it uses the heuristic that the qualifying noun in a compound noun (the one that's acting like an adjective) must be singular. Though breaking that rule is a rising common error in its own right.

So "settings plug in" is the same pattern as "couples break up" and gets flagged. Note that the message is only "might be a phrasal verb".

One possibility is that "plugin" is lowercase while "Settings" is uppercase, suggesting that "Settings" could be a 'name', but it could also be the first word in the sentence. So we'd have to check if it's at the start of the chunk. Another possibility is that "Settings" itself is part of a compound noun "Style Settings", which is reinforced by both words being in uppercase, and further reinforced by the words prior to that being in lowercase.

But even then it's ambiguous: "How do Apple Pencils plugin?"

I think the answer is that a small set of nouns make compounds in the plural so should be in a whitelist. Nobody says "cars park" or "trains station" but "settings plugin" sounds more natural than "setting plugin".

elijah-potter:
This looks to be resolved. Thanks @hippietrail!

PR #1502, opened 2025-07-03
fix: fix issue 1495

hippietrail:
# Issues 
#1495 

# Description

Fixes issue #1495 by hardcoding a single-entry whitelist of plural nouns that can form compound nouns, starting with "settings".

# How Has This Been Tested?

A unit test was added with the failing case: "Color schemes are available by using the Style Settings plugin."

There are numerous ways this could be built upon:
- Add a `dictionary.dict` annotation flag for nouns like "settings" that compound with plural forms.
- Add a whitelist as a module in `harper-core` that linters can call.
- Add more `.is_xxx` methods to `WordMetadata` and/or `TokenKind` such as `is_uppercase`, `is_first_word_in_sentence` - these wouldn't have panned out in this case - see the comment in the bug report. But maybe `is_compoundable_plural`?
- `Pattern` or `Expr` for `CompoundNoun` that has such heuristics built-in.

# Checklist

### DIFF
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -34,7 +34,7 @@ impl Linter for PhrasalVerbAsCompoundNoun {
         let mut lints = Vec::new();
         for i in document.iter_noun_indices() {
             // It would be handy if there could be a dict flag for nouns which are compounds of phrasal verbs.
-            // Instead let's use a few heuristics.
+            // Instead, let's use a few heuristics.
             let token = document.get_token(i).unwrap();
             // * Can't also be a proper noun or a real verb.
             if token.kind.is_proper_noun() || token.kind.is_verb() {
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -139,7 +139,13 @@ impl Linter for PhrasalVerbAsCompoundNoun {
                 }
 
                 // "dictionary lookup" is not a mistake but "couples breakup" is.
-                if prev_tok.kind.is_noun() && !prev_tok.kind.is_plural_noun() {
+                // But "settings plugin" is not.
+                if prev_tok.kind.is_noun() && !prev_tok.kind.is_plural_noun()
+                    || prev_tok
+                        .span
+                        .get_content(document.get_source())
+                        .eq_ignore_ascii_case_chars(&['s', 'e', 't', 't', 'i', 'n', 'g', 's'])
```
elijah-potter: Using `eq_ignore_ascii_case_chars` would be more legible. I don't think there would be any negative performance impact from using it either.

Issue #508, opened 2025-01-26
Flag double comparatives


hippietrail:
These consist of redundantly using both the comparative word "more" and the inflected adjectival ending "-er" together. And also irregular comparatives.

- **More** bigg**er** is **more better** → **Bigger** is **better**.

ccoVeille:
You reminded me these memes

![Screenshot_20250126_101528_Firefox.jpg](https://github.com/user-attachments/assets/5af1bf7f-77e8-4c3e-a846-3c0812d55941)

![bodybuilding-school-is-important-but-muscles-are-importanter-alessandra-roth.jpg](https://github.com/user-attachments/assets/5a869460-b30f-4cf5-8df6-ac58669127cc)

ccoVeille:
Joke aside, there might have a need to detect issues such as

- importanter
- anything prefixed with -er that doesn't respect the "more for anything that is a bit long"

Some might be caught by the spell checker, some might not.

hippietrail:
> Joke aside, there might have a need to detect issues such as
> 
> * importanter
> * anything prefixed with -er that doesn't respect the "more for anything that is a bit long"
> 
> Some might be caught by the spell checker, some might not.

I think this should be a separate lint. There are many cases where both more + adj and adj+er are correct. I think some may have even changed since I was a kid. I'll wait until I spot some real ones in the wild. I think that would be realisticer.

ccoVeille:
> I think this should be a separate lint.

I opened 

- #511

ccoVeille:
I worked on extracting this list of short adjective and their comparative

bigger - big
smaller - small
older - old
younger - young
taller - tall
shorter - short
longer - long
nearer - near
farther/further - far
higher - high
lower - low
faster - fast
slower - slow
stronger - strong
weaker - weak
better - good (irregular, but very frequent)
worse - bad (irregular, but very frequent)
happier - happy
easier - easy
heavier - heavy
prettier - pretty
busier - busy
funnier - funny
nicer - nice
finer - fine
later - late
earlier - early
colder - cold
warmer - warm
cooler - cool
darker - dark
lighter - light
brighter - bright
wider - wide
narrower - narrow
thicker - thick
thinner - thin
deeper - deep
shallower - shallow
louder - loud
quieter - quiet
harder - hard
softer - soft
sweeter - sweet
sourer - sour
bitterer - bitter
cleaner - clean
dirtier - dirty
drier - dry
wetter - wet
fairer - fair
gladder - glad
[...]

PR #1507, opened 2025-07-05
feat: detect "more faster", "most best"

hippietrail:
# Issues 
#508 

# Description

Detects "most fastest", "more better", etc.

Superlative annotation flag changed from `T` to `^` to match the comparative `>`. Mnemonic comment fields added.

As a bonus, the same flag, which already also added `er` to verbs to create agent nouns ("eat" → "eater") now sets the correct metadata too.

As another bonus, `SequenceExpr` now has a `::word_set()` constructor and matching `::then_word_set()` method.

Unit tests for all cases have been mined from real sentences found on GitHub.

# Checklist


ccoVeille:
Juste looking at the code very quickly.

I might have missed something do you detect "less faster" ?

https://github.com/search?q=%22less+faster%22&type=code

This could be addressed separately.

hippietrail:
> Juste looking at the code very quickly.
> 
> I might have missed something do you detect "less faster" ?
> 
> https://github.com/search?q=%22less+faster%22&type=code
> 
> This could be addressed separately.

Interesting! I think that should be a separate linter though since this one deals specifically with redundancy.

ccoVeille:
> > Juste looking at the code very quickly.
> > 
> > I might have missed something do you detect "less faster" ?
> > 
> > https://github.com/search?q=%22less+faster%22&type=code
> > 
> > This could be addressed separately.
> 
> Interesting! I think that should be a separate linter though since this one deals specifically with redundancy.

I opened #1546

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
```diff
@@ -254,6 +254,18 @@ Suggest:
 
 
 
+Lint:    WordChoice (126 priority)
+Message: |
+     104 | other languages. Statistics derived by analyzing it formed the basis for most
+         |                                                                          ^~~~~
+     105 | later part-of-speech tagging systems, such as CLAWS and VOLSUNGA. However, by
+         | ~~~~~ The degree of the adverb conflicts with the degree of the adjective.
```
elijah-potter, line 262: Good stuff.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -7135,6 +7135,17 @@ Suggest:
 
 
 
+Lint:    WordChoice (31 priority)
+Message: |
+    5528 | “Now he’s dead,” I said after a moment. “You were his closest friend, so I know
```
elijah-potter, line 7120: This doesn't feel right.

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
hippietrail: > Good stuff.

Actually that's a false positive but probably a very hard one to filter out.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
hippietrail: > This doesn't feel right.

I've seen this happen before. I think it's a side-effect of correcting some `dictionary.dict` entries... Yes I'm pretty sure `closest` wasn't marked as an adjective before and now that it is a latent false positive in the were/we're linter in being triggered.

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
elijah-potter: oh yeah! hmmm...

hippietrail: > oh yeah! hmmm...

Maybe we should keep track of really subtle and challenging false positives in a list somewhere?

PR #1519, opened 2025-07-07
fix: fixes a bug in `SequenceExpr` found when using the `Optional` `Expr` in a linter

hippietrail:
# Issues 

No issue but see this Discord thread where I found it and tried to figure it out: [1391388089372643489](https://discord.com/channels/1335035237213671495/1335035237213671498/1391388089372643489)

# Description

While working on #1514 using the newish `Optional` `Expr` for the first I found that some tests were not doing what I expected and I noticed they were actually crashing.

Some work with an LLM led to this fix. I don't necessarily trust the LLM's findings but it made my new linter behave as expected matching the results from a different `Expr` that matched equivalent runs of tokens.

# How Has This Been Tested?

I couldn't think of a good unit test, so just manual testing.

# Checklist

### DIFF
--- a/harper-core/src/expr/sequence_expr.rs
+++ b/harper-core/src/expr/sequence_expr.rs
@@ -50,15 +50,20 @@ impl Expr for SequenceExpr {
 
         for cur_expr in &self.exprs {
             let out = cur_expr.run(cursor, tokens, source)?;
-
-            window.expand_to_include(out.start);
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
Looks good to me. I'm not sure how we could test this directly. I think your new linter may be good enough as a more integration-ish test.

PR #1529, opened 2025-07-09
feat: phrase set corrections to group phrase corrections

hippietrail:
# Issues 

I'm not sure if there's also an issue but there is this Discussion: https://github.com/Automattic/harper/discussions/1527

# Description

Groups closely related phrase corrections together into a single lint with a single setting.

Especially for phrase corrections dealing with all inflections of a noun or verb, but other close variants will also fit such as "change tack" & "change of tack".

So instead of having two settings, one for correcting "definitive article" to "definite article", and a second for correcting "definitive articles" to "definite articles", they are now grouped under a single setting, generally the one with the uninflected name (singular for nouns, present/infinitive for verbs).

There's a simpler constructor for ones that only map single bad variants to single suggestions.

And there's a more complex one for many-to-many, such as correct "get rid off" and "get ride off" to "get rid of" for each conjugation of `to get`.

# How Has This Been Tested?

I've moved all the tests over from phrase corrections and they're all passing.

# Checklist


elijah-potter (changes requested):
Your code looks good. I wouldn't have done it any other way.

I would like for you to update the documentation to match these changes. Make sure the identifiers and macro invocations are updated in `packages/web/src/routes/docs/contributors/author-a-rule/+page.md` and the distinction between the two available macros is clearly explained.

PR #1533, opened 2025-07-10
feat(cli): add `annotate-tokens` subcommand

86xsk:
# Description

Adds a harper-cli command to annotate tokens. This should hopefully make it easier to view and analyze the way we parse tokens. My primary motivation for this was to make it easier to view how certain sentences are parsed via the POS tagger. To the best of my knowledge, currently the only way to get this information is via the `parse` command, whose output (unless processed externally) is not very human-readable.

Though this currently only supports UPOS annotations, I tried to make this command as easily extensible as possible. To add a new annotation method, one would just have add the respective variant to `AnnotationType`, and add that variant to the match statement in `Annotation::from_token()`.

Some potential improvements include:
- Allowing `AnnotationType` to contain arguments/parameters. This could allow for the merging of the `spans` command into `annotate-tokens` (as an `AnnotationType`).
- Ensuring that distinctive colors are generated in `get_color_for_enum_variant()`. The implementation currently uses `ariadne::ColorGenerator`, which is simple to use, but does not guarantee that the colors generated are distinguishable from one another.

# Demo

image

# How Has This Been Tested?
- `cargo test`

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -840,6 +840,8 @@ dependencies = [
  "hashbrown 0.15.4",
  "serde",
  "serde_json",
+ "strum",
+ "strum_macros",
 ]
 
 [[package]]
--- a/harper-cli/Cargo.toml
+++ b/harper-cli/Cargo.toml
@@ -19,6 +19,8 @@ harper-typst = { path = "../harper-typst", version = "0.49.0" }
 hashbrown = "0.15.4"
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = "1.0.140"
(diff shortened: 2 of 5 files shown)


elijah-potter (approved):
I've been wanting this functionality for a while now. Thanks for getting it in. Your implementation looks good to me. I see no reason to wait to merge it.

hippietrail:
🖼️ Pretty! I noticed that some words don't get any explanation at all while others get an explanation of `NONE` - I'll probably file an issue when I explore it a bit.

PR #1536, opened 2025-07-10
feat(core): add rule for redundant modal verbs

elijah-potter:
# Description

This rule looks for the essentially impossible sequence of `MD MD` in English text. As it stands, I have not been able to locate a false-positive.

# How Has This Been Tested?

Additional unit tests

# Checklist

### DIFF
--- a/harper-core/src/expr/sequence_expr.rs
+++ b/harper-core/src/expr/sequence_expr.rs
@@ -23,6 +23,13 @@ macro_rules! gen_then_from_is {
                 })
             }
 
+            #[doc = concat!("Adds an optional step matching a token where [`TokenKind::is_", stringify!($quality), "()`] returns true.")]
+            pub fn [< then_optional_$quality >] (self) -> Self{
+                self.then_optional(|tok: &Token, _source: &[char]| {
+                    tok.kind.[< is_$quality >]()
+                })
+            }
+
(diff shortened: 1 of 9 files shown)


harper-core/src/patterns/modal_verb.rs
```diff
@@ -0,0 +1,26 @@
+use super::{Pattern, WordSet};
```
hippietrail, line 1: I almost made a `ModalVerb` `Expr` a few times but I didn't because of a few potential tricky problems:

The obvious simple modal verbs are the single word ones that all work the same like can, could, may, might, must, shall, should, will, and would.

But dare, ought, and need are also in the lists but don't really behave in all the same ways.

Then there are the multi-word ones like "have to" and ... "need to" and "ought to" that are modal in meaning but behaving differently in sentence structure. And note there may be some kind of fuzzy overlap with those last two.

Partly I didn't know how to name a set of just the traditional single word ones that behave the same, vs the "... to" ones, vs a group containing them all.

Then there's their negates, which in the case of the simple ones, behave just like the positive ones: you can use "can't" anywhere you can use "can", etc. Well except if they're already followed by "not".

Then there's the fact that people leave the apostrophes out so often that they'd lead to many false negatives if "couoldnt", "shouldnt", etc. are not included.

I was leaning towards having a constructor that took some kind of argument about which ones you want included. Or having multiple constructors, like `::with_negatives()`, `::with_missing_apostrophes()`, etc.

Issue #1548, opened 2025-07-12
Sentence-final punctuation linter confused by `.exe`


hippietrail:
## Description

In the text `Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and ...` Harper complains "Unnecessary space at the end of the sentence."

<img width="917" height="135" alt="Image" src="https://github.com/user-attachments/assets/91ed9e87-c6ca-4ef7-9256-0f8a4446a124" />

It could check for text after the `.` but It's very common for people to put a space before the `.` instead of after it, which would cause false negatives. Perhaps a specific check that what follows the `.` looks like a file extension: three letters long either all lowercase or all uppercase?

elijah-potter:
This should be fixed now, no?

hippietrail:
> This should be fixed now, no?

Yes.

PR #1549, opened 2025-07-12
fix: don't flag `.exe` as space before full stop

hippietrail:
# Issues 

Solves #1548

# Description

The `spaces` linter would flag .exe in mid sentence as an unnecessary space at the end of a sentence.

Screenshot 2025-07-12 at 11 30 00 am

This fixes that by peeking at the start of the next sentence and preventing the flag is the next token is a 3-letter word with all uppercase or all lowercase alphabetic characters.

# How Has This Been Tested?

I added a unit test using the snippet that revealed the false positive, and a bunch I made up to check edge cases.

# Checklist


elijah-potter (commented):
I feel like this should exist at a higher level (somewhere in the `parse()` step of the `Document`). Would it be possible to condense period + word tokens into the latter if they're file extensions?

hippietrail:
> I feel like this should exist at a higher level (somewhere in the `parse()` step of the `Document`. Would it be possible to condense period + word tokens into the latter if they're file extensions?

Makes perfect sense. That's an area of the code I haven't delved into. I would probably use a version of my logic here. I count it only if it's 3 letters, no other symbols, and only if all lowercase or all uppercase. Of course file extensions can now be arbitrary but we'd be risking false positives and unforeseen edge cases. Otherwise have a whitelist of common file extensions. I'd also only count it as a "word" if preceded by whitespace.

<s>Do you want to take a stab at it?</s>

hippietrail:
> I feel like this should exist at a higher level (somewhere in the `parse()` step of the `Document`). Would it be possible to condense period + word tokens into the latter if they're file extensions?

I feel it shouldn't condense them when they're part of a filename of hostname, just when they follow whitespace (or start anchor) but so far can't see how to do that ...

hippietrail:
> I feel like this should exist at a higher level (somewhere in the `parse()` step of the `Document`). Would it be possible to condense period + word tokens into the latter if they're file extensions?

It took a few attempts but I got a version I'm happy with: https://github.com/Automattic/harper/pull/1602

Issue #1473, opened 2025-06-28
Should have error for "in" in "I'm in the 3rd floor"


thuan1412:
## Description

Give a detailed description of the grammatical error Harper should be able to find.

## Resources

If there are any resources online we can reference that described grammatical rules, let us know.

https://dictionary.cambridge.org/grammar/british-grammar/at-on-and-in-place#:~:text=to%20talk%20about%20a%20floor%20in%20a%20building

https://www.accelerateenglish.com/post/using-in-on-at-for-location#:~:text=For%20example%2C%20%22the%20floor%22%20is%20a%20flat%20surface%20and%20you%20can%20stand%20%22on%20the%20floor%22%20by%20touching%20the%20surface%20of%20the%20floor%20with%20your%20feet%20but%20the%20floor%20come%20up%20and%20surround%20you%20like%20a%20room%20does.

## Examples

Please provide at least three examples of the grammatical error in real text.

"I'm in the 3rd floor" -> "I'm on the 3rd floor"
"They live at the 3rd floor" -> "They live in the 3rd floor"
"My company is in the 11th floor" -> "My company is on the 11th floor"

## Potential False Positives

If there is any grammatical context in which the error may lie (for example, a word may always be invalid if preceded by a preposition), let us know. Provide a few examples, if possible.

elijah-potter:
It looks like this was implemented. Should we close this?

hippietrail:
> It looks like this was implemented. Should we close this?

yes `harper-core/src/linting/on_floor.rs`

PR #1555, opened 2025-07-13
feat: add lint for on the floor phrase

thuan1412:
# Issues 

https://github.com/Automattic/harper/issues/1473

# Description

This PR resolves issue https://github.com/Automattic/harper/issues/1473 by adding lint for the `on the floor` phrase.

# Demo

# How Has This Been Tested?

Unit-tests is passed.

# Checklist


harper-core/src/linting/on_floor.rs
```diff
@@ -0,0 +1,113 @@
...
+                "Corrects `{}` to `on` when talking about position inside a building",
+                incorrect_preposition
+            )
+            .to_string(),
+            priority: 63,
+        })
+    }
+
+    fn description(&self) -> &'static str {
+        "A rule that looks for incorrect uses of the preposition `in` or `at` when telling position inside a building. It suggests replacing it with `on the floor`."
```
hippietrail: The wording is a bit awkward. I asked a random LLM and it came up with
```
"This rule identifies incorrect uses of the prepositions `in` or `at` when referring to locations inside a building and recommends using `on the floor` instead."
```
Either way I'd probably put `nth` or pick an example floor number.

Otherwise it all looks good to me!

I like to find real-world examples to complement my made-up examples. Preferably from GitHub. Sometimes doing that reveals false positives you didn't think of. In fact that just happened to me. I didn't imagine any potential false positives but when I started searching these don't look wrong to me:
- The gate at the ground floor to prevent people going down when they are panicked.
- ... at the first floor level
- The elevator can stop either at the ground floor or the first floor.

I think they all sound OK with "on" as well. The second one sounds right with "at" because it goes with "level". The third one sounds right because "stop" goes with "at".

In fact in other sentences "at" seems OK if it's talking about where the lift/elevator is but sound wrong if it's talking about where a person is.

Sometimes I start to question my own native-speaker intuition!

thuan1412: Thanks for quick review. I've update the description.

Could you elaborate on the "Preferably from GitHub" point?

I've just checked from Grammarly about the 3 above example, the two first one are suggested to replace `at` to `on`. The `on` in the last example is correct. For future improvement, I think we will need to have separated lint for each preposition, it might quite complicated, let's address later.

thuan1412:
hmm, the lint check fails for the `harper-core/tests/text/The Great Gatsby.md`, I will check later

hippietrail:
> hmm, the lint check fails for the `harper-core/tests/text/The Great Gatsby.md`, I will check later

It will fail when anything changes, whether a good or bad change when you do `cargo test` the first time but will succeed the second time you do `cargo test`. But check the diffs to see if it caused a false positive or found a legit mistake, etc.

thuan1412:
@hippietrail Turns out it's a legit mistake. 

```
Lint:    WordChoice (63 priority)
Message: |
    4589 | Tom stopped beside the porch and looked up at the second floor, where two
         |                                            ^~ Corrects `at` to `on` when talking about position inside a building
    4590 | windows bloomed with light among the vines.
Suggest:
  - Replace with: “on”
```

I think we need a more comprehensive rule for this case 😅 . Do you have any suggestion?

thuan1412:
Asked ChatGPT about the rule. In most case, we use `on`. On some rare case, for `in` and `at`, the static context about the sentence is not enough, we need to know the semantic context, (what the sentence is meaning) to know.

Preposition | Usage Description | Acceptable Example | Notes
-- | -- | -- | --
on | Level of a building | "on the second floor" | ✅ Most common
in | Inside the structure/material of the floor | "pipes in the floor" | ❌ Not for levels
at | Point of arrival (very limited use) | "elevator stops at the third floor" | ❗ Rare and specific

hippietrail:
> @hippietrail Turns out it's a legit mistake.
> 
> ```
> Lint:    WordChoice (63 priority)
> Message: |
>     4589 | Tom stopped beside the porch and looked up at the second floor, where two
>          |                                            ^~ Corrects `at` to `on` when talking about position inside a building
>     4590 | windows bloomed with light among the vines.
> Suggest:
>   - Replace with: “on”
> ```
> 
> I think we need a more comprehensive rule for this case 😅 . Do you have any suggestion?

To me this is correct with "at" and wrong with "on". In fact "look on something" instead of "look at" is a common EuroEnglish mistake we haven't tackled yet.

hippietrail:
> Asked ChatGPT about the rule. In most case, we use `on`. On some rare case, for `in` and `at`, the static context about the sentence is not enough, we need to know the semantic context, (what the sentence is meaning) to know.
> 
> Preposition	Usage Description	Acceptable Example	Notes
> on	Level of a building	"on the second floor"	✅ Most common
> in	Inside the structure/material of the floor	"pipes in the floor"	❌ Not for levels
> at	Point of arrival (very limited use)	"elevator stops at the third floor"	❗ Rare and specific

The LLMs have extremely good "intuition" about English grammar and only make mistakes that are already overwhelmingly common among native speakers. But they are less good at explaining rules. The way you phrase the questions and the examples you show them seem to make a difference. I was even able to get one to admit it's defence of a rule did not match its own use! (-:

I trust them with questions about foreign languages I'm learning but I'm more skeptical about their answers about my native language.

I would say when the verb is "like be" it describes that location of the subject, it will be "on". Which is the obvious case we want to address when it's wrong, but in other cases, especially when there's a preposition that's an idiomatic part of a phrasal verb, it can be questionable.

I wish for a tool that can scan our (attempted?) grammar patterns in both lots of known-to-be-edited texts on the one hand vs lots of known-to-be-full-of-common-mistake texts on the other hand. Project Gutenberg is probably a good source of the former. Scraping Reddit or Medium would be great for the latter and scraping GitHub repos would be on-target.

thuan1412:
>To me this is correct with "at" and wrong with "on". In fact "look on something" instead of "look at" is a common EuroEnglish mistake we haven't tackled yet.

@hippietrail yes, "at" is correct. I'm thinking about adding a check for the words before the preposition. For example, "look up", "stop". I think we can't check for all possible cases, so let's just check of cases we can think of, and improve the rule later.

What do you think?

hippietrail:
> > To me this is correct with "at" and wrong with "on". In fact "look on something" instead of "look at" is a common EuroEnglish mistake we haven't tackled yet.
> 
> @hippietrail yes, "at" is correct. I'm thinking about adding a check for the words before the preposition. For example, "look up", "stop". I think we can't check for all possible cases, so let's just check of cases we can think of, and improve the rule later.
> 
> What do you think?

Yes I usually try to think of the most general checks, based on parts of speech (noun verb adjective) or properties (singular plural tense) of surrounding words. But sometimes that's not enough and I have to use a whitelist or blacklist. Sometimes I have to use all of those techniques. This is why you want to try to gather a bunch of real test cases. The worse case is when it's based on semantics rather than specific kinds of words, properties of words, or lists of words.

thuan1412:
hey @hippietrail , I've added some exceptions for the rule. All checks have passed now.

thuan1412:
> I like to find real-world examples to complement my made-up examples. Preferably from GitHub ....


btw, could you share how you find real-world example?

hippietrail:
> > I like to find real-world examples to complement my made-up examples. Preferably from GitHub ....
> 
> btw, could you share how you find real-world example?

I use Google with lots of `OR` operators and sometimes the `*` operator. I found this one quite difficult to Google for well though. I usually try many searches and restrict to `site:github.com` and sometimes widen to include Stack Overflow, Slashdot, and when I have to Medium and Reddit, which are chock full of mistakes.

Here's one of the ones I tried for this yesterday that was still open in a browser tab: `"in OR at the ground OR 1st OR first OR 2nd OR second OR 3rd OR third OR 4th OR fourth OR 5th OR fifth floor" site:github.com`

hippietrail:
> hey @hippietrail , I've added some exceptions for the rule. All checks have passed now.

I'm new to GitHub reviewing so at the risk of doing it wrong I'll see if I can mark it as passing my review and then we shall see what @elijah-potter thinks. No doubt there will be some false positives and negatives we'll hear about or find ourselves.

hippietrail (approved):

harper-core/src/linting/lint_group.rs
```diff
@@ -435,6 +436,7 @@ impl LintGroup {
         insert_expr_rule!(NounInsteadOfVerb, true);
         insert_struct_rule!(NumberSuffixCapitalization, true);
         insert_struct_rule!(OfCourse, true);
+        insert_struct_rule!(OnFloor, true);
```
elijah-potter: To take advantage of Harper's aggressive caching for `ExprLinter`s, change this to `insert_expr_rule` instead. It won't change behavior at all, but it will make incremental lints faster.

elijah-potter (changes requested):
I see no major problems here. You've written some pretty idiomatic code here. Once you make the one tweak, it's ready to merge.

harper-core/src/linting/lint_group.rs
thuan1412: Thanks @elijah-potter , I've changed to `insert_expr_rule `

PR #1565, opened 2025-07-15
feat: VerbInsteadOfNoun to complement NounInsteadOfVerb

hippietrail:
# Issues 
#402 

# Description

Flags verbs used where their related noun belongs. For now only after adjectives as in "good advise".

I moved the set of noun verb pairs into its own file both linters can share.
A couple of false positives are prevented but some others may require more context.
I removed the `J` adjective annotation from "gasoline" in the dictionary that seems to have come from Wiktionary. All nouns can be used as qualifiers and no other dictionary lists it as other than a noun. Was causing false positives in The Great Gatsby.

# How Has This Been Tested?

I created unit tests based on sentences mined from GitHub.

# Checklist


harper-core/src/linting/noun_verb_pairs.rs
```diff
@@ -0,0 +1,11 @@
+/// Common noun-verb pairs that are often confused
```
elijah-potter: Since these guys share code, it might make more sense to put them in their own module and expose them as a single rule using the `merge_linters!` macro.

harper-core/src/linting/verb_instead_of_noun.rs
```diff
@@ -0,0 +1,190 @@
...
+            &NOUN_VERB_PAIRS
+                .iter()
+                .map(|&(_, verb)| verb)
+                .collect::<Vec<_>>(),
+        ));
+        Self {
+            expr: Box::new(
+                SequenceExpr::default()
+                    .then_adjective()
+                    // .then(|tok: &Token, _: &[char]| tok.kind.is_adjective() && !tok.kind.is_noun())
```
elijah-potter: Why was this commented-out?

hippietrail: > Why was this commented-out?

Good question! Let me check... ah it turned out upon investigation that the sentence that prompted me to add that caveat was wrongly annotated in `dictionary.dict` - it had "gasoline" marked as an adjective as well as an adjective, which it obviously got from Wiktionary which also does so. But I checked Cambridge, Collins, Longman, Oxford, and Websters - all consider it a noun only. So I fixed the dictionary entry and took out the special logic but forget to remove it after checking it now worked without it.

harper-core/src/linting/noun_verb_pairs.rs
hippietrail: Will check that out!

hippietrail: Done. That might be interesting for combining phrase_corrections with custom linters for some confusable pairs.

PR #1566, opened 2025-07-15
feat: IIRC correctly→IIRC

hippietrail:
# Issues 
N/A

# Description

I noticed myself type "if IIRC" and a quick search revealed I'm not alone and that "IIRC correctly" is even more common.
This fixes both.

(I also had to add `IIRC` to `dictionary.dict`.)

# How Has This Been Tested?

I added a few unit tests using sentences found on GitHub.

I did have to `#ignore` a test affected by the case-matching bug in `replace_with_match_case` though. 

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -4753,6 +4753,7 @@ IDF/Og              # Israel's military
 IE/ONJ
 IED/N               # improvised explosive device
 IEEE/O              # standards organization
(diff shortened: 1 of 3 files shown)


ccoVeille:
Context: IIRC is an internet abbreviation that stands for if I recall correctly or if I remember correctly

elijah-potter (approved):

PR #1568, opened 2025-07-15
doc(core): document `lint_group` macros

elijah-potter:
# Issues 

Revealed while reviewing #1555 

# Description

I realized some macros in `lint_group.rs` were not properly documented. This makes their function clearer, as well as instructing rule authors on when to use each one.


harper-core/src/linting/lint_group.rs
```diff
@@ -360,6 +361,9 @@ impl LintGroup {
             };
         }
 
+        /// Add an `ExprLinter` to the group, setting it to be enabled by default.
+        /// While you _can_ pass an `ExprLinter` to `insert_struct_rule`, using this micro instead
```
hippietrail: Typo alert: `this micro`→`this macro`

Issue #1170, opened 2025-04-29
Ignore entire file


neo451:
## Description

I have markdown files that I want to be ignored entirely, `harper-ls` should also provide option to ignore whole file.

## Component

Which component of Harper would this feature affect? For example:
- Core grammar checking engine
- The `harper-ls` language server

## Use Cases

Describe specific scenarios where this feature would be useful. Include examples of text or situations where this feature would be particularly beneficial.

## Implementation Ideas

For harper-ls, I think I want either:

1. in server config specify a list of files to ignore, less flexible.
2. write  cues for harper-ls to read and ignore
   - For markdown files this can be a field in yaml head like `harper: false`
   - For language comments can be a comment at the top of the file like `---harper ignore file`.

## Existing Solutions

like for vim's builtin spell checker, I would just `set nospell`.

lhilden:
I would like to see an `Ignore file in Harper` menu item in the note context menu (3-dots or mouse right-click menu):
![Image](https://github.com/user-attachments/assets/f74997a5-79ec-4fe8-b5fc-d2efe8c35b1a)

mcecode:
Hi, it's not a complete solution to this issue but  the `harper.excludePatterns` config just landed in v0.58.0 and if you don't have many files you want to ignore (e.g., `["*/my-file-1.md", "*/my-file-2.md"]`) or if they're in a specific folder (e.g., `["**/ignored/*.md"]`), it could work as an interim solution while we don't have a way to ignore files in a more fine-grained way.

elijah-potter:
> in server config specify a list of files to ignore, less flexible.

@mcecode, I believe your solution fulfills this requirement. Do we want to tackle the other use-case, or should we close this issue?

mcecode:
> Do we want to tackle the other use-case, or should we close this issue?

As it got quite a few likes, I kept it open since I was contemplating how to tackle ignoring files in a more fine-grained way. Since OP mentioned `set nospell` from Vim, I was thinking we could have a similar command for `harper-ls` where a user could turn linting on or off for a file. It could be convenient and useful for people (like me) who like to write things down before getting feedback and looking for the mistakes they've made. That said, I haven't had much time recently to tackle things and make PRs, so I haven't gotten much progress. I think this could be closed since, as you mentioned, my solution did solve this somewhat. If ever I do get around to implementing my idea, I can just make a PR unrelated to this issue.

PR #1569, opened 2025-07-15
feat(obsidian): allow users to exclude files and paths

elijah-potter:
# Issues 

#1170 

# Description

This PR makes it possible to ignore files or entire directories from linting in the Obsidian plugin. This works by adding a field in the settings menu where users can set one or more globs to be ignored. 

# Demo

image

# How Has This Been Tested?

Manually.	

# Checklist

### DIFF
--- a/packages/obsidian-plugin/src/State.test.ts
+++ b/packages/obsidian-plugin/src/State.test.ts
@@ -7,7 +7,7 @@ function createEphemeralState(): State {
 	return new State(
 		(_) => Promise.resolve(),
 		() => {},
-    undefined
+		undefined,
 	);
 }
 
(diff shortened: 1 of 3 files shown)


elythron:
How is this feature intended to work? Does it fall back to Obsidian’s native spellcheck, or does it simply ignore the specified file or folder?

Either way, in my case, it doesn’t seem to work at all. I’ve tried several patterns:

```
TEST
TEST/
TEST/**
'TEST/**'
./TEST/
./TEST/**
TEST/noteTEST
TEST/noteTEST.md
```

Obsidian has the correct spellcheck languages configured. The `data.json` file (including `ignoredGlobs` entry) shows the correct pattern, but the files are still processed — Harper continues to function in the ignored notes.

I’ve also reproduced this in a brand new vault with no community plugins other than Harper.

PR #1576, opened 2025-07-16
perf(core): prefer `FirstMatchOf` when possible

86xsk:
# Description

Changes:
- (025883773492ee0a8fca51cf0be7bcf67cf79ece)
  - Adds `then_either` and `then_either_longest` to `SequenceExpr` to make it easier to write expressions with such logic. Refactors users of `SequenceExpr` to use these new methods where appropriate.
- (8f62f8724d36372656bbe2840da0d877e777644d)
  - Refactors `FirstMatchOf`'s push/add function so that its signature and name matches the equivalent function in `LongestMatchOf`. Updates calls to this function where necessary.
- (afd35f31a4cd0f708e9dbc3c7c0f644fd29ceadb)
  - Replaces some uses of `LongestMatchOf` with `FirstMatchOf` where a longest match seems unnecessary. This should improve performance to some extent, since `FirstMatchOf` is short-circuiting. Though I made sure all tests continue pass and reviewed each potential replacement individually, there is a chance that some of the replacements may be incorrect. 
- (b4a70b97384563b686249bcf1aff6a5873159add) 
  - Moves the current functionality of `OwnedExprExt::or()` to `OwnedExprExt::or_longest()`, and changes `OwnedExprExt::or()` to use `FirstMatchOf` instead of `LongestMatchOf`. Similarly to the aforementioned, adjusts callers of the `or()` function as appropriate.
- (c7fce86d2c102d732078552740d27517e4249960) 
  - Simplifies the iterator used in `FirstMatchOf`. 

# How Has This Been Tested?
- `cargo test`

# Checklist


hippietrail:
Looks great!

["Either" means "one of two options"](https://dictionary.cambridge.org/grammar/british-grammar/either). I actually renamed `LongestMatchOf` from its old name `Either` and `FirstMatchOf` from its old name which I forget but it was non-obvious and was a little used `Pattern` (before the change to `Expr`).

So I would suggest a name like "any" or "any_of" or something that expresses that it can take any number of options.

harper-core/src/expr/sequence_expr.rs
```diff
@@ -109,16 +109,20 @@ impl SequenceExpr {
     }
 
     /// Pushes an expression that will match any of the provided expressions.
-    pub fn then_either(mut self, exprs: Vec<Box<dyn Expr>>) -> Self {
+    ///
+    /// If more than one of the provided expressions match, this function provides no guarantee
+    /// as to which match will end up being used. If you need to get the longest of multiple
+    /// matches, use [`Self::then_longest_of()`] instead.
+    pub fn then_any_of(mut self, exprs: Vec<Box<dyn Expr>>) -> Self {
```
hippietrail, line 121: Nice!

I've also been thinking of good descriptions of when to use `FirstMatchOf` vs `LongestMatchOf`. I think the key is that you *need* `LongestMatchOf` when your `Pattern`/`Expr` doesn't have a fixed number of tokens it will match *and* they can overlap. It's like the concept of "greedy" in regular expressions.

It's most apparent if the `Expr` has optional or repeating parts, but I think it can come up in other places.

Maybe the rule of thumb might be just if the options/arms are obviously distinct, use `FirstMatchOf`, if it's complex use `LongestMatchOf`. What do you think?

86xsk: >I think the key is that you _need_ `LongestMatchOf` when your `Pattern`/`Expr` doesn't have a fixed number of tokens it will match _and_ they can overlap. It's like the concept of "greedy" in regular expressions.

I think that's an excellent way to define/concretize it. With that in mind, I now kind of want to see if I can write a `debug_assert!()` (or something of that nature) in `FirstMatchOf` to catch such cases. Adding the extra usage hints in the documentation certainly wouldn't hurt either.

Although, would it be wrong to say that the problematic ambiguity can occur even with a fixed number of tokens? (As an imaginary unrealistic example, perhaps a `FirstMatchOf` that matches both of something like `.then_period()` and `.then_punctuation()`?) Perhaps the problematic use of `FirstMatchOf` can be narrowed down to 'overlapping patterns, with one of them being longer than the other(s)', or even 'overlapping patterns' in general?

hippietrail: > I think that's an excellent way to define/concretize it. With that in mind, I now kind of want to see if I can write a `debug_assert!()` (or something of that nature) in `FirstMatchOf` to catch such cases. Adding the extra usage hints in the documentation certainly wouldn't hurt either.

Yes we're making a conscious effort to improve the documentation at the moment, including the code comments, and including improving identifier names when needed. Adding more debug asserts is also a great idea. Do you think it might be possible to come up with some kind of test/assert to differentiate `FirstMatchOf` and `LongestMatchOf`? Would it be of any value?

> Although, would it be wrong to say that the problematic ambiguity can occur even with a fixed number of tokens? (As an imaginary unrealistic example, perhaps a `FirstMatchOf` that matches both of something like `.then_period()` and `.then_punctuation()`?) Perhaps the problematic use of `FirstMatchOf` can be narrowed down to 'overlapping patterns, with one of them being longer than the other(s)', or even 'overlapping patterns' in general?

Yes it's definitely when a match is longer but so many patterns have quite simple alternatives that for newcomers it might not be obvious what it means. They might expect a longer word would match whereas a word of any length is still just one token. Maybe I'm overthinking it?

But you make me realize it may be possible to have another kind of ambiguity affecting token length that I hadn't considered. I always make a conscious effort to make my patterns exactly as narrow or wide as they need to be.

elijah-potter (dismissed):
I like this work! It cleans up the code quite nicely. You're absolutely right: For performance reasons, we should prefer `FirstMatchOf`. 

However, likely due to Amdahl's law, I don't see any improvement in `cargo bench` as compared to `master`.

PR #1578, opened 2025-07-17
feat: add letter-case flags to `WordMetadata`

hippietrail:
# Issues 
N/A

# Description

Adds letter-case flags to `WordMetadata` (plus other orthography flags.)

These are all for the "case-folded lexemes" in the dictionary, independent of the spelling in the document. So even at the start of a sentence or in all caps you can check how the words are capitalized in the dictionary, etc.

Also added a `harper-cli case-variants` command that will print all the words in the dictionary that fold multiple case variants into a single entry.

# How Has This Been Tested?

I added unit tests for camel case words and case-folded words.

# Checklist


harper-cli/src/main.rs
```diff
@@ -666,6 +670,33 @@ fn main() -> anyhow::Result<()> {
             println!("\nFound {} compound word groups", results.len());
             Ok(())
         }
+        Args::CaseVariants => {
+            let case_bitmask = OrthFlags::LOWERCASE
+                | OrthFlags::TITLECASE
+                | OrthFlags::ALLCAPS
+                | OrthFlags::LOWER_CAMEL
+                | OrthFlags::UPPER_CAMEL;
```
86xsk, line 678: Might it be worth it to add this value as a function or const of `OrthFlags`? It would likely make it easier to keep this value up to date if we decide to add more casing types in the future, especially if this value is likely to be used elsewhere in the codebase.

harper-cli/src/main.rs
```diff
@@ -666,6 +670,33 @@ fn main() -> anyhow::Result<()> {
...
+            let mut count = 0;
+            for word in dictionary.words_iter() {
+                if let Some(metadata) = dictionary.get_word_metadata(word) {
+                    let orth = metadata.orth_info;
+                    let bits = orth.bits() & case_bitmask.bits();
+                    // Is more than one flag set?
+                    if bits.wrapping_sub(1) & bits != 0 {
```
86xsk:
```suggestion
                    if bits.count_ones() > 1 {
```

Would be more readable in my opinion.

harper-cli/src/main.rs
```diff
@@ -666,6 +670,33 @@ fn main() -> anyhow::Result<()> {
...
+                        // Mask out all bits except the case-related ones before printing
+                        let masked_orth =
+                            OrthFlags::from_bits_truncate(orth.bits() & case_bitmask.bits());
```
86xsk:
```suggestion
                        let masked_orth = OrthFlags::from_bits_truncate(bits);
```

I think we can just reuse `bits` here.

harper-cli/src/main.rs
```diff
@@ -666,6 +670,33 @@ fn main() -> anyhow::Result<()> {
...
+                        println!(
+                            "{count} '{}' : {:?}",
+                            word.iter().collect::<String>(),
```
86xsk:
```suggestion
                            word.to_string(),
```

harper-core/src/spell/rune/attribute_list.rs
```diff
@@ -229,6 +233,111 @@ impl AttributeList {
     }
 }
 
+/// Gather metadata about the orthography of a word.
+fn check_orthography(word: &MarkedWord) -> OrthFlags {
+    use crate::char_ext::CharExt;
+    use crate::word_metadata::OrthFlags;
+
+    let mut ortho_flags = OrthFlags::default();
+    let mut saw_letter = false;
```
86xsk: I think this can be simplified slightly by determining `saw_letter` with 
```rust
let saw_letter = letter_count > 0;
``` 
(after the declaration of `letter_count`), rather than determining it in the loop.

harper-core/src/spell/rune/attribute_list.rs
```diff
@@ -229,6 +233,111 @@ impl AttributeList {
...
+            if is_first_char {
+                first_is_lower = true;
+            } else {
+                saw_lower_after_first = true;
+                if upper_to_lower {
+                    lower_to_upper = true;
+                }
+                upper_to_lower = true;
+            }
+        } else if c.is_uppercase() {
```
86xsk: ~~Not too sure about this one, but if we're only considering English letters, we might be able to use the simpler and more performant `is_ascii_{lower|upper}case()` methods instead. It hinges on whether there are casable(?) English characters that aren't covered by ASCII.~~

Never mind, certain letters with diacritics can be considered to be English-lingual and aren't covered by ASCII.

harper-core/src/word_metadata.rs
```diff
@@ -933,10 +956,59 @@ impl Default for DialectFlags {
...
+    // Starts with a lowercase letter but also contains uppercase letters.
+    LowerCamel = 1 << 3,
+    // Starts with an uppercase letter but also contains lowercase letters. (Superset of Titlecase.)
+    UpperCamel = 1 << 4,
+    // Contains at least one space.
+    Multiword = 1 << 5,
+    // Contains at least one hyphen.
+    Hyphenated = 1 << 6,
+    // Contains an apostrophe, so it's a possessive or a contraction.
+    Apostrophe = 1 << 7,
```
86xsk: I would recommend using `///` here so these comments are included in the documentation.

harper-core/src/word_metadata.rs
```diff
@@ -564,6 +571,23 @@ impl WordMetadata {
...
+    }
+    pub fn is_allcaps(&self) -> bool {
+        self.orth_info.contains(OrthFlags::ALLCAPS)
+    }
+    pub fn is_lower_camel(&self) -> bool {
+        self.orth_info.contains(OrthFlags::LOWER_CAMEL)
+    }
+    pub fn is_upper_camel(&self) -> bool {
+        self.orth_info.contains(OrthFlags::UPPER_CAMEL)
+    }
```
86xsk: I don't think this use of documentation comments is correct. rust-analyzer shows that the 'Orthographic queries' documentation is only applied to `is_lowercase`, and the other functions have no documentation.

harper-cli/src/main.rs
hippietrail: > Might it be worth it to add this value as a function or const of `OrthFlags`? It would likely make it easier to keep this value up to date if we decide to add more casing types in the future, especially if this value is likely to be used elsewhere in the codebase.

I usually hold off on adding methods like that until they seem needed. `harper-cli` is mostly an internal tool so doesn't quite qualify as "needed" (-:

harper-cli/src/main.rs
hippietrail: > Would be more readable in my opinion.

Cool! I knew of the C trick but I don't know the Rust APIs too well yet.

harper-cli/src/main.rs
hippietrail: Wow. I've been using `.iter().collect::<String>()` for months! But I think only in printf-debugging.

harper-core/src/word_metadata.rs
hippietrail: > I don't think this use of documentation comments is correct. rust-analyzer shows that the 'Orthographic queries' documentation is only applied to `is_lowercase`, and the other functions have no documentation.

Ah I haven't developed a feeling for how these things work yet.

harper-core/src/word_metadata.rs
```diff
@@ -933,10 +1007,59 @@ impl Default for DialectFlags {
     }
 }
 
+/// Orthography information.
```
elijah-potter: I think this is complex enough to be refactored into its own module.

harper-core/src/word_metadata.rs
```diff
@@ -1623,4 +1746,51 @@ mod tests {
...
+    }
+
+    #[test]
+    fn db_and_kw_symbols_are_lower_camel_case() {
+        // dB, kW
+        assert!(md("db").orth_info.contains(OrthFlags::LOWER_CAMEL));
+    }
+
+    #[test]
+    fn am_is_lowercase_and_titlecase_and_all_caps() {
```
elijah-potter: Good stuff. I want to integrate this into lints as soon as possible.

elijah-potter (changes requested):
YES! We desperately needed this. Good work. I have just a few notes.

Could you also add some test cases from: https://itsfoss.com/harper-grammar-checker?

harper-cli/src/main.rs
hippietrail: > I think we can just reuse `bits` here.

We don't use vanilla `bits` because the output gets flooded with `apostrophe` stuff that makes the signal harder to spot among the noise.

harper-core/src/word_metadata.rs
hippietrail: > I think this is complex enough to be refactored into its own module.

What do you have in mind?

harper-core/src/word_metadata.rs
hippietrail: > Good stuff. I want to integrate this into lints as soon as possible.

Keep in mind that it still leaves something to be desired in comparison to de-case-folding the dictionary. For instance we can mark that a word is `lowerCamel` or `UpperCamel` but not at which positions in the word the case changes happen.

harper-core/src/word_metadata.rs
elijah-potter: I would just put `Orthography` and its various function implementations in their on `orthography.rs` file.

harper-core/src/word_metadata.rs
```diff
@@ -9,8 +9,12 @@ use strum_macros::{Display, EnumCount, EnumString, VariantArray};
 use std::convert::TryFrom;
 
 use crate::spell::WordId;
+use crate::word_metadata_orthography::OrthFlags;
 use crate::{Document, TokenKind, TokenStringExt};
 
+/// This represents a "lexeme" or "headword" which is case-folded but affix-expanded.
```
elijah-potter, line 15: Good for clarification. Thank you

harper-core/src/token_kind.rs
```diff
@@ -115,7 +115,14 @@ impl TokenKind {
 
         // Generic word methods
         is_swear,
-        is_likely_homograph
+        is_likely_homograph,
+
+        // Orthography methods
```
elijah-potter, line 120: Good.

elijah-potter: In the future, I think I want to make it possible to customize the doc-comments for each of these macro-generated methods.

harper-core/src/word_metadata.rs
hippietrail: Especially since these files are right in the project root directory I didn't want to name it just `orthography.rs` as that has a broad meaning of 'correct writing' which could get confusing. It's only about the orthography aspect of the word metadata (which is really lexeme/headword metadata).

Issue #1226, opened 2025-05-04
Use of `Span` in `nominal_phrase.rs` does not match documentation


hippietrail:
While looking for a lint in `nominal_phrase.rs` that failed when changing some of the metadata fields, I found the use of `Span` very confusing.

As far as I knew from previous work on Harper, `Span`s were runs of text, and this is how they are documented in `span.rs`
```rs
/// A window in a [`char`] sequence.
///
/// Although specific to `harper.js`, [this page may clear up any questions you have](https://writewithharper.com/docs/harperjs/spans).
#[derive(Debug, Clone, Copy, Serialize, Deserialize, Default, PartialEq, Eq)]
pub struct Span {
    pub start: usize,
    pub end: usize,
}
```

So that's how I thought it was being used in `nominal_phrase.rs` but printing out the spans resulted in nonsensical words and parts of words. After some stumbling I realized we were dealing with runs of `Token`s but using `Span` to do so.

(The LLM I was using had the same confusion and disagreed with my theory until I showed it the output!)

So either:
a) `nominal_phrase.rs` is misusing `Span` or
b) The documentation of `Span` is wrong

But it must be a) because `Span` has these functions:
```rs
    /// Get the associated content. Will return [`None`] if any aspect is
    /// invalid.
    pub fn try_get_content<'a>(&self, source: &'a [char]) -> Option<&'a [char]> {
```
```rs
    /// Get the associated content. Will panic if any aspect is invalid.
    pub fn get_content<'a>(&self, source: &'a [char]) -> &'a [char] {
```
Which return nonsense when used in `nominal_phrase.rs`

Would the right fix be to make a new cousin of `Span` such as `TokenSpan`? Maybe it ought to use `&[Token]` directly without a special type?

This is kinda at the edge of both my rust knowledge and my knowledge of the Harper codebase.

hippietrail:
This was fixed in #1595

PR #1595, opened 2025-07-20
refactor: typed spans

86xsk:
# Issues 
(Partially) related to #1226 

# Description

Adds a generic type parameter to `Span` (via `PhantomData`).

Currently, `Span` appears to be used for 3 distinct types throughout the codebase (`char`, `Token`, and `u8`). With these differences not being clearly communicated via type signature (and oftentimes not even documentation), it can be very easy to confuse different uses of `Span`. These changes should hopefully make things clearer, make it more difficult to accidentally use one type of span as another, and reduce the potential burden of having to describe and redescribe these nuances in the documentation.

This also makes it possible to implement specific functionality for certain types of `Span`, like the `to_char_span()` function implemented for `Span<token>`.

Surprisingly, these changes only revealed a single case of `Span` type confusion in the codebase: https://github.com/Automattic/harper/blob/9bbe9b7051d03beb91b0c626174915b6314ffcb0/harper-core/src/patterns/nominal_phrase.rs#L107-L107

(`span` is a token span, but it's used to index/subslice `doc.get_source()`, which returns a char slice. This has been fixed in this PR.)

# How Has This Been Tested?

- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/span.rs
+++ b/harper-core/src/span.rs
@@ -208,7 +208,14 @@ impl<T> Copy for Span<T> {}
 
 #[cfg(test)]
 mod tests {
-    type UntypedSpan = super::Span<()>;
+    use crate::{
+        Document,
+        expr::{ExprExt, SequenceExpr},
+    };
+
+    use super::Span;
+
+    type UntypedSpan = Span<()>;
 
     #[test]
     fn overlaps() {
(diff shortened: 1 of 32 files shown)


harper-tree-sitter/src/lib.rs
```diff
@@ -62,7 +62,7 @@ impl TreeSitterMasker {
     /// Visits the children of a TreeSitter node, searching for comments.
     ///
     /// Returns the BYTE spans of the comment position.
```
elijah-potter, line 64: The last line of this comment is now redundant.

harper-core/src/span.rs
```diff
@@ -1,54 +1,92 @@
...
+impl<T> Span<T> {
+    /// An empty [`Span`].
+    pub const EMPTY: Self = Self {
+        start: 0,
+        end: 0,
+        span_type: PhantomData,
+    };
+
+    /// Creates a new [`Span`] with the provided start and end indices.
+    ///
```
elijah-potter, line 39: I much appreciate all the additional documentation you've brought to this module. Thank you.

elijah-potter (approved):
Looks good! I have just the one potential edit, but it's not worth the effort to change before merging. This has been something that's been itching at the back of my head for a while now, so I'm thrilled to see it get into the code. Thank you so much.

PR #1605, opened 2025-07-21
feat(core): create first draft of rule for missing articles

elijah-potter:
# Issues 

# Description

Draft

# Demo

# How Has This Been Tested?

# Checklist


harper-core/src/linting/missing_article.rs
```diff
@@ -0,0 +1,209 @@
...
+    use crate::linting::tests::{assert_no_lints, assert_top3_suggestion_result};
+
+    use super::MissingArticle;
+
+    #[test]
+    fn fixes_good_day() {
+        assert_top3_suggestion_result(
+            "I had good day.",
+            MissingArticle::default(),
+            "I had good day.",
```
hippietrail, line 108: Ironically, Is the article missing?

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -274,6 +274,17 @@ Message: |
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     138 | finding it very nice, (it had, in fact, a sort of mixed flavour of cherry-tart,
+         |                                                         ^~~~~~~ Consider adding an article before this noun.
```
hippietrail, line 280: You don't want to add a/an before mass nouns. "A stuff", "an information", "a knowledge", etc.
But you must also keep in mind that most mass nouns also have countable senses: "a beer", "an area", "a flavour", etc. These can now all be fully queried.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -594,6 +617,16 @@ Message: |
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     317 | “Perhaps it doesn’t understand English,” thought Alice; “I daresay it’s a French
+         |                                ^~~~~~~ Consider adding an article before this noun.
```
hippietrail, line 623: Oh I have probably not yet annotated all of the names of languages as mass nouns. Many are both mass and countable like "German" but not "English". (Linguists do talk about Englishes but non-linguists don't)

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -655,6 +688,17 @@ Message: |
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     347 | and beg for its dinner, and all sorts of things—I can’t remember half of
+         |                                                                  ^~~~ Consider adding an article before this noun.
```
hippietrail, line 694: This is different. "Half" is only a countable noun. I don't know how to classify this off the top of my head. I'm thinking "half of" as a phrase is a quantifier... Wow discussions online seem to say this usage of "half" is as a mass noun after all.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -830,6 +874,16 @@ Suggest:
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     412 | advisable to go with Edgar Atheling to meet William and offer him the crown.
+         |                                             ^~~~~~~ Consider adding an article before this noun.
```
hippietrail, line 880: This definitely is different. Proper nouns are already "definite" and so don't take articles unless it's part of the name like in "The Gambia" or surnames when referring to the whole family like in "The Simpsons".
And of course in a humorous style: "I met two Toms and a William".

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -887,6 +941,17 @@ Suggest:
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     544 | hurried off at once: one old Magpie began wrapping itself up very carefully,
+         |                                           ^~~~~~~~ Consider adding an article before this noun.
```
hippietrail, line 947: Here we've run into the fact that verbs in their -ing form can behave as both "gerunds" and "present participles". Gerunds behave as nouns and present participles behave as adjectives, but in either case they still also act like verbs in that they can take objects when they're transitive verbs.

I think in this case it's a gerund because it's acting as a noun that's the object of the verb "began". But I'm not 100% sure and would probably consult two LLMs (-:

My hunch is that the best way forward would be to specifically handle them. I better get to work on making sure all the verb form properties are fully propagated. I think `.is_verb_progressive_form` should be fully propagated though since -ing forms are fully regular.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -987,6 +1062,16 @@ Message: |
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     624 | When I used to read fairy-tales, I fancied that kind of thing never happened,
+         |                     ^~~~~ Consider adding an article before this noun.
```
hippietrail, line 1068: We really need more support for hyphenated terms and further down the line we're going to need support for compound nouns that use spaces too.

Maybe you can already do this using `NominalPhrase` *but* I've been thinking for months how best to handle the full nominal phrase including the optional determiner and optional adjectives vs the core nominal phrase including just a run of nouns. Not sure if this always means a compound noun or not.

For runs of nouns as a unit whatever the right term is, normally the last one in the group is the "head" and bears the inflection: "man hole covers" and not "men hole cover" but there's a set of special ones like "passers by" and "sons in law". They probably can include possessives in the middle though. And what happens with gerunds in the middle vs present-participles acting as adjectives at the beginning might need some thought and experimentation.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -1022,6 +1118,17 @@ Suggest:
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     679 | another snatch in the air. This time there were two little shrieks, and more
+     680 | sounds of broken glass. “What a number of cucumber-frames there must be!”
+         |                  ^~~~~ Consider adding an article before this noun.
```
hippietrail, line 1125: This reveals two problems. We've already covered that "glass" is both a mass noun and countable noun. Here it's a mass noun.

But even if it were only countable and needed correction, the article would actually belong before the adjective that precedes the noun: "of a broken glass", not "of broken a glass".

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -1159,6 +1276,17 @@ Message: |
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+     764 | over; and the moment she appeared on the other side, the puppy made another rush
+     765 | at the stick, and tumbled head over heels in its hurry to get hold of it; then
+         |                           ^~~~ Consider adding an article before this noun.
```
hippietrail, line 1283: In this case "head over heels" is an idiom so the rules are different. Idioms act as single words. Do not confuse them with "idiomatic" which means "as used naturally by native speakers" which covers other things that are not fixed phrases.

I've been thinking of addressing this problem like this:
- Mark idioms in the dictionary with a new annotation.
- When building the `FstDictionary` by parsing the `dictionary.dict` and applying the affix/annotation rules, add a property to each word that is either the first or last word of an idiom.

This way, in linters we can check that the following word is not the start of an idiom or the previous word is not the end of an idiom.

It surely needs further thought.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -1491,6 +1630,17 @@ Message: |
 
 
 
+Lint:    Miscellaneous (0 priority)
+Message: |
+    1226 | up into a sort of knot, and then keep tight hold of its right ear and left foot,
+         |                                                                            ^~~~ Consider adding an article before this noun.
```
hippietrail, line 1636: Here "foot" and even "left foot" are part of a list making up a noun phrase that as a group has the determiner "its". This gets deeper and deeper into syntax parsing!

For now you can see the adjective "left" makes it not require an article.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
hippietrail: To follow up on my comment about verb properties, all the properties of regular verbs can be queried already, they're generated by the annotations that derive the inflections. But for irregular verbs we need 4 or 5 more annotations for manually entered irregular inflections. I might do this today.

PR #1630, opened 2025-07-25
chore(dict): Add Forgejo noun

JadedBlueEyes:
(This is just the result of `just addnoun`, probably needs moving to the middle of the dict)

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -52610,6 +52610,7 @@ Famicom/NOgS        # game console
 FireWire/gS         # trademark
 Firestore/g
 Foddy/Sg
+Forgejo/gO          # programming forge / git host
 Frankl/Sg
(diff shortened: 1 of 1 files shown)


hippietrail (approved):
It should be collated between these around line 52,612:
```
Foddy/Sg
Frankl/Sg
```
That section is marked:
```
# This section contains mostly programming and technical terms
# Some words may be very domain-specific, proprietary, or trademarked
```
These niche things should probably have comments though. I have no idea what the above two are but the new one is some kind of container-related thing. Maybe I should update the comment section to advise including a comment. Maybe we could also do so for `just addnoun`?

JadedBlueEyes:
Forgejo is a self-hosted code forge like GitLab - https://forgejo.org/ :) 

I'll update this when I get back to my computer

JadedBlueEyes:
Updated now!

Issue #1631, opened 2025-07-25
Feature: try to detect "conform"/"confirm" related typos


patrakov:
## Description

The words "conform" and "confirm" are both valid, yet the letters "o"/"i" that differentiate them are on adjacent keys, thus creating a potential for typos.

I have just made this typo:

> I conform that I can connect.

Grammarly caught it. I would like Harper to catch the intended meaning, too.

hippietrail:
> ## Description
> The words "conform" and "confirm" are both valid, yet the letters "o"/"i" that differentiate them are on adjacent keys, thus creating a potential for typos.
> 
> I have just made this typo:
> 
> > I conform that I can connect.
> 
> Grammarly caught it. I would like Harper to catch the intended meaning, too.

That's pretty impressive. I would've expected it needs semantic context and was going to ask if Grammarly uses AI. But then I noticed Chrome flags it too. 

<img width="288" height="66" alt="Image" src="https://github.com/user-attachments/assets/644abc65-ff37-4026-9ce4-365446f2411a" />

So maybe there are some things to look out for re prepositions and complements etc.

hippietrail:
It looks like this was fixed with #1632

PR #1632, opened 2025-07-25
feat: conform that→confirm that

hippietrail:
# Issues 

#1631

# Description

Tackles the most obvious/easy case of typing "conform" when you mean to type "confirm".

So far it only detects the typo when followed by "that". "That" is rare after most forms of "conform", "conformed", and "conforms"; but is less rare after "conforming":
> is there any example of a case that isn't fully conforming that is supported today?

I false positives or false negatives are found in the wild please post issues.

# How Has This Been Tested?

I made unit tests from sentences found on GitHub.

There's also a unit test for the known false positive though it's marked `#ignore` until we can detect it.

# Checklist


patrakov:
Maybe it makes sense to add "this" as an alternative word that can follow "confirm" but not "conform".

hippietrail:
> Maybe it makes sense to add "this" as an alternative word that can follow "confirm" but not "conform".

Sounds reasonable. I'll do some research... It turns out the vast majority of conform/conformed/conforms/conforming + it on GitHub are intending the meaning "make it conform" and very few are typos.

Either not typos or not enough info to tell:
- simply conform it to Codable protocol and create a request
- It is just a matter of creating a type, conforming it to my FirebaseFetchable protocol, ...
- This extension takes input Object Placement data from a "Spot" and conforms it to a surface using a directable raycast.
- @ViewState creates a ViewState structure and conforms it to the ActionSendable protocol.
- improved readability of code and conformed it to pep8.
- Because Void is not decodable I've created this empty struct Nothing and conformed it to decodable.

Actual typos:
- As you said, I conformed it worked well in PDF.

Other contexts where it would be a false positive:
- Even when features become obsolete or non conforming it still makes sense to include the rules for their rendering.

That said, Chrome does flag the one actual typo I could find, so maybe a verb after "it" might be enough: 
image

PR #1639, opened 2025-07-26
feat: that typo for than

### LINKED ISSUE #506: feat: Flag using "that" instead of "than".
This is probably a typo due to muscle memory of typing the more common word. But it's possible non-native speakers could pick it up from reading English full of typos.

- It's bigger **that** I expected it to be. → It's bigger **than** I expected it to be.

The heuristic is that it comes after a comparative like "more" or "-er" forms of adjectives.

Examples harvested from the wild:
- so no different **that** right now → so no different **than** right now

Potential false positives:
- Explain better that to attach ticket from alert you need "ticketing provider"
- but is there an something different that each is meant to be used?
- this can be achieved through several ways, and likely some more that I haven't thought of
- My point was more that if the compact index API returns 404
- so it's probably more that Croatian had an easier test
- but I have 3 or 4 more that I did from home

hippietrail:
# Issues 

Fixes #506

# Description

Fixes cases where "that" is a typo for "than".

# How Has This Been Tested?

Unit tests made from sentences found on GitHub.

# Checklist


harper-core/src/linting/lint_group.rs
```diff
@@ -97,6 +97,7 @@ use super::somewhat_something::SomewhatSomething;
 use super::spaces::Spaces;
 use super::spell_check::SpellCheck;
 use super::spelled_numbers::SpelledNumbers;
+use super::that_than::ThatThan;
```
ccoVeille, line 101: Can't it be grouped with ThenThan rule ?

The unit test might be different, but the way to detect might be the same ?

I'm asking because I'm surprised that working on ThatThan doesn't bring ideas, exception, fixes for ThenThat.

I mean all the true positive examples you are adding to the test of ThanThat could be added to ThenThan by replacing that by then, and see how it behaves.

harper-core/src/linting/that_than.rs
```diff
@@ -0,0 +1,240 @@
...
+                    return false;
+                }
+                let adj_lower = tok.span.get_content_string(src).to_lowercase();
+                let is_better_or_later = adj_lower == "better" || adj_lower == "later";
+                !is_better_or_later
+            })
+            .t_ws()
+            .t_aco("that")
+            .t_ws()
+            // .then_any_word();
```
ccoVeille: This was left behind when you had to use an exception for "way"

harper-core/src/linting/that_than.rs
```diff
@@ -0,0 +1,240 @@
...
+            .then(|tok: &Token, src: &[char]| {
+                if !tok.kind.is_word() {
+                    return false;
+                }
+                let adj_lower = tok.span.get_content_string(src).to_lowercase();
+                let is_way = adj_lower == "way";
+                !is_way
+            });
```
ccoVeille: Don't you have a need for a method like this that would be

.any_words_except("way", "foo", "bar)

Here I'm using the variadic common pattern I would use in Go,but I'm unsure it exists in Rust.

harper-core/src/linting/that_than.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    #[test]
+    #[ignore = "not handled because 'better' results in false positives"]
+    fn fix_better_that() {
+        assert_suggestion_result(
+            "Examples of how different cards perform far better that others.",
+            ThatThan::default(),
+            "Examples of how different cards perform far better than others.",
+        );
+    }
```
ccoVeille, line 117: See https://github.com/Automattic/harper/pull/1639#discussion_r2233747831

I feel like detecting this in ThenThan could be interesting

            "Examples of how different cards perform far better then others."

harper-core/src/linting/lint_group.rs
hippietrail: > Can't it be grouped with ThenThan rule ?
> 
> The unit test might be different, but the way to detect might be the same ?
> 
> I'm asking because I'm surprised that working on ThatThan doesn't bring ideas, exception, fixes for ThenThat.
> 
> I mean all the true positive examples you are adding to the test of ThanThat could be added to ThenThan by replacing that by then, and see how it behaves.

Nah it's all different because "than" and "then" are different parts of speech which makes the heuristics different, and people switch "than" and "then" both ways but while people use "that" for "than" they don't do the opposite.

harper-core/src/linting/that_than.rs
hippietrail: > Don't you have a need for a method like this that would be
> 
> .any_words_except("way", "foo", "bar)
> 
> Here I'm using the variadic common pattern I would use in Go,but I'm unsure it exists in Rust.

I actually made one and in fact one is kinda made by a macro which makes lots of methods, but it was subtly different and I kept getting unexpected results, so I backed that all out and did it the simple straightforward way and then it worked first time.

Most of the `.then` methods are derived from `.is` methods that can come up from either `TokenKind` or `WordMetadata` or both, but you can also add custom ones at the various levels.

harper-core/src/linting/lint_group.rs
ccoVeille: > > Can't it be grouped with ThenThan rule ?
> > 
> > The unit test might be different, but the way to detect might be the same ?

OK 👍

> > I'm asking because I'm surprised that working on ThatThan doesn't bring ideas, exception, fixes for ThenThat.
> > 
> > I mean all the true positive examples you are adding to the test of ThanThat could be added to ThenThan by replacing that by then, and see how it behaves.
> 
> Nah it's all different because "than" and "then" are different parts of speech which makes the heuristics different, and people switch "than" and "then" both ways but while people use "that" for "than" they don't do the opposite.

I think you know that but people don't 😜😂🤣

https://github.com/search?q=%2F%28better%7Clonger%7Cgreater%29+then+%28others%7Cme%7Chim%7Cher%7Cmine%7Cyours%29%2F&type=code

ccoVeille: 🤦‍♂️ I'm reporting this as if ThenThan was not there ...

hippietrail: > > > Can't it be grouped with ThenThan rule ?
> > > The unit test might be different, but the way to detect might be the same ?
> 
> OK 👍
> 
> > > I'm asking because I'm surprised that working on ThatThan doesn't bring ideas, exception, fixes for ThenThat.
> > > I mean all the true positive examples you are adding to the test of ThanThat could be added to ThenThan by replacing that by then, and see how it behaves.
> > 
> > 
> > Nah it's all different because "than" and "then" are different parts of speech which makes the heuristics different, and people switch "than" and "then" both ways but while people use "that" for "than" they don't do the opposite.
> 
> I think you know that but people don't 😜😂🤣
> 
> https://github.com/search?q=%2F%28better%7Clonger%7Cgreater%29+then+%28others%7Cme%7Chim%7Cher%7Cmine%7Cyours%29%2F&type=code

Those all look like "then" instead of "than" to me. Maybe I'm missing something?

ccoVeille: They are. As I said, I reported them as if ThenThan rule was missing.

There are fine.

harper-core/src/linting/that_than.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+impl Default for ThatThan {
+    fn default() -> Self {
+        let adjective_er_that_nextword = SequenceExpr::default()
+            .then(|tok: &Token, src: &[char]| {
+                if !tok.kind.is_comparative_adjective() {
+                    return false;
+                }
+                let adj_lower = tok.span.get_content_string(src).to_lowercase();
```
elijah-potter: This is pretty expensive because it requires an allocation. Try using `eq_ignore_ascii_case_str`.

elijah-potter (dismissed):

PR #1662, opened 2025-07-29
chore: noun attrs, words from trending GitHub readmes

hippietrail:
# Issues 
N/A

# Description

New words from trending GitHub project readmes.
Marking nouns for plurality and countability.

# Checklist


elijah-potter (approved):

86xsk:
Might be worth it to throw in:
`quaternion/N0g`

(Not sure if I should open a PR just to add a single word.)

hippietrail:
> Might be worth it to throw in: `quaternion/N0g`
> 
> (Not sure if I should open a PR just to add a single word.)

I'm surprised that's not in there. But I've been surprised before. (-: The `0` isn't necessary for regular countable nouns by the way since the original dictionary format only had a plural flag and assumed nouns without it were singular, which didn't work for words like `aircraft`, `biceps`, `fish`, `sheep`, and probably a tiny handful of others. It was easier (cleaner?) to make the fallback to regular singular work than to add `0` to 40,000 entries. But since it has a regular plural it does need `/S` so I'd annotate it `NgS` but I think there's also tons of `NSg`.

elijah-potter (approved):

PR #1668, opened 2025-07-30
feat: quality-of-life methods for `SequenceExpr`

hippietrail:
# Issues 
N/A

# Description

Add handy methods to `SequenceExpr`:
- `then_word_except`
- `then_kind_except`
- `then_kind_both`
- `then_kind_either`
- `then_kind_any`
- `then_kind_is_but_is_not`
- `then_kind_is_but_is_not_except`

They do combinations of checking one or more `TokenKind` predicates are true or false plus those with `except` in their name also take an array of words that are exceptions that will prevent an otherwise successful `then_*` from being true.

- Refactor various linters to use the new methods.
- Refactor other `TokenKind` predicate use in linters.
-  Move `SpanVecExt` linting assertion helper to be accessible to all linters.

# How Has This Been Tested?

I added a couple of unit tests for some of the new methods, but not all of them.

All existing tests pass but some of the linters that were modified seemed to lack any tests so I'm a bit concerned there's a small possibility of changed logic that isn't tested for regressions.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -36777,7 +36777,7 @@ oven/~NVgS
 ovenbird/NSg
 ovenproof/J
 ovenware/Nmg
-over/~JNPVgYS
+over/~JNPgYS
 overabundance/Ng
 overabundant/J
 overachieve/VZGd>S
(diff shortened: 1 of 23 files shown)


elijah-potter (dismissed):
I like these changes. The examples you've already moved to this new syntax are much clearer now. Thanks!

elijah-potter (approved):

PR #1671, opened 2025-07-31
feat: flag mass nouns used as plurals

### LINKED ISSUE #955: Should flag mass nouns (non-count nouns) used in the plural.
## Description

English has a concept on nouns called "countability". English speakers that haven't studied grammar or linguistics or taught English as a foreign language are often unaware as it is intuitive and comes naturally.

Non-native speakers often have no such intuition and treat all English nouns as being able to be made into plurals. Often their language has a cognate which does have a plural.

Even native speakers who are aware of it often overanalyse what the rules are and assume you can always make a mass noun plural for senses meaning "portions of" or "different kinds of" because these are true for many concrete nouns.

Many nouns have countable senses and uncountable senses.

Examples of nouns which can never be plural:
❌ advices
❌ evidences
❌ furnitures
❌ informations
❌ infos
❌ knowledges

Examples of mass nouns which can be plural for "portions":
✅ Give me two beers please.
✅ Should we get two rices or share one large rice?

Examples of mass nouns which can be plural for "different kinds":
✅ I like mild cheeses but not stinky cheeses.
✅ Alien intelligences may be so different that you won't even recognize them.

Examples of nouns with a mass noun sense and a plural sense:
✅ The hurricane caused a lot of damage.
❌ The hurricane caused a lot of damages.
✅ The judge told the company that they had to pay damages!

## Resources

If there are any resources online we can reference that described grammatical rules, let us know.

## Examples

- MuJoCo simulation for common **furnitures**
- A GitHub Action aiming to help in the process of the fellows **evidences** referenda.
- The existing query evidence doesn't return handled **evidences**. The issue may be because we're not storing the evidence here
- An advice generator application that uses the Advice Slip API to generate **advices**
- We propose adding support for selecting and using multiple **knowledges** within a single session.

## Potential False Positives

- Many. It's safest to only flag nouns we are sure have no countable sense.
- Some nouns will have a verb counterpart that in the third-person singular present looks identical to a plural noun: Water **damages** electronics.

hippietrail:
# Issues 
Fixes #955 

# Description

Flags nouns used in the plural that are mass nouns that don't have plurals: "furnitures", "advices", "softwares", "knowledges", "source codes", etc.

At present it doesn't flag such words if they also happen to be legit verbs, only ones which are not in the dictionary at all.

It doesn't yet offer suggestions but I'll add that.

It doesn't try to fix agreeing words such as "these" before or "are" after.

# Demo
Screenshot 2025-07-31 at 11 41 27 am
Screenshot 2025-07-31 at 11 42 20 am

# How Has This Been Tested?

I included a few unit test to cover the various kinds of regular plurals.

# Checklist


harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -161,18 +161,6 @@ Message: |
 
 
 
-Lint:    Spelling (63 priority)
-Message: |
```
elijah-potter, line 165: Nice.

PR #1677, opened 2025-08-01
fix(core): typo in phrase correction message

jj15asmr:
# Description
I recently came across Harper, and while source-diving, found a small typo in the message for the `AfterAWhile` phrase correction. I'm actually not familiar with Rust, but it's only some text, so I hope I did everything correctly 😅.

# How Has This Been Tested?
It hasn't — but it doesn't seem like this message is asserted against anywhere.

# Checklist


hippietrail:
Excellent! I must've made a search & replace error on that one.

elijah-potter (approved):

PR #1681, opened 2025-08-02
fix: don't flag repeated words if either is hyphenated

hippietrail:
# Issues 
N/A

# Description

I noticed that the repeated word linter was flagging words where one was actually part of a hyphenated term rather than a true repeated word: 
image
"In" is not a repeat of "opt-in".

So I modified the linter to check for a hyphen immediately before the first word or immediately after the second.

# How Has This Been Tested?

I added a unit test based on the sentence that alerted me to this problem plus a second unit test where there are hyphenated words on either side of the normal word.

# Checklist

### DIFF
--- a/harper-core/src/linting/repeated_words.rs
+++ b/harper-core/src/linting/repeated_words.rs
@@ -42,6 +42,15 @@ impl Linter for RepeatedWords {
                 let word_a = document.get_span_content(&tok_a.span);
                 let word_b = document.get_span_content(&tok_b.span);
 
+                let prev_tok = document.get_token_offset(idx_a, -1);
+                let next_tok = document.get_token_offset(*idx_b, 1);
+
+                if prev_tok.is_some_and(|t| t.kind.is_hyphen())
+                    || next_tok.is_some_and(|t| t.kind.is_hyphen())
+                {
+                    continue;
+                }
+
                 if (tok_a.kind.is_preposition()
                     || tok_a.kind.is_conjunction()
                     || !tok_a.kind.is_likely_homograph()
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/repeated_words.rs
```diff
@@ -42,6 +42,15 @@ impl Linter for RepeatedWords {
...
 
+                let prev_tok = document.get_token_offset(idx_a, -1);
+                let next_tok = document.get_token_offset(*idx_b, 1);
+
+                if prev_tok.is_some_and(|t| t.kind.is_hyphen())
+                    || next_tok.is_some_and(|t| t.kind.is_hyphen())
+                {
+                    continue;
+                }
+
```
86xsk, line 53: I would consider moving this check above the preceding `get_span_content()`s to avoid the calls to `get_span_content()` when not necessary.

Also, it might make the code more concise if the other `continue` was integrated into this `if` block, perhaps like so:

```rust
if prev_tok.is_some_and(|t| t.kind.is_hyphen())
    || next_tok.is_some_and(|t| t.kind.is_hyphen())
    || chunk[idx_a + 1..*idx_b]
        .iter()
        .any(|t| !t.kind.is_whitespace())
{
    continue;
}
```

86xsk (approved):
Just a couple nitpicks, but other than that LGTM. I've encountered this bug from time to time, and it's nice to finally have it fixed.

harper-core/src/linting/repeated_words.rs
hippietrail: Oh well spotted! I'm usually careful to put this kind of logic before the `get_content` but I've missed it this time.

I usually find myself integrating too much logic into a block but in this case I was thinking it would be a mistake making it harder to read and probably producing the exact same machine code anyway.

hippietrail: By the way, does this negative homograph test look like a mistake or just counterintuitive?
```rs
                if (tok_a.kind.is_preposition()
                    || tok_a.kind.is_conjunction()
                    || !tok_a.kind.is_likely_homograph()
                    || self.is_special_case(word_a)
                    || tok_a.kind.is_adverb()
                    || tok_a.kind.is_determiner())
                    && word_a.to_lower() == word_b.to_lower()
                {
```

hippietrail: Well that must've changed something. The tests in the `.rs` file are unaffected but it's broken something in the integration tests. But I find those much harder to debug...

elijah-potter: I think this is fine as-is. Calls to `get_span_content()` themselves are essentially free (they get optimized out) since they just transform a Span into a slice. Subsequent memory access _does_ have a significant cost.

Issue #1682, opened 2025-08-02
False positive in "them is" agreement


hippietrail:
<img width="746" height="83" alt="Image" src="https://github.com/user-attachments/assets/82db560d-59c3-46b3-a467-b5dc61963156" />

In this case the subject is "understanding them" and not "them".

hippietrail:
This is not happening in the current Harper release.

PR #1692, opened 2025-08-05
fix(core): false-positive from #1682

elijah-potter:
# Issues 

#1682

# Description

Fixes a false-positive in #1682 by using the new NP chunker.

# How Has This Been Tested?

Additional unit test.

# Checklist

### DIFF
--- a/harper-core/src/linting/pronoun_inflection_be.rs
+++ b/harper-core/src/linting/pronoun_inflection_be.rs
@@ -51,6 +51,14 @@ impl PronounInflectionBe {
         map.insert(arent, "isn't");
 
         let is = SequenceExpr::default()
+            .then(|tok: &Token, _: &[char]| {
+                tok.kind
+                    .as_word()
+                    .as_ref()
+                    .and_then(|m| m.as_ref().and_then(|m| m.np_member))
+                    .unwrap_or_default()
+            })
+            .then_whitespace()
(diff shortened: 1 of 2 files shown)


harper-core/tests/text/linters/this and that.snap.yml
```diff
@@ -1,9 +0,0 @@
-Lint:    Agreement (30 priority)
-Message: |
-       2 | As such, disambiguating them is necessary.
-         |                              ^~ Make the verb agree with its subject.
```
hippietrail, line 4: ❤️

harper-core/src/linting/pronoun_inflection_be.rs
```diff
@@ -51,6 +51,14 @@ impl PronounInflectionBe {
         map.insert(arent, "isn't");
 
         let is = SequenceExpr::default()
+            .then(|tok: &Token, _: &[char]| {
+                tok.kind
+                    .as_word()
+                    .as_ref()
+                    .and_then(|m| m.as_ref().and_then(|m| m.np_member))
```
hippietrail, line 58: Ooh both the `.and_then` and the `np_member` are quite novel to me. I'd seen the latter but wasn't sure how to make use of it. The former I don't recall seeing before.

Since you were recently working on the documentation, did you cover these? Maybe they're "advanced"?

Issue #1497, opened 2025-07-02
French Spacing


phikal:
I insert two spaces at the end of sentences, as this makes sentence-wise navigation in Emacs easier, but also as a habit.  Harper objects to this and complains:

> There are 2 spaces where there should be only one.

According to <https://writewithharper.com/docs/rules>, I can disable the "Spaces" linter, but that would miss spacing in other places.  Would it be possible to add a secondary rule or just tolerate french spacing in general?

ccoVeille:
I'm French and I have never heard about French spacing 🤣

But I assume it's like French fries 😅, that are Belgium for us.

Anyway, I agree with the idea of two trailing spaces as it's a thing in Markdown syntax.

https://www.markdownguide.org/basic-syntax/#line-break-best-practices

And yes, I second it should be a separate setting.

phikal:
> I'm French and I have never heard about French spacing 🤣

According to https://en.wikipedia.org/wiki/History_of_sentence_spacing#French_and_English_spacing, this might be the best phrase.  I am referring to this practice:

> Adding two spaces after a period is called French spacing. French spacing was quite common in books before the 19th century. Later it became the norm for typewritten copy.

But otherwise I am glad that you agree.

ccoVeille:
Now, I've read this. I feel like what I suggested about Markdown is totally wrong.

French Spaces is about doing this:

Summer is hot.`<SPACE><SPACE>`Winter is cold.`<SPACE><SPACE>`Fall is rainy.

So this

Summer is hot.  Winter is cold.  Fall is rainy.


While Markdown trailing spaces is about doing this.

Summer is hot. Winter is cold,`<SPACE><SPACE>`
but beautiful. Fall is rainy.

Which is rendered like this

Summer is hot. Winter is cold,  
but beautiful. Fall is rainy.

While it's rendered like this with a single space

Summer is hot. Winter is cold, 
but beautiful. Fall is rainy.

hippietrail:
I had also never heard of French spacing and the description sounded weird but the way @ccoVeille described it in MarkDown I realized I've used this exact formatting feature many times myself.

So it should be part of the Markdown format handling. And testing shows me that it is already. It's flagging double spaces between words but not at the end of lines in Markdown.

So are you asking for this in other formats than Markdown? That doesn't seem right to me.

phikal:
[...]

PR #1693, opened 2025-08-05
feat(core): make French spaces configurable

elijah-potter:
# Issues 

#1497 

# Description

> "I think the best solution would be to relax the Spaces rule to ignore French spaces, then add a new rule that specifically disallows French spacing. We'll have both rules enabled by default to maintain existing behavior. Those who wish to use French spaces can simply disable the latter rule." - Me on #1497

This is exactly what this PR does.

# How Has This Been Tested?

Additional unit tests.

# Checklist


harper-core/src/linting/spaces.rs
```diff
@@ -10,7 +10,13 @@ impl Linter for Spaces {
         let mut output = Vec::new();
 
         for sentence in document.iter_sentences() {
-            for space in sentence.iter_spaces() {
+            for space_idx in sentence.iter_space_indices() {
+                if space_idx == 0 {
+                    continue;
```
ccoVeille, line 15: Please add a comment here explaining this one is ignored but handled in French space rule

PR #1694, opened 2025-08-05
feat(core): create rule to detect duplicate punctuation

elijah-potter:
# Description

Detects erroneous use of extra punctuation. 

# How Has This Been Tested?

Additional unit tests + LLM assisted fuzzing.

# Checklist


harper-core/src/linting/duplicate_punctuation.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    #[test]
+    fn flags_double_slash() {
+        assert_lint_count("This // is a typo.", DuplicatePunctuation, 1);
+    }
+
+    #[test]
+    fn flags_triple_slash() {
+        assert_lint_count("Path error: ///tmp.", DuplicatePunctuation, 1);
+    }
```
ccoVeille: Maybe adding a test about

https://example.com

harper-core/src/linting/duplicate_punctuation.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    #[test]
+    fn flags_triple_question() {
+        assert_lint_count("Why???", DuplicatePunctuation, 1);
```
ccoVeille: This one could be used in a dialog to emphasis the surprise. I'm not saying it's valid in English, but it's common on the internet

harper-core/src/linting/duplicate_punctuation.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    #[test]
+    fn flags_double_question() {
+        assert_lint_count("Really??", DuplicatePunctuation, 1);
+    }
+
+    #[test]
+    fn flags_mixed_qbang_pair() {
+        assert_lint_count("What?!", DuplicatePunctuation, 1);
```
ccoVeille: Maybe this one is an ASCII version of interrobang character.

https://en.wikipedia.org/wiki/Interrobang

I wouldn't report it.

harper-core/src/linting/duplicate_punctuation.rs
hippietrail: This is an example of why some people would want to turn on or off individual lints. Multiple `?` is a formal error but much writing is informal. But in this linter one preference would turn on or off all duplicate punctuation, you wouldn't be able to allow `??` and flag everything else.

harper-core/src/linting/duplicate_punctuation.rs
hippietrail: As with `??` this is common but informal so some will want to flag anything informal and some won't. It's not "duplicate" though. Maybe "punctuation sequence" or "punctuation clusters" would be a more accurate name?

ccoVeille: I feel like detecting this should be in the current rule

- ,,
- ;;
- ::
- //
- .,
- :,
- ;:
- ,.
- ?.
- !.
- and other combination

And these in another 

- ??
- ?!
- !!!

So the rule can be disabled.

Also what about `...` does the rule reports it? Don't you have a rule for using `…` instead ?

ccoVeille: Also, be careful about emoticons

Some will match your rule

- :/
- ?:

There might be more to identify and handle separetly

harper-core/src/linting/punctuation_clusters.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    #[test]
+    fn flags_double_semicolon() {
+        assert_lint_count("He hesitated;; then spoke.", PunctuationClusters, 1);
+    }
+
+    #[test]
+    fn flags_double_colon() {
+        assert_lint_count("Choices:: A or B.", PunctuationClusters, 1);
```
ccoVeille: This could match some language notation for classes. I don't know if it shouldn't be in another rule

harper-core/src/linting/duplicate_punctuation.rs
ccoVeille: What is the status about this?

harper-core/src/linting/punctuation_clusters.rs
hippietrail: Methods and namespaces in various languages.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -405,6 +405,16 @@ Message: |
 
 
 
+Lint:    Formatting (63 priority)
+Message: |
+     204 | > Alice’s Right Foot, Esq., Hearthrug, near the Fender, (with Alice’s love).
+         |                          ^~~ Condense this punctuation cluster to a single mark.
```
hippietrail: These should not be coalesced! The period/dot/full stop is ambiguous in English. Here it's a correct part of an abbreviation. I'm not sure how these are handled. I know they can be in the dictionary but I think there's also a condense function on document that only handles 2-3 cases IIRC.

harper-core/tests/text/linters/Computer science.snap.yml
```diff
@@ -1002,6 +1002,16 @@ Suggest:
 
 
 
+Lint:    Formatting (63 priority)
+Message: |
+     394 | Johnson and Frederick P. Brooks Jr., members of the Machine Organization
+         |                                   ^~~ Condense this punctuation cluster to a single mark.
```
hippietrail: Same problem re abbreviations.

harper-core/tests/text/linters/Difficult sentences.snap.yml
```diff
@@ -16,6 +16,16 @@ Suggest:
 
 
 
+Lint:    Formatting (63 priority)
+Message: |
+      51 | Be back by ten o'clock!.
+         |                       ^~~ Condense this punctuation cluster to a single mark.
```
hippietrail: This probably is catching a true mistake in the source material though. Nice!

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -1410,6 +1410,15 @@ Message: |
 
 
 
+Lint:    Formatting (63 priority)
+Message: |
+    1149 | . . . I was standing beside his bed and he was sitting up between the sheets,
+         | ^~~~~~ Condense this punctuation cluster to a single mark.
```
hippietrail: This is probably a legit old-fashioned orthography for the ellipsis?

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -1410,6 +1410,15 @@ Message: |
...
+Suggest:
+  - Replace with: “.”
```
hippietrail, line 1361: Ah but this is wrong. If it's going to correct it needs to suggest ellipses: `...` or `…`

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -4972,6 +5341,27 @@ Suggest:
...
+
+
+Lint:    Formatting (63 priority)
+Message: |
+    3795 | song of it. . . . High in a white palace the king’s daughter, the golden girl. .
+         |                                                                              ^~~~
+    3796 | . .
+         | ~~~ Condense this punctuation cluster to a single mark.
+Suggest:
+  - Replace with: “.”
```
hippietrail, line 5269: A bit unexpected. Not sure if it should suggest both ellipsis and full stop for two dots, or just ellipsis.

ccoVeille (changes requested):
For now, as a safeguard I'll mark the PR with request changes

harper-core/src/linting/punctuation_clusters.rs
```diff
@@ -0,0 +1,239 @@
+use std::collections::HashSet;
+
+use crate::{
+    Document, Punctuation, Span, TokenKind,
+    linting::{Lint, LintKind, Linter, Suggestion},
+};
+
+/// Flags clusters of punctuation that should be collapsed to a single mark
+/// (e.g. `!!`, `?!?`, `//`, `.,`, `; :`, etc.).
```
hippietrail, line 9: `.,` will probably come up in contexts like a `,` after `etc.` - this has always felt wrong either way to me. We should look into what various style guides have to say.

harper-core/src/linting/punctuation_clusters.rs
```diff
@@ -0,0 +1,239 @@
...
+                let suggestions = uniq
+                    .into_iter()
+                    .map(|c| Suggestion::ReplaceWith(vec![c]))
+                    .collect::<Vec<_>>();
+
+                lints.push(Lint {
+                    span,
+                    lint_kind: LintKind::Formatting,
+                    suggestions,
+                    message: "Condense this punctuation cluster to a single mark.".into(),
```
hippietrail, line 114: Maybe some variation among "choose (just) a/one single one/mark"? "Condense" sounds a bit technical and might confuse?

harper-core/src/linting/punctuation_clusters.rs
```diff
@@ -0,0 +1,239 @@
...
+            }
+        }
+
+        lints
+    }
+
+    fn description(&self) -> &str {
+        "Detects consecutive or mixed punctuation marks that should be reduced \
+         to a single comma, semicolon, colon, slash, question mark, \
+         exclamation mark, period, or ampersand."
```
hippietrail, line 130: Maybe some wording along the lines of "Unusual or redundant sequences/clusters/groupings where a single one would be clearer"?

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
hippietrail: Similar to my new comment on `etc.,` - I'm not confident - maybe there are multiple schools of thought on this?

harper-core/tests/text/linters/The Great Gatsby.snap.yml
hippietrail: This is now on my to-do list to support this type of ellipsis properly.

PR #1700, opened 2025-08-06
feat: condense TL;DR

hippietrail:
# Issues 
N/A

# Description

Previously the common internet abbreviation `TL;DR` (or `tl;dr`) was flagged as one or two spelling errors.

There has been an entry in `dictionary.dict` for at least a couple of weeks, but spellcheck is word-by-word and `;` is not treated as a character that can be part of a word.

This PR condenses the abbreviation into a single word token that will then behave with the spellchecker so the letter cases can be corrected etc.

# How Has This Been Tested?

I added several unit tests for different letter case variants.

# Checklist

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -11,7 +11,7 @@ use crate::patterns::WordSet;
 use crate::punctuation::Punctuation;
 use crate::spell::{Dictionary, FstDictionary};
 use crate::vec_ext::VecExt;
-use crate::{FatStringToken, FatToken, Lrc, Token, TokenKind, TokenStringExt};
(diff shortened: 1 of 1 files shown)


harper-core/src/document.rs
```diff
@@ -994,4 +1047,33 @@ mod tests {
         assert!(doc.tokens[22].kind.is_unlintable());
         assert!(doc.tokens[23].kind.is_close_round());
     }
+
+    #[test]
+    fn condesne_tldr_uppercase() {
```
elijah-potter: Typo. Otherwise it's ready to merge.

hippietrail: Fixed in new commit that also allows plurals (but not possessives/contractions)

PR #1712, opened 2025-08-07
feat(core): create initial ToTwoToo rule

elijah-potter:
# Description

This PR introduces a new rule that detects misuse and confusion between `to` and `too`.
# How Has This Been Tested?

Extensive unit tests + fuzzing.

# Checklist


harper-core/src/linting/to_two_too/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+            "We're going too the conference.",
+            ToTwoToo::default(),
+            "We're going to the conference.",
+        );
+    }
+
+    #[test]
+    fn fixes_too_a() {
+        assert_suggestion_result(
+            "We're going too a concert.",
```
ccoVeille, line 128: I'm curious about

"He goes too far with bets"

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -3340,6 +3340,16 @@ Message: |
 
 
 
+Lint:    Typo (31 priority)
+Message: |
+    2961 | watching the setting sun, and thinking of little Alice and all her wonderful
+    2962 | Adventures, till she too began dreaming after a fashion, and this was her
+         |                      ^~~ Use the infinitive marker `to` here instead of the adverb `too`, which indicates excess degree.
```
hippietrail: You will be able to check the verb form to prevent this if my verb form PR makes it in. `to` can only be before the lemma/infinitive/citation form, `to begin` in this case.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -597,6 +597,24 @@ Suggest:
 
 
 
+Lint:    Typo (31 priority)
+Message: |
+     411 | “Things went from bad to worse,” suggested Miss Baker.
+         |                       ^~ Use `too` (with two `o`’s) when indicating excess or addition.
```
hippietrail: You want to check if the word is an adjective in what's called the 'positive form' only. In this case `bad`. This functionality is already there. But I think there will still be false positives because `to` can be a preposition before noun phrases and noun phrases can start with adjectives in the positive form.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -1701,6 +1719,16 @@ Suggest:
 
 
 
+Lint:    Typo (31 priority)
+Message: |
+    1331 | escort, a persistent undergraduate given to violent innuendo, and obviously
+         |                                          ^~ Use `too` (with two `o`’s) when indicating excess or addition.
```
hippietrail: Ah this is what I just mentioned above: `to` + `adjective` + `noun`.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
hippietrail: > You will be able to check the verb form to prevent this if my verb form PR makes it in.

Now possible!

harper-core/tests/text/linters/Computer science.snap.yml
```diff
@@ -900,6 +900,16 @@ Message: |
 
 
 
+Lint:    Typo (31 priority)
+Message: |
+     313 | synthesis and manipulation of image data. The study is connected to many other
+         |                                                                  ^~ Use `too` (with two `o`’s) when indicating excess or addition.
```
hippietrail: Here the `too` associates with `connect` rather than to `many`. There may be various ways to tackle this. We might want to have a system for which prepositions can "transitivize" intransitive verbs at some point.

hippietrail (commented):
I still don't know my way around GitHub reviews. I've been commenting on this file rather than "reviewing" it...

PR #1713, opened 2025-08-07
fix: canonicalize workspace_root in configuration

### LINKED ISSUE #1707: Setting `userDictPath` does not change the user dictionary location
**Describe the bug**
harper 0.55.0 does not recognize a modified user dictionary location after changing the value of `userDictPath`.

**To Reproduce**
1. Set a custom `userDictPath` in neovim
```lua
vim.lsp.config('harper_ls', {
    settings = {
        ["harper-ls"] = {
            userDictPath = vim.fn.stdpath("config") .. "/spell/en.utf-8.add",
        }
    },
})
vim.lsp.enable('harper_ls')
```
3. Add words to the user dictionary
4. Harper still marks them as mispelled

**Expected behavior**
Changing the value of `userDictPath` change the location of the user dictionary

**Screenshots**

**Platform**
Neovim 0.11.3

**Additional context**
This bug is not present is harper 0.54.0 and previous versions.

The default user dictionary location `$XDG_CONFIG_HOME/harper-ls/dictionary.txt` is still recognize even after `userDictPath` is set.

kinnison:
# Issues 

Fixes #1707

# Description

If the LSP client does not provide any mechanism to determine the root of the workspace, we need to resolve things relative to the current directory, but previously we tried that as `.` which it turns out `.try_resolve_in()` does not like.

This changes the behaviour to always `.canonicalize()` the root of the workspace so that subsequent path resolution starts from an absolute path.

# Demo

N/A ?

# How Has This Been Tested?

The original poster of #1707 confirmed that this fixed their issue.

# Checklist

### DIFF
--- a/harper-cli/src/input.rs
+++ b/harper-cli/src/input.rs
@@ -32,7 +32,7 @@ impl Input {
     /// Gets a human-readable identifier for the input. For example, this can be a filename, or
     /// simply the string `"<input>"`.
     #[must_use]
-    pub(super) fn get_identifier(&self) -> Cow<str> {
+    pub(super) fn get_identifier(&self) -> Cow<'_, str> {
         match self {
             Input::File(file) => file
                 .file_name()
--- a/harper-core/src/char_string.rs
+++ b/harper-core/src/char_string.rs
@@ -9,10 +9,10 @@ pub type CharString = SmallVec<[char; 16]>;
 /// Extensions to character sequences that make them easier to wrangle.
 pub trait CharStringExt {
     /// Convert all characters to lowercase, returning a new owned vector if any changes were made.
-    fn to_lower(&self) -> Cow<[char]>;
+    fn to_lower(&self) -> Cow<'_, [char]>;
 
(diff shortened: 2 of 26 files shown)


kinnison:
The precommit fail appears to be a new clippy lint, do you want me to fix it in this PR too?

kinnison:
@elijah-potter 

I've added two chore commits, one which deals with clippy errors which can be auto-fixed and one which deals with mismatched lifetime elisions.

elijah-potter (approved):
Looks good to me! Thanks for dealing with the Rust `1.89.0` fall out and thank you for fixing this so quickly.

kinnison:
> Looks good to me! Thanks for dealing with the Rust `1.89.0` fall out 

You're welcome - this wasn't that much work and let me practice using Helix's diagnostics jump functionality some more 😀 

> and thank you for fixing this so quickly.

I'm just sorry that I didn't spot the problem before it caused someone to suffer.

Issue #1696, opened 2025-08-05
Add new annotations for verb forms in `dictionary.dict`


hippietrail:
Linters sometimes need to know what form a verb is:
- eat / walk - lemma/plain/basic/dictionary/citation/"present"
- ate / walked - simple past/preterite/"past"
- eaten / walked - past participle/"past"
- eats / walks - 3rd person singular present/"present"
- eating / walking - continuous/progressive/"present"

Most of these except lemma and the distinction between simple past and past participle are already handled for deriving most regular verbs: `/d` creates a "past" form in `-ed` since regular verbs share a single form for simple past and past participle. `/S` creates a third person singular present form in `-s` or `-es`. `/G` creates a progressive form in `-ing`.

But for irregular verbs and also regular verbs with spellings that the affix system doesn't handle, such as when a final consonant needs to be doubled, each form is just added to the dictionary with the generic `/V` marking it as a verb of some kind.

Also each verb in the `WordMetadata` has a "verb form" field which can only be set to one of the above types. But as we already know, all irregular verbs have one form that covers two functions. But for irregular verbs a past form may even coincide with the lemma/present form: I run / I ran / I have run. I read / I read / I have read.

So this means we should change the "verb form" from a plain enum to a bitfield enum as we use for `Dialect` and `OrthFlags`.

I'm proposing these annotation flags, which are currently unused:
- `b` for the lemma or **b**ase form
- `t` for the simple pas**t** or pre**t**eri**t**e form
- `T` for the pas**t** par**t**iciple form
- `H` for the t**h**ird person singular present form
- `6` for the `-ing` progressive/continuous form. (`6` looks a bit like a rotated `g`).

I've already started work on this.

hippietrail:
This is all completed as of #1730

PR #1730, opened 2025-08-11
Implement full verb form annotations

hippietrail:
# Issues 

Implements #1696 

# Description

Verb forms (inaccurately known as "tense") properties fully propagated.

It turns out that the expectation that all past participles and present participles can function as adjectives is not true after all: "cared" for instance. This has affected a couple of current tests.

The best fix is unclear. Perhaps separate annotation flags for verbs where the assumptions are true versus false. Perhaps defaulting to true and needing to create the individual form entries separately so they don't get the adjective annotation. It seems this strongly correlates with whether verbs are transitive or intransitive, which we don't yet track. So maybe we should handle both together.

- "Lemma" for the basic form of the verb (aka headword, dictionary form, citation form)
- Just "past" for regular verbs ("walked", "asked", etc.)
- separate "preterite"/"simple past" and "past participle" for irregular verbs ("ate"/"eaten", "ran"/"run", "did"/"done", etc.)
- A single form of a verb can have multiple functions: "I run" and "I read" use the same verb form as "I have run" and "I have read" = lemma + preterite ("present" + "past")

Also:
- POS tag snapshots now annotate all verb forms.
- Adjective positive degree is now also tracked.
- Improved error messages for `annotations.json` parse failures.
- Simplified `everyday.rs` linter to use these changes.
- New rust lifetime lint annotations.

# How Has This Been Tested?

All tests pass though I disabled between one and three due to the fact that past participles are not always adjectives as previously expected.

Please file issues for these and any other new false positives and false negatives!

I added a couple of unit tests for inflected forms of other POSes as well as the new verb form properties.

# Checklist


harper-core/src/linting/lint_group.rs
```diff
@@ -650,6 +650,24 @@ mod tests {
         group.all_descriptions_html();
     }
 
+    #[test]
+    fn fix_low_hanging_fruit_msg() {
+        assert_suggestion_result(
```
elijah-potter: Should these be `assert_no_lint`?

harper-core/src/spell/mutable_dictionary.rs
```diff
@@ -30,13 +30,13 @@ pub struct MutableDictionary {
...
-        MutableDictionary::from_rune_files(
-            include_str!("../../dictionary.dict"),
-            include_str!("../../annotations.json"),
-        )
-        .expect("Curated dictionary should be valid."),
+    MutableDictionary::from_rune_files(
+        include_str!("../../dictionary.dict"),
+        include_str!("../../annotations.json"),
     )
+    .map_err(|e| format!("Failed to load curated dictionary: {}\nThis is likely due to a problem with the dictionary files.\n\nError details: {}", if cfg!(test) { "(running in test mode)" } else { "" }, e))
```
elijah-potter: This feel awkward. What's going on?

harper-core/src/spell/rune/attribute_list.rs
```diff
@@ -36,10 +36,21 @@ impl AttributeList {
...
+        match human_readable {
+            Ok(parsed) => parsed.into_normal(),
+            Err(e) => {
+                let error_msg = format!(
+                    "Failed to parse JSON: {}\nError occurred at line: {}, column: {}",
+                    e,
+                    e.line(),
+                    e.column()
+                );
+                eprintln!("{error_msg}");
```
elijah-potter: We should not print on this path. If more information must be conveyed, include it in the Error type.

elijah-potter (commented):
I have some questions, but this is otherwise good work, and I'm glad the verb forms are getting some attention.

elijah-potter (commented):
I have some questions, but this is otherwise good work, and I'm glad the verb forms are getting some attention.

harper-core/src/linting/lint_group.rs
hippietrail: Well spotted! Thanks!

harper-core/src/spell/mutable_dictionary.rs
hippietrail: I had trouble wrangling the JSON changes for the attributes while experimenting with what would work best and there was a hard-coded error. This looks like part of what the LLM wrote while improving that. Let me take a look at it... Right I'll change it to include the specific error info but cut out all the wordy stuff...

harper-core/src/spell/rune/attribute_list.rs
hippietrail: I've used the LLM to change this since it's not an area of Rust I've explored...

harper-core/src/spell/rune/error.rs
```diff
@@ -12,8 +14,22 @@ pub enum Error {
...
+        column: usize,
+    },
+    #[error("An error occurred with a condition: {0}")]
     Matcher(#[from] matcher::Error),
 }
+
+impl From<SerdeJsonError> for Error {
+    fn from(e: SerdeJsonError) -> Self {
+        Error::MalformedJSON {
+            message: Box::leak(e.to_string().into_boxed_str()),
```
elijah-potter: I've noticed that LLMs don't seem to care about leaking memory. We do. I'd suggest you replace the `&'static str` with an owned `String`.

hippietrail: > I've noticed that LLMs don't seem to care about leaking memory. We do. I'd suggest you replace the `&'static str` with an owned `String`.

Oh wow I better watch out for that! Thanks.

elijah-potter (approved):
Thanks for iterating on this. I think it's ready to be merged!

Issue #1731, opened 2025-08-11
`AnchorStar` and `AnchorEnd` do not match start and end of **chunks** as expected.


hippietrail:
These are needed to work on **chunks** because **chunks** are the subdivisions of a document over which `PatternLinter`s/`ExprLinter`s run. I don't think **chunk** is a properly defined concept though. Instead, the `AnchorStart` and `AnchorEnd` refer to the concept of a **token stream**, which I think is also not a defined concept:
```
/// A [`Step`] which will match only if the cursor is over the first word-like of a token stream.
/// It will return that token.
```
The importance of the chunk concept is seen in `lint_group.rs`:
```
impl Linter for LintGroup {
    fn lint(&mut self, document: &Document) -> Vec<Lint> {
        let mut results = Vec::new();

        // Normal linters
        for (key, linter) in &mut self.linters {
            if self.config.is_rule_enabled(key) {
                results.extend(linter.lint(document));
            }
        }

        // Pattern linters
        for chunk in document.iter_chunks() {
            let Some(chunk_span) = chunk.span() else {
                continue;
            };

            let chunk_chars = document.get_span_content(&chunk_span);
            let config_hash = self.hasher_builder.hash_one(&self.config);
            let key = (chunk_chars.into(), config_hash);

            let mut chunk_results = if let Some(hit) = self.chunk_expr_cache.get(&key) {
                hit.clone()
            } else {
                let mut pattern_lints = Vec::new();

                for (key, linter) in &mut self.expr_linters {
                    if self.config.is_rule_enabled(key) {
                        pattern_lints.extend(run_on_chunk(linter, chunk, document.get_source()));
                    }
                }

                // Make the spans relative to the chunk start
                for lint in &mut pattern_lints {
                    lint.span.pull_by(chunk_span.start);
                }

                self.chunk_expr_cache.put(key, pattern_lints.clone());
                pattern_lints
            };

            // Bring the spans back into document-space
            for lint in &mut chunk_results {
                lint.span.push_by(chunk_span.start);
            }

            results.append(&mut chunk_results);
        }

        results
    }
```
A **lint group** is an arbitrary bunch of lints, and crucially is used for the set of all lints enabled by the user when linting a document through the LSP or any of the plugins.

You can see that the way a document is linted is in two steps:
1. First run all of the "normal" linters over the document as a whole.
2. Iterate over each **chunk** and run the *pattern linters** (including `Expr`s) over each **chunk**.

[...]

PR #1736, opened 2025-08-11
test: tests for how I would expect `AnchorStart` and `AnchorEnd` to work

hippietrail:
# Issues 
#1731 

# Description

Provides a test for each of `AnchorStart` and `AnchorEnd` for what I expect them to be able to do.

# Checklist


harper-core/src/expr/anchor_start.rs
```diff
@@ -38,4 +40,44 @@ mod tests {
...
+            })
+        }
+
+        fn description(&self) -> &str {
+            "Testing `AnchorStart`."
+        }
+    }
+
+    #[test]
+    fn flags_single_token() {
```
elijah-potter, line 76: The problem here is that the expression is only allowed to view individual chunks at a time, so it can only see `Hello`, `world!`, and `START two three four five.` separately. Because the other two chunks have other punctuation (and your `Expr` doesn't match punctuation, it doesn't match the other two.

I can see how this is confusing. Would it be better if we change this to this?

```
/// A [`Step`] which will match only if the cursor is over the last non-whitespace character in stream.
```

```
/// A [`Step`] which will match only if the cursor is over the last word token in the stream.
```

harper-core/src/expr/anchor_end.rs
```diff
@@ -48,4 +50,46 @@ mod tests {
...
     }
+
+    pub struct End {
+        expr: Box<dyn Expr>,
+    }
+
+    impl Default for End {
+        fn default() -> Self {
+            Self {
+                expr: Box::new(SequenceExpr::default().then_any_word().then(AnchorEnd)), // Only replaces the very first word
```
elijah-potter, line 61: Having the `AnchorEnd` at the end of the expression is the intended usage. This is correct.

harper-core/src/expr/anchor_start.rs
hippietrail: Ah actually a chunk is delimited by `,` so it should be this:
```rs
        assert_suggestion_result(
            "Hello, world! One two three four five.",
            Start::default(),
            "START, START! START two three four five.",
        );
```

hippietrail:
Closing this one to be replace by a draft of a work-in-progress implementation to address the issues that I've stuck on...

PR #1758, opened 2025-08-16
feat: condense R&D to a single word token

hippietrail:
# Issues 

I thought I filed a feature request for this some time ago, but I can't find it. Maybe I mentioned it in Discord or elsewhere?

# Description

Condenses the 3 tokens in `R` `&` `D` into a single word token.
Doesn't care about the case of the letters.
Doesn't allow spaces between the characters.

Before: 
image

# How Has This Been Tested?

I added unit tests for uppercase, lowercase, mixed case, and a negative test to make sure spaces aren't allowed.

# Checklist

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -139,6 +139,7 @@ impl Document {
         self.condense_latin();
         self.condense_filename_extensions();
         self.condense_tldr();
+        self.condense_r_ampersand_d();
         self.match_quotes();
 
         let chunker = burn_chunker();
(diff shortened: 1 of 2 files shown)


ccoVeille:
I can think about 
- Q&A : Questions & Answers
- B&B: Bed & Breakfast
- M&M's : chocolate 
- M&S: Mark & Spencer
- C&A: clothes

But there are many other use cases

https://github.com/search?q=language%3AMarkdown+%2F%28%3F-i%29%5BA-Z%5D%26%5BA-Z%5D%2F&type=code

Maybe the rule could be less specific than focusing on R&D only

ccoVeille (approved):

PR #1761, opened 2025-08-16
Fix 1734

### LINKED ISSUE #1734: Documentation on Dictionaries is unclear and should be clarified
**Describe the bug**

In the [Dictionaries](https://writewithharper.com/docs/integrations/language-server#Dictionaries) section of the documentation, it is unclear that user dictionary is only created when the "Add to global dictionary" quick action is used (thanks to @kinnison for clarifying this on Discord). In the later section on Static Dictionaries, it is stated that the dictionary is bundled into the binary. This leads to the assumption that the creation of dictionary.txt is a deprecated feature, and the user should not expect to find it. 

**Expected behavior**
- 
- The documentation should clarify that the file is created only when the user adds a word to the dictionary
- Change the action from "Add to Global Dictionary" to "Add to User Dictionary" to make it consistent with the expected behavior. 

**Platform**

Documentation

kinnison:
# Issues 

This PR closes #1734 

# Description

In order to deal with the items raised in #1734, we change user-visible strings from 'global dictionary' to 'user dictionary' to match the LSP server config, verbs, and documentation.   We also clarify in the documentation that the user dictionary will only be created when words are added to it.

# Demo

N/A (unless you really want)

# How Has This Been Tested?

I fired up an editor using the harper-ls built from this tree and confirmed that the action text read `Add "fooble" to the user dictionary.`

# Checklist

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -209,7 +209,7 @@ impl Backend {
         );
 
         let mut global_dictionary =
-            global_dictionary.context("Unable to load the global dictionary.")?;
+            global_dictionary.context("Unable to load the user dictionary.")?;
         global_dictionary.add_dictionary(Arc::new(
             file_dictionary.context("Unable to load the file dictionary.")?,
         ));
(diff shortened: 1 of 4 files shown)


kinnison:
@elijah-potter afaict the failed test wasn't my fault but I can't trigger a retry.

PR #1775, opened 2025-08-18
fix(core): don't push Harper version

elijah-potter:
# Issues 

Closes #1774 

# Description

This retains the update notice, but removes the header which contains the Harper version. We never used this data. I honestly don't remember why we added the offending line.

@joethei, let me know if there are any other changes I should make in this PR to be in compliance with the developer policies. Thanks!

Once this PR is merged, I'll immediately run a release to push the fixes out to users.

# Checklist

- [x] I have performed a self-review of my own code
- [ ] I have added tests to cover my changes

Issue #1774, opened 2025-08-18
Obsidian Developer Policies violation


joethei:
Hi @elijah-potter,

We (the Obsidian team) noticed that your plugin has client-side telemetry code posting user data to your own servers from within Obsidian.

Please note that this violates our [Developer policies](https://docs.obsidian.md/Developer+policies)

The offending code appears to be part of your update notice:
<https://github.com/Automattic/harper/blob/c37fa1a437279ed8449f75a80178c55f29e4df80/packages/obsidian-plugin/src/index.ts#L9-L14>

Plugins should not submit user data to third-party telemetry services from within Obsidian, even if the data collection is opt-in.
Please remove the offending code within 30 days and create a new release, otherwise this plugin will be removed from the community plugins directory.

Thanks,
— The Obsidian team

ccoVeille:
Everyone dislike this auto update check 😁

elijah-potter:
Hey @joethei, I understand. I'm sorry, I didn't realize that would be outside the terms. I'll remove that right away and push out an update as soon as possible. Thanks for letting me know!

Issue #1725, opened 2025-08-10
Harper-ls is connecting to 192.0.66.196


tonkoshkurik:
**Describe the bug**
Offline, privacy-first grammar checker is trying to connect to some server. Why?

**To Reproduce**
Install harper to nvim with mason, have a Lulu firewal enabled. 

**Expected behavior**
I expect it to be offline 

**Screenshots**

<img width="662" height="455" alt="Image" src="https://github.com/user-attachments/assets/8fbf4d75-40cd-44b2-bf32-1cfbd8c8bfe6" />

**Platform**
Neovim

mcecode:
Hi, I believe that IP is the IP for <https://writewithharper.com> and that the request is done to retrieve and print debug info in the logs that people can provide when they file issues.

You can start `harper-ls` with the `--skip-version-check` flag to prevent this request from being made.

See #1211 for a discussion about this.

ccoVeille:
Remote connection was removed with

- https://github.com/Automattic/harper/pull/1775

mcecode:
> Remote connection was removed with
> 
>     * [fix(core): don't push Harper version #1775](https://github.com/Automattic/harper/pull/1775)

I think that only removes it for the Obsidian plugin, not `harper-ls`. `harper-ls` still does the version check by default without the `--skip-version-check` flag.

mcecode:
Hi @tonkoshkurik, the version check has been removed in v0.59.0. `harper-ls` will no longer make any remote requests, so you can remove the `--skip-version-check` flag from your config once you upgrade.

PR #1779, opened 2025-08-19
docs(harper-ls): document version check

mcecode:
# Issues 

None in particular

# Description

The version check was introduced in #1204, and since then has created some friction (#1211, #1725, #1774) . It's recently been removed in the Obsidian plugin (#1775), but still remains in `harper-ls`. Personally, I don't mind it since I understand that it doesn't log anything. However, unless it really helps with debugging, it might also be worth considering removing it from `harper-ls` in the future. Especially if it creates further friction with more privacy-focused users. That said, I think documenting it first would be good (I took some direct quotes from #1334's description) so users are aware of it and don't get surprised by it, hence, this PR.

# How Has This Been Tested?

Viewed the change locally

# Checklist

### DIFF
--- a/packages/web/src/routes/docs/integrations/language-server/+page.md
+++ b/packages/web/src/routes/docs/integrations/language-server/+page.md
@@ -304,3 +304,7 @@ These configs are under the `markdown` key:
 
 Want your language added?
 Let us know by [commenting on this issue](https://github.com/Automattic/harper/issues/79).
+
+## Skipping the Version Check
+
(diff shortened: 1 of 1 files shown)


elijah-potter:
Thanks for going through the effort of documenting this. However, I think you're right: It doesn't really help with debugging and it feels counterproductive to our mission. If you want to open a PR that removes the version check, I'll approve it.

mcecode:
Closed in favor of #1782.

PR #1781, opened 2025-08-19
feat: quiet⇔quite

### LINKED ISSUE #1778: Should flag mixing up "quite" and "quiet"
Using the adjective "quiet" instead of the adverb "quite" is very common both among native and non-native English speakers.

- Rendering videos 145 frames, with lightx loras for 2.1 i experience reboots **quiet** often.
- ... has been already reported multiple times and I find it **quiet** dumb that it still exists
- It seems that framework like chakra-ui are **quiet** good in opening dropdowns/modals in such enviroment.

The reverse, using "quite" instead of "quiet" may also occur but isn't as common I feel.

hippietrail:
# Issues 

Closes #1778

# Description

Fixes typos "quiet" when it should be "quite", and vice versa.

Added a new `then_kind_any_except` method to `SequenceExpr`.

# How Has This Been Tested?

Unit tests have been added using sentences from GitHub and one made-up sentence.

There are likely to be false positives, so please report any you see!

# Checklist


harper-core/src/linting/quite_quiet.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl ExprLinter for QuiteQuiet {
+    fn expr(&self) -> &dyn Expr {
+        self.expr.as_ref()
+    }
+
+    fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {
+        let text = toks.span()?.get_content_string(src).to_lowercase();
+
+        if text.ends_with(" quite") {
```
elijah-potter: This is fragile because it doesn't account for other kinds of whitespace (like newlines).

hippietrail: Good point!

harper-core/src/linting/quite_quiet.rs
```diff
@@ -0,0 +1,206 @@
...
+                TokenKind::is_noun,
+                &["here", "up"],
+            );
+
+        let negative_contraction_quiet = SequenceExpr::default()
+            .then(|tok: &Token, src: &[char]| {
+                if !tok.kind.is_verb() || !tok.kind.is_apostrophized() {
+                    return false;
+                }
+                let norm_lower = tok.span.get_content_string(src).to_lowercase();
```
elijah-potter: This is very expensive in the hot path.

hippietrail: > This is very expensive in the hot path.

I think I can fix that ...

elijah-potter (approved):
This looks good to me. Thanks for making those changes!

elijah-potter (approved):

hippietrail:
> This looks good to me. Thanks for making those changes!

Thanks for spotting the oversights!

PR #1794, opened 2025-08-22
feat: would've never/would never have→never would have

### LINKED ISSUE #1130: Flag would've/could've never happened - never would've/could've is correct
## Description

Some people use the illogical "would've never happened" or "could've never happened".
Both the modal verbs and the "never" act on what comes next, so logically they mean "there was a possibility that this never happened".
But the writer usually means "It definitely would not happen under these circumstances".

## Resources

- https://www.reddit.com/r/grammar/comments/16gt3l7/could_never_have_or_could_have_never/
- https://www.reddit.com/r/grammar/comments/evkw71/i_would_have_never_i_would_never_have_or_i_never/

## Examples

 - Having a conversation **would have never** been easier with Ramen!
- This **would have never** come about without the help and encouragement of many people, too numerous to mention here.
- Thanks for the help, think I **would have never** find it out alone.

## Potential False Positives

None spring to mind.

hippietrail:
# Issues 
Fixes #1130 

# Description

Normalizes "would have never" and "would never have" to "never would have".
Also works with "could".
Handles contractions, including without the apostrophe.

In the feature request I was going to normalize to "would never have" but research seems to suggest that "never would have" has historically been more popular over a longer period of time.

But I could change the logic to only flag "would have never" and suggest both "would never have" and "never would have", if that's what people prefer.

(It could be more flexible again if Harper supported options for linters.)

# How Has This Been Tested?

I made unit tests using mostly examples sourced from GitHub.

# Checklist


elijah-potter:
> (It could be more flexible again if Harper supported options for linters.)

It's on my radar. I'm still mulling over what the most elegant solution would look like.

elijah-potter (approved):

PR #1796, opened 2025-08-22
feat(harper-ls): Replace all in document with...

### LINKED ISSUE #998: Add a "fix all" button for repeated errors in a document.
## Description

Sometimes one mistake that a linter picks up is repeated over and over throughout a document. Perhaps throughout a codebase: 

Image

It would be nice to have a button to fix all occurrences of the same error.

## Component

Definitely applies to the VS Code extension, but I'm pretty sure that's due to how things like code actions and such work at a lower level. I haven't really looked into this area of the code.

kinnison:
Find all similar lints and allow applying them all across the document in one code action. This should fix #998.

# Issues 

Fixes #998 

# Description

This adds additional quickfix actions which perform a particular lint suggestion document-wide.  The intent being to resolve #998 in a reasonable fashion.

# Demo

replace-all

# How Has This Been Tested?

I added the locally built `harper-ls` to my `helix` configuration and tested across a variety of markdown documents doing global replacements of the same and differing lengths, including situations where the action was similar but the lint different (eg. replacing a double-space in the middle of a sentence vs. after a full-stop).  All seemed good.

Also, `cargo test --all` didn't complain.

Right now I'm not particularly pleased with *how* this is implemented, so I'm interested in the views of @elijah-potter in particular.  I feel there must be a neater way to do this.

# Checklist

### DIFF
--- a/harper-ls/src/diagnostics.rs
+++ b/harper-ls/src/diagnostics.rs
@@ -25,6 +25,7 @@ pub fn lints_to_diagnostics(
 
 pub fn lint_to_code_actions<'a>(
     lint: &'a Lint,
+    others: &[&'a Lint],
     uri: &'a Uri,
     document: &Document,
     config: &CodeActionConfig,
@@ -48,7 +49,27 @@ pub fn lint_to_code_actions<'a>(
                     ),
                 };
 
-                Some(CodeAction {
+                let all_edits = if others.is_empty() {
+                    None
+                } else {
+                    Some(
+                        Some(TextEdit {
+                            range,
+                            new_text: replace_string.clone(),
+                        })
+                        .into_iter()
+                        .chain(others.iter().map(|other| {
+                            let range = span_to_range(source, other.span);
+                            TextEdit {
+                                range,
+                                new_text: replace_string.clone(),
+                            }
+                        }))
+                        .collect::<Vec<_>>(),
+                    )
+                };
+
+                let this_edit_action = CodeAction {
                     title: suggestion.to_string(),
                     kind: Some(CodeActionKind::QUICKFIX),
                     diagnostics: None,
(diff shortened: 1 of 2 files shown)


elijah-potter:
Here's how I see this.

We don't want to overwhelm the user with options. It should not take a long time for them to find the code action they're looking for. The more reading they have to do, the longer it will take for that to happen. To that end, we want to show a just a few, high-impact code actions at a time.

If we want to stick to surfacing this functionality via a code action, we should show at most one additional action at a time when a user hovers. In other words, if the user hovers over an error related to too many spaces, we should show one code action to fix the specific location, and one to fix all in the document. If they hover over a spelling mistake, we should show the same 3 that we always do, but only one additional one to apply the most likely suggestion for the whole document.

From an architectural standpoint, I'd love to see the logic contained in this PR moved to `harper-core` so we can implement similar functionality for other platforms.

kinnison:
Fair enough - I'll give this some deeper thought, thanks for that review @elijah-potter - Are the suggestions in the lint in order of likelihood of correctness?

elijah-potter:
> Are the suggestions in the lint in order of likelihood of correctness?

If you have access to the full `Lint` struct, you should use the one with the lowest `priority`.

elijah-potter:
I misunderstood. Yes, go with just the first suggestion.

elijah-potter:
@kinnison, do you have any updates for us on this?

kinnison:
> @kinnison, do you have any updates for us on this?

I'm so sorry for the radio silence - I've been in the process of buying a new house so I've not had any time to work on this patch.

I think I need to just work out the neatest way to limit to the best suggestion; but I've not had time to look at the code since late august :(

I'd be comfortable if you just wanted to chuck away my patch and redo-from-start; or if you're happy waiting then I'll see what I can do in a week or two.

elijah-potter:
Absolutely no worries! Buying a house is a huge step in life. Please focus on that.

I think we should keep this PR open so it remains visible to potential contributors. I'm going to change it to a draft, however. When you've had a chance to look at the problem once more, mark it as ready for review. Thanks again for all the work!

kinnison:
I have been moving house and it has taken all my energy so I don't think I'll be able to continue with this effort, sorry.

I'll try and come back to harper once everything has settled, but given current timelines that could be many months yet :(  Sorry again.

elijah-potter:
No worries! Thanks for making the effort. Perhaps someone else can pick up where you left off.

PR #1798, opened 2025-08-23
Added a new togglable "Restore Active Status" setting

iBenMoore:
Added a new togglable "Restore Active Status" setting to set whether to restore the Harper linter active/inactive status between sessions.

# Issues 

# Description
Added a new togglable "Restore Active Status" setting to set whether to restore the Harper linter active/inactive status between sessions.

# Demo

# How Has This Been Tested?
Tested closing and opening Obsidian vault with and without the new "Restore Active Status" enabled, and with and without the current Harper state being enabled.  The new toggle is defaulted to off and the plugin behaves as it did previously where Harper is always enabled when Obsidian is opened.  But if the new toggle is set to on and Harper is in an inactive state when Obsidian is closed, that inactive state will be preserved when the vault is next opened.

# Checklist

- [ x ] I have performed a self-review of my own code


iBenMoore:
Will resubmit with precommit fixes.

PR #1799, opened 2025-08-23
Added a new togglable "Restore Active Status" setting

iBenMoore:
Added a new togglable "Restore Active Status" setting to set whether to restore the Harper linter active/inactive status between sessions.

# Issues 

# Description
Added a new togglable "Restore Active Status" setting to set whether to restore the Harper linter active/inactive status between sessions.

# Demo

# How Has This Been Tested?
Tested closing and opening Obsidian vault with and without the new "Restore Active Status" enabled, and with and without the current Harper state being enabled. The new toggle is defaulted to off and the plugin behaves as it did previously where Harper is always enabled when Obsidian is opened. But if the new toggle is set to on and Harper is in an inactive state when Obsidian is closed, that inactive state will be preserved when the vault is next opened.

# Checklist

- [ x ] I have performed a self-review of my own code

### DIFF
--- a/packages/obsidian-plugin/src/HarperSettingTab.ts
+++ b/packages/obsidian-plugin/src/HarperSettingTab.ts
@@ -87,12 +87,12 @@ export class HarperSettingTab extends PluginSettingTab {
 		});
 
 		new Setting(containerEl)
-			.setName('Restore Active Status')
-			.setDesc('Whether to restore the Harper active/inactive status between sessions.')
+			.setName('Activate Harper')
+			.setDesc('Enable or disable harper with this option.')
 			.addToggle((toggle) =>
-				toggle.setValue(this.settings.lintEnabledRestore).onChange(async (value) => {
-					this.settings.lintEnabledRestore = value;
-					await this.state.initializeFromSettings(this.settings);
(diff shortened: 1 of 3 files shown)


elijah-potter:
Sorry it took me so long to get a look at this. I like your implementation, but I'm curious as to why we wouldn't want it to _always_ restore the active status. Why should it be configurable?

iBenMoore:
I'd be good with it always restoring the previous active status as well, I just threw in the option to be safe in case someone else didn't like the idea. I can take out the option if you'd like and resubmit.

Another idea instead of removing that new option, I could alter it to just show the current active state, giving another place where active state can be toggled on or off.  Some other plugins have an option in the settings page like that.

I'm good either way, let me know what you think.

elijah-potter:
> Another idea instead of removing that new option, I could alter it to just show the current active state, giving another place where active state can be toggled on or off. Some other plugins have an option in the settings page like that.

I really like this idea. It makes the most intuitive sense.

iBenMoore:
Closing this PR, will submit another with the changes discussed.

iBenMoore:
PR #1825 submitted with discussed changes.

PR #1803, opened 2025-08-23
feat: condense I/O and other pairs into a single token

hippietrail:
# Issues 
N/A

# Description

Made a general condense delimited pair function that is now used for "R&D", "Q&A", and "I/O".

Later I went through the Wiktionary categories of English terms with & or / and added all the ones that were familiar to me.

Added `is_slash` to `TokenKind`

# How Has This Been Tested?

I added a unit test using test from one of the GitHub repo readmes that contain an example.

# Checklist


ccoVeille:
Can it support M&M's ?

hippietrail:
> Can it support M&M's ?

Not in this form because it currently works with one punctuation with one letter either side. M&M's would need a special condensing function.

There is support for arbitrary patterns as used in linters to be used in condensing functions, but I haven't got my head around those yet.

ccoVeille:
OK, I created

- https://github.com/Automattic/harper/issues/1807

PR #1820, opened 2025-08-27
Add LuaDoc Support

lukasmwerner:
# Description
Improve the compatibility of parsing for lua files.  

# Demo
This code now properly ignores [annotated](https://luals.github.io/wiki/annotations/) lines and only parses the actual documentation lines. 
```lua
---@class NodeProps
---Spacing between child elements in [VStacks](lua://VStack) or [HStacks](lua://HStack)
---@field spacing? number
---Spacing inside the element
---Max 4 elements
---
---Example Usage:
---`padding = {1}` {Top|Bottom|Left|Right}
---`padding = {1, 2}` {Top|Bottom, Left|Right}
---`padding = {1, 2, 3, 4}` {Top, Right, Bottom, Right}
---@field padding? number[]
---Sets background color of the element
---@field backgroundColor? string
---Sets the font family to the specified font
---@field fontFamily? string
---Sets the font size to the specified pixel size
---@field fontSize? number
---Sets the fixed width of the element
---@field width? number
---Sets the fixed height of the element
---@field height? number

--- Slide is the base primitive of each slide of the slideshow.
--- Pass in other elements such as [Text](lua://Text) or
--- [VStack](lua://VStack) to present content to the slideshow. Each call to
--- Slide makes a new slideshow.
---@param children SlideNode[]
function Slide(children)
end
```

# How Has This Been Tested?
Tested on local lua codebases. 

# Checklist

### DIFF
--- a/harper-comments/src/comment_parser.rs
+++ b/harper-comments/src/comment_parser.rs
@@ -1,7 +1,7 @@
 use std::path::Path;
 
 use crate::comment_parsers;
-use comment_parsers::{Go, JavaDoc, JsDoc, Solidity, Unit};
+use comment_parsers::{Go, JavaDoc, JsDoc, Lua, Solidity, Unit};
 use harper_core::Token;
 use harper_core::parsers::{self, MarkdownOptions, Parser};
 use harper_core::spell::MutableDictionary;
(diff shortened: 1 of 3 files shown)


elijah-potter (changes requested):
Since there is such significant overlap with the `Unit` and `Go` parsers, I think we should refactor the new `Lua` parser to work abstractly, and replace the underlying `Unit` and `Go` implementations with it. I would modify the construct to accept a function pointer that maps from the line to an accept/reject boolean for whether to keep it.

PR #1824, opened 2025-08-28
chore: various improvements to phrasal verb linter

hippietrail:
# Issues 
N/A

# Description

- a compound noun at the start of a chunk followed by a preposition will no longer be flagged
- a compound noun followed by "that" or "which" will no longer be flagged
- exceptions added for "markup" and "plugin" as parts of compounds
- `annotations.json` now marks `-ing` forms of verbs as mass nouns and adjectives
- several dictionary entries needed to be tweaked
- existing tests were converted to use the more concise `assert_no_lints`

Unfortunately, a couple of quite subtle false positives are now flagged in the snapshots.
But some hitherto hidden false negatives are revealed and some false positives are fixed.

# How Has This Been Tested?

Several new unit tests are added using the sentences that revealed the problems this PR fixes.

# Checklist


elijah-potter (approved):
Looks like an improvement to me!

PR #1825, opened 2025-08-29
Restore Harper linter active/inactive status between sessions.

iBenMoore:
# Issues 

# Description
Restore Harper linter active/inactive status between sessions.
Also added an 'Activate Harper' option in settings as one more way to enable or disable Harper's active state.

# Demo

# How Has This Been Tested?
Tested closing and opening Obsidian vault with and without the current Harper state being enabled to validate current state is restored between sessions.
Tested toggling the new 'Activate Harper' setting to validate it successfully changes the Harper active state and the status bar indicator. 

# Checklist

### DIFF
--- a/packages/obsidian-plugin/src/HarperSettingTab.ts
+++ b/packages/obsidian-plugin/src/HarperSettingTab.ts
@@ -87,12 +87,12 @@ export class HarperSettingTab extends PluginSettingTab {
 		});
 
 		new Setting(containerEl)
-			.setName('Restore Active Status')
-			.setDesc('Whether to restore the Harper active/inactive status between sessions.')
+			.setName('Activate Harper')
+			.setDesc('Enable or disable harper with this option.')
 			.addToggle((toggle) =>
-				toggle.setValue(this.settings.lintEnabledRestore).onChange(async (value) => {
-					this.settings.lintEnabledRestore = value;
-					await this.state.initializeFromSettings(this.settings);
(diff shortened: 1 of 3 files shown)


packages/obsidian-plugin/src/HarperSettingTab.ts
```diff
@@ -86,6 +86,16 @@ export class HarperSettingTab extends PluginSettingTab {
 				});
 		});
 
+		new Setting(containerEl)
+			.setName('Activate Harper')
+			.setDesc('Enable or disable harper with this option.')
```
elijah-potter, line 91: Make sure you capitalize `Harper`.

packages/obsidian-plugin/src/State.ts
```diff
@@ -28,6 +29,7 @@ export default class State {
 	private onExtensionChange: () => void;
 	private ignoredGlobs?: string[];
 	private editorViewField?: StateField<MarkdownFileInfo>;
+	private lintEnabled: boolean;
```
elijah-potter, line 32: On older installs of Obsidian, this value may be undefined. Would that break anything? If not, make sure you label the field as potentially undefined.

elijah-potter (commented):
Good stuff. I gave it a whirl and it seems to work. I have a few nitpicks before we can merge this.

iBenMoore:
Got it, sending updated PR.

PR #1829, opened 2025-08-29
feat: add correction of degrees kelvin to -> kelvin

lukasmwerner:
# Description
Corrects usage of `°K` -> `K`

# How Has This Been Tested?

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_corrections/mod.rs
@@ -669,6 +669,13 @@ pub fn lint_group() -> LintGroup {
             "Ensures `just deserts` is used correctly, preserving its meaning of receiving an appropriate outcome for one's actions.",
             LintKind::Spelling
         ),
+        "KelvinDegreesSymbol" => (
+            ["°K"],
+            ["K"],
+            "Use just K when discussing units of Kelvin. No degree symbol needed.",
+            "Replaces degree symbol followed by K to just the kelvin symbol",
+            LintKind::Redundancy
+        ),
         "KindOf" => (
             ["kinda of"],
             ["kind of", "kinda"],
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2281,7 +2281,7 @@ dependencies = [
 
 [[package]]
 name = "harper-brill"
-version = "0.59.0"
+version = "0.60.0"
 dependencies = [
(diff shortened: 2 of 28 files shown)


ccoVeille:
TIL something, thanks

lukasmwerner:
I was only passively aware of this but I was going through the AP style guide and was reminded about it.

lukasmwerner:
Just fixed this up and added tests. 

On a side note, I feel it is necessary to update the documentation on authoring a rule that all descriptions need to be unique. In the process of making the test case I realized the rule I was authoring in fact needed to be two separate rules. However, since I just lazily copied the rule and put the same description, it took me a bit to realize that the description needed to be different.

harper-core/src/linting/phrase_corrections/mod.rs
```diff
@@ -300,6 +300,13 @@ pub fn lint_group() -> LintGroup {
             "Corrects the eggcorn `day in age` to `day and age`, which properly means the current era or time period.",
             LintKind::Eggcorn
         ),
+        "DegreesKelvin" => (
+            ["degrees kelvin", "degrees Kelvin"],
```
hippietrail: What about singular if it's just one degree kelvin?

elijah-potter: This should be a quick fix. Just add a new line for `DegreeKelvin`

lukasmwerner: I believe I just added that? Should be ready for final review.

elijah-potter (approved):
Looks good to me. Thanks @lukasmwerner!

hippietrail:
> Just fixed this up and added tests.
> 
> On a side note, I feel it is necessary to update the documentation on authoring a rule that all descriptions need to be unique. In the process of making the test case I realized the rule I was authoring in fact needed to be two separate rules. However, since I just lazily copied the rule and put the same description, it took me a bit to realize that the description needed to be different.

Good point. Often you can get two different "rules" into a single linter. The pro is that the user only has to consider one config and the con is the code is more complex.

I've also been thinking about a way to combine multiple linters into a `LintGroup` in a new way where they'd both be visible to the user under a single setting but to the programmer as separate. I don't know how possible this is with how `LintGroup` currently works.

PR #1833, opened 2025-08-30
feat: condense loan phrases from foreign languages into single tokens

hippietrail:
# Issues 
No issue as such but prompted by @elijah-potter 's question in Discord: https://discord.com/channels/1335035237213671495/1335036485765828668/1411099860807057418

Though I'd been thinking of implementing something like this in recent weeks as well.

# Description

Condenses all foreign loan phrases that are in our dictionary into single tokens, including `en masse`, `kung fu`, etc.

I renamed the `condense_latin` method since there's more Latin in the new method now.
I rearranged the order of the condensing functions with the helper for `SequenceExp` ones after all its uses. I put a comment before each of those to make it clearer that each consisted of a `fn_uncached_*_pattern()`, a `thread_local!` `*_EXPR`, and a `condense_*` function.

I experimented with building the list of multi-word terms to condense by checking all multi-word entries in the dictionary to see which of them included at least one word not in the dictionary in its own right. This revealed a few surprising English entries that fit that pattern too, but since we want to initialize the condensing pattern statically we can't use the `FstDictionary`.
I discovered two typos in the dictionary while working on this.
One false positive in the snapshots is resolved by this as a bonus.

# How Has This Been Tested?

I've added a unit test that checks that a two-word and a three-word loan phrase both get tokenized.

# Checklist

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -136,7 +136,8 @@ impl Document {
         self.condense_dotted_initialisms();
         self.condense_number_suffixes();
         self.condense_ellipsis();
-        self.condense_latin();
+        self.condense_dotted_latin();
+        self.condense_loan_phrases();
         self.condense_filename_extensions();
         self.condense_tldr();
         self.condense_ampersand_pairs();
@@ -438,11 +439,13 @@ impl Document {
         self.tokens.remove_indices(remove_these);
     }
 
+    // Dotted Latin expressions such as etc. vs. et al.
+
     thread_local! {
-        static LATIN_EXPR: Lrc<FirstMatchOf> = Document::uncached_latin_expr();
+        static DOTTED_LATIN_EXPR: Lrc<FirstMatchOf> = Document::uncached_dotted_latin_expr();
     }
 
-    fn uncached_latin_expr() -> Lrc<FirstMatchOf> {
+    fn uncached_dotted_latin_expr() -> Lrc<FirstMatchOf> {
         Lrc::new(FirstMatchOf::new(vec![
             Box::new(
                 SequenceExpr::default()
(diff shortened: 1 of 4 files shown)


harper-core/src/document.rs
```diff
@@ -458,27 +461,75 @@ impl Document {
...
+                "per diem",
+                "per se",
+                "prima facie",
+                "pro rata",
+                "quid pro quo",
+                "sui generis",
+                "tai chi",
+                "tom yam",
+                // "vis-à-vis",
+                "",
```
ccoVeille: This empty entry is strange, no?

harper-core/src/document.rs
```diff
@@ -458,27 +461,75 @@ impl Document {
...
+                "Bézier curve",
+                "bona fide",
+                // "cul-de-sac",
+                "de facto",
+                "de jure",
+                "de minimis",
+                "déjà vu",
+                "deja vu",
+                "en masse",
+                // "foo bar baz",
```
ccoVeille: This is about meta syntaxic variables.

https://en.wikipedia.org/wiki/Metasyntactic_variable

Is there a rule about them ?

Same question for Loren Ipsum ?

https://en.wikipedia.org/wiki/Lorem_ipsum

I'm sorry I ask without checking.

Is there a need for meta syntaxic and Lorem Ipsum in this rule, or in another ?

harper-core/src/document.rs
```diff
@@ -458,27 +461,75 @@ impl Document {
...
+        static LOAN_PHRASES_EXPR: Lrc<FirstMatchOf> = Document::uncached_loan_phrases_expr();
+    }
+
+    fn uncached_loan_phrases_expr() -> Lrc<FirstMatchOf> {
+        Lrc::new(FirstMatchOf::new(
+            [
+                "ad nauseam",
+                "alma mater",
+                // "avant-garde",
+                "Bézier curve",
```
ccoVeille: This one looks strange.

The plural should also be considered.

And when I see this one I can think about many others that are in the same logic.

Doppler effect
Turing machine
Gaussian distribution
Poisson distribution

But also some with a possessive

Euler's identity
Coulomb's law
Newton's laws
Parkinson's disease




Is there a need for these "Eponymous collocations" ?

 And is this rule the right one for this edge case?

harper-core/src/document.rs
hippietrail: Woah! Thanks for bringing that to my attention.

harper-core/src/document.rs
hippietrail: > This one looks strange.
> 
> The plural should also be considered.

Good point. That might be evidence that I took it too far. In this case just adding `Bézier` as a proper noun might be the better solution.

> And when I see this one I can think about many others that are in the same logic.
> 
> Doppler effect Turing machine Gaussian distribution Poisson distribution

The idea is this is for multi-word terms that contain one word that isn't in `dictionary.dict` and shouldn't be. Those are either already in there or are definitely the kind of proper nouns that belong there.

> But also some with a possessive
> 
> Euler's identity Coulomb's law Newton's laws Parkinson's disease
> 
> Is there a need for these "Eponymous collocations" ?

This is not what this linter is about. I think there's a term for them that's not springing to mind right now...

> And is this rule the right one for this edge case?

ccoVeille: > > This one looks strange.
> > 
> > The plural should also be considered.
> 
> Good point. That might be evidence that I took it too far. In this case just adding `Bézier` as a proper noun might be the better solution.
> 
> > And when I see this one I can think about many others that are in the same logic.
> > 
> > Doppler effect Turing machine Gaussian distribution Poisson distribution
> 
> The idea is this is for multi-word terms that contain one word that isn't in `dictionary.dict` and shouldn't be. Those are either already in there or are definitely the kind of proper nouns that belong there.

I agree, these proper noun should be simply in the dictionary.

So it means that this line shouldn't be in this rule.

> > Is there a need for these "Eponymous collocations" ?
> 
> This is not what this linter is about.

By linter, you meant this rule,
or you are talking about Harper itself as a linter?

> I think there's a term for them that's not springing to mind right now...

So you think there would be q need for them?

harper-core/src/document.rs
ccoVeille: You resolved without providing an explanation 😬😭

harper-core/src/document.rs
hippietrail: > > This is not what this linter is about.
> 
> By linter, you meant this rule, or you are talking about Harper itself as a linter?

Oh sorry I'm so used to writing linters that I forgot this is actually a condenser on `Document` rather than a linter.

> > I think there's a term for them that's not springing to mind right now...
> 
> So you think there would be q need for them?

harper-core/src/document.rs
```diff
@@ -459,27 +462,73 @@ impl Document {
...
-            remove_indices.extend(m.start + 1..m.end);
-            self.tokens[m.start].span = self.tokens[m.into_iter()].span().unwrap();
-            edit(&mut self.tokens[m.start]);
-        }
+    thread_local! {
+        static LOAN_PHRASES_EXPR: Lrc<FirstMatchOf> = Document::uncached_loan_phrases_expr();
+    }
+
+    fn uncached_loan_phrases_expr() -> Lrc<FirstMatchOf> {
+        Lrc::new(FirstMatchOf::new(
```
ccoVeille, line 533: If you like Latin, here it is

ad hoc
​ad hominem
​ad infinitum
​ad lib
​ad valorem
​alter ego
​ante bellum
​a posteriori
​a priori
​carpe diem
​caveat emptor
​caeteris paribus
​cogito, ergo sum
​cum laude
​curriculum vitae
​deus ex machina
​dramatis personae
​e.g. (exempli gratia)
​et al. (et alii)
​et cetera
​ex officio
​ex post facto
​id est (i.e.)
​in absentia
​in memoriam
​in toto
​in vino veritas
​in vitro
​lapsus linguae
​mea culpa
​mea maxima culpa
​modus operandi
​non sequitur
​nota bene (N.B.)
​opus magnum
​persona non grata
​post mortem
​pro bono
​pro forma
​quo vadis?
​rigor mortis
​sine qua non
​status quo
​tabula rasa
​tempus fugit
​veni, vidi, vici
​vice versa

These two are also Latin, but not multi words:

veto
sic

elijah-potter:
My gut doesn't love the idea of merging words just because they're part of a larger foreign phrase. Is there a better solution that allows us to keep them as their own tokens?

hippietrail:
> My gut doesn't love the idea of merging words just because they're part of a larger foreign phrase. Is there a better solution that allows us to keep them as their own tokens?

Yes but it would probably be down to you to code it (-: make the spellchecker aware of multi-word terms using something like a sliding window that can check two-word and possibly three-word terms.

We could probably add a flag to `WordMetadata` meaning "is used in multi-word terms" so we can prune a lot of more expensive checks.

It would need to work with both compounds with spaces between words and with hyphens between words, and accept any kind of whitespace as a space.

The good part is the dictionary already works with multi-word terms.

ccoVeille:
@hippietrail was anything blocking with this PR?

hippietrail:
> @hippietrail was anything blocking with this PR?

Yes. Elijah had valid concerns with my implementation and my proposed "proper fix" probably requires deeper knowledge about how more parts of Harper work than I yet possess. I also don't know if Elijah thinks that's the way forward or not.

hippietrail:
Superseded by [#2879](https://github.com/Automattic/harper/pull/2879)

Issue #1832, opened 2025-08-30
feat: Mixing up "to" and "too"


hippietrail:
**The Error**
This is a very common error in English. It may even be more common among native speakers than learners.
Though the word "two" is pronounced the same as "to" and "too", it's very rare for "two" to get mixed up with the other words.

We do fix a few of the low-hanging fruit in this category in `map_phrase_linter` and/or `map_phrase_set_linter`, but we don't do well yet in the general case.

**Examples:**
1. This isn't **to** hard **too** do.
2. I was just thinking just by joining the server would actually be too hard **too** do.
3. With the current behavior its way **to** easy to move the folder to the Side and drag your items to pinned by accident.

PR #1839, opened 2025-08-31
feat: to/too mixups in set phrases

hippietrail:
# Issues 

Help out with #1832

# Description

Looks for idioms and set phrases that should use "too" where instead "to" has been used.
Also one case of the opposite.
Also moved and expanded "spoke to soon" from `phrase_corrections` to `phrase_set_corrections`.

I only included those that got more than a few search hits.
For others which got plenty of search hits but also lots of false positives, I included comments so people won't hastily try to add them.

# How Has This Been Tested?

I added one unit test for each pattern that's flagged, all or most sourced from GitHub.

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -353,7 +353,7 @@ pub fn lint_group() -> LintGroup {
             "Corrects `rise the question` to `raise the question`.",
             LintKind::Grammar
         ),
-        "ToToo" => (
+        "ToTooIdioms" => (
             &[
                 (&["a bridge to far"], &["a bridge too far"]),
                 (&["cake and eat it to"], &["cake and eat it too"]),


harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -353,6 +353,49 @@ pub fn lint_group() -> LintGroup {
             "Corrects `rise the question` to `raise the question`.",
             LintKind::Grammar
         ),
+        "ToToo" => (
```
elijah-potter: I'm worried this would get confused with the `ToTwoToo` rule. Maybe `IdiomaticToToo`?

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -353,6 +353,49 @@ pub fn lint_group() -> LintGroup {
...
+                (&["going to far"], &["going too far"]),
+                (&["gone to far"], &["gone too far"]),
+                (&["went to far"], &["went too far"]),
+
+                // "in to deep" has many false positives
+                (&["life's to short", "lifes to short"], &["life's too short"]),
+                (&["life is to short"], &["life is too short"]),
+
+                // "one to many" has many false positives
+                (&["put to fine a point"], &["put too fine a point"], ),
```
elijah-potter, line 373: I'm wondering if we can take a more data driven approach to this file as a whole. Is there a public dataset of these kinds of corrections?

harper-core/src/linting/phrase_set_corrections/mod.rs
hippietrail: > I'm worried this would get confused with the `ToTwoToo` rule. Maybe `IdiomaticToToo`?

I think there's a tension between names which will be useful for users since they're used in the options, and names which don't conflict for the programmers.

I think we were working on these at the same time. All uses of "to" and "too" are idiomatic. The proper solution is surely to merge them by way of a `LintGroup`. I think but am not sure that there's not a way to group lints together so that only the name of the group is exposed to user? I looked into that a week ago but forget.

(But see my other comment. I forget this PR was a `phrase_corrections` one.)

As a sidenote, while "two" is a third homophone for these, I've never seen it mixed up for the others in either direction. People seem to find one difference a lot more salient than the other.

harper-core/src/linting/phrase_set_corrections/mod.rs
hippietrail: Oh yeah sorry. In my previous comment I forget the way I made this PR! For now I think `ToTooIdioms` and/or `TooToIdioms` or maybe ...`InIdioms`, but I still think putting them behind a single setting is best, even when some are phrase corrections and some are a dedicated linter. (Not `Idiomatic` since that term has a broader meaning than `idiom`.)

Since these ones are all in idioms I used the Wiktionary article for "too", which links all the idioms it's a part of. I checked them one by one since some are too marginal to include.

You could consider that a kind of "data driven". Idioms are special so finding lists of them already assembled is probably the way. For fixing "to" vs "too" in the general case outside of idioms a more "big data" approach analysing neighbouring word properties would be the best. As for all general grammar linters, especially syntax. It would require a good corpus of good English and a good corpus of bad English as well, ideally.

elijah-potter:
I don't want labeling to stop us from merging this. I can't think of an immediate solution to the `LintGroup` problem. Can I change the name to `ToTooIdioms` and hit "merge"?

hippietrail:
> I don't want labeling to stop us from merging this. I can't think of an immediate solution to the `LintGroup` problem. Can I change the name to `ToTooIdioms` and hit "merge"?

Yes. Absolutely!

Issue #1838, opened 2025-08-31
Flag using "cant" instead of "can't"


hippietrail:
**The Error**

This is one of many apostrophe errors that are usually out of laziness rather than a real typo or spelling mistake. Unlike in "it's" most people know it should be there but leave it out for quicker typing in casual style.

But "cant" is also a legit word for a kind of jargon or secret language. More rarely, [it can also be a verb](https://en.wiktionary.org/wiki/cant#Verb).

**Examples (2-3):**
1. why i **cant** install vrcftacetracking
2. hi everyone, i **cant** install any app in ish, i trying use a proxy, but it don't work with proxy
3. [Bug] **cant** accept connection in log repeating although no scripts are calling it

**Potential Edge Cases**

A robust implementation that tries to rule out legit uses will be all about edge cases. A simpler solution could flag every instance but explain when it's not a mistake in the message.

elijah-potter:
Can we close this issue?

hippietrail:
> Can we close this issue?

Yes. I thought putting `implements` with the issue number did that automatically? Or does it have to be 'fixes' or 'closes' only?

PR #1840, opened 2025-08-31
feat: implements #1838

hippietrail:
# Issues 

Implements #1838 

# Description

Flags uses of "cant" where it should be "can't" and tries to avoid flagging uses of "cant" that could refer to a kind of jargon or secret language.

Also adds a helper function that finds the only matching token to be used inside the `match_to_lint` of `SequenceExpr` that fails if zero or multiple match.

# How Has This Been Tested?

A bunch of unit tests are included with both simple examples and real sentences from GitHub.

There are probably false negatives and possibly false positives in places using the jargon sense. Please file issues for any found.

# Checklist


harper-core/src/linting/cant.rs
```diff
@@ -0,0 +1,153 @@
...
+            span: token.span,
+            lint_kind: LintKind::Enhancement,
+            suggestions: vec![Suggestion::replace_with_match_case_str(cannot, jargon)],
+            message: "`Cant` is secret language or jargon. If that's not what you mean you should use `can't` here.".to_string(),
+            priority: 127,
+        })
+    }
+
+    fn description(&self) -> &'static str {
+        "Suggests correcting `cant` to `can't`."
```
ccoVeille, line 75: It can also be cannot

Issue #1190, opened 2025-04-30
Better documentation for the dictionary format


RunDevelopment:
I just wanted to edit a property of word in the dictionary and there is basically no documentation (that I could find) on the format of `dictionary.dict`. The [website](https://writewithharper.com/docs/contributors/dictionary) doesn't say much.

## Format

I assume each line has the form:

```
<word> "/" <tags> <space>* ( "#" <any>* )? <new-line>
```

But it would be nice if that was explained.

## Tags

Similarly, the docs for tags isn't great. Pointing towards 800 lines of JSON isn't documentation...

It's also not explained what happens when a tag appears multiple times. Are tags single characters (so having the same character twice does nothing) or can multiple characters represent a single tag (morse-code/Huffman-encoding style)?
I'm asking, because the word "nitroglycerine" has two `@` tags.

And does the order of tags matter?

# Suggestions

Since `affixes.json` also contains comments for every entry, we could use these entries to store a short description and then generate documentation from it. For example, I wrote a small script that turns `affixes.json` into this markdown:

<details>

Properties:

- `1`: (singular) noun property
- `2`: proper noun property
- `3`: linking verb property
- `4`: verb property
- `5`: adjective property
- `6`: swear word property
- `7`: conjunction property
- `8`: pronoun property
- `9`: plural noun property
- `~`: common word property
- `+`: preposition property
- `-`: determiner property
- `j`: adverb property
- `s`: subject case property
- `o`: object case property
- `f`: first-person property
- `n`: second-person property
- `t`: third-person property
- `c`: comparative property
- `u`: superlative property
- `*`: singular third-person subject pronoun property
- `:`: plural pronoun property
- `m`: auxiliary verb property
- `<`: American property
- `!`: GB property
- `@`: CA property
- `_`: AU property

Prefixes:
- `K`: 'pro-' prefix
- `E`: 'dis-' prefix
- `U`: 'un-' prefix
- `C`: 'de-' prefix
- `A`: 're-' prefix
- `F`: 'con-' prefix
- `I`: 'in-' prefix

Suffixes:

- `L`: '-ment' suffix
- `Y`: '-ly' suffix
- `H`: '-ieth' suffix
- `T`: '-(i)est' suffix
- `R`: '-(i)er' suffix
- `V`: '-ive' suffix
- `N`: nominalization suffixes: -ion, -ication, -en
- `Z`: '-(i)(e)rs' suffix
- `P`: '-(i)ness' suffix
[...]

PR #1851, opened 2025-09-02
feat: support roman numerals

hippietrail:
# Issues 

Not filed as an issue but Roman numerals came up in discussion in #1190

# Description

Adds support for Roman numerals. Not as a `TokenKind` since words, especially `I` can be simultaneously an English word and a Roman numeral, and only analysing the context can reveal which. Instead they are added as part of the orthography flags in the `WordMetadata` that also tracks whether dictionary entries are capitalized, hyphenated, apostrophized, use camel case, etc.

The code first checks that only the letters MDCLXVI are used and they can't be mixed case. If that test passes then a more complex check is done to make sure that the letters are in a sensible order. We could probably go with just the second test but my hunch was that it would be slower or have a higher potential for edge cases. I can remove the preliminary test if that way is preferred.

In the POS tags snapshots they are marked with `#r` in line with `#` for regular numbers and `#d` for decades.

# Demo
image

# How Has This Been Tested?

Unit tests have been added to enforce all the above constraints.

# Checklist


hippietrail:
> I like this. Good stuff.
> 
> In a separate PR, it might be a good idea to skip spellchecking on Roman numbers whose length is greater than or equal to two.

Or skipped if the roman numerals property is set but none of the POS properties perhaps.

elijah-potter:
Ooh yes, I like that too.

harper-core/src/spell/rune/attribute_list.rs
```diff
@@ -333,9 +333,101 @@ fn check_orthography(word: &MarkedWord) -> OrthFlags {
...
+
+    if !check_roman_group(&mut chars, 'i', 'v', 'x') {
+        return false;
+    }
+
+    if chars.next().is_some() {
+        return false;
+    }
+
+    println!("Valid roman numeral: {}", s);
```
elijah-potter: On second thought, we don't want these functions to print anything.

hippietrail: > On second thought, we don't want these functions to print anything.

Woah good catch! I'm on it...

PR #1854, opened 2025-09-03
docs: add path to `stats.txt` for Linux and macOS

hippietrail:
# Issues 
N/A but it came up in this Discord thread: 

# Description

Add the path to `stats.txt` for Linux and macOS in the website documentation.

# Checklist

### DIFF
--- a/packages/web/src/routes/docs/contributors/local-stats/+page.md
+++ b/packages/web/src/routes/docs/contributors/local-stats/+page.md
@@ -28,7 +28,7 @@ The `stats.txt` file is formatted into lines (so it is easy to open in append-mo
 {"kind":{"Lint":{"kind":"Spelling","context":[{"content":"Teasting","kind":{"kind":"Word","value":null}}]}},"when":1743696288,"uuid":"bd955190-a4d9-4f3e-b7df-d4bf6f12a415"}
 ```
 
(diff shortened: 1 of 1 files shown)


mcecode:
Why not add where it is on Windows too? I think it should be in `C:\Users\<username>\AppData\Local\harper-ls\stats.txt` on most systems.

hippietrail:
> Why not add where it is on Windows too? I think it should be in `C:\Users\<username>\AppData\Local\harper-ls\stats.txt` on most systems.

Only because I didn't know. I was hoping somebody would come along with the info. So thanks!
I'll add it as `%APPDATA%/harper-ls/stats.txt` so let me know if that's wrong and I'll replace it with the long version. (Reasoning is that `%APPDATA%` feels similar to `~`)

mcecode:
> Only because I didn't know. I was hoping somebody would come along with the info. So thanks! I'll add it as `%APPDATA%/harper-ls/stats.txt` so let me know if that's wrong and I'll replace it with the long version. (Reasoning is that `%APPDATA%` feels similar to `~`)

When using environment variables, I think the most accurate would be `%FOLDERID_LocalAppData%/harper-ls/stats.txt` since that's what [`dirs`](https://codeberg.org/dirs/dirs-rs#features) uses and it's also what we use on our [`harper-ls` docs](https://writewithharper.com/docs/integrations/language-server#File-Local-Dictionary).

hippietrail:
> %FOLDERID_LocalAppData%/harper-ls/stats.txt

Thank you once again!

elijah-potter (approved):

Issue #1753, opened 2025-08-14
"I've looking" should suggest "I'm looking" and "I've been looking"


hippietrail:
I noticed this one in our own recent issues/discussions:

<img width="254" height="30" alt="Image" src="https://github.com/user-attachments/assets/a130a752-90d7-4bc7-b294-e0dd9fc48fb5" />

I make similar word omission errors pretty often myself.

We should probably look for (`have` or -`'ve` contractions) followed by a verb in the "progressive form" and it seems like there's two good suggestions:

- I've looking into → I've been looking into
- I've looking into → I'm looking into

hippietrail:
@elijah-potter, was this resolved with #1860?

elijah-potter:
I believe we can close this.

PR #1860, opened 2025-09-03
feat(core): wrote `IveGerund` rule

elijah-potter:
# Issues 

#1753

# Description

Adds a rule for addressing the grammatical error outlined in #1753

# How Has This Been Tested?

Additional unit tests.

# Checklist


harper-core/src/linting/ive_gerund.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use super::{ExprLinter, Lint, LintKind, Suggestion};
+
+pub struct IveGerund {
+    expr: Box<dyn Expr>,
+}
+
+impl Default for IveGerund {
+    fn default() -> Self {
+        let expr = SequenceExpr::aco("I've")
```
ccoVeille: "I/you/we/they have looking" can also be found

https://github.com/search?q=%2Fhave%20looking%2F&type=code

ccoVeille: Also simple tick 

https://github.com/search?q=%2Fi%E2%80%99ve+looking%2F&type=code

"I’ve looking"

https://github.com/mcgarrah/mcgarrah.github.io/blob/1c0af2cec529e611db4610521a5a31a26cb3c254/_posts/2014-07-01-wordpress-plugins.md?plain=1#L17

ccoVeille: And "has looking"

https://github.com/search?q=%2Fhas+looking%2F&type=code

harper-core/src/linting/ive_gerund.rs
```diff
@@ -0,0 +1,126 @@
...
```
ccoVeille: `We've`, `they've` and `you've` can be found

https://github.com/search?q=%2F%5B%5Ei%5D%27ve+looking%2F&type=code

harper-core/src/linting/ive_gerund.rs
hippietrail: I've been thinking about the best way to handle contractions in the system generally for weeks. Currently they have no annotation in `dictionary.dict`. They probably call for something like a vec/slice of `WordMetadata` per "word" or such. In the meantime: I/we/you/they+have/had & he/she/it+has/had would be the full pattern.

harper-core/src/linting/ive_gerund.rs
hippietrail: Probably all the `'d` variants too for past perfect.

harper-core/src/linting/ive_gerund.rs
```diff
@@ -0,0 +1,126 @@
+use crate::expr::Expr;
+use crate::expr::SequenceExpr;
+use crate::{Token, TokenKind};
+
+use super::{ExprLinter, Lint, LintKind, Suggestion};
+
+pub struct IveGerund {
```
hippietrail: A gerund is an -ing form of a verb acting as a noun. This is actually a present participle, which is the -ing form of a verb in compound tenses (or adjectives which doesn't concern us here). And it's likely to apply to I've, I'd, we've, we'd, you've, you'd, he's, he'd, she's, she'd, it's, it'd, they've, and they'd.

The normal names for the compound tenses are `present perfect` "have eaten" and `past perfect` "had eaten". So we're dealing with a mixup of a `perfect tense` and a `present participle`. While `gerund` is "too nounish" we could chose one of the "verbish" common names for the -ing form: `continuous` or `progressive`. So maybe `PerfectContinuous` or `PerfectProgressive` would be good names for a broader version of the linter?

harper-core/src/linting/ive_gerund.rs
```diff
@@ -0,0 +1,126 @@
...
+                Suggestion::replace_with_match_case("I'm".chars().collect(), span.get_content(src)),
+                Suggestion::InsertAfter(" been".chars().collect()),
+            ],
+            message: "Use present progressive (`I'm …`) or present perfect progressive (`I've been …`) instead of `I've …ing`.".to_string(),
+            priority: 31,
+        })
+    }
+
+    fn description(&self) -> &'static str {
+        "Detects the ungrammatical pattern `I've` followed directly by a gerund (e.g., `I've looking`) and suggests either the present progressive (`I'm …`) or the present perfect progressive (`I've been …`)."
```
hippietrail: Definitely avoid the word `gerund` in the message and definition. That would be a grammar checker spreading bad grammar (-: `continuous`, `progressive`, or `-ing form`. You've nailed the other terminology!

harper-core/src/linting/ive_gerund.rs
```diff
@@ -0,0 +1,126 @@
...
+            "I'm looking into it.",
+        );
+    }
+
+    #[test]
+    fn offers_both_suggestions() {
+        assert_good_and_bad_suggestions(
+            "I've looking into it.",
+            IveGerund::default(),
+            &["I'm looking into it.", "I've been looking into it."],
```
hippietrail: ❤️

harper-core/src/linting/ive_gerund.rs
ccoVeille: But this one might be more general about using simple quote or tick or back tick for contraction.

But people might disable a rule about contraction, so here this should be detected anyway

harper-core/src/linting/ive_perfect_progressive.rs
```diff
@@ -0,0 +1,170 @@
+use crate::expr::Expr;
+use crate::expr::SequenceExpr;
+use crate::{Token, TokenKind};
+
+use super::{ExprLinter, Lint, LintKind, Suggestion};
+
+pub struct IvePerfectProgressive {
```
ccoVeille: While the code is fine the name if the rule is a bit misleading.

It now supports:
- other pronouns than `I`
- other verb contraction than `'ve`

hippietrail: I was thinking it doesn't really matter in the code but I think you're right because it's also going to be the name of an option.

ccoVeille: Yes, that's my concern

harper-core/src/linting/lint_group.rs
```diff
@@ -58,6 +58,7 @@ use super::in_on_the_cards::InOnTheCards;
 use super::inflected_verb_after_to::InflectedVerbAfterTo;
 use super::its_contraction::ItsContraction;
 use super::its_possessive::ItsPossessive;
+use super::progressive_needs_be::ProgressiveNeedsBe;
```
ccoVeille: Shouldn't it be moved aside pronoun_*

Here I feel the rule is added at the wrong place, assuming they are sorted alphabeticaly

harper-core/src/linting/lint_group.rs
```diff
@@ -444,6 +445,7 @@ impl LintGroup {
         insert_struct_rule!(HowTo, true);
         insert_expr_rule!(HyphenateNumberDay, true);
         insert_expr_rule!(IAmAgreement, true);
+        insert_expr_rule!(ProgressiveNeedsBe, true);
```
ccoVeille: Sane remark about the position among the list

harper-core/src/linting/mod.rs
```diff
@@ -57,6 +57,7 @@ mod it_is;
 mod it_would_be;
 mod its_contraction;
 mod its_possessive;
+mod progressive_needs_be;
 mod left_right_hand;
```
ccoVeille: Same

harper-core/src/linting/mod.rs
```diff
@@ -185,6 +186,7 @@ pub use inflected_verb_after_to::InflectedVerbAfterTo;
 pub use initialism_linter::InitialismLinter;
 pub use its_contraction::ItsContraction;
 pub use its_possessive::ItsPossessive;
+pub use progressive_needs_be::ProgressiveNeedsBe;
```
ccoVeille: Same

ccoVeille (commented):
Nitpicking

PR #1869, opened 2025-09-05
feat: add colour to `harper-cli nominal-phrases` and `just getnps-colour`

hippietrail:
# Issues 
N/A

# Description

The new colour versions will print out the whole document with all nominal phrases highlighted.

It also now accepts either a text string or filename on the commandline.

The just command works with `-color` and `-colour`.

# Demo
image

# How Has This Been Tested?
Manually

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -17,8 +17,8 @@ use harper_comments::CommentParser;
 use harper_core::linting::{LintGroup, Linter};
 use harper_core::parsers::{Markdown, MarkdownOptions, OrgMode, PlainEnglish};
 use harper_core::{
-    CharStringExt, Dialect, Document, TokenKind, TokenStringExt, WordMetadata, remove_overlaps,
-    word_metadata_orthography::OrthFlags,
+    CharStringExt, Dialect, Document, Span, TokenKind, TokenStringExt, WordMetadata,
+    remove_overlaps, word_metadata_orthography::OrthFlags,
 };
 use harper_literate_haskell::LiterateHaskellParser;
 #[cfg(feature = "training")]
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -591,8 +591,15 @@ newest-dict-changes *numCommits:
     });
   });
 
-getnps a:
-  cargo run --bin harper-cli -- nominal-phrases "{{a}}"
+# List the nominal phrases in the input string or file.
+getnps text:
```
elijah-potter, line 595: Let's just make color the default.

harper-cli/src/main.rs
```diff
@@ -746,14 +752,62 @@ fn main() -> anyhow::Result<()> {
...
+                        print!("{}", txt);
+                    }
+                }
+
+                // Highlighted nominal phrase
+                let span = Span::new(start, end);
+                let txt = doc.get_span_content_str(&span);
+
+                if colour {
+                    print!("\x1b[33m{}\x1b[0m", txt);
```
elijah-potter: Does these character cause noise on terminals that don't support color? Since this is a debug tool, do we even care about those terminals?

elijah-potter (changes requested):
I'm a huge fan of this. It's _way_ better than the original. Is there any reason we should keep the old method?

hippietrail:
> I'm a huge fan of this. It's _way_ better than the original. Is there any reason we should keep the old method?

Only that I thought you might have some use for it in its old form in your own tools etc.

By the way, I believe the `NominalPhrase` `Expr`/`Pattern` would do a different job. I'd love to interleave them in alternating colours. I'm not sure how to do that one yet though.

harper-cli/src/main.rs
hippietrail: > Does these character cause noise on terminals that don't support color? Since this is a debug tool, do we even care about those terminals?

VT100 has been a defacto standard for almost 50 years. Almost every terminal supports it even if by simply ignoring it. Sometimes they're a pain when redirecting or piping, but if anyone would want to do that, the old non-visual output format is what they should be using.

justfile
hippietrail: > Let's just make color the default.

I'll hold off until you reply to the other comment thread(s) here with a final decision.

harper-cli/src/main.rs
elijah-potter: Good to know. Let's keep that bit as-is.

elijah-potter:
> Only that I thought you might have some use for it in its old form in your own tools etc.

Nope. Let's just make color the default. I can't think of a use-case for the old version, so feel free to strip that out entirely unless you disagree.

PR #1871, opened 2025-09-05
feat: Added conditionals to spell check, allowing certain kinds of words to be ignored by spell check.

ethqnol:
# Description
- Added spell check config struct to manage conditionals in spell check
- LintGroupConfig has a field for spell check config
- Spell check config is modified through fn lint
- Linter trait now has "update_spell_check_config" but returns () by default except for the spell check linter

# How Has This Been Tested?
Added tests to check new spell check rules. 

# Checklist

### DIFF
--- a/harper-core/src/linting/lint_group.rs
+++ b/harper-core/src/linting/lint_group.rs
@@ -106,7 +106,7 @@ use super::since_duration::SinceDuration;
 use super::somewhat_something::SomewhatSomething;
 use super::sought_after::SoughtAfter;
 use super::spaces::Spaces;
-use super::spell_check::SpellCheck;
+use super::spell_check::{SpellCheck, SpellCheckConfig};
 use super::spelled_numbers::SpelledNumbers;
 use super::that_than::ThatThan;
 use super::that_which::ThatWhich;
@@ -245,6 +245,7 @@ impl Hash for LintGroupConfig {
 /// Each child can be toggled via the public, mutable [Self::config] object.
 pub struct LintGroup {
     pub config: LintGroupConfig,
+    pub spell_check_config: SpellCheckConfig,
     /// We use a binary map here so the ordering is stable.
     linters: BTreeMap<String, Box<dyn Linter>>,
     /// We use a binary map here so the ordering is stable.
@@ -263,6 +264,7 @@ impl LintGroup {
     pub fn empty() -> Self {
         Self {
             config: LintGroupConfig::default(),
+            spell_check_config: SpellCheckConfig::default(),
             linters: BTreeMap::new(),
             expr_linters: BTreeMap::new(),
             chunk_expr_cache: LruCache::new(NonZero::new(10000).unwrap()),
@@ -310,6 +312,7 @@ impl LintGroup {
     /// The other lint group will be left empty after this operation.
     pub fn merge_from(&mut self, other: &mut LintGroup) {
         self.config.merge_from(&mut other.config);
+        // Note: spell_check_config is not merged, as each LintGroup should maintain its own
 
         let other_linters = std::mem::take(&mut other.linters);
         self.linters.extend(other_linters);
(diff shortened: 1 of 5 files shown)


elijah-potter:
It looks like you'll need to run `cargo fmt`.

hippietrail:
Sorry it's 1am here so I'm not reading the code at my best. I see the capitalization check on the first letter but I'm not seeing the abbreviation check. Be aware that "abbreviation" can be interpreted in various ways but the broad interpretation encompasses acronyms and initialisms and with and without periods combined or not, and all caps or not.

At some point we may want to have settings grouped how various style guides and authorities prefer them, so we probably shouldn't pick one specific narrow meaning for "abbreviation".

Sorry for not putting this in an inline code review.

I think is a great idea by the way!

ethqnol:
No worries! thanks for the feedback! I currently busy with some thing in school, but I'll check back in and figure out why the tests are failing in a little bit.

ethqnol:
> Sorry it's 1am here so I'm not reading the code at my best. I see the capitalization check on the first letter but I'm not seeing the abbreviation check. Be aware that "abbreviation" can be interpreted in various ways but the broad interpretation encompasses acronyms and initialisms and with and without periods combined or not, and all caps or not.
> 
> At some point we may want to have settings grouped how various style guides and authorities prefer them, so we probably shouldn't pick one specific narrow meaning for "abbreviation".
> 
> Sorry for not putting this in an inline code review.
> 
> I think is a great idea by the way!

More on this, here's my function for checking abbreviations:

```
fn check_abbrv(chars: &[char]) -> bool {
    if chars.is_empty() {
        return false;
    }

    if chars.iter().all(|&c| c.is_uppercase() || c.is_numeric()) {
        return true;
    }

    // Find valid suffixes at the end
    let mut found_lowercase = false;
    let mut lowercase_start_idx = chars.len();

    for (i, &c) in chars.iter().enumerate().rev() {
        if c.is_lowercase() {
            lowercase_start_idx = i;
            found_lowercase = true;
        } else if found_lowercase {
            break;
        }
    }

    if !found_lowercase {
        return chars.iter().all(|&c| c.is_uppercase() || c.is_numeric());
    }

    let lowercase_part: String = chars[lowercase_start_idx..].iter().collect();

    let allowed_suffixes = ["d", "s", "'s", "ed"];
    let is_valid_suffix = allowed_suffixes
        .iter()
        .any(|&suffix| lowercase_part == suffix);

    if !is_valid_suffix {
        return false;
    }

    chars[..lowercase_start_idx]
        .iter()
        .all(|&c| c.is_uppercase() || c.is_numeric())
}

```

The idea is to essentially assume that any completely capitalized word (and also allow suffixes at the end) are abbreviations.

hippietrail:
> The idea is to essentially assume that any completely capitalized word (and also allow suffixes at the end) are abbreviations.

Ah thanks for that! I would recommend not naming it "abbreviations" in that case since it doesn't cover `etc.`, `mb`, `Mon`, `min` and the like. A name like "all caps" would be more specific. And your intent can be explained in comments.

harper-core/src/linting/lint_group.rs
```diff
@@ -245,6 +245,7 @@ impl Hash for LintGroupConfig {
 /// Each child can be toggled via the public, mutable [Self::config] object.
 pub struct LintGroup {
     pub config: LintGroupConfig,
+    pub spell_check_config: SpellCheckConfig,
```
elijah-potter, line 248: The purpose of `LintGroup` was to provide an easy interface for merging the inputs and outputs of individual linters. The curated lint group ended up being a special case of that.

I don't want to expose the implementation details of the constituent linters. Is there a way we can maintain the `LintGroup`?

In any case, we try to avoid exposing config that isn't directly related to a grammatical rule through the `LintGroup` it seems like this should be exposed as a separate option.

ethqnol: Hmm... I'll think of another approach

PR #1886, opened 2025-09-08
feat: addicting→addictive

hippietrail:
# Issues 
N/A

# Description

Corrects `addicting` to `addictive` when used as an adjective.
Doesn't do so when `addicting` is used as a verb.

Got a linter with `AnchorEnd` working!

- Made `SequenceExpr` ctors for `any_word` and `unless`.
- Tidied the order of the methods a bit for consistency.

# How Has This Been Tested?

There's unit tests for the various cases that should and shouldn't be flagged.

# Checklist

### DIFF
--- a/harper-core/src/expr/sequence_expr.rs
+++ b/harper-core/src/expr/sequence_expr.rs
@@ -84,6 +84,8 @@ impl Expr for SequenceExpr {
 impl SequenceExpr {
     // Constructor methods
 
+    // Single word token methods
+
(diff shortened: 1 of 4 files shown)


elijah-potter (approved):
Hmm... The more you know!

hippietrail:
> Hmm... The more you know!

I'm assuming many younger people will turn this one off. I was considering a similar one for "deceiving" vs "deceptive" but that one seems less clear cut.

Issue #1873, opened 2025-09-05
"TeX" is linted


Andrew15-5:
**What got flagged?**

`TeX`

**Why is this incorrect?**

Typst (😏) and even LaTeX are not linted. Then why should TeX be? Same with LuaTeX, and probably some others.

**Example of correct usage:**

TeX is a typesetting system.

hippietrail:
TeX is in there: 

<img width="523" height="134" alt="Image" src="https://github.com/user-attachments/assets/5f7fe77b-13be-499a-8634-c278e5e3c01b" />

I'll add `LuaTeX` in the specialized section. Will be in the next curation.

Andrew15-5:
> TeX is in there

Is in not released yet? I'm on latest.

hippietrail:
> > TeX is in there
> 
> Is in not released yet? I'm on latest.

It was added 3 months ago. But I can confirm it gets flagged as a spelling error so that's an actual bug you've found! I knew we had similar bugs with some all-caps acronyms etc. but hadn't seen it before on a mixed-case word.

Andrew15-5:
It is still being linted.

hippietrail:
This just came up again when I was curating the dictionary and not changing any code. After a lot of frustration it turns out the order of entries in the dictionary matters. There are different behaviours if `Tex` is before or after `TeX` in the dictionary.

I believe this should not be the case. I think it's a bug.

Andrew15-5:
The "TeX" is no longer linted, so you should open a new issue and reference this one. This issue is fixed and closed. Also, it's not clear what are the differences.

PR #1874, opened 2025-09-05
chore: mostly annotating verbs and nouns

hippietrail:
# Issues 
Fixes #1873

# Description

Applying new verb annotations.
Continuing with mass noun annotations.

"LuaTeX" added for #1873

Also simplifies the POS tag snapshot format so that `P` alone is used for verbs that have the same form for preterite (simple past) and past participle. Previously this was only for regular verbs.

# Checklist

- [x] I have performed a self-review of my own code
- [ ] I have added tests to cover my changes

PR #1889, opened 2025-09-08
chore: add & update several TeX-related proper nouns

Andrew15-5:
# Issues

https://github.com/Automattic/harper/issues/1873

# Description

Updated: TeX, LyX
Added: ConTeXt, LaTeX, pdfTeX

- https://en.wikipedia.org/wiki/ConTeXt
- https://en.wikipedia.org/wiki/LaTeX
- https://en.wikipedia.org/wiki/PdfTeX

See https://github.com/hippietrail/harper/pull/4.

Supersedes and closes https://github.com/hippietrail/harper/pull/4, which was left open when https://github.com/Automattic/harper/pull/1874 got merged.

# Demo

# How Has This Been Tested?

No testing.

# Checklist

Cc @hippietrail


Andrew15-5:
Looks like ConTeXt gives some false positives. It actually is not being linted, but just like LaTeX, it's not linted because there is a different "latex" word present. This should be throwing an error without correct proper noun.

elijah-potter:
I can confirm that this fixes #1873.

elijah-potter (approved):

PR #1892, opened 2025-09-09
feat: `harper-cli normalize-dictionary`

hippietrail:
# Issues 
N/A

# Description

(Work in progress)

- Adds a command to normalize `dictionary.dict`.
- (Will also add a command to audit `dictionary.dict` and `annotations.json`.)

# How Has This Been Tested?
Manual testing

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -996,7 +996,7 @@ fn normalize_annotation_flags(flag_str: &str) -> String {
             if let Some(which) = poses_with_props
                 .iter()
                 // The first entry in each pos-with-props vector is always its POS
-                .position(|pos_with_props| poses.contains(&pos_with_props.first().unwrap()))
+                .position(|pos_with_props| poses.contains(pos_with_props.first().unwrap()))
             {
                 poses_with_props[which].push(flag);
             } else {
@@ -1049,6 +1049,6 @@ fn normalize_annotation_flags(flag_str: &str) -> String {
     tildes
         .into_iter()
         .chain(poses_with_props.into_iter().flatten())
-        .chain(other_flags.into_iter())
+        .chain(other_flags)
         .collect::<String>()
 }


harper-cli/src/main.rs
```diff
@@ -891,3 +939,101 @@ fn file_dict_name(path: &Path) -> PathBuf {
...
+            pos_map.get_mut(&'\0').unwrap().push(*flag);
+        }
+    });
+
+    let mut result = String::new();
+    // get the pos in order then append it and its values to the result string
+    pos_order.iter().for_each(|pos_flag_char| {
+        result.push(*pos_flag_char);
+        // result.extend(pos_map.get(pos_flag_char).unwrap());
+        // let props = pos_map.get(pos_flag_char).unwrap();
```
ccoVeille: Commented code means it was a left behind debug, or some idea that was tried and is unresolved for now

harper-cli/src/main.rs
```diff
@@ -891,3 +939,101 @@ fn file_dict_name(path: &Path) -> PathBuf {
...
+    let mut pos_map: HashMap<char, Vec<char>> = HashMap::new();
+
+    let pos_tags = "NOVJRIPCD";
+    let noun_props = "09gmw";
+    let verb_props = "lAbdGtT6h";
+    let pron_props = "aso123F";
+    let adj_props = "^cuY*.:";
+    let det_props = "qM5";
+    let noun_and_verb_props = "S";
+    let verb_and_adj_props = ">";
```
ccoVeille: So it means you have to maintain annotations.json and this at the same time?

I might be wrong, but if not, I'm a bit worried.

harper-cli/src/main.rs
hippietrail: > Commented code means it was a left behind debug, or some idea that was tried and is unresolved for now

This is still a draft. I should've put "work in progress" in the description.

harper-cli/src/main.rs
hippietrail: > So it means you have to maintain annotations.json and this at the same time?
> 
> I might be wrong, but if not, I'm a bit worried.

I think I'll add fields to `annotation.json` to enable pulling these out programmatically. Still developing this feature...

ccoVeille: This is a pre-AI-era image I made ages ago, and I cherish

hippietrail: (-: Feedback is still cool.

hippietrail:
I can't find a way to get rid of all the merge commits from this PR, other than making a new clean PR. So #2133 replaces this PR.

PR #1895, opened 2025-09-09
feat: toggleable spell check rules

ethqnol:
# Description
- Added a spell check rule that ignores all caps words when enabled (targeted to let spell check ignore acronyms/abbreviations)
- Implemented as a modifier linter `SpellCheckIgnoreAllCaps` that configures existing `SpellCheck` behavior
- Supports ALL CAPS words with common suffixes: `'s`, `'d`, `ed`, `s`, `es` (e.g., APIs, CPUs, API'd, GPSes)
- Uses existing config structure (users toggle via `"SpellCheckIgnoreAllCaps": true` in their config)

# How Has This Been Tested?
- Unit tests have been added
- Integration tests added

# Checklist


harper-core/src/linting/spell_check.rs
```diff
@@ -24,11 +25,48 @@ impl<T: Dictionary> SpellCheck<T> {
...
+    fn is_all_caps(&self, word: &[char]) -> bool {
+        if word.len() <= 1 {
+            return false;
+        }
+
+        let word_str: String = word.iter().collect();
+        let word_str = word_str.as_str();
+
+        // Check for allowed single suffixes only: 's, 'd, ed, s
+        let suffixes = ["'s", "'d", "ed", "s"];
```
hippietrail: Worth including "es" too?

ethqnol: Oh true, perhaps like "GPSes" for example?

hippietrail: > Oh true, perhaps like "GPSes" for example?

Yeah I'm pretty sure I write "SMSes".

elijah-potter:
What is the consensus? Should this be enabled by default?

ethqnol:
> What is the consensus? Should this be enabled by default?

I can see an argument for both. I'd maybe say disabled by default? That way it doesn't interfere with people's current Harper configs.

ethqnol:
> > What is the consensus? Should this be enabled by default?
> 
> I can see an argument for both. I'd maybe say disabled by default? That way it doesn't interfere with people's current Harper configs.

@elijah-potter what do you think?

harper-core/src/linting/lint_group.rs
```diff
@@ -580,9 +604,17 @@ impl Linter for LintGroup {
     fn lint(&mut self, document: &Document) -> Vec<Lint> {
         let mut results = Vec::new();
 
+        // Check for spell check flags before running linters
```
elijah-potter, line 671: This is another case where we're adding specialized logic to the `LintGroup`, which should be general. Could we set it by passing a boolean into the `LintGroup::curated()` method call? We could then expose config to the user as an actual checkbox (or similar).

I apologize for the back-and-forth. Thanks for responding to feedback so quickly!

ethqnol: Hi! Thanks for your response! I'll see if I can implement that. Do note that I have not looked into the frontend settings config code; is there a checkbox (or similar) feature in that pipeline already?

Anyways I'll take a look and get back to you ASAP. I've got some important tests this week so it might be a bit delayed.

elijah-potter:
@ethqnol, any updates on this? If not, I'm tempted to close this or convert it to draft.

PR #1899, opened 2025-09-11
feat(#1897): transposition costs one instead of two

### LINKED ISSUE #1897: Use Damerau–Levenshtein distance to correctly score character swaps
As noted in #1411, a lot of correction suggestions for words with swapped characters don't contain the correct variant.

**Proposed Solution**
By using the [Damerau-Levenshtein Distance](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance) as a scoring function, you can solve all these cases *and* remove special-cased logic for words like 'the' and 'ie' vs 'ei' swaps.

**Examples**
|word a|word b|levenshtein|damerau-levenshtein|
|-------|-------|------|-----|
|the|teh|2|1|
|recieved|received|2|1|

**Component**
[x] Core engine

**Additional Context**
I've had a quick peek at the code, and noticed you use some fancy crate that uses automata to quickly calculate the distances.
Either you have to propose the change upstream, fork the crate and implement the new metric yourself, or just remove the dependency and use the regular memoized implementation of the algorithms.

grantlemons:
# Issues 
Closes #1897

# Description
Consider the cost of a transpositon (swap) of two characters to be one instead of two.
This is a boolean when constructing the Levenshtein Builder.

# How Has This Been Tested?
Added tests to verify expected behavior from the two versions of the Levenshtien Builder.

# Checklist

### DIFF
--- a/harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
+++ b/harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
@@ -118,8 +118,8 @@ Message: |
       69 | curtseying as you’re falling through the air! Do you think you could manage it?)
 Suggest:
   - Replace with: “curtsy”
+  - Replace with: “courtesy”
   - Replace with: “curse”
-  - Replace with: “cursed”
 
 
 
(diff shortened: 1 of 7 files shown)


elijah-potter:
Thanks for taking care of that, @grantlemons. I'm glad it was that easy. 

I fixed the failing tests. For future context, the `test_most_lints` "tests" is a series of snapshot tests. To regenerate the snapshots, all you need to do is run `cargo test`. It will "fail", but you can run it a second time to confirm that the snapshots were overwritten.

PR #1909, opened 2025-09-11
feat(chrome-ext): show hints randomly

elijah-potter:
# Description

There are lots of subtle ways Harper helps me out in the browser. I want our users to be aware of them. This PR adds a small hint section to the suggestion popup. It only shows up sometimes, and the hint is selected randomly from the `hints.json` file.

Let me know if there are any hints which should be added.

# Demo

image

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/harper-core/tests/text/linters/The Constitution of the United States.snap.yml
+++ b/harper-core/tests/text/linters/The Constitution of the United States.snap.yml
@@ -1156,7 +1156,7 @@ Message: |
 Lint:    Formatting (63 priority)
 Message: |
      455 | Oath or Affirmation:-- "I do solemnly swear (or affirm) that I will faithfully
-         |                     ^~ A sequence of hyphens is not an en dash.
+         |                     ^~ Replace these two hyphens with an en dash (–).
 Suggest:
   - Replace with: “–”
 


ethqnol:
Here are a couple of ideas!

- Use an em-dash (---) to set apart a phrase or indicate an abrupt change in thought.
- Modern style guides recommend using only one space after a period, not two.
- For a true ellipsis, use the dedicated character (…) instead of three separate periods (...).
- Use a hyphen (-) to join words into a single concept, like 'state-of-the-art'.
- To make text bold in Markdown, surround it with two asterisks (**text**). For italics, use one asterisk (*text*).
- You can create a blockquote for quoting text by starting the line with a > symbol.
- To create a hyperlink in Markdown, use the format [link text](https://example.com).
- Organize your document with headings by starting a line with one or more hash symbols (#).
- Don't agree with a suggestion? Click 'Ignore' and Harper will learn from your writing style.
- You can add specialized terms, names, or acronyms to your personal dictionary in Harper's settings.
- Check your language preferences in Settings to get suggestions for different dialects (e.g., American vs. British English).

hippietrail:
image
This example reminds me that the `message` for the hyphen vs. em dash linter says something like "Two hyphens is not an em dash", which can be interpreted as saying the opposite. I would change that linter to include a friendlier/more helpful message as part of this PR.

Which reminds me of another kind of "hint", that linters could optionally be able to include a longer text explaining why they're flagging this lint and/or why they're suggesting this replacement.

hippietrail:
> Here are a couple of ideas!
> 
> * For a true ellipsis, use the dedicated character (…) instead of three separate periods (...).

Even as a lover of typography and Unicode I find this too opinionated. To me they're both true ellipses. The single-character one probably got into Unicode because one of the legacy encodings already encoded it separately? I haven't double checked but I read that the Chicago Manual of Style recommends spaced ellipses - you can see these in our testing snapshot for The Great Gatsby.

> * Use a hyphen (-) to join words into a single concept, like 'state-of-the-art'.

This is more subtle in reality. It depends on the part-of-speech of the compound. I believe it's always true when a phrase is used as an attributive adjective. I'm slightly less confident that it's never done for verbs. But whether it's done for nouns or phrases used predicatively seems to depend on the individual word/phrase. Further research is needed.

> * To make text bold in Markdown, surround it with two asterisks (**text**). For italics, use one asterisk (_text_).

Neat idea! We could make ones like these depend on the file format! By the way, for italics you stated asterisks but showed underscores.

> * Don't agree with a suggestion? Click 'Ignore' and Harper will learn from your writing style.

"Learn" might be too strong in this era of AI. Perhaps "adapt to" or "comply with"?

> * You can add specialized terms, names, or acronyms to your personal dictionary in Harper's settings.

The same or a separate tip should mention the per-file dictionaries. I forget if we ended up adding support for per-project/workspace dictionaries?

> * Check your language preferences in Settings to get suggestions for different dialects (e.g., American vs. British English).

Great idea! This makes me think the spellcheck linter should be able to check if a failed word would be correct in another dialect and advise on how to change dialects. Even better if it could go ahead and change it for you too!

elijah-potter:
> The same or a separate tip should mention the per-file dictionaries. I forget if we ended up adding support for per-project/workspace dictionaries?

This doesn't apply to the Chrome extension.

@ethqnol I like a lot of your suggestions. I've added them. I don't think we should include hints that are specific to Markdown, since the Chrome extension often runs in contexts that don't use Markdown.

PR #1916, opened 2025-09-12
test(core): examples from #1902

### LINKED ISSUE #1902: "To" wrongly flagged to change to "too" before an adverb that precedes a present tense verb.
**What got flagged?**

Image

We do want to flag "I ate to quickly". But "to" can be used with the lemma form of a verb (aka "present tense") even with an intervening adverb: "I have to quickly do something".

The verb properties in `WordMetadata` and `TokenKind` should now be mature enough to make this easy to implement.

Note that "to" can't be used before verbs in forms other than the lemma and also can't be used before modal verbs. The `kind_is_but_is_not` helper methods should make this possible but note that not all auxiliary verbs are modal verbs, so maybe use the `ModalVerb` `Expr` instead...

(Off the top of my head so some references to code might be slightly wrong.)

elijah-potter:
# Issues 

Closes #1902 

# Description

I've added tests from #1902 to try to replicate the problem. They pass, but I'm not sure why. Has the issue somehow been resolved?

### DIFF
--- a/harper-core/src/linting/to_two_too/mod.rs
+++ b/harper-core/src/linting/to_two_too/mod.rs
@@ -393,4 +393,20 @@ mod tests {
             ToTwoToo::default(),
         );
     }
+
+    #[test]
+    fn no_lint_to_quickly_find_a_factory() {
+        assert_no_lints(
+            "To quickly find a factory, look for a map.",
+            ToTwoToo::default(),
+        );
+    }
+
+    #[test]
+    fn no_lint_llm_as_judge_to_automatically_score() {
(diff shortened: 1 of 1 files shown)


hippietrail:
> I've added tests from #1902 to try to replicate the problem. They pass, but I'm not sure why. Has the issue somehow been resolved?

It must've been! I'm building from HEAD all the time so I must've hit that just before the too/to changes. I can now verify it doesn't happen. I had a quick look but it's 1am here and there's a lot of too/to files so I'm going to call it fixed.

PR #1922, opened 2025-09-13
fix(#1876): merge dialects from curated, file, user, and workspace dicts

### LINKED ISSUE #1876: Not recognizing "Calibre" or "calibre" even after adding to neovim spell file
**What got flagged?**
Calibre or calibre

**Why is this incorrect?**
Took created by Kovid Goyal and I added it to my spell file

**Example of correct usage:**
See min 8:11 in this video:
https://youtu.be/xTFAbuvcF0A?t=491&si=q_3Vqr1ZSyJUJBat

grantlemons:
# Issues 
Closes #1876

# Description
- Use the `WordMetadata::or` method to combine the metadata of words in `MergedDictionary`
- Add the user's configured dialect to their user, workspace, and file dictionaries in `harper-ls`
- Add test in `harper-core/linters/spell_check.rs`

# How Has This Been Tested?
Created unit test and built and tested on an actual file.

# Checklist

### DIFF
--- a/harper-core/src/spell/dictionary.rs
+++ b/harper-core/src/spell/dictionary.rs
@@ -1,4 +1,5 @@
 use blanket::blanket;
+use std::borrow::Cow;
 
 use super::FuzzyMatchResult;
 use super::WordId;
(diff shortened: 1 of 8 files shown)


elijah-potter:
This also solves the issue I've had with the word `Matt`.

harper-core/src/spell/dictionary.rs
```diff
@@ -33,11 +33,11 @@ pub trait Dictionary: Send + Sync {
     ) -> Vec<FuzzyMatchResult<'_>>;
     fn get_correct_capitalization_of(&self, word: &[char]) -> Option<&'_ [char]>;
     /// Get the associated [`WordMetadata`] for any capitalization of a given word.
-    fn get_word_metadata(&self, word: &[char]) -> Option<&WordMetadata>;
```
elijah-potter, line 36: Is this PR possible without changing this? Cloning `WordMetadata` may become more expensive in the near future.

grantlemons: I don't think so. `Metadata::or` creates a new Metadata object that is owned by `MergedDictionary::get_word_metadata`, so we can't return a reference.

grantlemons: We could return a `Cow`?

grantlemons: I changed it to `Cow`

elijah-potter (approved):
This is a more elegant solution than I could have come up with. Thanks @grantlemons!

PR #1928, opened 2025-09-14
chore: curate dict; improve POS tag snapshots

hippietrail:
# Issues 
N/A

# Description

- Curation of mostly noun, verb, and adjective annotations.
- POS snapshots didn't consider that words like "wood"/Wood"
  can be common noun, proper noun, mass noun, and countable
  noun all at once, so logic that simplified that output failed to
  denote singular for words that were proper _and_ common.
- Denote _probable contractions_ in the POS snapshots with `K`.
  There isn't an annotation flag for contractions and they should
  probably be handled by splitting them into two tokens eventually.
  But for now, we check the Orthography flags and if the dictionary
  entry includes an apostrophe we denote it as a contraction so
  there'll be false positives for things like "Baha'i". But it helps find
  words missing from the dictionary or wrongly annotated since
  most words the POS snapshots denoted with `W?` actually are
  in the dictionary but are contractions with no annotations.

# How Has This Been Tested?

Comparing the old POS snapshots against the new ones.

# Checklist


hippietrail:
> Looks good to me. I'll think about whether we truly want to split contractions or if there's an alternate solution.

Yeah I was thinking of other ways to do it but from the bit of checking I did that seems to be the standard or most common approach. But I'm keen to see what other solution you might come up with!

Issue #1878, opened 2025-09-07
Simple Harper PWA


jlarmstrongiv:
**What problem does this solve?**
Some people just want to use Harper as a standalone [PWA](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps), not install a chrome or vscode extension.

**Proposed Solution**
Just a simple standalone webpage with the harper editor plus the [`vite-plugin-pwa`](https://www.npmjs.com/package/vite-plugin-pwa).  Nothing else.

**Examples**
Many other providers offer standalone apps:
- https://hemingwayapp.com/desktop
- https://prowritingaid.com/download/desktop

By offering a simple PWA, Harper can have a similar standalone app experience without worrying about desktop app code signing, etc.

**Component**
[ ] Core engine
[x] Plugin/Extension
[ ] Other: _____

**Additional Context**
More about PWAs https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps

I usually just copy and paste into the editors for corrections.  I don’t need file management, etc.

elijah-potter:
We have a version of our website's demo available standalone at https://writewithharper.com/editor

I don't see why we couldn't hook up `vite-plugin-pwa`. I am a little hesitant to do it and call it day, since I'd prefer to provide a more comprehensive experience. Would a Tauri app make sense, since it would leave room for hooking into an accessibility API?

KraXen72:
@elijah-potter please let me know if i should open a new issue about this, but please add line wrapping to writewithharper.com/editor. I pasted in a long note of mine to check for grammar and since the lines are so long, the underlines de-synced from the actual text.

the note is 2.7K words, 14K characters, in case it matters.

elijah-potter:
@KraXen72, I'm looking into setting up the website's editor to use the same backend as the Chrome extension, so the demo more accurately represents the real experience. If we swapped out the demo text editor with, say, [ProseMirror](https://prosemirror.net/), would it still suit your use-case?

KraXen72:
oh yeah, definitely.

i've found the chrome extension not very suitable for me since i don't need it that often, but every once in a while when i write something longer-form, i'd love to paste it into the harper demo editor, and quickly fix up the mistakes once i'm out of the writing phase and in the editing phase.

[...]

Issue #1877, opened 2025-09-07
Use curly, not straight quotes


jlarmstrongiv:
**Describe the bug**
Haper should use “curly” or “typographer’s” quotes, not "straight" or "dumb" quotes in suggestions.

**To Reproduce**
1. Go to https://writewithharper.com/
2. Click on the second suggestion, Word Choice - “That s”
3. Click the button Replace "That s" with "That's"
4. Harper erroneously uses straight quotes, not curly qutoes

**Expected behavior**
Haper should use “curly” or “typographer’s” quotes, not "straight" or "dumb" quotes.

**Screenshots**

<img width="778" height="389" alt="Image" src="https://github.com/user-attachments/assets/8d70627a-f787-4467-8351-c64ce53dba81" />

**Platform**
Only tested in Chrome

**Additional context**
https://www.typewolf.com/cheatsheet

hippietrail:
It should definitely be an option. Same for apostrophes and single quotes, which can be tricky for matching pairs to decide which should be `‘` vs `’` since ASCII apostrophe and single quote are the same character.

elijah-potter:
This should be fixed by #1920.

jlarmstrongiv:
@elijah-potter is that enabled in the web and standalone editors?  Or is this option only enabled when linting programmatically?

elijah-potter:
Once we've pushed out another release, it should be enabled on all platforms.

elijah-potter:
Oh, I think I may have misinterpreted what you were asking about. You mean in the replacements, not the message. I need to revisit this.

hippietrail:
Related: #2220

hippietrail:
🌲

PR #1942, opened 2025-09-15
feat(web): apply feedback from #1878

elijah-potter:
# Issues 

Addresses feedback from #1878 

# Description

In #1878, some additional features were requested from the website's editor. Namely:

1. Ignoring lints
2. Jumping to a selected lint
3. Expanding / collapsing all lints in the sidebar

I've also added a button to ignore all lints.

# Demo

image

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/packages/web/src/lib/Editor.svelte
+++ b/packages/web/src/lib/Editor.svelte
@@ -115,9 +115,6 @@ function toggleCard(i: number) {
 		const ns = new Set(openSet);
 		ns.add(i);
 		openSet = ns;
-		if (lints[i]) {
-			jumpTo(lints[i]);
-		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


elijah-potter:
@jlarmstrongiv, should ignored lints persist between sessions? I think not, at least for the demo.

elijah-potter:
Demo here: https://writewithharper.fly.dev/

jlarmstrongiv:
Love it!  The collapse/expand all is fantastic.

I also would not expect ignoring lints to persist between sessions.  

Though, for ignore, I would keep it consistent in the popovers and sidebar.  Both would have say the suggestions on the right and the ignore button on the left.

Also, how does the jumping to a selected lint work?

I am curious about other options like straight vs curly quotes https://github.com/Automattic/harper/issues/1877#event-19668715516

Anyway, I really appreciate these changes!  They really improve my experience using Harper.

elijah-potter:
> Also, how does the jumping to a selected lint work?

Are you asking how to make it happen, or how the code works?

From a technical standpoint, we're just using a `Range` to select the error. When the selection changes to something off-screen, browsers scroll to bring it into view.

PR #1943, opened 2025-09-15
feat(core): add rule for unnecessary double-negatives

elijah-potter:
# Description

I've added a rule that looks for common double-negatives and corrects them to encourage clarity. I'm mostly interested if this is reasonable from a philosophical perspective. _Should_ Harper include this?

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -280,7 +280,7 @@ pub fn lint_group() -> LintGroup {
             LintKind::Eggcorn
         ),
 
-        // General litotes (double negatives) → direct positive
+        // General litotes (double negatives) → direct positive suggestions
         "LitotesDirectPositive" => (
             &[
                 ("not uncommon", "common"),
@@ -298,9 +298,9 @@ pub fn lint_group() -> LintGroup {
                 ("not unreasonable", "reasonable"),
                 ("not impossible", "possible"),
             ],
-            "Prefer a direct form over litotes like `not uncommon`.",
-            "Rewrites common double negatives (litotes) to a clearer direct positive.",
-            LintKind::WordChoice
+            "Consider a direct form instead of litotes like `not uncommon`.",
+            "Offers direct-positive alternatives when double negatives might feel heavy.",
+            LintKind::Style
         ),
 
         // Redundant degree modifiers on positives (double positives) → base form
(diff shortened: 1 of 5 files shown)


harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -279,6 +279,42 @@ pub fn lint_group() -> LintGroup {
...
+                ("not unclear", "clear"),
+                ("not irrelevant", "relevant"),
+                ("not unpredictable", "predictable"),
+                ("not inadequate", "adequate"),
+                ("not unpleasant", "pleasant"),
+                ("not unreasonable", "reasonable"),
+                ("not impossible", "possible"),
+            ],
+            "Prefer a direct form over litotes like `not uncommon`.",
+            "Rewrites common double negatives (litotes) to a clearer direct positive.",
```
hippietrail: The message should be less opinionated since these are not mistakes but stylistic choices. Perhaps something like "Consider..." etc.

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -279,6 +279,42 @@ pub fn lint_group() -> LintGroup {
...
+            LintKind::WordChoice
+        ),
+
+        // Redundant degree modifiers on positives (double positives) → base form
+        "RedundantSuperlatives" => (
+            &[
+                ("more optimal", "optimal"),
+                ("most optimal", "optimal"),
+                ("more ideal", "ideal"),
+                ("most ideal", "ideal"),
```
hippietrail, line 313: I think I or maybe somebody else made a lint for "more preferable". Yes I can see it now (-: Maybe these should be rolled in together. 

Image

harper-core/src/linting/phrase_set_corrections/mod.rs
elijah-potter: Done.

hippietrail (approved):

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -279,6 +279,43 @@ pub fn lint_group() -> LintGroup {
...
+                ("not infrequent", "frequent"),
+                ("not inaccurate", "accurate"),
+                ("not unclear", "clear"),
+                ("not irrelevant", "relevant"),
+                ("not unpredictable", "predictable"),
+                ("not inadequate", "adequate"),
+                ("not unpleasant", "pleasant"),
+                ("not unreasonable", "reasonable"),
+                ("not impossible", "possible"),
+                ("more preferable", "preferable"),
```
ccoVeille, line 300: Here are some examples I found, feel free to consider or ignore them.

not unconditionally 
not unneccessarily 
not unhelpful
not unmodified
not unlimited
not unreasonable
not unlike
not unexpected
not unkindly
not unhappy
not incorrect
not incorrectly

These are examples I found in open-source projects

Issue #1898, opened 2025-09-10
harper-ls crash when using harper-typst


rnxpyke:
**Describe the bug**
When editing typst documents, I consistently trigger crashes in the language server.

Here is the log output for my reproduction case:
```
Server initialized!
2025-09-10T20:25:57.572439Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:57.572554Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.158761Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.158837Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.349549Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.704354Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.704460Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.704485Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.716909Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.787026Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.853239Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.853303Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.853342Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:58.870311Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:59.172042Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:59.248271Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:59.571447Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
2025-09-10T20:25:59.571554Z ERROR harper_ls::backend: Settings must contain a "harper-ls" key.
[...]

PR #1948, opened 2025-09-16
test: add failing example for markdown comment parser

rnxpyke:
# Issues 
something similar to #1898 seems to also be an issue with the markdown parser.

# Description
This is a minified version of a new fuzzing run failure on the markdown comment parser.

# How Has This Been Tested?
This is the test :)

# Checklist

### DIFF
--- a/harper-comments/src/comment_parsers/jsdoc.rs
+++ b/harper-comments/src/comment_parsers/jsdoc.rs
@@ -145,15 +145,21 @@ fn parse_inline_tag(tokens: &[Token]) -> Option<usize> {
         return None;
     }
 
+    if tokens.len() <= 3 {
+        return None;
+    }
+
     let mut cursor = 3;
 
-    while !matches!(
-        tokens.get(cursor),
-        Some(Token {
-            kind: TokenKind::Punctuation(Punctuation::CloseCurly),
-            ..
-        })
-    ) {
+    while cursor < tokens.len()
+        && !matches!(
+            tokens.get(cursor),
+            Some(Token {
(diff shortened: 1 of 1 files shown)


elijah-potter:
I've gone ahead and pushed a possible fix. Would you mind doing a bit more fuzzing to confirm?

rnxpyke:
@elijah-potter the harper-core markdown parser seems fine, at least after 10_000_000 fuzzing iterations.
I added a fuzzing target just for this case in the fuzzing PR.
however, i found another markdown comment issue, which is a problem at least for the java parser.
You can see it in the new commit I attached.
If you want to merge the fix as is, we can move this java case to another issue/PR.

harper-comments/src/comment_parser.rs
```diff
@@ -123,3 +123,28 @@ impl Parser for CommentParser {
...
+    use super::CommentParser;
+    use harper_core::parsers::{MarkdownOptions, StrParser};
+
+    #[test]
+    fn hang() {
+        use std::sync::mpsc::channel;
+        use std::thread;
+        use std::time::Duration;
+
+        let (tx, rx) = channel::<()>();
```
elijah-potter, line 134: Why are you doing it inside a seperate thread?

rnxpyke: because this is a timeout issue. the seperate thread is to not stall CI forever due to infinite-loops or similar.

elijah-potter: Ah.

elijah-potter:
I finally had a chance to go in and fix this problem. Thanks again for reporting it!

PR #1949, opened 2025-09-16
fuzz: add example fuzzing targets

rnxpyke:
# Description
this PR adds `cargo-fuzz` fuzzing targets for harper parsers.
Ideally, we wouldn't just fuzz the parser, but also the linting step with the resulting tokens to cover the `harper-core` code.

Notice that I had to turn of  `lto = true` in the workpace Cargo.toml, so maybe we want to figure out how fuzzing can coexist with lto before merging.

Related: #1898 #1948 

# Demo
Example of fuzzing run finding a crash: https://asciinema.org/a/ShdOcwthkwwn72JyGpRk9emOu

# Checklist


rnxpyke:
hm, seems like we can use an ENV var in the fuzzing call: https://github.com/rust-fuzz/cargo-fuzz/issues/384

rnxpyke:
@elijah-potter would be nice to know what you think about this approach

PR #1964, opened 2025-09-18
feat:many more variants of "for/to all intents and purposes"

hippietrail:
# Issues 
N/A

# Description

I happened upon "all **extents** and purposes" somewhere and was going to add it to the existing phrase correction for "all **intensive** purposes". But while researching I found out that there are a bunch of other variants out in the wild.

I also learned that "to all intents and purposes" is as legit as "for all intents and purposes" and is actually the original variant. It's mostly now a UK variant so I could make this linter dialect-aware if anybody thinks that's a good idea.

Other variants include using the wrong preposition and words including "intense", "intensive", "intended" and "intrinsic", and omitting the "and".

Oddly, this flags "to all Intents and Purposes" in the US Constitution snapshot even though that's a legit accepted variant, all `Expr`s used are case insensitive, and the whitespace `Expr` should accept linefeeds, which seems to be the trigger in this case.

# How Has This Been Tested?

A bunch of new unit tests are added, covering as many variants as I could find in the wild in GitHub, StackOverflow, and Slashdot.

# Checklist


harper-core/tests/text/linters/The Constitution of the United States.snap.yml
```diff
@@ -1637,6 +1637,18 @@ Message: |
 
 
 
+Lint:    Nonstandard (50 priority)
+Message: |
+     668 | proposing Amendments, which, in either Case, shall be valid to all Intents and
+         |                                                             ^~~~~~~~~~~~~~~~~~~
+     669 | Purposes, as Part of this Constitution, when ratified by the Legislatures of
+         | ~~~~~~~~ The time period is redundant because it repeats what's already specified by the AM/PM indicator.
```
elijah-potter, line 1645: I feel like this is OK, since it's weird and rare capitalization.

elijah-potter (approved):

harper-core/tests/text/linters/The Constitution of the United States.snap.yml
hippietrail: > I feel like this is OK, since it's weird and rare capitalization.

It's not the capitalization that's causing it though. It's the line break. I know because I added a test case with a `\n` for the whitespace in that spot.

But line breaks should be treated as whitespace right?

PR #1977, opened 2025-09-21
feat:fell→feel

hippietrail:
# Issues 
N/A

# Description

"Fell" is a very common typo for "feel". Seemingly even more among non-native English speakers.
I expected it would be more common as a typo for "felt" but I found no evidence for that.

There's a couple of words which signal the error if they come before "fell" and a larger set of words that signal it if they come after "fell".

There's a couple of edge cases, hopefully more false negatives than false positives. Please report any false positives. The error `message` tries not to be too opinionated, just in case.

# How Has This Been Tested?

I added at least one unit test for each variant I could find, all using sentences from GitHub.

# Checklist


harper-core/src/linting/feel_fell.rs
```diff
@@ -0,0 +1,248 @@
...
+use crate::linting::{ExprLinter, Lint, LintKind, Suggestion};
+
+pub struct FeelFell {
+    expr: Box<dyn Expr>,
+}
+
+impl Default for FeelFell {
+    fn default() -> Self {
+        let with_word_before = SequenceExpr::default()
+            .then_word_set(&["didn't", "doesn't"])
```
elijah-potter, line 14: Could we include `does not` and `did not`?

hippietrail: > Could we include `does not` and `did not`?

I actually did extensive searches with `not` and for some reason nothing useful turned up.

Issue #2005, opened 2025-09-25
False positive: "to" in "all the way to advanced usage" wrongly flagged to change to "too"


hippietrail:
**What got flagged?**
The text that was incorrectly flagged.

<img width="374" height="68" alt="Image" src="https://github.com/user-attachments/assets/43eab5fb-98d3-4355-a336-45b0a778d095" />

> all the way to advanced usage, like

**Why is this incorrect?**
The "to" in this snippet is about the end point of "all the way", not a qualifier for "how advanced".

hippietrail:
This is probably the same thing: 

<img width="784" height="69" alt="Image" src="https://github.com/user-attachments/assets/d15b37a3-0d42-4622-9256-7a8648a03a8e" />

> you'll have access to over 400 integrations and a

elijah-potter:
This was fixed by #2022

PR #2022, opened 2025-09-26
fix(core): address problems with prepositional phrases

### LINKED ISSUE #2018: False positive: "too" incorrectly flagged to change to "to"
**What got flagged?**

Image

> we may have gotten a bit too theme-happy. We have over one

**Why is this incorrect?**

"theme-happy" is a compound adjective, and adjectives can follow "too".

The current heuristics probably check for a single noun following the "too" to determine it should be "to".

elijah-potter:
# Issues 

Closes #2018

# How Has This Been Tested?

Additional unit tests

# Checklist


elijah-potter:
Also fixed #2005

PR #2023, opened 2025-09-26
alskda

elijah-potter:
# Description

Introduces a new rule that detects missing nouns when they're implied by a determiner.

# How Has This Been Tested?

Additional unit tests.

# Checklist


harper-core/src/linting/determiner_without_noun.rs
```diff
@@ -0,0 +1,79 @@
...
+use super::{ExprLinter, Lint, LintKind, Suggestion};
+
+pub struct DeterminerWithoutNoun {
+    expr: Box<dyn Expr>,
+}
+
+impl Default for DeterminerWithoutNoun {
+    fn default() -> Self {
+        let expr = SequenceExpr::default()
+            .then(|tok: &Token, _: &[char]| tok.kind.is_determiner())
```
hippietrail, line 14: You'll want to weed out determiners that are also pronouns, especially "this" and "that".
I can't remember if "my" etc. are marked as determiners. They should be and I think they are. If so a couple of those are also pronouns, I think "her" and "its" off the top of my head. `.kind_is_but_is_not()` might suffice to address this.

elijah-potter:
I actually feel like those aren't exceptions to this rule. Could you give me a specific sentence that _is_ flagged, but shouldn't be?

PR #2028, opened 2025-09-28
Criteria phenomena

TomasoAlbinoni:
# Issues 

# Description
New linter that checks if 'criteria' or 'phenomena' is used as singular.

# Demo

# How Has This Been Tested?
Using just lint filename

# Checklist

### DIFF
--- a/harper-core/src/linting/criteria_phenomena.rs
+++ b/harper-core/src/linting/criteria_phenomena.rs
@@ -4,7 +4,7 @@ use crate::expr::Expr;
 use crate::expr::SequenceExpr;
 use crate::linting::LintKind;
 use crate::patterns::WordSet;
-use crate::{CharStringExt, Lint, Lrc, Token, TokenStringExt};
+use crate::{Lint, Lrc, Token, TokenStringExt};
 
 /// Linter that checks if 'criteria' or 'phenomena' is used as singular.
 pub struct CriteriaPhenomena {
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/lint_group.rs
```diff
@@ -483,6 +484,7 @@ impl LintGroup {
         insert_expr_rule!(Months, true);
         insert_expr_rule!(MostNumber, true);
         insert_expr_rule!(MultipleSequentialPronouns, true);
+        insert_expr_rule!(CriteriaPhenomena, true);
```
hippietrail: It's not crucial but try to keep these in alphabetical order of Linter name. We've had a couple of times entries have become duped during merges after changing the order and ordering them makes it effortless to spot.

harper-core/src/linting/criteria_phenomena.rs
```diff
@@ -0,0 +1,120 @@
...
+                .to_ascii_lowercase();
+
+            if second_word == "criteria" {
+                suggestions.push(Suggestion::ReplaceWith("criterion".chars().collect()));
+            } else if second_word == "phenomena" {
+                suggestions.push(Suggestion::ReplaceWith("phenomenon".chars().collect()));
+            }
+        }
+
+        Some(Lint {
```
hippietrail, line 55: It's just a tiny nitpick but if something changes the linter in the future to have other than 3 tokens it would still return `Some(Lint)` just without any suggestions. I would include returning the lint insdide the if block. You can return `None` otherwise, or often it's possible to use `?` when checking the preconditions to obviate that...

harper-core/src/linting/criteria_phenomena.rs
```diff
@@ -0,0 +1,120 @@
...
+        let plural_words = Lrc::new(WordSet::new(&["criteria", "phenomena"]));
+
+        let singular_modifiers = Lrc::new(WordSet::new(&["this", "that", "a", "one"]));
+
+        Self {
+            expr: Box::new(
+                SequenceExpr::default()
+                    .then(singular_modifiers.clone())
+                    .then(
+                        SequenceExpr::default()
```
hippietrail: This `SequnceExpr` shouldn't  be needed. You can just chain the the two `.then...` inside it right after the first one.

TomasoAlbinoni:
> Some very minor nitpicks.

@hippietrail Thanks very much for the feedback!

harper-core/src/linting/criteria_phenomena.rs
```diff
@@ -41,20 +38,19 @@ impl ExprLinter for CriteriaPhenomena {
...
-        if matched_tokens.len() == 3 {
-            let second_word = matched_tokens[2]
-                .span
-                .get_content(source)
-                .to_string()
-                .to_ascii_lowercase();
+        let second_word = matched_tokens[2]
+            .span
+            .get_content(source)
+            .to_string()
```
hippietrail: I believe this will cause an allocation that can be avoided by using the string comparison methods available directly on the result of `get_content()`. (This was part of my original review but the Great Firewall caused a problem.)

hippietrail (commented):
One of my comments that got blocked by the Great Firewall.

elijah-potter (approved):
This looks great! I can imagine extending this down the road to cover even more plural confusions.

PR #2048, opened 2025-10-02
feat: add a bunch of noun-verb confusions

jvoisin:


harper-core/src/linting/noun_verb_confusion/mod.rs
```diff
@@ -10,9 +10,16 @@ pub(crate) const NOUN_VERB_PAIRS: &[(&str, &str)] = &[
     ("advice", "advise"),
     ("belief", "believe"),
     ("breath", "breathe"),
+    ("device", "devise"),
     ("effect", "affect"), // "Effect" is also a verb meaning "to bring about". "Affect" is a noun in psychology.
     ("emphasis", "emphasize"), // TODO how to handle "emphasise" as well as "emphasize"?
+    ("envelop", "envelope"),
     ("intent", "intend"),
+    ("pour", "poor"),
```
hippietrail, line 18: "Poor" is not a noun, it's an adjective which can occasionally be used as a noun in "the poor". As such the heurisitcs which are designed to disambiguate nouns and verbs, but not adjectives.

Also you've switched the columns. As you can see from the previous entries, it should be noun first and then verb. I'm surprised your new unit tests worked. Did you add new unit tests?

harper-core/src/linting/noun_verb_confusion/mod.rs
```diff
@@ -10,9 +10,16 @@ pub(crate) const NOUN_VERB_PAIRS: &[(&str, &str)] = &[
...
+    ("pray", "prey"),
+    ("resume", "resumé"),
+    ("waste", "waist"),
+    ("write", "rite"),
```
hippietrail, line 22: Except for "device" and "devise", these all have the verb first and the noun second, where it should be noun first and verb second. An easy way to remember is the name of the file as noun before verb.

"resume" as a noun has three spellings in English: `resume`, `résumé`, and `resumé`. Without further changes this module isn't designed to work with words that have multiple spellings. With my original "emphasis" vs "emphasize" I chose the most universal one and added a comment about the limitation. I don't think `resumé` is the most univerally accepted spelling in English but some research could prove otherwise.

hippietrail (changes requested):
These absolutely need unit tests. And from experience I strongly recommend finding examples in the wild rather than contriving examples which may not turn out to be representative. I always start by searching on GitHub.

PR #2072, opened 2025-10-09
fix(core): detect acronyms in `AnA`

### LINKED ISSUE #1702: False positive: "a LAN" is wrongly flagged to change to "an LAN"
Image

It seems the logic that determines whether an acronym/initialism/abbreviation is possibly pronounced as a word is deficient.

See also #1718

### LINKED ISSUE #1718: False positive: "a" flagged before "MAC address"
Image

"MAC" is universally pronounced as a true acronym, i.e. it's a homophone of "Mack". As such it takes the indefinite article "a".

See also #1702

86xsk:
# Issues 

Fixes #1718, Fixes #1702

# Description

Reduce incorrect lints from `AnA` by attempting to detect when an initialism is an acronym.

The logic assumes that the initialism is an acronym if its first two letters are not consonants. Though this seems to work for the cases covered in the tests, there must exist cases where this is too simple and not correct. It will likely need refinement as time goes on and issues are found.

# How Has This Been Tested?
- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/linting/an_a.rs
+++ b/harper-core/src/linting/an_a.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 
 use itertools::Itertools;
 
+use crate::char_ext::CharExt;
 use crate::linting::{Lint, LintKind, Linter, Suggestion};
 use crate::{Document, TokenStringExt};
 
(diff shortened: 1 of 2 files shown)


harper-core/src/linting/an_a.rs
```diff
@@ -191,6 +191,20 @@ fn starts_with_vowel(word: &[char]) -> bool {
     )
 }
 
+fn is_likely_acronym(word: &[char]) -> bool {
+    fn is_upper_vowel(c: char) -> bool {
+        matches!(c, 'A' | 'E' | 'I' | 'O' | 'U')
```
elijah-potter: Could you refactor this into a method on `CharExt` that allows us to query if a `char` is a vowel?

PR #2073, opened 2025-10-10
Add LaTeX support

### LINKED ISSUE #56: LaTeX Support

shreyasminocha:
# Issues 

Resolves #56.

# Description

Hacked together a quick MVP of a Harper LaTeX parser based on [TexLab's crate](https://github.com/latex-lsp/texlab/blob/master/crates/parser).

Early comments, suggestions, and pointers are welcome.

## Known issues

- [x] incorrect lint spans after multi-byte codepoints start appearing
- [x] a newline followed by two (indent) spaces in the middle of a sentence is treated as two spaces
- [x] there's a lint for replacing `---` but IMO it shouldn't apply to LaTeX since it's rendered as an em dash anyway
- [x] non-breaking~spaces have the wrong `TokenKind`s
- [x] \`\`quotation marks'' have the wrong `TokenKind`s
    - [x] single
    - [x] double
- [x] heading start markers
- [x] paragraph breaks
- [x] hrefs
- [ ] `\textit{word}s` is treated as [word, s]
- [ ] lints in comments
- [ ] …?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2861,6 +2861,7 @@ dependencies = [
  "itertools 0.14.0",
  "parser",
  "paste",
+ "regex",
  "syntax",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1793,7 +1793,7 @@ dependencies = [
 [[package]]
 name = "distro"
 version = "0.0.0"
-source = "git+https://github.com/latex-lsp/texlab#5a7c553eb3ed0db6bbd136cbd189300bf15c946f"
+source = "git+https://github.com/latex-lsp/texlab?tag=v5.24.0#5a7c553eb3ed0db6bbd136cbd189300bf15c946f"
 dependencies = [
  "anyhow",
  "rustc-hash 2.1.1",
@@ -2860,7 +2860,6 @@ dependencies = [
  "harper-core",
  "itertools 0.14.0",
  "parser",
- "rowan",
  "syntax",
 ]
 
@@ -4132,7 +4131,7 @@ dependencies = [
 [[package]]
 name = "parser"
 version = "0.0.0"
-source = "git+https://github.com/latex-lsp/texlab#5a7c553eb3ed0db6bbd136cbd189300bf15c946f"
+source = "git+https://github.com/latex-lsp/texlab?tag=v5.24.0#5a7c553eb3ed0db6bbd136cbd189300bf15c946f"
 dependencies = [
  "distro",
  "log",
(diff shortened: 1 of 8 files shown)


hippietrail:
I would say that comments should be linted, since that's what's explicity linted in programming languages. But I'm not sure which of the markup languages we support have comments and whether those are linted?

shreyasminocha:
How do you recommend I handle the french spaces and dash lints?

For the former, one solution might be to replace [newline, space+] in the harper tokens with just a single space token (with a span size of 1) but I'm not sure if that would cause any other issues.

The latter I don't think really applies to LaTeX either since `---` is semantically equivalent to `—`. I could do a string replace on the original document, but that would mess up the spans. I could also just replace three consecutive hyphen (harper) tokens that are preceded and followed by non-hyphens with an em dash token.

elijah-potter:
I'm sorry I didn't see this until now!

I think the appropriate solution for the french space problem would be to simply emit `Space(1)`. Leave the span size as-is. The repeated space linter reads the `LintKind` value, not the span content.

As for the `---` problem: In your parser, you can include a post-processing pass that converts any triple-hyphen sequence into an Em-dash token which would be ignored by the offending rule.

TL;DR: add a second pass over the parser that goes through and adjusts the output tokens to the semantically correct form. If the TeX code represents an em-dash, emit an em-dash. If it represents a single space, emit a single space.

lerboe:
@shreyasminocha what's the status on this? I can also help out if there is something I can help out with :) would love to try out harper with latex support

shreyasminocha:
planning to get to it some time this month!

rishubn:
Would really love this

shreyasminocha:
This is an awkward moment to realize that the TexLab crate is GPL3-licensed whereas Harper is Apache2-licensed. It seems like the two are [incompatible in this direction](https://www.apache.org/licenses/GPL-compatibility.html)?

I can see if we can swap it for the MIT-licensed [tree-sitter-latex](https://github.com/latex-lsp/tree-sitter-latex), which, like the TexLab crate, seems to be maintained by @pfoerster.

elijah-potter:
It really sucks that TexLab's license is incompatible, especially after you did all of that work. 

Rather than using a Tree-sitter parser, I've had a go at this problem with one I wrote myself. See #2689.

shreyasminocha:
> It really sucks that TexLab's license is incompatible, especially after you did all of that work.
> 
> Rather than using a Tree-sitter parser, I've had a go at this problem with one I wrote myself. See #2689.

All good, happy to see it on the verge of happening.

Issue #1945, opened 2025-09-16
Reduce binary size


jvoisin:
Currently, as of harper 0.64.0, the  harper-ls-aarch64-unknown-linux-gnu binary on my machine weights 48M, which is a tad enormous for a command-line utility processing text. It's more than:
- a (minimal) [Linux distribution (alpine-virt)](https://www.alpinelinux.org/downloads/)
- the typst binary, which is doing parsing, processing, complex rendering pdf generation, … and their binary even embeds fonts!
-  two times the size of the emacs binary

I know that "disk is cheap", but not everyone is using brand new hardware with terabytes of space, and even if it were the case, due to the rapid release schedule of harper, it ends up taking a significant amount of space on Linux distribution packages mirrors/archives. 

The fact that the [release page](https://github.com/Automattic/harper/releases/tag/v0.64.0) shows that the .tar.gz archive weights only ~7M hints that the dictionaries aren't compressed, which would be a quick win.

elijah-potter:
Most of the bloat comes from the sheer number of programming languages we support. For Typst alone, we need to bundle the entire Typst parser (which weighs ~500 KB). Our binary size is much larger than the source code because we pull a good number of dependencies to parse these programming languages.

Harper's dictionary resides with the `.text` section of the binary, which weighs about 5 MB. I might be able to compress that down.

In any case, I'll see what I can do to lighten the load on our distributors.

laundmo:
It could make sense to have a distribution without programming language support, for usecases which don't need it, such as perhaps the browser extension.

elijah-potter:
The browser extension does not include programming language support. I believe the total WebAssembly module size is just a megabyte or two.

laundmo:
Makes sense - so there already is a way to exclude that at compile time. Should mean its possible for a build to be made for other platforms which excludes them - unless thats not done using features but instead relies on conditional compilation checking for wasm?

elijah-potter:
The parsers for additional programming languages mostly reside in `harper-comments`. Our WebAssembly module consumes `harper-core` directly, and it doesn't include `harper-comments`. It would be trivial to expose each language as a feature on the crate. The relevant code is in `harper-comments/src/comment_parser.rs`. If anyone would like to open PR, go ahead.

laundmo:
> It would be trivial to expose each language as a feature on the crate. 

[...]

PR #2085, opened 2025-10-16
feat: create `harper-thesaurus`

86xsk:
# Description

Creates `harper-thesaurus`, laying the initial groundwork for providing synonyms in Harper.

This uses Moby Thesaurus II, a public domain thesaurus.

# TODO
- [x] Develop a way for callers to get a list of synonyms sorted by the frequency of their use. Hopefully making it easier to get and provide relevant synonyms.
- [x] Integrate with the `BoringWords` linter, providing suggestions for alternative words.
- [x] Explore methods for compacting thesaurus data.
  - [x] Compression
  - [x] Filtering out entries

# How Has This Been Tested?
- `cargo test`

# Checklist

### DIFF
--- a/.github/workflows/build_web.yml
+++ b/.github/workflows/build_web.yml
@@ -22,7 +22,6 @@ jobs:
       - uses: redhat-actions/buildah-build@v2
         with:
           image: web
-          layers: true
           containerfiles: |
             Dockerfile
           build-args: |


hippietrail:
Since Harper is a linter at heart it needs to begin with flagging a run of text. Do you envisage a thesaurus for changing any word to a synonym or for flagging certain specific words and then suggesting synonyms? So far the latter seems to fit Harper but not the former.

I've been toying with this myself for certain kinds of buzzwords and overused words that have become fashionable:

- "iconic" - mostly now used to mean "special" but sometimes has its real meaning.
- "diminish" - suddenly the hot new word for "reduce" and other words with similar meanings. Also still used in its original meaning.
- "uplift" - a verb almost the antonym of "diminish". I don't think it has a common original meaning. The adjective "uplifting" has always been common though.
- "utilize" / "utilise" - now almost always used as a fancy way of saying "use" though still sometimes seen it its correct sense in British writing.

Some require some heuristics to decide if they're OK or should be flagged. Or to show a different `message` depending on how strong the clues are that it's rightly/wrongly used. Sometimes we might be able to reword the context in a linter more than what a regular thesaurus.

The swear word replacement linter is kind of in this category too.

86xsk:
> Since Harper is a linter at heart it needs to begin with flagging a run of text. Do you envisage a thesaurus for changing any word to a synonym or for flagging certain specific words and then suggesting synonyms? So far the latter seems to fit Harper but not the former.

That's a good point. At the moment my goal with this is to create a basic backbone for a thesaurus and to try experimenting by integrating it with a linter. I'm hoping that will provide a better idea as to how practical/useful this feature would be.

Currently, I'm only envisioning it as providing a basic common thesaurus API that linters could call into to provide synonyms when they deem beneficial. (In the future, if there's interest in it, I could see `harper-ls` being able to provide synonyms for words on request too, much like other language servers can provide refactoring actions on demand. But as you mention, that might be out of scope for the project, and it's certainly outside the scope of this PR.)

One concern I currently have is the large size of the thesaurus. The source text file is ~24MB, and I'm not sure how workable and/or justifiable that is, especially if the feature isn't used much. I guess it might be possible to make it an optional feature, or perhaps to compress the file in some way too. I'd be curious to see how it affects the binary size and benchmarks once I can integrate it with a linter.

hippietrail:
> > Since Harper is a linter at heart it needs to begin with flagging a run of text. Do you envisage a thesaurus for changing any word to a synonym or for flagging certain specific words and then > One concern I currently have is the large size of the thesaurus. The source text file is ~24MB, and I'm not sure how workable and/or justifiable that is, especially if the feature isn't used much. I guess it might be possible to make it an optional feature, or perhaps to compress the file in some way too. I'd be curious to see how it affects the binary size and benchmarks once I can integrate it with a linter.

That's pretty big. Last time I looked at the size of the Harper curated dictionary it was less than 60,000 lines... in fact it's 754,147 bytes on disk right now. At the moment the dictionary has some amount of affix compression while I'll assume the thesaurus has no compression. But since they are text files presumably with words referencing other words, it shouldn't be too hard to come up with a format where most words appear as full strings only once and the rest are references as string slices. Also every word could just have a numeric ID and thence reference each other by ID, which may or may not be smaller than a string slice.
Also the thesaurus need not live in memory like the dictionary and could live just on disk. I'm sure it would be possible to implement it once without worrying about efficiency and then improve it with new schemes over time.

elijah-potter:
I adore the idea of doing stuff with a thesaurus, but (like @hippietrail), I'm concerned about the filesize. 

My initial reaction is that we could have synonym lookups be a separate feature from the linter and happen on-demand over the network? 

We could also provide multiple versions of the Harper binary. One would include only the linter with a limited dataset (call it `harper-ls-small`) and other would include the thesaurus (call it `harper-ls-large`). This would also give us space to add features like autocomplete.

I eagerly await your thoughts.

86xsk:
> My initial reaction is that we could have synonym lookups be a separate feature from the linter and happen on-demand over the network?
> 
> We could also provide multiple versions of the Harper binary. One would include only the linter with a limited dataset (call it `harper-ls-small`) and other would include the thesaurus (call it `harper-ls-large`). This would also give us space to add features like autocomplete.

Those are both very interesting ideas that I haven't considered. I'll certainly have to keep them in mind.

I've done some more development on this locally, and I've found that it increases the size of `harper-ls` by about the same size as the thesaurus file itself (~24MB). I'm curious to explore options regarding reducing the file size, perhaps by filtering out entries we're unlikely to need, or by compressing the file in some way.

In terms of shipping these changes, I guess my current development is more aligned with the second idea you mentioned, with the thesaurus being an optional feature that can be enabled for a build. (Though admittedly I haven't thought too much about that yet, as I've been more focused on just getting some decent output :laughing:)
image

Unfortunately I haven't been able to dedicate too much time to this as of late, hence progress has been somewhat slow. If there's interest, I could clean up and push the changes I have locally at this point. Though currently there's still the aforementioned quality of output issues, along with what seem to be some noticeable performance regressions (pending further testing).

elijah-potter:
I'd love to see what you have been cooking up. As this is a draft PR, I see no problem with your pushing up your work, even if it's unfinished.

elijah-potter:
Random thought: Instead of storing the full words to create the network, you could store the `WordId`. Since it is much smaller, it could compress the full thesaurus down by quite a bit. It would obviously have to be a compile step, however. I understand that would make things much more complex.

86xsk:
> Random thought: Instead of storing the full words to create the network, you could store the `WordId`. Since it is much smaller, it could compress the full thesaurus down by quite a bit. It would obviously have to be a compile step, however. I understand that would make things much more complex.

If my (slightly sleep-deprived) shell scripting is correct, it unfortunately doesn't look like it would provide too much space savings:
image

Currently I'm thinking of:
- Compressing/serializing the file in some way. Merely compressing the file with zip brings it down to about ~9MB, so I think there's some decent potential for savings there. Though whether it will be enough is another question.
- Removing entries that we might not need. For instance, by removing entries for words that don't exist in the curated dictionary. Now that I think about it, I think this should technically be possible to do at build-time too, so we can avoid making destructive changes to the thesaurus file.
- Rather than making the thesaurus an optional feature that's built-in to the binary, we could simply load a thesaurus file from disk at runtime, much like the user dictionary. However, I'm not sure how workable this would be for the non-native/web builds, since I'm not too experienced on that side of things.

(I still haven't done too much research into this to be honest, though it is my next step. I've been experimenting with trying to inflect the synonyms based on the source word, but I'm starting to feel like that might be better left for a future PR, in the interest of keeping this one reviewable.)

elijah-potter:
> Removing entries that we might not need. For instance, by removing entries for words that don't exist in the curated dictionary. Now that I think about it, I think this should technically be possible to do at build-time too, so we can avoid making destructive changes to the thesaurus file.

I think this is a great first step. We can also remove any entries we don't consider "common" according to the metadata.

86xsk:
I implemented compression of the thesaurus at build time, which reduces the size of the included thesaurus data to ~4MB.

I'll take a look at filtering out entries for words that don't exist in the curated dictionary next, and I should probably have that ready within a day or two.

86xsk:
Down to 3.18MiB by removing entries for words that don't exist in the curated dictionary. If that's still a little large, I could try filtering out synonyms that don't exist in the curated dictionary too.

86xsk:
I guess this is pretty much ready for review at this point. The quality of the provided synonyms can still be a little hit or miss, but I think the changes required to meaningfully improve that would be too large for this PR.

One thing I'm not sure about is whether the word-frequency sorting should be kept or not. Here are the current results with and without it:

With freq sorting
- **very**: awful,truly,genuine,powerful,just
- **interesting**: seductive,juicy,spicy,readable,attractive
- **several**: respective,diverse,discrete,numerous,various
- **most**: extreme,superior,principal,senior,much
- **many**: different,sundry,plenty,multiple,much

Without freq sorting
- **very**: just,genuine,especial,bona fide,awful
- **interesting**: irresistible,glamorous,flirtatious,coquettish,attractive
- **several**: divergent,distributional,dissonant,discrete,discordant
- **most**: maximum,maximal,genius,extreme,much
- **many**: affluent,sundry,plenty,multiple,much

It seems to provide a slight benefit in some cases, but I don't know if it's worth including another file. The dataset is also probably not ideal. It's just something I threw together from a small Wikipedia dump, since I couldn't find a permissively licensed word-frequency dataset. If anyone has any better ideas for a dataset, I'd certainly be open to trying it out too.

harper-core/Cargo.toml
```diff
@@ -50,3 +51,4 @@ harness = false
 [features]
 default = []
```
elijah-potter: I feel that the thesaurus flag should be enabled by default.

harper-thesaurus/src/thesaurus.rs
```diff
@@ -0,0 +1,120 @@
...
+            .collect()
+    })
+}
+
+pub struct Thesaurus {
+    /// Contains the words in the thesaurus and their corresponding synonyms, both as indices into
+    /// [`Self::deduped_word_set`].
+    entries: HashMap<usize, Vec<usize>>,
+    /// Contains (and holds ownership of) all words that occur in the thesaurus, deduped.
+    deduped_word_set: IndexSet<String>,
```
elijah-potter, line 34: Clever! Saves memory and speeds things up.

harper-thesaurus/build.rs
```diff
@@ -0,0 +1,65 @@
+#![warn(clippy::pedantic)]
```
elijah-potter, line 1: I really appreciate the build script. It is well-written.

harper-thesaurus/build.rs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+    } else {
+        // Compress without any filtering.
+        zstd::stream::copy_encode(reader, writer, zstd::zstd_safe::max_c_level())
+            .expect("Able to write compressed thesaurus");
+    }
+}
+
+/// Get all unique words contained in the curated dictionary.
```
elijah-potter: Harper's dictionary expands each entry into a number of inflected words. By reading it literally, you're missing a TON of entirely valid words. That results in a severely pared-down thesaurus (after you do your pruning).

Honestly, I don't think you need to prune the thesaurus based on the dictionary at all.

harper-core/src/dict_word_metadata.rs
```diff
@@ -752,6 +752,43 @@ impl DictWordMetadata {
         *self = self.or(other);
         self
     }
+
+    /// How different is this word from another?
+    pub fn difference(&self, other: &Self) -> u8 {
```
elijah-potter: Would it be possible to generate this function from within `generate_metadata_queries` macro? I noticed that some queries are missing.

elijah-potter (changes requested):
I'm so sorry it took me so long to review this. With the holidays and my support rotation... it fell under my radar. 

Overall, this PR is phenomenal. Really amazing work. I can't emphasize it enough. There are a few tweaks needed to perfect it, but it is very close. I've done rough versions of these tweaks on my copy and the suggestions are amazing. I can't wait to find places to use the thesaurus in more places. Thank you!

harper-thesaurus/build.rs
86xsk: Yeah, trying to reimplement dictionary parsing felt iffy from the start, it's not very DRY, it tries to read a file outside the crate, and it causes oversights like this. I initially tried to add `harper-core` and as build dependency and reuse the logic from there, but that causes a circular dependency.

The savings were smaller than expected too. I think I'll just remove the pruning code for the sake of keeping things simple.

86xsk:
~There seems to be a bug causing an infinite loop while running the tests. I'll take a look and I should probably have a fix later today or tomorrow. It likely has something to do with the changes in `replace_with_match_case`.~

fixed

elijah-potter (dismissed):
Fantastic revision. I had to go into the `Dockerfile` for the website and add `clang` as a dependency. If that causes issues for contributors' environments, we might need to rethink how the compression works. For now, I think it's fine to merge.

Thanks!

jvoisin:
There are two low-hanging fruits I think to reduce the size of thesaurus.txt:

1. Trim down the amount of synonyms. For example, the word `abandon` has around 350, it's ridiculous.
2. Construct at compile-time an id→word structure, and reference words by their id. This can even be compressed with lzma/bzip2 if needed/wanted.

86xsk:
For what it's worth, the thesaurus is currently being compressed at build-time (with Zstd) to 3.84 MiB.

I could probably take a look at reducing it further at some point or another if that's still problematic. The hard part is that such reductions are likely to come with complexity or reduced functionality. (For instance, removing synonyms might reduce the capability to provide the most appropriate synonyms based on context.)

Right now I think there's more to be gained by tinkering with dependencies and build options. A day or two ago, I was tinkering with the build profiles, and I found that adding...
```toml
panic = "abort"
lto = true
```
...to the release profile reduced the size of harper-ls from 60.56MiB to 53.53MiB. Not a lot, but the gain is still bigger than the size of the curated dictionary and the (compressed) thesaurus combined.

jvoisin:
7MiB is enough space to store Doom twice, I wouldn't call this "not a lot" :D

There is already #1945 open to track harper's size and suggest *diets*.

PR #2101, opened 2025-10-22
fix(chrome-ext): use Harper-specific HTML tag to avoid CSS conflicts

### LINKED ISSUE #2074: [Chrome Extension] Shadow DOM host causes visual artifacts on some sites
**Describe the bug**
Styles defined on a web page are able to target shadow DOM host element that is appended to the page body by the Harper Chrome extension.

Since the element in question is a `div`, it may become targeted with fairly generic selectors such as `body > div`.

https://github.com/Automattic/harper/blob/649d81cfadb74809b357d8d1be2308d82b5ea6a2/packages/lint-framework/src/lint/RenderBox.ts#L16-L17

**To Reproduce**
Open this page with Harper extension enabled.

```html
<!DOCTYPE html>
<html>
<head>
  <title>Demo</title>
  <style>
    body > div {
      background: rebeccapurple;
      display: flex;
      height: 100px;
      width: 100px;
    }
  </style>
</head>
<body>
  This is a demo.
</body>
</html>
```

A colored square appears at the center of the screen. Disabling the extension and refreshing the page hides this.

**Expected behavior**
Extension should not affect how the page is typically rendered.

**Screenshots**
Image

Image

**Platform**
Chrome Extension
Firefox Extension

**Additional context**
I originally noticed this on https://hacktoberfest.com/ , the example is adapted from the `body > div` selector used there. There the artifact is a white vertical line visible across the page.

I'm unsure if this can ever be fully eliminated – what stops someone from using even a more generic CSS selector such as `body > *` on a page to style everything? Something like `all: unset` or `all: initial` could be useful, but [some further investigation is probably in order](https://stackoverflow.com/questions/20727108/how-to-not-apply-a-webpages-css-on-the-div-i-am-inserting-into-it-via-content-s).

As a first line of defence, it makes sense to me that the shadow DOM host should be an element other than div, perhaps a custom element that makes it harder to target by tag name.

Also it's rather evident that the content script is also ran on pages where Harper is inactive. This probably merits creating another issue as ideally the content script shouldn't be affecting pages where Harper isn't allowed/needed, potentially lessening the amount of visual artifacts that appear in the wild.

elijah-potter:
# Issues 

Resolves #2074

# Description

See issue for problem details.

Almost all of Harper's UI is rendered inside a shadow DOM, which normally protects us from CSS conflicts. The _host_ of the shadow DOM, however, is not protected. To mitigate the problem, as @Ingramz suggested, I've assigned the host a custom HTML tag.

# How Has This Been Tested?

Manually on hacktoberfest.com

# Checklist

### DIFF
--- a/packages/lint-framework/src/lint/RenderBox.ts
+++ b/packages/lint-framework/src/lint/RenderBox.ts
@@ -13,7 +13,7 @@ export default class RenderBox {
 	private shadowHost: HTMLElement;
 
 	constructor(parent: Node) {
-		this.shadowHost = document.createElement('div');
+		this.shadowHost = document.createElement('harper-render-box');
 		parent.appendChild(this.shadowHost);
 	}
 


Ingramz:
Looks good to me.

I was wondering how `document.createElement` works on unrecognized element names and turns out it creates an instance of `HTMLUnknownElement`, so not a big deal.

PR #2104, opened 2025-10-23
fix(obsidian): problems when converting from Rust char indexes

### LINKED ISSUE #2098: Emojis in notes cause Obsidian plugin spell checker to mis-replace misspelled words due to incorrect text offset calculation
**Describe the bug**
In the Obsidian plugin, when a note contains multiple emojis, the spell checker mis-replaces misspelled words.  
It appears that the plugin’s text offset calculation does not correctly account for the extra space (multi-byte width) that emojis occupy.  
When several emojis appear before a misspelled word, their combined width shifts the internal cursor position. As a result, when accepting a correction, the replacement is applied at the wrong location — typically overwriting or inserting text a few characters away from the intended target.

---

**To Reproduce**
1. Open Obsidian and create a new note.
2. Type a line that includes **multiple emojis** before a misspelled word, for example:  
   `This is a 😊🚀🔥 smple test.`
3. Wait for the spell checker to detect “smple” and offer a correction (e.g., “simple”).
4. Accept the suggestion.
5. Observe that the correction is applied in the wrong place — not replacing “smple” but altering another word or inserting incorrectly, due to emoji spacing offsets.

---

**Expected behavior**
The spell checker should correctly replace the misspelled word (“smple” → “simple”), regardless of how many emojis or multibyte Unicode characters appear earlier in the same line.

---

**Screenshots**
Image

---

**Platform**
- App: **Obsidian**
- Plugin: _(insert plugin name here)_
- Obsidian version: **1.9.14**
- Operating System: **Linux**

---

elijah-potter:
# Issues 

Fixes #2098

# Description

When Harper's core engine (Rust) needs to interact with the ProseMirror editor embedded inside Obsidian, it needs to convert between Unicode scalar values and UTF-16 code points. The spans did not go along with this properly. I've gone ahead and reworked the conversion. 

# How Has This Been Tested?

Manually + additional unit tests.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/ProtocolClient.ts
+++ b/packages/chrome-plugin/src/ProtocolClient.ts
@@ -51,8 +51,16 @@ export default class ProtocolClient {
 		return (await chrome.runtime.sendMessage({ kind: 'getDomainStatus', domain })).enabled;
 	}
 
-	public static async setDomainEnabled(domain: string, enabled: boolean): Promise<void> {
-		await chrome.runtime.sendMessage({ kind: 'setDomainStatus', enabled, domain });
+	/** Set whether Harper is enabled for a given domain.
+	 *
+	 * @param overrideValue dictates whether this should override a previous setting.
+	 * */
+	public static async setDomainEnabled(
+		domain: string,
+		enabled: boolean,
+		overrideValue = true,
+	): Promise<void> {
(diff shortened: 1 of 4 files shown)


elijah-potter:
After further consideration, I've come to the conclusion that bug resides deeper, inside `harper-wasm`, and therefore also affects the Chrome extension. I've made fixes accordingly.

PR #2114, opened 2025-10-27
2056 -- Spell Checking Mode Option

### LINKED ISSUE #2056: Feature Idea: Spell Checking Mode (to control timing)
Currently the spell checking is run in real time. This is functionally fine, but appears a bit visually distracting when looking at the typed text. Typically for me, incorrectly spelled words are worth correcting _after_ finishing typing the current sentence, so I feel this would allow a more natural flow.

**Proposed Solution**
An option in settings to configure the timing for when spell checking is applied. This would allow the user to choose when they'd like it to happen.

**Example**
Concept for a new settings option:

Image

 
As an idea, this would allow a few options. Firstly the current default behaviour can be maintained.

Then alternatively, only check spelling after the spacebar is pressed. Then it would show any incorrectly spelled words after each is fully typed. Maybe even this could be expanded to include end of sentence punctuation.

Or another option would be only check spelling after typing stops (e.g. after 1-2 seconds without a keystroke).

 
**Component**
- [ ] Core engine
- [x] Plugin/Extension
- [ ] Other: _____

 
**Additional Context**
I feel this would be a very handy option. Crucially to maintain the current default for those who prefer the status quo, and then new options for fine-tuning when spell checking is applied.

Thank you for your continued work on this excellent addon!

LoganBrinsmead:
# Issues 

Closes #2056 

# Description

This adds a new "spell checking" option to the Harper options. The "default" choice defaults to the current Harper behavior where incorrect words are highlighted immediately as they're being typed. "Space" option only suggests a correction to the previously typed word after hitting the spacebar, and finally the "Stop" option waits two seconds after the user stops typing before calculating and applying suggestions.

# Demo

Will add video demonstrations to this PR very soon. 

# How Has This Been Tested?

Tested so far by typing in multiple websites including Github, blank page, and reddit and changes work across my tests.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/background/index.ts
+++ b/packages/chrome-plugin/src/background/index.ts
@@ -312,7 +312,7 @@ async function handleSetSpellCheckingMode(req: SetSpellCheckingModeRequest): Pro
 		throw new Error(`Invalid spell checking mode: ${req.spellCheckingMode}`);
 	}
 	await setSpellCheckingMode(req.spellCheckingMode);
-	
+
 	return createUnitResponse();
 }
 async function handleOpenReportError(req: OpenReportErrorRequest): Promise<UnitResponse> {


elijah-potter:
@LoganBrinsmead, would you mind marking this as a draft until it's ready for review? If it is ready, say so and I'll review it as soon as I can. Thanks!

PR #2116, opened 2025-10-27
feat(vscode): support more platforms

### LINKED ISSUE #2110: VSCode Extension fails to install into Dev Container (alpine arm64)
**Describe the bug**

Just as Visual Studio Code reports in the interface:

```The 'elijah-potter.harper' extension is not available in Visual Studio Code for the Alpine ARM 64.```

**To Reproduce**
1. Go to 'Extensions' in VSCode and find `Harper` Extension on the marketplace.
2. Click on 'Install in Dev Container: Existing Dockerfile'. Please note that the `existing Dockerfile` text may be different depending how you run the dev containers.
3. Look at the bottom right corner. There will be a pop-up.

**Expected behavior**
The extension is installed and usable inside the devcontainer.

**Screenshots**

Image

**Platform**
Visual Studio Code on MBA M2

**Additional context**
I run orbstack instead of docker if that helps.

mcecode:
# Issues 

Resolves #2110

# Description

Updates the VS Code CI to support all [platforms that VS Code supports](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions) except `web`.

# Checklist

### DIFF
--- a/.github/workflows/vscode_plugin.yml
+++ b/.github/workflows/vscode_plugin.yml
@@ -18,6 +18,9 @@ jobs:
           - os: windows-latest
             rust_target: x86_64-pc-windows-msvc
             code_target: win32-x64
+          - os: windows-latest
+            rust_target: aarch64-pc-windows-msvc
+            code_target: win32-arm64
           - os: macOS-latest
             rust_target: x86_64-apple-darwin
             code_target: darwin-x64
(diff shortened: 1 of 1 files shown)


mcecode:
@elijah-potter, while working on this, I was thinking that maybe we can turn off the other workflows for PRs and just leave the just checks since that's the only thing that has to do with actually linting and testing things. At the very least, I think we can turn off the VS Code workflow for PRs since, like the binaries workflow, most of the work there is just using `cross` to compile `harper-ls`. There shouldn't be any problems as long as the VS Code tests pass.

elijah-potter:
@mcecode, I think you're right. I'll disable them for PRs but keep them for commits to `master`.

PR #2136, opened 2025-11-04
feat: `harper-cli audit-dictionary`

hippietrail:
# Issues 
N/A

# Description

Adds a command to `harper-cli` to audit the curated dictionary.

- Checks that every flag defined in `annotations.json` is used
- Checks for flags not defined in `annotations.json` but occurring in `dictionary.dict`
- Checks for the same flag used multiple times in a single entry in `dictionary.dict`

# Demo
image

# How Has This Been Tested?
Manually, using the current `dictionary.dict` and `annotations.json`, and also with a modified `annotations.json` to force errors that weren't already in `dictionary.dict`.

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -704,15 +704,16 @@ fn main() -> anyhow::Result<()> {
                 .map(|c| (c, 0))
                 .collect();
 
-            let mut duplicate_flags = 0;
-            let mut unknown_flags = 0;
-            let mut unused_flags = 0;
+            // let mut duplicate_flag_total = 0;
+            let mut duplicate_flags = std::collections::HashMap::new();
+            let mut unknown_flags = std::collections::HashMap::new();
+            let mut unused_flag_total = 0;
 
             let dict_path = dir.join("dictionary.dict");
             let dict_content = fs::read_to_string(&dict_path)
                 .map_err(|e| anyhow!("Failed to read dictionary: {e}"))?;
 
-            for line in dict_content.lines() {
+            for (line_num, line) in dict_content.lines().enumerate() {
                 if line.is_empty()
                     || line.starts_with('#')
                     || line.chars().all(|c| c.is_ascii_digit())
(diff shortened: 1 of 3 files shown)


elijah-potter:
I'd love to include this as part of CI. How can we make that happen?

hippietrail:
> I'd love to include this as part of CI. How can we make that happen?

I wish I knew. I experimented with CI stuff months ago and found that I'm really bad at it.

But maybe you can ask some more specific questions since you seem to have a good grasp of CI?

elijah-potter:
I mean a trivial solution would be to simply add an entry to the matrix in `.github/workflows/just_checks.yml`.

I'm not sure that's the best solution, though. It just feels a little overkill. Perhaps it can be a part of the `just test-rust` step?

hippietrail:
> I mean a trivial solution would be to simply add an entry to the matrix in `.github/workflows/just_checks.yml`.
> 
> I'm not sure that's the best solution, though. It just feels a little overkill. Perhaps it can be a part of the `just test-rust` step?

Hmm I don't even know what `cargo test -q` does.
Oh it runs all the tests in the rust source code.
I guess we would just add a test in `harper-cli/main.rs` or `harper-cli/tests.rs`?

elijah-potter:
> I guess we would just as a test in harper-cli/main.rs or harper-cli/tests.rs?

That's one option. I think it would be better if it was a test in `harper-core`, hard-coded to look at the canonical files. If you'd rather keep it flexible, I think it would be fine to call `harper-cli` from inside `test-rust`.

hippietrail:
> > I guess we would just as a test in harper-cli/main.rs or harper-cli/tests.rs?
> 
> That's one option. I think it would be better if it was a test in `harper-core`, hard-coded to look at the canonical files. If you'd rather keep it flexible, I think it would be fine to call `harper-cli` from inside `test-rust`.

Flexible seems good. Keeping the code in `harper-cli` and calling it in a test from another module makes sense to me.

hippietrail:
I've just been thinking about this more and playing with the code a bit.
- `harper-cli` includes `harper-core` but not vice versa
- so `harper-core` can't call functions in `harper-cli`
- `harper-cli` can call functions in `harper-core` so we could move the auditing code there and still be able to call it from `harper-cli`
- But tests are not compiled in to releases
- So we'd have to have the audit code living in `harper-core`, which feels like bloat
- We could try to launch the `harper-cli` process from a test in `harper-core` but that would a) have a race condition where we'd need to ensure that specific test would have to depend on `harper-cli` already being built and b) would need to know a path for `harper-cli`

This might well have an elegant solution using features of Rust and/or cargo that I haven't learned yet.
Or maybe the proper solution is in CI without such a shortcut?
Or maybe move the audit code to `harper-core` but behind a feature flag? So `harper-cli` and the CI test would use a core with the feature flag but `harper-ls` and `harper-wasm` would not use it?
Or maybe some kind of integration test??

elijah-potter:
I think it would be sufficient to simply add it as a part of the `check-rust` recipe in the `justfile`

Issue #1991, opened 2025-09-24
Single source of truth for `LintKind` colours


hippietrail:
**What problem does this solve?**
Currently the mapping from `LintKind` to colour, currently only used when Harper interacts with the web, is duplicated in two TypeScript files:
- `packages/lint-framework/src/lint/lintKindColor.ts`
- `packages/web/src/lib/lintKindColor.ts`

(The two files are identical by the way)

It seems to me the colours should:
1. have a single source of truth
2. should be available to users of `harper-core`, not just web-based ones.

**Proposed Solution**
1. The colours could be part of the `LintKind` enum in `harper-core/src/linting/lint_kind.rs`.
2. Or they could be external to the enum but associated with its elements externally as some kind of map, probably in the same source file.

How to make it available to the TypeScript components is a more important question to answer. I can think of two ways:
1. Generate the two `lintKindColor.ts` files as part of the build process.
2. Provide some kind of API or APIs through which the TypeScript code can fetch the entire mapping of `LintKind`s to colours, or fetch the specific colour for a given `LintKind`.

At the moment the TypeScript code defines the most efficient way possible as a const literal:
```ts
const LINT_KIND_COLORS = {
	Agreement: '#228B22', // Forest green
	...
} as const;
```

I'm not sure how essential that is. Naive implementations of an API to fetch the colours not be as efficient. Generating the TypeScript files at build time would maintain the efficiency. Non-naive implementations of an API would fall somewhere between.

I played with a few ways to implement an API a few weeks ago and found it quite a bit more cumbersome than I expected. Somebody more skilled at Rust than me can surely do better.

**Examples**
Here's what I was playing with, to use the colours in a summary when linting on the commandline. But I'm sure there would be many uses for people making tools that use `harper-core` or `harper.js`

<img width="1126" height="114" alt="Image" src="https://github.com/user-attachments/assets/a9176f66-5416-4b72-9266-fbaa2e56ed59" />

**Component**
- [x] Core engine
- [x] Plugin/Extension
- [ ] Other: _____

hippietrail:
Reopening since there's still two sources of truth: One in `harper-cli/src/lint.rs` and the other in `packages/lint-framework/src/lint/lintKindColor.ts`

I'm experimenting with this now.

hippietrail:
[...]

PR #2138, opened 2025-11-04
Harper cli lint many

### LINKED ISSUE #2025: `just lint`/`harper-cli lint` should accept a directory and output a report tallying lints by type etc

It would help a lot for dogfooding.
I have a collection of test files that I've been keeping since the beginning, plus a collection of READMEs from GitHub repos that have been trending over the past few weeks.

You could also use it with the existing switches to enable one specific lint, then go through them to see how many are rightly flagged vs how many are false positives, etc.

I'm already working on this but wanted to file a PR in case anybody had the same idea or has suggestions for it.

**Component**
- [x] Core engine
- [ ] Plugin/Extension
- [ ] Other: _____

hippietrail:
# Issues 
Resolves #2025 

# Description

- `harper-cli lint` now accepts multiple arguments, and they can now be directories.
- Shows summaries of numbers of each `LintKind` and each rule for each file, using the `LintKind` colours.
- Shows final summary of numbers of `LintKind` and rule, plus the combination of the two, plus also the tallies of the spellchecking.
- Doesn't recursively walk directories, only files immediately within each directory pass as an argument.
- Colours are hard-coded here until #1991 is solved.

Please critique any and all aspects of this implementation: things I should do a different way, things I should add, things I should change, etc.

# Demo
Screenshot 2025-11-05 at 1 31 35 am

# How Has This Been Tested?
Manual testing.

# Checklist

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -353,6 +353,15 @@ pub fn lint_group() -> LintGroup {
             "Corrects common misspellings of the idiom `get rid of`.",
             LintKind::Typo
         ),
+        "HolyWar" => (
+            &[
+                (&["holey war", "holly war"], &["holy war"]),
+                (&["holey wars", "holly wars"], &["holy wars"]),
+            ],
+            "Literally for religious conflicts and metaphorically for tech preference debats, the correct spelling is `holy war`.",
+            "Corrects misspellings of `holy war`.",
+            LintKind::Malapropism
+        ),
         "HowItLooksLike" => (
             &[
                 (&["how he looks like"], &["how he looks", "what he looks like"]),
--- a/harper-core/src/linting/phrase_set_corrections/tests.rs
+++ b/harper-core/src/linting/phrase_set_corrections/tests.rs
@@ -934,6 +934,27 @@ fn got_ride_of() {
     );
 }
 
+// HolyWar
+
+#[test]
+#[ignore = "Known failure due to replace_with_match_case working by character index"]
+fn correct_holy_war() {
+    assert_suggestion_result(
+        "I know it is Holly War about idempotent in HTTP and DELETE",
+        lint_group(),
+        "I know it is Holy War about idempotent in HTTP and DELETE",
+    );
+}
+
+#[test]
+fn correct_holly_wars() {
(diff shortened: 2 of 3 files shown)


hippietrail:
> The `main.rs` file is starting to get overwhelmingly complex. Would you mind spending some time refactoring your work into one or more individual files? Right now, it's isn't immediately clear why some of these functions exist or which command they are relevant to.

Agreed. I actually like doing it in two steps because this way we can see the diff when it was in one file before the refactor. I've done just the minimal splitting of `lint()` into its own file.

I also fixed `just lint` do work with multiple inputs, which was missing in the original PR.

> Overall, I really like the experience of the command itself. I've already started using it productively :). Thanks!

Please critique anything about it, add features, or file issues for anything missing.

For instance I think some things are probably missing from the final report, I'll probably add a `--quiet` switch that hides the less essential output. I couldn't get the logic to decide to show or hide the Ariadne output when lines are too long to work well enough. I'm not necessarily happy with the commandline options and/or their order. I tried to add to them minimally to make the diff easier to inspect firstly.

hippietrail:
`cargo test` passes fine on my laptop but not `just precommit`, which I don't run very often when I'm only working on Rust stuff, and I don't really understand the output yet...
```
Reading harper.js.api.json

Deleting old output from ./markdown
Writing harper.js package
Deleting old output from ./html
Deleting old output from ../web/static/docs/harperjs
Rendering HTML...
parallel not found, falling back to sequential processing
node:fs:440
    return binding.readFileUtf8(path, stringToFlags(options.flag));
                   ^

Error: ENOENT: no such file or directory, open '{}'
    at Object.readFileSync (node:fs:440:20)
    at file:///Users/hippietrail/harper/packages/harper.js/renderPage.js:19:21
    at ModuleJob.run (node:internal/modules/esm/module_job:377:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:689:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '{}'
}

Node.js v24.9.0
error: Recipe `build-harperjs` failed with exit code 1
```

elijah-potter:
The issue with parallel missing should be resolved now that you've merged master. Let me know if that is not the case.

PR #2143, opened 2025-11-05
fix(obsidian): ignore files

Vinzent03:
# Issues 

This got mentioned here #1569 after merge

# Description

It previously used the deprecated [editorViewField](https://github.com/obsidianmd/obsidian-api/blob/23947b58d372ea02225324308e31d36b4aa95869/obsidian.d.ts#L1414), but now uses the [editorInfoField](https://github.com/obsidianmd/obsidian-api/blob/23947b58d372ea02225324308e31d36b4aa95869/obsidian.d.ts#L1263) directly from the obsidian import.

But the actual issue was that the editorViewVield was passed to the constructor, but not assigned to the attribute of the class.

# How Has This Been Tested?

I ran the plugin in a testing vault, and it worked perfectly fine.

# Checklist


elijah-potter (approved):
Thanks! This looks good to me. I'd love to write a test for this, but I'm not sure how. Are you aware of any prior art?

Either way, I think this is good to merge.

PR #2149, opened 2025-11-07
Hotkey

### LINKED ISSUE #1904: Hotkey improvements
If a word is incorrect but quite an obvious typo, it makes sense to either auto correct it or let to user choose to instantly correct it without checking the options. A hotkey makes sense for this.

**Proposed Solution**
A configurable hotkey to auto accept the most likely correct word of the last incorrect word.

Another option could be a hotkey to accept all most likely correct words instead of just the last one.

**Component**
[ ] Core engine
[x] Plugin/Extension
[ ] Other: _____

**Additional Context**
I'm using the Chrome extension currently.

LoganBrinsmead:
# Issues 

closes #1904

I am remaking this PR and the SpellChecking PR as I was having too many problems building lint-framework that should be resolved by just making new PRs.

# Description

This change adds a hotkey that corrects the last incorrect word as well as an option to change the hotkey in the settings menu. 

To change the hotkey, click the "Modify Hotkey" button which will then listen for your hotkey change and display it in the textarea next to the button. To be abundantly clear, this is a WIP PR so far and should not be merged as there are still some issues to iron out.

# Demo

Screenshot of the new option in the settings menu: 
Screenshot from 2025-09-26 13-27-29

# How Has This Been Tested?

Tested in blank.page where the hotkey is confirmed to correct the last incorrect word, as well as the Github editor, reddit text editor, and in Gmail. Confirmed to be working in all of these.

# Checklist

### DIFF
--- a/packages/lint-framework/src/lint/LintFramework.ts
+++ b/packages/lint-framework/src/lint/LintFramework.ts
@@ -183,7 +183,7 @@ export default class LintFramework {
 
 						const previousBox = this.lastBoxes[closestIdx];
 						const suggestions = previousBox.lint.suggestions;
-						if (suggestions) {
+						if (suggestions.length > 0) {
 							previousBox.applySuggestion(suggestions[0]);
 						} else {
 							previousBox.ignoreLint?.();


LoganBrinsmead:
@elijah-potter hey! I was gone for a moment but now I'm back trying to get this feature in Harper again. I was previously getting the lint-framework errors where it was failing to build on local and in Actions, so I remade the branch locally to ensure that everything builds fine, which it does, but it seems that it is once again having issues building lint-framework in actions. Do you know what the issue may be? My code is up to date with Harper's master. Thank you!

elijah-potter:
I'm so sorry it took me so long to take a look at this. It looks like, in `LintFramework.ts`, you pull in the Chrome plugin's `ProtocolClient.ts`. This causes `tsc` to lint the content of the latter file. This breaks things because `lint-framework` is an entirely different project, which the Chrome plugin depends on. I'd refactor your implementation to use dependency injection instead.

packages/chrome-plugin/src/options/Options.svelte
```diff
@@ -223,6 +277,20 @@ async function exportEnabledDomainsCSV() {
       <div class="space-y-5">
         <div class="flex items-center justify-between">
           <div class="flex flex-col">
+            <span class="font-medium">Apply Last Suggestion Hotkey</span>
+            <span class="font-light">Hotkey to apply the most likely suggestion to previously incorrect lint.</span>
```
elijah-potter: I think this description could be reworded.

packages/chrome-plugin/src/options/Options.svelte
```diff
@@ -223,6 +277,20 @@ async function exportEnabledDomainsCSV() {
       <div class="space-y-5">
         <div class="flex items-center justify-between">
           <div class="flex flex-col">
+            <span class="font-medium">Apply Last Suggestion Hotkey</span>
```
elijah-potter: This title's style does not match the others. See the "Activation Key"'s heading.

packages/chrome-plugin/src/options/Options.svelte
```diff
@@ -223,6 +277,20 @@ async function exportEnabledDomainsCSV() {
...
+          <Textarea readonly bind:value={buttonText} />
+          <Button size="sm" color="light" style="background-color: {isBlue ? 'blue' : ''}" bind:this={modifyHotkeyButton} on:click={() => {startHotkeyCapture(modifyHotkeyButton); isBlue = !isBlue}}>Modify Hotkey</Button>
+
+        </div>
+      </div>
+
+      <div class="space-y-5">
+        <div class="flex items-center justify-between">
+          <div class="flex flex-col">
+            <span class="font-medium">User Dictionary</span>
```
elijah-potter: It appears that these titles may have been accidentally duplicated? There are two "User Dictionary" headings.

packages/chrome-plugin/src/options/Options.svelte
```diff
@@ -223,6 +277,20 @@ async function exportEnabledDomainsCSV() {
       <div class="space-y-5">
         <div class="flex items-center justify-between">
           <div class="flex flex-col">
+            <span class="font-medium">Apply Last Suggestion Hotkey</span>
+            <span class="font-light">Hotkey to apply the most likely suggestion to previously incorrect lint.</span>
+          </div>
+          <Textarea readonly bind:value={buttonText} />
+          <Button size="sm" color="light" style="background-color: {isBlue ? 'blue' : ''}" bind:this={modifyHotkeyButton} on:click={() => {startHotkeyCapture(modifyHotkeyButton); isBlue = !isBlue}}>Modify Hotkey</Button>
```
elijah-potter, line 286: I think it would be more intuitive if the text itself changed while the hotkey capture was listening. Maybe have it say something like, "Press your desired key combination now". Or, something shorter. Either way, as it is right now, it isn't super obvious that when it's listening.

packages/chrome-plugin/src/options/Options.svelte
```diff
@@ -13,6 +14,8 @@ let dialect = $state(Dialect.American);
 let defaultEnabled = $state(false);
 let activationKey: ActivationKey = $state(ActivationKey.Off);
 let userDict = $state('');
+let modifyHotkeyButton: Button;
+let hotkey: Hotkey = $state({ modifiers: ['Ctrl'], key: 'e' });
```
elijah-potter, line 18: I really like this as a default key combination. It just feels... right.

packages/lint-framework/src/lint/LintFramework.ts
```diff
@@ -141,6 +151,43 @@ export default class LintFramework {
...
+				const hasCtrl = event.ctrlKey === hotkey.modifiers.includes('Ctrl');
+				const hasAlt = event.altKey === hotkey.modifiers.includes('Alt');
+				const hasShift = event.shiftKey === hotkey.modifiers.includes('Shift');
+
+				const match = key === expectedKey && hasCtrl && hasAlt && hasShift;
+
+				if (match) {
+					event.preventDefault();
+					event.stopImmediatePropagation();
+					const previousBox = this.lastBoxes[this.lastBoxes.length - 1];
```
elijah-potter: Because of how the `lastBoxes` are pulled together, you can't make any guarantees on their ordering. As a result, this strategy of just applying the last one results in somewhat random behavior. 

You may want to instead iterate through them and use the functions in `Box.ts` to find the nearest one to the cursor. We do this already when selecting a box to open when the "Activation Key" is triggered.

Take a look at the `PopupHandler.openClosestToCaret` function.

packages/chrome-plugin/src/options/Options.svelte
LoganBrinsmead: Thanks for the review! I changed it to "Applies suggestion to last highlighted word.", does that work? I thought that  "last highlighted word" might make more sense to a user than "previously incorrect lint"

packages/chrome-plugin/src/options/Options.svelte
LoganBrinsmead: fixed!

packages/chrome-plugin/src/options/Options.svelte
LoganBrinsmead: whoops, fixed!

packages/chrome-plugin/src/options/Options.svelte
LoganBrinsmead: Thanks, it should now say "Press desired hotkey combination now." after clicking the modify button

packages/lint-framework/src/lint/LintFramework.ts
LoganBrinsmead: thanks, I took a look and used the methods from Box to get the closest box to the caret. How does it look now? I totally agree with your assessment here and I think it should behave more inline with what would be expected now.

elijah-potter:
I compiled and ran your branch, loaded it into Chrome and spammed `Ctrl+e` on the demo page. This is what happened:

image

It appears there is some kind of confusion between the span that should be replaced and the actual suggestion.

packages/lint-framework/src/lint/LintFramework.ts
```diff
@@ -183,7 +183,7 @@ export default class LintFramework {
 
 						const previousBox = this.lastBoxes[closestIdx];
 						const suggestions = previousBox.lint.suggestions;
-						if (suggestions.length > 0) {
+						if (suggestions) {
```
elijah-potter: The suggestions are not guaranteed to be non-empty. This could result in an undefined value getting passed to `applySuggestion`. Please revert this change.

PR #2152, opened 2025-11-08
feat:far be it for→far be it from

hippietrail:
# Issues 
N/A

# Description

I heard somebody say "Far be it from me" in a YouTube video and was going to write a linter to correct that to "Far be it for me", which is what I've always used.
But first I did my research, and discovered that I've been using the wrong form all my life (-:
So instead I wrote a linter that corrects "far be it for" to "far be it from".

While looking for examples on GitHub I found out that some people using it without either "from" or "for". In those cases the linter still flags it but doesn't suggest a correction. Instead, the message suggests the user rephrase the sentence.

# How Has This Been Tested?

I added a set of unit tests from sentences found on GitHub.

# Checklist


harper-core/src/linting/far_be_it.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn far_be_it_for_the_software() {
+        assert_suggestion_result(
+            "Far be it for the software to give any indication of that fact.",
+            FarBeIt::default(),
+            "Far be it from the software to give any indication of that fact.",
+        );
+    }
+
+    #[test]
+    #[ignore = "No punctuation between '... so far' and 'be it ...'"]
```
elijah-potter, line 107: This test is not helpful and it adds noise. Would you mind removing it?

elijah-potter (changes requested):
Minor nitpick, but this is otherwise ready to merge.

harper-core/src/linting/far_be_it.rs
hippietrail: > This test is not helpful and it adds noise. Would you mind removing it?

You think? It's a false positive that isn't handled yet with sentences that contain "so far be it", which rightly should have a comma but often enough the comma is left out. Maybe my ignore message is no good?

elijah-potter: I understand. But isn't it out of scope for this rule? If you think otherwise, I'll merge this as-is.

hippietrail: > I understand. But isn't it out of scope for this rule? If you think otherwise, I'll merge this as-is.

My thinking was that it's in scope because it's a reasonably common false positive and future editors of the module should be aware of it and a clever one might be able to detect and de-ignore it?

PR #2162, opened 2025-11-10
feat: add suggestions to mass plurals linter

hippietrail:
# Issues 
N/A

# Description

I was adding "wear and tears" to this linter when I noticed it doesn't do `Suggestion`s.
So I added those too.

Note that it doesn't check or suggest anything in relation to plural determiners before the pluralized mass noun. So it will correct "these softwares" to "these software" rather than "this software". That could/should be a separate linter.

I also refactored a tiny bit, mostly making identifiers a bit clearer.

# How Has This Been Tested?

I added a unit test which checks both "wear and tears" and the suggested correction to "wear and tear".

# Checklist


elijah-potter (approved):
Looks good. No comments. Thanks for extending it!

Issue #2164, opened 2025-11-10
Should flag questions lacking a question mark.


hippietrail:
**The Error**

**Examples:**
1. Do you often see questions like this.
2. Can they be fixed
3. Why does it happen!
4. Am I the only one who is annoyed.

**Potential Edge Cases**
Detecting questions is not as straightforward as you might think when the `?` is missing.
They mostly start a present of past form of "be" or "do", a modal verb such as "can" or "should", or an interrogative word such as "what" or "how". In all the verb cases, the negative contractions are also possible.

But this isn't enough on its own. Many statements also fit this pattern.

Maybe the interrogatives have to precede one of those types of verbs, and maybe a pronoun in subject form has to follow said verb?

**Implementation**
It might be easier to implement if there were a version of `ExprLinter` that worked on sentences rather than "chunks".

hippietrail:
Isn't this 🌲evergreen - yes it is?

PR #2165, opened 2025-11-10
feat: allow `ExprLinter` to work on sentences as well as chunks

hippietrail:
# Issues 
No issue but this is the result of my thoughts on #2164

# Description

Currently `ExprLinter` works on "chunks", where a chunk can be thought of as a clause, or a section of text between commas.

This means if you want to work on whole sentences, you either have to instead use `Linter` and not be able to use patterns/expressions, or when a sentence contains a comma, you'll only see parts of a sentence at once.

In the case of the potential missing-question-mark linter, that could result in an `ExprLinter` thinking it's looking at the beginning or ending of a sentence (an interrogative or ending with something other than a `?`)

This PR adds a way to optionally choose to run an `ExprLinter` over sentences instead of chunks. The default is still chunks and no existing code needs to be changed to retain the default behaviour.

# How Has This Been Tested?

I added unit tests to use the same toy `ExprLinter` on either chunks or sentences with a pattern containing a comma that fails for chunks and succeeds for sentences.

Please go ahead and critique my implementation!

# Checklist

### DIFF
--- a/harper-core/src/linting/find_fine.rs
+++ b/harper-core/src/linting/find_fine.rs
@@ -3,6 +3,7 @@ use crate::expr::Expr;
 use crate::expr::SequenceExpr;
 use crate::patterns::InflectionOfBe;
 
+use super::expr_linter::Chunk;
 use super::{ExprLinter, Lint, LintKind, Suggestion};
 
 pub struct FindFine {
@@ -23,6 +24,8 @@ impl Default for FindFine {
 }
 
 impl ExprLinter for FindFine {
+    type Unit = Chunk;
+
     fn expr(&self) -> &dyn Expr {
         self.expr.as_ref()
     }


harper-core/src/linting/expr_linter.rs
```diff
@@ -64,13 +82,77 @@ where
...
+                suggestions: vec![Suggestion::ReplaceWith("foo and bar".chars().collect())],
+                ..Default::default()
+            })
+        }
+
+        fn description(&self) -> &str {
+            "A test"
+        }
+
+        fn get_unit(&self) -> Unit {
```
elijah-potter: Since the linter could return different values for different invocations, I'm not sure this is the best strategy.

Ideally, and I'm not sure if you could reuse this work for this, we would have two distinct traits, say `ChunkExprLinter` and `SentenceExprLinter`.

hippietrail: I did at first think I'd split it into two linters but I also thought making the change smaller and keeping the logic in one place (DRY) might help, such as the `_with_context` enhancement - which should combine with this smoothly.

What are your thoughts on parameterizing it?

elijah-potter: If we do it this way, we'd have to refactor a significant portion of the `lint_group`'s caching system to account for sentences. My main concern is that the `get_unit` function can return different values on each invocation.

I agree with your sentiment around keeping things DRY. Maybe using an [associated type](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html) would be a happy medium?

hippietrail: > associated type

I'll be investigating that today then!

hippietrail: > If we do it this way, we'd have to refactor a significant portion of the `lint_group`'s caching system to account for sentences. My main concern is that the `get_unit` function can return different values on each invocation.
> 
> I agree with your sentiment around keeping things DRY. Maybe using an [associated type](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html) would be a happy medium?

I spent a few hours on this and feel like there's some advanced Rust stuff I've seen in videos but that is not yet within my reach that might help here.
The point is that each linter will only work on chunks or sentences and never both. Doing so would be an error but the way I coded it up in the PR has no way to enforce that.
I wanted to not have to change every `ExprLinter`. Maybe some way of making `ExprLinter` and `ExprSentenceLinter` as newtypes of a parameterized "abstract class" for want of a better term, that could somehow encapsulate `iterate_chunks` vs `iterate_sentences` into some kind of type or trait, but I'm not at a skill/knowledge level yet in Rust where I can even discern if that's possible, let alone implement it.

Any ideas?

hippietrail: I thought I should've been able to find a way to implement it sentence iteration without having to change every `ExprLinter` but I gave up and went with a way that modifies them all.

Is it going to avoid the problems with the caching?

harper-core/src/linting/lint_group.rs
```diff
@@ -349,13 +350,16 @@ impl LintGroup {
     pub fn add_expr_linter(
```
elijah-potter: Since this is now restricted to only Chunk-based ExprLinters, I think the naming of this function and the associated `expr_linter` field should assume more appropriate names.

Something else to consider: should we have a specialized cache for sentence linters as well?

harper-core/src/linting/expr_linter.rs
```diff
@@ -5,12 +5,41 @@ use crate::{Document, LSend, Token, TokenStringExt};
...
+
+impl DocumentIterator for Chunk {
+    type Unit = Chunk;
+
+    fn iter_units<'a>(document: &'a Document) -> Box<dyn Iterator<Item = &'a [Token]> + 'a> {
+        Box::new(document.iter_chunks())
+    }
+}
+
+impl DocumentIterator for Sentence {
```
elijah-potter, line 27: This implementation tickles my brain in a good way. I like the implications, especially the extensibility if there end up being additional levels of granularity.

harper-core/src/linting/expr_linter.rs
elijah-potter: It looks like your solution solves the caching problem. I think it's good that we have to declare it in every linter, since that makes it easy to tell which is which at a glance.

harper-core/src/linting/lint_group.rs
hippietrail: Oh I think I did rename that to `add_chunk_expr_linter` in a previous attempt. Smaller diffs make PR reviewing easier though so changing it now is probably good.

As for the caching that's a part of the code I haven't looked at at all. I don't know if the associated types propagate the same way that parameterized types do but I think so right? And I don't know if the caching code is inherently affected by that type?

harper-core/src/linting/expr_linter.rs
hippietrail: Yeah me too!

hippietrail:
> Great stuff. This PR is almost perfect. I like the test cases, the implementation, with just a few nits about naming. Once this is merged, I'd love to see a follow-up with an actual sentence-based ExprLinter so we can have one running in production as soon as possible.

The quintessential linters that would be affected by this are surely the Oxford comma linters, which are now just Linter's rather than ExprLinter's but fail if a list follows a clause? that ends with a comma.

But that's probably a big job. I had some other problem in mind when I first thought of sentence iterators but I forget what that was...

harper-core/src/linting/lint_group.rs
elijah-potter: Don't worry about the caching for the sentence linters. That's a problem to be solved in a separate PR. For now, let's just make the naming a bit clearer so we can merge this.

PR #2169, opened 2025-11-10
Add Julia support

rturquier:
# Issues 
#79

# Description

Add support for [Julia](https://julialang.org/).   
This is my first time contributing code to an open-source project – please let me know if I do anything wrong!

# Demo

# How Has This Been Tested?

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -2229,7 +2229,6 @@ Clorets/g
 Clorox/NgV
 Closure/g
 Clotho/Og
-Cloudflare/Og
 Clouseau/Og
 Clovis/Og
 Clyde/Og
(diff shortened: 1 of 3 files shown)


rturquier:
Oh no, it looks like it might not be as easy as I first thought. I wrote a test, and docstrings are not spell-checked (only simple inline comments). 
It looks like something similar happened with Python, and this might be too difficult for me (I have no experience with Rust or TreeSitter). @elijah-potter, what would be the way forward now?

elijah-potter:
Honestly, you might be able to get away with copy-pasting `harper-python` into `harper-julia` and using grep to make it work for Julia with the Julia Tree-sitter library.

rturquier:
Ok, I'll try that then.

rturquier:
…I tried, and it's not that easy. I initially thought this would just take a couple of hours, but now that it's turned out to be more complicated, I think I currently don't have the skills or time to make this work.

elijah-potter:
I think I might be able to knock this out. Do you mind if I push a few commits here to build off of your work? For me, the hard part is building out test cases, so you've already done most of the work here.

rturquier:
Yes please, feel free to take it from here!

PR #2173, opened 2025-11-10
feat: add `match_to_lint_with_context` to `ExprLinter`

### LINKED ISSUE #1562: Feature: detect conflicting numerals
I wrote:

> A few minutes ago, there was an outage due to two hosts being down at the same time.

Then I realized that there were more than two hosts down:

> A few minutes ago, there was an outage due to several two hosts being down at the same time.

I.e., I added "several", but forgot to remove "two". Harper didn't detect that, but Grammarly did.

### LINKED ISSUE #2055: Feature requst: A way to access the context tokens before & after what `match_to_lint` gets passed
In a `Linter` you have access to all the tokens but you have to manually find which subsequence of the tokens match what you are primarily looking for.

In an `ExprLinter` you can build an `Expr`/`Pattern` which will match what you are primarily looking for automatically - but you don't have access to the tokens before or after it.

Sometimes you want to start looking for a main pattern of tokens but then need to examine the surrounding context to decide if it's what you're really looking for.

A `Linter` can do this but you lose the ability to use `Expr`s.

You can work around this because an `ExprLinter` does have access to the full source. You can re-tokenize the parts of the source before and after the `matched_tokens` but this is wasteful as it can be slow and it has already been done.

Another way to work around it is to include the context around the main pattern *in* the pattern - but this is very difficult to reason about well. At least for me.

Perhaps we could have a companion method to `match_to_lint` or perhaps we need a new trait with a `match_to_lint_with_context`?

Perhaps this can already be done and I'm just not aware of it?

hippietrail:
# Issues 
Resolves #2055

Also resolves #1562 

# Description

Sometimes in an `ExprLinter` you want to take into account the tokens immediately before or after the `matched_tokens` that `match_to_lint` receives but making them as part of the expression/pattern is too difficult and/or makes the code difficult to follow.

This adds `match_to_lint_with_context` which provides `before` and `after` `&[Token]` slices as well as the `matched_tokens` slice.

An `ExprLinter` can override whichever of the two methods it wishes. This way none of the existing linters needs to be modified.

# How Has This Been Tested?

Unit tests are included that call the old method without context, that call the new method with context, and that take the context into account to produce lints.

I also implemented #1562 as a pretty good test for the new functionality.

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2225,6 +2225,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "fuzz"
+version = "0.0.0"
+dependencies = [
+ "harper-comments",
+ "harper-core",
+ "harper-html",
+ "harper-literate-haskell",
+ "harper-typst",
+ "libfuzzer-sys",
+]
+
+[[package]]
 name = "gemm"
 version = "0.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3369,6 +3381,16 @@ name = "libc"
 version = "0.2.174"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1171693293099992e19cddea4e8b849964e9846f4acee11b3948bcc337be8776"
+
+[[package]]
+name = "libfuzzer-sys"
+version = "0.4.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5037190e1f70cbeef565bd267599242926f724d3b8a9f510fd7e0b540cfa4404"
+dependencies = [
+ "arbitrary",
+ "cc",
+]
 
 [[package]]
 name = "libloading"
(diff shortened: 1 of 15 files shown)


elijah-potter (changes requested):
I like the way you've implemented this. It is simple and it works. 

It looks like you've implemented a linter to test that the logic works as expected. At first glance, readers (such as myself) may not see that the code resides in a test module. Would you mind changing the name for each of those structs to be of the form `Test*`?

The pipeline is somewhat complex, and I can't think of any downstream consequences for this change. To be sure, though, would you implement a real rule (or modify an existing one) to use the provided context?

hippietrail:
> The pipeline is somewhat complex, and I can't think of any downstream consequences for this change. To be sure, though, would you implement a real rule (or modify an existing one) to use the provided context?

I've implemented a linter #1562 which required some additions and fixes. My favourite part is that you can even use a `Pattern` or `Expr` in the context!

elijah-potter (approved):
Looks great! I added a random commit to fix an out-of-date comment in `lint_group.rs`. Thanks for the linter as well, I can't wait to start toying around with utility functions for the pre-and-post context windows.

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -135,6 +135,23 @@ Suggest:
 
 
 
+Lint:    Grammar (127 priority)
+Message: |
+      57 | lessons in the schoolroom, and though this was not a very good opportunity for
+      58 | showing off her knowledge, as there was no one to listen to her, still it was
```
elijah-potter: After merging `master` and updating snapshots, it looks like there is a problem that appears with your other PR that changes `no`'s metadata. Mind taking a look?

hippietrail: > After merging `master` and updating snapshots, it looks like there is a problem that appears with your other PR that changes `no`'s metadata. Mind taking a look?

Is that related to #2200? I'll dig into it at the coffee shop...

PR #2177, opened 2025-11-11
feat(core): create linter to detect improper use of `need to <nominal>`

elijah-potter:
# Description

In a recent [blog post](https://elijahpotter.dev/articles/improving-rust-compile-times-by-71-percent) I included some grammatical errors which were not caught by Harper.
Specifically, I included the sentence, "We simply don't need to compiler to do as much work anymore". The word "to" should be "the". 
This PR introduces a rule to handle that case, as well as most other cases where "need to" is written where "need the" is more appropriate.

# How Has This Been Tested?

Extensive unit testing and LLM-assisted fuzzing.

# Checklist

### DIFF
--- a/harper-core/src/linting/lint_group.rs
+++ b/harper-core/src/linting/lint_group.rs
@@ -481,8 +481,6 @@ impl LintGroup {
         insert_expr_rule!(CautionaryTale, true);
         insert_expr_rule!(ChangeTack, true);
         insert_expr_rule!(ChockFull, true);
-        insert_expr_rule!(NeedToNoun, true);
-        insert_expr_rule!(VerbToAdjective, true);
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/lint_group.rs
```diff
@@ -474,7 +475,8 @@ impl LintGroup {
         insert_expr_rule!(CautionaryTale, true);
         insert_expr_rule!(ChangeTack, true);
         insert_expr_rule!(ChockFull, true);
-        insert_struct_rule!(NeedToNoun, true);
+        insert_expr_rule!(NeedToNoun, true);
+        insert_expr_rule!(VerbToAdjective, true);
```
hippietrail: It looks like these got out of alphabetical order.

PR #2184, opened 2025-11-12
fix(core): improve the initialisms linter for all-caps inputs

### LINKED ISSUE #2181: Expanding AFAIK shouldn't (only) offer title case.
**Describe the bug**
Expanding the initialism `AFAIK` only offers an expansion in title case, "As Far As I Know".

What it should really do is decide whether to capitalize "As" based on whether it's at the start of a sentence or not. But it can probably also just offer an all-lowercase expansion and the sentence capitalization linter would take care of whether to suggest capitalizing "as".

Something that could be added to this and probably elsewhere is an analysis of the context. If all/most of the surrounding text is all-caps, title case, or all-lowercase for stylistic reasons.

**Screenshots**
Image

elijah-potter:
# Issues 

Hopefully resolves #2181

# Description

All caps inputs for the initialisms linter would produce title-case outputs.
I've moderately improved the situation by only matching the case for the first letter.
It doesn't completely solve the issue, since certain words, like "I", aren't properly capitalized.

@hippietrail, I'm curious if you have an improvement upon my solution.

# How Has This Been Tested?

See additional unit test.

# Checklist


harper-core/src/linting/initialisms.rs
```diff
@@ -118,4 +118,13 @@ mod tests {
             "Really excited for this.",
         );
     }
+
+    #[test]
+    fn issue_2181() {
+        assert_suggestion_result(
+            "AFAIK, we don't currently have an issue for it.",
+            lint_group(),
+            "As far as i know, we don't currently have an issue for it.",
```
hippietrail, line 127: Ah I see what you mean. That's a bit unfortunate. Maybe we can have a function that converts a run of text to "normal" case, similar to how we have `title_case.rs`
A beginner version could do nothing other than capitalize every instance of the word "i".
A more advanced version could do a lookup on every word and look at the orth flags to try to work out if a word has a single canonical lettercase spelling.

elijah-potter: I was thinking along those same lines. Since we have a separate rule for capitalization, shall we leave this PR as-is and return to it if it's a problem?

hippietrail: > I was thinking along those same lines. Since we have a separate rule for capitalization, shall we leave this PR as-is and return to it if it's a problem?

Yes. Good approach.

Issue #1827, opened 2025-08-29
Add General Ruleset Preset Toggles


lukasmwerner:
**What problem does this solve?**
As the ruleset in harper grows it would be really nice to be able to enable and disable large sets of rules that belong or do not belong to a style guide. 

**Proposed Solution**
There could be a rule group that enables the rules for AP Style and another set of rules for Chicago etc. 

**Examples**
Basic dropdown similar to dialect selector that picks the rules that are pre-enabled.

lukasmwerner:
Please keep this open.

lukasmwerner:
not stale.

PR #2185, opened 2025-11-13
feat: correct lo and behold, and jack-o'-lantern

lukasmwerner:
# Description
I was drafting a blog post and realized that Harper was missing these corrections. 

# How Has This Been Tested?
WIP

# Checklist


hippietrail:
It should be `jack-o'-lantern` with the apostrophe following the `o` since it represents an elided `f` in the word `of`.

"Lo and behold" used to be in the predecessor to the current `phrase_corrections` and disappeared some time around the transition. I remember pointing it out when I noticed but it wasn't put back for whatever reason?

lukasmwerner:
It seems that I need this correction in the web edition as I didn't make that mistake in the pr itself but just the title. That should be fixed now.

hippietrail:
> It seems that I need this correction in the web edition as I didn't make that mistake in the pr itself but just the title. That should be fixed now.

Yes I only noticed that after I'd sent my comment. I also thought the spelling you included "jack o'lantern" was an error. So I checked all the major online dictionaries: American Heritage, Cambridge, Collins, Longman, OED, Oxford, and Merriam-Webster. Interestingly it's not in any of them except for Collins. Being in any one of them is sufficient IMHO. It makes me wonder about at some point marking dictionary entries somehow so that users can allow or block specific spellings or sets of spellings by authority etc.

Just some observations (-:

harper-core/src/linting/phrase_corrections/mod.rs
```diff
@@ -697,6 +697,20 @@ pub fn lint_group() -> LintGroup {
...
+            "Corrects contraction of `jack o lantern` to `jack o'lantern` or `jack-o'-lantern`",
+            LintKind::Punctuation
+        ),
+        "JackOLanterns" => (
+            ["jack o lanterns", "jackolanterns", "jack-o-lanterns"],
+            ["jack-o'-lanterns", "jack o'lanterns"],
+            "Use contraction when referring to `jack-o'-lanterns`",
+            "Corrects contraction of `jack o lanterns` to `jack o'lanterns` or `jack-o'-lanterns`",
+            LintKind::Punctuation
+        ),
```
hippietrail: Oh I only just realized. In `phrase_corrections` like this will create two settings that have to be enabled/disabled individually. One for the singular and one for the plural. If you move them to the newer `phrase_set_corrections` you can put them both together with just a single setting.

hippietrail (changes requested):

lukasmwerner:
I got this one from the Associated Press's style guide originally. But agreed that we will want to start adding rule-sets as suggested in #1827 so users will be able to start to follow specific groups of rules.

harper-core/src/linting/phrase_corrections/mod.rs
lukasmwerner: That's super neat! I'll also have to update my previous rule(#1829) to that new correction style as well.

elijah-potter:
@lukasmwerner, is this ready for review?

lukasmwerner:
I'm still needing to do testcases but otherwise this is done.

elijah-potter:
@lukasmwerner, it looks like most style guides recommend you _always_ include hyphens in `jack-o'-lantern`. Did you find a competing report?

lukasmwerner:
I haven't seen others but I was going off of AP style where they were using hyphenation. However, I was also going off of Wikipedia which states: ['A jack-o'-lantern (or jack o'lantern)'](https://en.wikipedia.org/wiki/Jack-o%27-lantern).

But if that isn't a strong enough case I'm happy to remove those cases.

elijah-potter:
I don't think the cases need to be removed. I just think they should all map to the hyphenated version.

elijah-potter:
Let me know if there's anything I can do to help with those test cases.

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -268,6 +268,33 @@ pub fn lint_group() -> LintGroup {
...
+                ("not inadequate", "adequate"),
+                ("not unpleasant", "pleasant"),
+                ("not unreasonable", "reasonable"),
+                ("not impossible", "possible"),
+                ("more preferable", "preferable"),
+                ("not online", "offline"),
+                ("not offline", "online"),
+            ],
+            "Consider the direct form.",
+            "Offers direct-positive alternatives when double negatives might feel heavy.",
```
hippietrail, line 314: The message doesn't work for `("more preferable", "preferable"),` which is about redundancy rather than double negative, so it should not be in this group.

hippietrail (changes requested):

harper-core/src/linting/phrase_set_corrections/mod.rs
lukasmwerner: i just merged that in from master.

hippietrail: > i just merged that in from master.

I thought it might be something like that. Not sure if that's merge doing something funny or if master had it in the wrong place?

elijah-potter:
I'm going to close this PR. If you'd like to take another crack at it, feel free to reopen!

PR #2186, opened 2025-11-13
feat: correct regular inflected forms that should be irregular

hippietrail:
# Issues 
N/A

# Description

Corrects incorrect regular inflected forms that should be irregular:
- womans → women
- buyed → bought
- gooder → better

This includes when the incorrect regular form required doubling the last letter of the verb lemma:
- resetted → reset

It also corrects incorrect forms that are regular but that involve changes to endings:
- calfs → calves
- skys → skies
- tomatos → tomatoes

I reworked a number of dictionary entries because the `S` flag simultaneously makes noun plurals and verb 3rd person singular present forms, but while `mans` is a legit verb it's not a legit noun for instance.

**Draft** for now because:

This linter reveals a bug in the assertion functions such as `assert_top3_suggestion_result`.

I've noticed that for this linter, the order in which suggestions are generated is inconsistent between runs. I'm not sure if this happens for all linters but has just never surfaced before, or if it's due to this linter's use of [phf](https://crates.io/crates/phf).

I'm using phf because my LLM suggested it was the best way to make sure the table of irregular forms is only built once. phf was already used in the Harper project in `named_entities.rs`, something to do with "web atoms". (I don't understand the relationship yet.) But it wasn't used in `harper-core` until now.

Which brings me to my next question:

Also I'm looking for tips on how to unify the irregular verb tables between this linter and the `SimplePastToPastParticiple` linter.
I'm guessing a JSON file but I'm not sure how/where to implement the parsing of it to make sure it only happens once.

# How Has This Been Tested?

Unit tests are included.

# Checklist


hippietrail:
Here's another example for when I get back to this:
- I **sticked** to vim-lsp only for a relatively short period, as soon after I decided to try a more ambitious plugin

hippietrail:
Looks like I messed this one up with the AI too. `cargo test` worked locally but 170 commits can't be right.

PR #2193, opened 2025-11-13
feat(core): make it clearer that two `that` tokens can always be condensed

### LINKED ISSUE #2132: False positive: "that that" repetition and other similar constructions
**What got flagged?**
Somewhat frequently, I run into a construction like this:
> I know that that answer is correct

Which as far as I know is perfectly grammatical, if a bit odd written down.

This also occurs with 
- "had had" (ie, "the medicine had had no adverse effects")
- "is is" (ie, "What it is is interesting")

However, Harper will flag this as repetition:
image

I don't want to disable the word repetition rule entirely, since it's quite useful for "the" repeats or the like, but it also comes up frequently enough to be a mild hindrance to ignore every time. 

**Why is this incorrect?**
In this sentence, the second "that" refers to a previously-mentioned thing (in this case an answer) and the first one is a helper word (I'm sure there's a more accurate explanation but I think it suffices for this issue).

Similar explanations exist for had had and is is

**Example of correct usage:**
I have no idea how one would lint this properly, but neither Google Docs nor Grammarly flags any of the example sentences.

elijah-potter:
# Issues 

Resolves #2132

# Description

In #2132, it became clear that Harper would always suggest to condense `that that` into a single token, no matter which context it would appear.
This is actually correct behavior, since the second `that` is always redundant.
To resolve the problem, I've created a new phrase correction for this special case.
Hopefully the provided message is clearer.

# How Has This Been Tested?

Additional unit test + manually.

# Checklist

### DIFF
--- a/packages/chrome-plugin/tests/hn.spec.ts
+++ b/packages/chrome-plugin/tests/hn.spec.ts
@@ -29,6 +29,9 @@ test('Hacker News wraps correctly', async ({ page }) => {
 	await page.waitForTimeout(2000);
 	await page.reload();
 
+	// Needed because this element has a variable height and may offset the highlight boxes by an unknown amount.
+	await page.locator('.toptext').evaluate((el) => el.remove());
+
 	const editor = getTextarea(page);
 	await replaceEditorContent(
 		editor,
@@ -48,6 +51,9 @@ test('Hacker News scrolls correctly', async ({ page }) => {
 
 	await page.waitForTimeout(2000);
 	await page.reload();
+
+	// Needed because this element has a variable height and may offset the highlight boxes by an unknown amount.
+	await page.locator('.toptext').evaluate((el) => el.remove());
 
 	const editor = getTextarea(page);
 	await replaceEditorContent(


hippietrail:
I would not consider this to be a case of redundancy. In formal styles you would keep both "that"s while in formal style you would very much avoid redundancy.

harper-core/src/linting/phrase_corrections/mod.rs
```diff
@@ -1222,6 +1222,13 @@ pub fn lint_group() -> LintGroup {
             "Did you mean the definite article?",
             "Fixes especially common misspellings of the word `the`",
             LintKind::Typo
+        ),
+        "RedundantThat" => (
+            ["that that"],
+            ["that"],
+            "The second `that` is redundant.",
+            "There is never a situation where `that that` cannot be condensed into a single token.",
+            LintKind::Redundancy
```
hippietrail: This wording strikes me as too strong. It makes it sound like it's an error to say "I thought that that was true", which could mislead users.
An ideal implementation would detect whether to handle it as a `LintKind::Redundancy` or `LintKind::Style`. Since we can't do that, the message should be more informative and the `LintKind` should be weakened to `LintKind::Repetition` in my opinion.

elijah-potter: Good call. I've adjusted the wording. What do you think now?

PR #2195, opened 2025-11-13
The Belarusian language support (at least just spellchecking for it)

ssvb:
# Issues 

https://github.com/Automattic/harper/discussions/2155

# Description
This is not intended to be merged. It's just the most simple test to see whether the Harper engine is able to shoulder the load of supporting one more language in addition to English without degrading performance and without increasing the memory footprint too much.

The Belarusian language uses a different Cyrillic alphabet, so it doesn't clash with the existing English language support. The Belarusian dictionary has been taken from https://github.com/375gnu/spell-be-tarask

I could have added a simple Belarusian linter rule as well, but would prefer to keep it simple until the seemingly unreasonable memory footprint issue is resolved. The Belarusian wordlist has ~20MB size in a plain text format. Using the `fst-bin` command line tool (which accompanies the `fst` crate used by Harper) the Belarusian spellchecker dictionary can be compressed to a merely ~1.1MB binary. At least that's the theory.

# Demo

# How Has This Been Tested?

Used the `just lint` command to check short text files, such as "Ths is an test. Гта тэст."
And also the Firefox plugin to check the same short texts in https://textarea.online

# Checklist

### DIFF
--- a/harper-core/annotations.json
+++ b/harper-core/annotations.json
@@ -616,6 +616,3956 @@
 			"target": [],
 			"base_metadata": {},
 			"rename_ok": true
+		},
+		"а": {
+			"#": "-",
+			"kind": "prefix",
+			"cross_product": true,
+			"replacements": [
+				{
+					"remove": "",
+					"add": "не",
+					"condition": "."
+				}
+			],
+			"target": [],
+			"base_metadata": {},
+			"rename_ok": true
+		},
+		"н": {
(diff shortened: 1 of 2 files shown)


elijah-potter:
This is super cool. I'd love to help with any modularity changes that would need to be made to make this production-ready. How would you characterize the result of your experiment?

PR #2200, opened 2025-11-15
fix: `is_quantifier` matched all determiners, not just quantifiers

### LINKED ISSUE #1189: "no" isn't tagged as a determiner
"no", just like "some" and "many", is a [determiner](https://en.wiktionary.org/wiki/no#Determiner) but currently not tagged in word metadata.

hippietrail:
# Issues 
Found this while looking into #1562
Also fixes #1189

# Description

It turned out that `is_quantifier` and any functions derived from it were only checking if a word/token was a determiner, not also whether they were the right kind of determiner.

Also marking more quantifiers in `dictionary.dict`: "least", "no"

# How Has This Been Tested?

Two new unit tests to make sure it does match a quantifier and does not match a determiner that is not a quantifier.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -35546,7 +35546,7 @@ nitroglycerine/Nmg!@_
 nitrous/~Ng
 nitwit/NgS
 nix/~NgSVGd
-no/~PNSg
+no/~PNSgDq
 nob/~NSVY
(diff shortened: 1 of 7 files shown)


elijah-potter (approved):
Good catch. I always like to add these kinds of unit tests when I encounter problems like this. Thanks!

PR #2209, opened 2025-11-17
feat: Sneaked vs snuck

hippietrail:
# Issues 
N/A

# Description

I reimplemented this based on the way I implemented the vicious circle vs vicious cycle linter.
In fact I asked an agent to do that. Sneaked vs snuck is simpler since it doesn't have to worry about a common spello like "viscious circle/cycle" in the other linter.

I simplified and refactored the code the agent produced a decent amount. But it's still worth giving code that agents were involved in closer scrutiny.

# How Has This Been Tested?

There are some unit tests for lettercase variants and I picked two sentences from GitHub to use for real-world tests.

# Checklist

# See Also

vicious circle vs cycle #2384


harper-core/src/linting/sneaked_snuck/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+    Some(Lint {
+        span: tokspan,
+        lint_kind: LintKind::Usage,
+        suggestions: vec![Suggestion::replace_with_match_case_str(target_word, word)],
+        message: format!("Use `{}` instead of `{}`.", target_word, source_word),
+        ..Default::default()
+    })
+}
+
+macro_rules! impl_expr_linter {
```
elijah-potter, line 62: I'm not sure how I feel about this pattern. It could work well, or not. Let's go for it and see how it affects things down the road.

hippietrail: I feel the same. I initially was coding sneak/snuck a different way that didn't pan out but then realized that in the meantime the similar `ViciousCircle`/`ViciousCycle` module had been approved and merged so I rewrote 'snuck' based on a simplified version of the 'vicious' logic.

The difference is that 'vicious' has to deal with the common eggcorn 'viscous' at the same time as dealing with the preference for 'cycle' or 'circle'.

In the long run we'll see what strategy is adopted for mutually-exclusive rules and may come back and change these.

PR #2210, opened 2025-11-17
feat: `harper-cli` warn setting unknown rules

hippietrail:
# Issues 
N/A

# Description

When testing new linters you're working on, you'll often use `harper-cli lint` with the `--only` or `--ignore` options. But if you misspelled any option it'll just silently ignore it and you'll be left confused about what you've done wrong.

This is also common if you have more than one build and you accidentally run the wrong one and see nothing for your new lint, making you think your lint is at fault when it's a build without the new lint.

This is now addressed.

# Demo
image

# How Has This Been Tested?
Manually

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -225,12 +225,18 @@ fn main() -> anyhow::Result<()> {
                 linter.set_all_rules_to(Some(false));
 
                 for rule in rules {
+                    if !linter.contains_key(&rule) {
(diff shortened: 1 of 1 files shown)


hippietrail:
> I've run into the issue you've described on a number of occasions. Thanks for addressing it!

Well don't forget to get around to checking out my batch `lint` PR then. I use it constantly now!

PR #2214, opened 2025-11-18
feat: If I would've done→If I had done

hippietrail:
# Issues 
N/A

# Description

A linter for the very common mistake in American English:
"If I would've known I wouldn't've dont it" → "If I had known I wouldn't've done it"

And its common relative outside American English:
"If I had've known I wouldn't've done it".

It supports
- contractions: would've, had've
- misspelled/lazy contractions: wouldve, hadve
- non-contracted versions: would have, had have
- common mistakes: would of, had of

It will also match the same pattern which is not a mistake when it is not used in a counterfactual (known as "third conditional".) The user message is worded to guide the user when to apply it.

# How Has This Been Tested?

There are unit tests for all the variants listed about.
There are ignored unit tests for the false positives when not counterfactual so that future improvements to this linter know what to aim for.

# Checklist

### DIFF
--- a/harper-core/src/linting/if_wouldve.rs
+++ b/harper-core/src/linting/if_wouldve.rs
@@ -81,7 +81,7 @@ impl ExprLinter for IfWouldve {
 
             Some(Lint {
                 span,
-                lint_kind: LintKind::Usage,
+                lint_kind: LintKind::Nonstandard,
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
Looks good to me. Thanks!

PR #2241, opened 2025-11-25
test: tool to check the lists of linters in `lint_group.rs` and `mod.rs`

hippietrail:
# Issues 
N/A

# Description

I've been thinking of a tool to make sure all the linter lines in `linting/mod.rs` and `linting/lint_group.rs` are
- in alphabetical order
- not dupes
- use `struct` vs `expr` correctly
- etc.

This is a start on that. It's already helped me spot things that look wrong.

For instance:
```
mod bought;
mod call_them;
mod cant;
```
In `linting/mod.rs`. And in `linting/lint_group.rs`:
```
use super::bought::Bought;
use super::cant::Cant;
```
and:
```
        insert_expr_rule!(Bought, true);
        insert_expr_rule!(Cant, true);
```
No other mentions of `call_them` or `CallThem`!

Is it an experiment that got staged by accident with some other linter's PR? Or were the missing lines in those two files an unnoticed mistake? Note that these are kinds of errors that the Rust compiler and the unit tests do not pick up!

I actually thought there was another linter that overlapped with this `CallThem` that handled other pronoun or NP cases, but maybe that's one of my unfinished experiments? (-:

Note that I'm not sure the best way to check `insert_struct_rule` vs `insert_expr_rule` yet but when I started manually checking through the list I think I stopped after seeing at least four that were wrong.

Any idea or other feedback appreciated!

# Demo

Screenshot 2025-11-29 at 11 53 35 pm

image

# How Has This Been Tested?

Manually

# Checklist


justfile
```diff
@@ -685,3 +685,6 @@ runsnapshots:
 
   cd harper-core
   cargo test -- test_pos_tagger test_most_lints
+
+checklinters:
```
elijah-potter: Could we wire this into the `check-rust` recipe alongside the `auditdictionary` check?

hippietrail: Yep. I think I might have to update it to reflect some recentish-changes.
By the way, I think it doesn't find all cases due to some linters not using the standard boilerplate. Will look at that too.

Issue #2009, opened 2025-09-25
Solidify versioning plan and publish 1.0


imjoshin:
**What problem does this solve?**

We're currently at v0.60.0 / pre-1.0. To many, this can be a scary commitment to make in a product. I don't see any documentation about versioning methods, so I'm curious if anyone has any thoughts about harper being production-ready, or docs I can reference around versioning.

**Proposed Solution**

[As discussed in Discord](https://canary.discord.com/channels/1335035237213671495/1335036549976293448/1411795026802905270), there's nothing stopping this library from publishing a 1.0.

elijah-potter:
We plan to run a 1.0 release later this week.

elijah-potter:
Before we can, I'd love to get your sign off on the versioning policy (in #2257). What do you think, @imjoshin? Should anything be added or removed?

elijah-potter:
Are we good to close this?

elijah-potter:
Yup

PR #2257, opened 2025-11-26
docs: define a versioning policy

elijah-potter:
# Issues 

#2009

# Description

This PR introduces a versioning policy, viewable on the "About" page of the website.
@imjoshin, I'd love your thoughts.

### DIFF
--- a/packages/web/src/routes/docs/about/+page.md
+++ b/packages/web/src/routes/docs/about/+page.md
@@ -19,3 +19,23 @@ Harper takes advantage of decades of natural language research to analyze exactl
 If something is off, Harper lets you know.
 
 In a way, Harper is an error-tolerant parser for English.
+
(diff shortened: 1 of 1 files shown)


elijah-potter:
@imjoshin, thanks for taking a look!

 It seems to me, because we target consumers more directly than developers, that Harper's versioning needs are a little different from Gatsby's. If the synchronization becomes a practical problem, I'm happy to revise before the next breaking release.

imjoshin:
> @imjoshin, thanks for taking a look!
> 
>  It seems to me, because we target consumers more directly than developers, that Harper's versioning needs are a little different from Gatsby's. If the synchronization becomes a practical problem, I'm happy to revise before the next breaking release. 

That seems right to me. Thank you!

PR #2263, opened 2025-11-29
feat: more fast → faster etc.

### LINKED ISSUE #1509: Flag "more + adjective" when comparative exists
Here is a list that could be used to flag improper usage

 * more big
 * more small
 * more old
 * more young
 * more tall
 * more short
 * more long
 * more near
 * more far
 * more high
 * more low
 * more fast
 * more slow
 * more strong
 * more weak
 * more good
 * more bad
 * more happy
 * more easy
 * more heavy
 * more pretty
 * more busy
 * more funny
 * more nice
 * more fine
 * more late
 * more early
 * more cold
 * more warm
 * more cool
 * more dark
 * more light
 * more bright
 * more wide
 * more narrow
 * more thick
 * more thin
 * more deep
 * more shallow
 * more loud
 * more quiet
 * more hard
 * more soft
 * more sweet
 * more sour
 * more bitter
 * more clean
 * more dirty
 * more dry
 * more wet
 * more fair
 * more glad
 * more sad
 * more mad
 * more safe
 * more brave
 * more wise
 * more simple
 * more noble
 * more able
 * more gentle
 * more humble
 * more subtle
 * more clever
 * more cruel
 * more kind
 * more sharp
 * more dull
 * more quick
 * more dear
 * more cheap
 * more rich
 * more poor
 * more wild
 * more mild
 * more calm
 * more proud
 * more slim
 * more plump
 * more steep

_Originally posted by @ccoVeille in https://github.com/Automattic/harper/issues/508#issuecomment-2795127236_

This issue is somehow the opposite of

- https://github.com/Automattic/harper/issues/511

Both of them were identified when discussing about

- https://github.com/Automattic/harper/issues/508

hippietrail:
# Issues 
Resolves #1509 

# Description

Flags "more/most <adjective>" when the adjective has an inflected comparative/superlative form.

Note that this is not correcting an error and is not even always an improvement since native English speakers don't prefer some inflected adjective forms and they can seem clunky, awkward, or even wrong.

Because of this I've modified the `lint_descriptions_are_clean` test to turn off this one test. In the future it will likely make sense to replace this special case with an API etc. when there are other suggestions which won't mean that a description isn't clean.

I made `lint_descriptions_are_clean` in `lint_group.rs` easier to follow while I was at it by renaming variables and adding a nice comment.

While I was figuring that out, I rearranged the methods of `LintGroup` to group all the constructors together at the top before the other methods.

This linter's own `message` lets the user know it's not an error.

I also had to modify some dictionary entries that used the `>` annotation to make agent nouns from verbs ("time" → "timer" etc.) since they also assign the adjective property, resulting in false positives.

# How Has This Been Tested?

Unit tests should cover all the variants and known false positives. Most are made up but false positives come from GitHub.

# Checklist


harper-core/src/linting/lint_group.rs
```diff
@@ -815,21 +823,42 @@ mod tests {
...
+    /// 2. A linter is being too aggressive in flagging text that is actually correct English
+    ///    in the context of another linter's description.
     #[test]
     fn lint_descriptions_are_clean() {
-        let mut group = LintGroup::new_curated(FstDictionary::curated(), Dialect::American);
-        let pairs: Vec<_> = group
+        let lints_to_check = LintGroup::new_curated(FstDictionary::curated(), Dialect::American);
+
+        let mut enforcer_config = LintGroupConfig::new_curated();
+        enforcer_config.set_rule_enabled("MoreAdjective", false);
```
elijah-potter: I don't love the idea of disabling specific rules from this test. Our goal here is to make sure our descriptions are clean. If we want to ignore this rule becuase it's more of a stylistic choice, rather than a grammatical error, I think we should be ignoring specific `LintKind`s instead.

hippietrail: That's totally fair of course. But one problem is that it often feels more like fixing an actual grammatical error, as in "more big" or "more long", whereas it seems like a stylistic choice for words like "funner" or "oftener".

It might be best solved by a new annotation flag or two for adjectives?

hippietrail: That's totally fair of course. But one problem is that it often feels more like fixing an actual grammatical error, as in "more big" or "more long", whereas it seems like a stylistic choice for words like "funner" or "oftener".

It might be best solved by a new annotation flag or two for adjectives?

hippietrail: That's totally fair of course. But one problem is that it often feels more like fixing an actual grammatical error, as in "more big" or "more long", whereas it seems like a stylistic choice for words like "funner" or "oftener".

It might be best solved by a new annotation flag or two for adjectives?

hippietrail: A couple of other options:
- Make it depend on the `LintKind` with `LintKind::Style` not being used for checking descriptions. And documented as not considered errors.
- Make it depend on `priority` with priorities lower than a certain threshold not being linted in descriptions.
- Add a new field to `Lint` such as `not_considered_error: bool` or `enum Severity { Error, Suggestion }`
- Add a new constructor to `LintGroup` that specifically only enables linters which check for actual errors and disables any that are too stylistic.

elijah-potter:
Okay, I went and changed the `LintGroup` config to how I imagine it should be. I also eliminated a few false-positives by adjusting the dictionary. What do you think?

elijah-potter (approved):

PR #2266, opened 2025-11-30
fix: flag "piece of advise"

### LINKED ISSUE #2041: To advise/an advice
**The Error**
To advise someone is to give them advice. Advise, with an S, is the verb, while advice, with a C, is the noun.

**Examples (2-3):**
1. "Let me give you a piece of advise"
2. "I advice you to do otherwise"

**Potential Edge Cases**
None

hippietrail:
# Issues 
Resolves #2041

# Description

We now detect a verb being used as a noun if it's preceded by a preposition as well as an adjective.

This resolves example 1 of #2041. Example 2 is already flagged.

# How Has This Been Tested?

A unit test is added using the sentence from the issue.

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloca"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5a7d05ea6aea7e9e64d25b9156ba2fee3fdd659e34e41063cd2fc7cd020d7f4"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


harper-core/src/linting/noun_verb_confusion/verb_instead_of_noun.rs
```diff
@@ -24,7 +24,7 @@ impl Default for VerbInsteadOfNoun {
         Self {
             expr: Box::new(
                 SequenceExpr::default()
-                    .then(UPOSSet::new(&[UPOS::ADJ]))
+                    .then(UPOSSet::new(&[UPOS::ADJ, UPOS::ADP]))
                     .then_whitespace()
```
ccoVeille, line 28: Would it work for other "non countable plural"?

Furniture a piece of furniture
Information a piece of information
News a piece of news
Toast a piece of toast
Junk/Trash a piece of junk/trash
Land/Property a piece of land

Maybe it's a matter of adding affixes

hippietrail: I think you're confusing two properties of "advice". This fixes the linter which flags mixing it up with "advise".

There is already another linter which fixes treating "advice" as countable.

Feel free to stress test both of them and post any problems!

ccoVeille: You might be right, as often 🤪😅

elijah-potter (approved):

Issue #1449, opened 2025-06-24
Verb tense agreement/consistency


gogades:
## Description

Harper does not detect that verb tense usage is consistent and in agreement with the rest of the sentence.

## Resources

These links provides several good examples and explanations: [https://nmu.edu/writingcenter/tenses](https://nmu.edu/writingcenter/tenses) [https://owl.purdue.edu/owl/general_writing/grammar/verb_tenses/verb_tense_consistency.html](https://owl.purdue.edu/owl/general_writing/grammar/verb_tenses/verb_tense_consistency.html)

## Examples

The brown fox will ran thru the meadow. "will" and "ran" are not in agreement.

About noon the sky darkened, a breeze sprang up, and a low rumble announces the approaching storm. "darkened" and "sprang" are past tense, "announces" is present tense. 

When the show ended, the audience leaves the theater. "ended" and "leaves" are not in agreement.

## Not related
Harper is fantastic, I love it!

hippietrail:
Yes you are not the first to point this out. There are a couple of PRs coming shortly that start to address this.
It's actually trickier than you expect. I improved the noun and pronoun data quite a bit and verb metadata is already sketched out. After that it will probably just be a case of looking out for false positives and false negatives. 🤞

elijah-potter:
> Harper is fantastic, I love it!

Thanks!

> Harper does not detect that verb tense usage is consistent and in agreement with the rest of the sentence.

I'm going to have a crack at this today. I'll let you know how it goes.

elijah-potter:
It looks like we're going to need to go through and tag each verb in our dictionary. It's on my radar.

hippietrail:
> It looks like we're going to need to go through and tag each verb in our dictionary. It's on my radar.

At the moment the annotations for `S` `d` and `G` are used to generate regular 3rd person singular present, regular past, and progressive. So most words will already be marked.

Except for two things.

1. For irregular verbs there need to be an equivalent of each of those **plus** an extra one for past participle ("eaten" is different to "ate"). Off the top of my head I think we only have the one for 3rd person singular present.

2. I recently added or filled in the missing parts of the equivalent logic for adjective comparative and superlative forms and noun plural and mass/countable properties. In those cases to avoid having to add a new annotation to every single singular noun and positive degree adjective, I made is_singular and is_countable true either if marked so or if none of the contradicting annotations mark it as not so.

The same can be done for verbs.

hippietrail:
[...]

PR #2268, opened 2025-11-30
feat: will ran → will run / ran

hippietrail:
# Issues 
Example 1. of #1449

# Description

Work in progress towards correct wrong verb form after "will" or "shall".

# How Has This Been Tested?

Unit tests from the issue and from GitHub READMEs, not yet passing.

# Checklist

### DIFF
--- a/harper-core/src/linting/will_non_lemma.rs
+++ b/harper-core/src/linting/will_non_lemma.rs
@@ -1,8 +1,11 @@
 use crate::{
-    Lint, Token, TokenStringExt,
+    CharStringExt, Lint, Token, TokenStringExt,
     expr::{Expr, SequenceExpr},
     irregular_verbs::IrregularVerbs,
-    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk},
+    linting::{
+        ExprLinter, LintKind, Suggestion,
+        expr_linter::{Chunk, preceded_by_word},
+    },
     spell::Dictionary,
 };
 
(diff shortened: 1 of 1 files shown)


elijah-potter:
I found some false-positives: 

- "Around November 2023, shortly after the will was read, Eleanor started asking about using the cottage."
- "Vivian’s will was straightforward. The house, all its contents, and her savings were to be left to me."

I'm not sure what can be done to detect when "will" is being used as a noun. Thoughts?

hippietrail:
> I found some false-positives:
> 
> * "Around November 2023, shortly after the will was read, Eleanor started asking about using the cottage."
> * "Vivian’s will was straightforward. The house, all its contents, and her savings were to be left to me."
> 
> I'm not sure what can be done to detect when "will" is being used as a noun. Thoughts?

The immediate solution is to check for "will" and abort the rule. They're like fractals. You block all false positives at the cost of a smaller number of false negatives and repeat until there's either nothing left uncovered, or only really really marginal stuff (-: Now the immediate fix to check if will is a noun is to check if it's preceded by a determiner. I don't think possessive nouns are currently marked as determiners but they fill the same role. I think the best way to check for those would be to check `.is_apostrophized()` which I think I exposed up from `OrthFlags`.

I've been thinking about either making a module for POS disambiguation heuristics, or maybe one module per POS or POS pair that can be ambiguous. Or maybe a file somewhere in the meantime where we can collect ideas for a while and then use those ideas a bit later to design a module(s).

If you've ever played with parser generators, especially LALR ones have special sets of tokens. I know one is called "follow" which is the set of tokens that can gramatically follow the current token. I've been thinking about whether we can do something like that. Probably per-POS for starters but maybe it could be generalized to suffix/ngram of N previous POSes, or specialized to per word, especially for key words that are very common and have multiple POS, like "that" for instance.

elijah-potter:
Have you considered checking the Brill tagger's output? It may have flagged it correctly as a NOUN.

hippietrail:
> Have you considered checking the Brill tagger's output? It may have flagged it correctly as a NOUN.

Was able to find an even simpler fix.

elijah-potter:
Would you mind adding these as tests and making sure they work as expected?

`I also reminded her that the will specifies equal responsibility`

`She meticulously planned everything, and her will reflected that.`

`Now, his will stipulates everything is split 50/50.`

hippietrail:
> Now, his will stipulates everything is split 50/50.

Great tests! They initially failed but accommodating them was straightforward.

PR #2269, opened 2025-11-30
refactor: new `SequenceExpr` and `CharStringExt` methods

hippietrail:
# Issues 
N/A

# Description

While working on #1449 I noticed a pattern of `.then(|tok, src|)` calls that only need the `tok`'s `kind` and don't need the `src` at all. So I thought it might be worth having a `.then_` method that allows adding just a predicate for the `tok.kind`.

While thinking up a name for it I noticed `.then_strict()` also only acts on the `tok.kind` so I renamed it to `.then_kind()` and named my new method `.then_kind_where()`.

While applying it everywhere in the codebase that would benefit, I identified a few other patterns that would work with gaps in the set of `.then_kind_` methods, so I added those as well.

In turn, some benefited from filling in gaps in the `CharStringExt` methods, which I also added.

The result should be easier-to-read code without unused parameters and with more specific method names that document what's happening more clearly than `&&`, `||`, and parentheses on their own.

# Demo

# How Has This Been Tested?

I didn't add any new tests but I made sure all existing tests continue to pass.
(This does not mean nothing could've slipped through the cracks as I may have modified a path that lacked a test.)

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloca"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5a7d05ea6aea7e9e64d25b9156ba2fee3fdd659e34e41063cd2fc7cd020d7f4"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "allocator-api2"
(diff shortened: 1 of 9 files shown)


elijah-potter (approved):
This looks good to me. I don't see any obvious problems, and I think I'll be using these methods myself. Thanks!

PR #2285, opened 2025-12-02
Irregular verbs

hippietrail:
# Issues 
N/A

# Description

- Provides a module that handles irregular verbs.
- Adds a JSON file listing known irregular verbs.
- JSON file supports comments as strings to note why certain words are omitted, etc.
- JSON loading code is copied as closely as possible from the way that `dictionary.dict` and `annotations.json` are loaded.
- Comes with a converted `SimplePastToPastParticiple` linter that uses it.
- Will also be used with the in-progress `WillNonLemma` (#2268) and `ExtraneousDidPast` linters.
- Will provide a model for implementing a similar module for irregular noun plurals.

# How Has This Been Tested?

Unit tests are included for the one API so far needed.

# Checklist


harper-core/src/irregular_verbs.rs
```diff
@@ -0,0 +1,120 @@
+use lazy_static::lazy_static;
```
elijah-potter, line 1: This module is not directly related to linting. Could we instead move it next to the dictionary or perhaps simply higher up in the hierarchy?

harper-core/irregular_verbs.json
```diff
@@ -0,0 +1,127 @@
+[
+	"// comments can appear in the line before an entry",
```
elijah-potter, line 2: I adore the idea of having a dedicating file for this, similar to the dictionary. I'd like to centralize all data files, though. Would you mind moving this next to the `dictionary.dict`?

harper-core/src/irregular_verbs.rs
hippietrail: Most of the dictionary/annotation/metadata logic seems to be in `/src/` with some in `/spell/`. Since the latter would definitely be wrong I moved `irregular_verbs.rs` to `/src/` - and it's now joined by `irregular_nouns.rs`

harper-core/irregular_verbs.json
hippietrail: Done. And there's also `irregular_nouns.json` now.

harper-core/src/irregular_nouns.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        Ok(Self { nouns })
+    }
+
+    pub fn get() -> Arc<Self> {
+        (*NOUNS).clone()
+    }
+
+    pub fn get_plural_for_singular(&self, singular: &str) -> Option<&str> {
```
elijah-potter, line 57: I expect that we'll want to eventually have versions of these plugins that accept and return `char` arrays. I think we're fine for now, but it's something to think about.

harper-core/src/irregular_verbs.rs
```diff
@@ -0,0 +1,120 @@
...
+                // Strings are used for comments to guide contributors editing the file
+                serde_json::Value::String(_) => {}
+                _ => {}
+            }
+        }
+
+        Ok(Self { verbs })
+    }
+
+    pub fn get() -> Arc<Self> {
```
elijah-potter: Can we name this function similarly to the curated dictionary's equivalent?

harper-core/src/lib.rs
```diff
@@ -11,6 +11,8 @@ mod edit_distance;
 pub mod expr;
 mod fat_token;
 mod ignored_lints;
+pub mod irregular_nouns;
```
elijah-potter: Let's just export the structs directly. Having the extra layer of indirection is redundant.

harper-core/src/irregular_verbs.rs
hippietrail: I did start with the same `curated()` which I didn't think made sense for this. But then again they are actually curated!

harper-core/src/lib.rs
hippietrail: > Let's just export the structs directly. Having the extra layer of indirection is redundant.

At first I wasn't sure what this meant, and neither was the LLM. But have a look - I think I got it how you want it.

harper-core/src/irregular_nouns.rs
hippietrail: Yes I've been thinking of various things that can be added but adding when they're needed seems pragmatic.

hippietrail: Yes I've been thinking of various things that can be added but adding when they're needed seems pragmatic.

PR #2290, opened 2025-12-03
refactor: made a new `lint_group` from two mass noun linters

hippietrail:
# Issues 
N/A

# Description

I was investigating something our mass noun linting doesn't flag and realized there are two linters:
`MassPlurals` and `NounCountability`.

I've put them together into a new `lint_group`. It's the first one that takes an extra parameter (dictionary in this case) which makes it a good example for future contributors.

# How Has This Been Tested?

I added a couple of unit tests to the `mod.rs` to make sure the two member linters work via the new lint group,

# Checklist


harper-core/src/linting/mass_nouns/mod.rs
```diff
@@ -0,0 +1,93 @@
+mod mass_plurals;
+mod noun_countability;
+
+use mass_plurals::MassPlurals;
+use noun_countability::NounCountability;
+
+pub use merge_rule_hidden::MassNouns;
+
+mod merge_rule_hidden {
```
elijah-potter: Where does the need for module come from? Why not just expose the struct as-is?

hippietrail: > Where does the need for module come from? Why not just expose the struct as-is?

Still learning Rust and still learning Harper internals. Didn't have a good handle on all the hows and whys of `mod` used as it was inside the macro.

elijah-potter (approved):

PR #2297, opened 2025-12-04
feat(core): create rule to title-case headings

### LINKED ISSUE #108: feat: Title Case Conversion on a Code action.
Would be pretty cool to be able to convert any selected text into titles kinda like titlecaseconverter.com. Only issue is that there are a variety of standards commonly used. 

I'll take a look to see if I can pull this off myself in a week or so.

elijah-potter:
# Issues 

Sort-of resolves #108

# Description

For a long time, we've had a module for making text title-case. I've used it on [my blog](https://elijahpotter.dev) for quite some time.

I wanted to enforce title case on all my documents. To do that, I refactored the module to allocate less frequently and I've drafted a rule to make headings title case if they aren't already.

Note that I had to add a new `TokenKind` for heading starts. A heading is considered to be the length from the heading start token to the next paragraph break. I've modified the Markdown parser to emit these new tokens. 

Additionally, I've exposed the ability force an entire linting call in `harper.js` to be treated as a heading. I've used this functionality to allow specialized linting of header tags in the Chrome extension (see the second screenshot).

# Demo

image

image

image

# How Has This Been Tested?

Manually, plus many unit and integration tests.

# Checklist


lukasmwerner:
Amazing! Even better implementation that I originally had in mind!

PR #2299, opened 2025-12-05
feat: implement #1036 join prefix to word

### LINKED ISSUE #1036: Flag verb prefixes written as separate words or hyphenated
## Description

It's quite common these days for people to write prefixes of verbs as separate words.
It's probably more common to write them prefixed, but that's not always an error, only sometimes.

A decent implementation will check the word is a verb and check the combined form is also a verb. If the latter is not true then a prefix written separated can be changed to hyphenated.

### Separate word
- plugin, can't **pre transform** assets imported with query modifiers
- [can't **pre select** multiple options · Issue #2652](https://github.com/Dogfalo/materialize/issues/2652)
- [Can't **re import** python tensorflow module](https://stackoverflow.com/questions/52012012/cant-re-import-python-tensorflow-module)

### Hyphenated
- Can't **re-send** messages to users with safety number changes without marking them as verified.
- SynchronizedMap can't **de-serialized**
- [Can't **un-set** PIN code :( · Issue #221](https://github.com/atc1441/ATC_MiThermometer/issues/221)
- [[Feature][Config UI] **Dis-associate** the scope config with a ...](https://github.com/apache/incubator-devlake/issues/6251)

## Potential False Positives

There are things to watch out for to tell if hyphenated prefixes are OK or not.
For prefixes written as separate words it should be rarer. A few common prefixes may also be independent words. Some might rarely come before verbs in correct English?

hippietrail:
# Issues 

Resolves #1036 

# Description

Looks for prefixes that can be joined to their base word according to the dictionary.
If written with a space, it checks the dictionary to see if it can be hyphenated or joined.
If written with a hyphen it only checks if it can be joined.
It first checks if it's in the dictionary in its current form, in which case it won't be flagged.

Some prefixes are also words in their own right, and those tend to have more false positives. I've made it very easy to add new exceptions.

This is a really common occurrence so I had to adjust a number of existing tests and/or linter descriptions.

I added a couple of new constructors and shorthands to `SequenceExpr` and propagated the former through the other linters.

# Demo
image

# How Has This Been Tested?

Some unit tests are made up and some are from GitHub project READMEs.

# Checklist


harper-core/tests/run_tests.rs
```diff
@@ -68,7 +68,7 @@ macro_rules! create_org_test {
 }
 
 create_test!(whack_bullets.md, 1, Dialect::American);
-create_test!(preexisting.md, 0, Dialect::American);
```
elijah-potter, line 71: You should just remove this test and the underlying file.

harper-core/tests/run_tests.rs
```diff
@@ -86,7 +86,7 @@ create_test!(pr_452.md, 2, Dialect::American);
 create_test!(hex_basic_clean.md, 0, Dialect::American);
 create_test!(hex_basic_dirty.md, 1, Dialect::American);
 create_test!(misc_closed_compound_clean.md, 0, Dialect::American);
-create_test!(yogurt_british_clean.md, 0, Dialect::British);
+create_test!(yogurt_british_clean.md, 1, Dialect::British);
```
elijah-potter: You should edit the underlying file to remain clean.

hippietrail:
> I mistakenly hit "approve" when I meant to hit "request changes". Sorry for the noise.

No problem. Thanks for the guidance!

PR #2310, opened 2025-12-08
feat: more abbreviation expansions

hippietrail:
# Issues 
N/A

# Description

Adds a few more abbreviation expansions:
- decl → declaration, declarator
- fwd → forward
- param → parameter

args → argument was already linted but I added unit tests for it too.

But there is one problem:

The `lints_sol_clean_correctly` test on file `harper-comments/tests/language_support_sources/clean.sol` contains this line:
```
     * @param p This is a parameter
```
This `param` gets flagged. But should it? Is that line considered code or comment?
The test is expected to have no lints:
```
create_test!(clean.sol, 0);
```
But since it's clean I guess I shouldn't bump that to `1`
But if this `param` shouldn't be corrected I'm also reluctant to change the file to
```
     * @parameter p This is a parameter
```

Anyone have any advice?

# How Has This Been Tested?

A unit test from GitHub or elsewhere on the Intenet for each variant.

# Checklist


elijah-potter:
It looks like the `@param` notation is a kind of doc-comment. I wonder if this could be solved by using the `mark_inline_tags` function from the JsDoc parser as a post-processing step in the Solidity parser.

hippietrail:
> This looks good, but I don't want to merge this until we get that test passing. I don't want to break things for those writing Solidity with Harper.

Of course. Exactly why it's a draft. Thanks for the feedback!

hippietrail:
> This looks good, but I don't want to merge this until we get that test passing. I don't want to break things for those writing Solidity with Harper.

It turns out the Solidity parser was actually calling the Markdown parser instead of the JsDoc parser!
Found and fixed with the help of the free LLM in Windsurf.

Issue #233, opened 2024-10-19
Not working on writewithharper


ocrumbs:
![Screenshot from 2024-10-19 09-07-44](https://github.com/user-attachments/assets/3ea8a0e7-2da0-489b-942d-2140ecf41b82)

On the https://writewithharper.com page, I cannot get any suggestion for mistakes. May be I am using it wrong!

ocrumbs:
Thanks for the free software. Highly appreciated.

I really do wish to replace Grammarly. I work with lot of non English speaking developers (myself included). I want to suggest Harper to all of them, so I need you all to help me help you.

With other things you mentioned, it works great. I wish one day Harper replaces Grammarly. :muscle:

elijah-potter:
Hey @ocrumbs. Lead developer here.

I'm always trying to cover as many cases as possible on the algorithm side. To that end, we could use as much test data as possible. Since we don't collect directly from users, this becomes a little more difficult.

Would you mind sending us a copy of the various things you tried that Harper _didn't_ give great suggestions for? 

Thanks for trying it out. We'll get there.

ocrumbs:
Thanks @elijah-potter, let me compile that list. I will get back here.

ocrumbs:
These are the most recurring types grammatical pitfalls I have encountered.

![Image](https://github.com/user-attachments/assets/2d3fc9cb-9aa3-43ba-aae9-e5c3995d14d4)

hippietrail:
> [#433](https://github.com/Automattic/harper/issues/433) and [#485](https://github.com/Automattic/harper/issues/485) are related.

#485 is related but more specific. #433 is a fair bit more technical. This one here would mostly affect beginners just learning English, any wrong conjugation/form of verbs.

PR #2333, opened 2025-12-14
feat: fix "I walks" / "he walk" etc.

hippietrail:
# Issues 
Implements the first part of #233

# Description

Fixes disagreement of grammatical number between pronoun and verb:
- I walks → I walk
- He walk → He walks

- In the course of implementing this I found many annotations missing from the dictionary.
- I also added a bunch of missing irregular verbs.
- I added a couple of methods to `SequenceExpr` but the main one would be the new constructor `SequenceExpr::with()` which is a short way to express `SequenceExpr::default().then()`
- I propagated `::with()` through all the other linters I could find appropriate for it.

# Demo
image

# How Has This Been Tested?
- Unit tests covering the relevant examples in #233 
- Unit tests to cover positive and negative cases of every personal pronoun.
- Unit tests for known edge cases.

# Checklist


elijah-potter (changes requested):
This looks like very good work. Unfortunately, I found a few false positives that I'd love to see turn into passing tests:

- "On April 10th, I suggested she consider a smaller, more intimate gathering."
- "I suggested she sell it and use the proceeds to help with her relocation expenses, or perhaps rent a similar camera while in Barcelona."
- "I’ve offered her alternative options – suggesting she rent a similar cottage herself, or even offering to cover part of the cost – but she always deflects, insisting she deserves to use *this* cottage."
- "He donned his heavy oilskins and descended the winding staircase, his boots echoing in the hollow tower."

hippietrail:
> This looks like very good work. Unfortunately, I found a few false positives that I'd love to see turn into passing tests:
> 
>     * "On April 10th, I suggested she consider a smaller, more intimate gathering."
> 
>     * "I suggested she sell it and use the proceeds to help with her relocation expenses, or perhaps rent a similar camera while in Barcelona."
> 
>     * "I’ve offered her alternative options – suggesting she rent a similar cottage herself, or even offering to cover part of the cost – but she always deflects, insisting she deserves to use _this_ cottage."
> 
>     * "He donned his heavy oilskins and descended the winding staircase, his boots echoing in the hollow tower."

Excellent finds! Quite surprising that none of the texts I tested on used these very common constructions. I supposed it's more literary than what's in GitHub readmes.

The first three all involve a verb which makes the following "that" optional. Another one I could think of is "insisted". There are probably others that will reveal themselves.

The last one is an example of one of the words which is not an irregular verb but which has its forms spread over multiple `dictionary.dict` entries due to involving a consonant doubling as part of the ending changing. As we discussed related to `DerivedFrom`. The entry for the past form was lacking its annotation flags. There are probably still some others I've missed that will turn up.

elijah-potter:
I've found a few more false-positives. I think there may something more primitive at play here that can't be addressed by adding exceptions.

- "Surely, he cannot offer the same sum as the developers."
- "Am I the asshole for insisting she return the dress?"
- "It’s also important to recognize that your family's pride in you is a genuine reflection of your value."

hippietrail:
> I've found a few more false-positives. I think there may something more primitive at play here that can't be addressed by adding exceptions.

Yeah exceptions are kind of a last resort. Looking at this now... each is for a different reason:

1. Cannot was not marked as an auxiliary in the dictionary.
2. The verbs that take the subjunctive didn't allow present participle form.
3. Checking for a preposition before the pronoun will indicate that it's part of a clause and can take "is".

elijah-potter:
I'm very sorry to keep pushing this back. We're getting _very_ close, and I really want this rule to be excellent, particularly because it is so helpful. I was only able to find three more false-positives. 

- "She sought out Mrs. Hawthorne, the village’s oldest resident, a woman known for her vast knowledge of local history and her unsettlingly accurate intuition."
- "I admire your dedication to consistently drafting players who are actively trying to lose you points."
- "When I reiterated the conditions I'd previously set, she hung up on me."

I suspect these are simply small problems in the dictionary.

hippietrail:
> I'm very sorry to keep pushing this back. We're getting _very_ close, and I really want this rule to be excellent, particularly because it is so helpful. I was only able to find three more false-positives.
> 
> * "She sought out Mrs. Hawthorne, the village’s oldest resident, a woman known for her vast knowledge of local history and her unsettlingly accurate intuition."
> * "I admire your dedication to consistently drafting players who are actively trying to lose you points."
> * "When I reiterated the conditions I'd previously set, she hung up on me."
> 
> I suspect these are simply small problems in the dictionary.

No apologies needed! This is the best way - using real-world data to find the edges you can't anticipate.

The middle one is easily the hairiest for us! "lose you points" where "to lose" is a ditransitive verb that takes two objects. The first object, "you", is a pronoun which doesn't change form for subject vs object (like "it" but unlike every other personal pronoun). "Points" is here a plural noun but in form is also a 3rd person singular present verb". The 2nd person pronoun + 3rd person verb "you points" is the kind of thing this kind of linter looks for.

----
Actually it wasn't too hard. We can be sure there will be persistent edge cases. So put it through the wringer. Look for false negatives as well as false positives - but that is more difficult.

PR #2337, opened 2025-12-15
feat: oldest in the books→oldest in the book

### LINKED ISSUE #821: Flag "oldest * in the books" instead of "oldest * in the book"
## Description

Another one I just heard for the first time. I'd call it a hypercorrection adding more logic to an idiomatic phrase.

## Resources

Definitions of the correct form from real dictionaries

- [Collins](https://www.collinsdictionary.com/dictionary/english/its-the-oldest-trick-in-the-book)
- [Cambridge](https://dictionary.cambridge.org/dictionary/english/oldest-trick-in-the-book)

## Examples

- (Stack Overflow) - This is the **oldest Delphi mistake in the books** and I'm sure you've made it before (we all have), and I'm sure you recognise it when you see it.
- (Stack Overflow) - ... **oldest trick in the books**, a restart and it works all the times(for now).
- (GitHub) - // (The **oldest trick in the books** - TTY / TIOCSTI stuffing)
- (Stack Overflow) - Once the OS is started the MBR is typically protected for virus reasons - this is one of the **oldest virus tricks in the books** - goes back to ...
- (Stack Exchange) - Ok, I realized now that I was making the **oldest mistake in the books** with my code, dividing my v by 2 instead of dividing it by 5.
- (Stack Exchange) - One of the **oldest chromatic alterations in the books** is the raising of the leading tone in final cadences when the diatonic seventh degree
- (Slashdot) - **The oldest trick in the books**. Unix IS proven technology.
- (Slashdot) - He enables the **oldest tricks in the books**, create fear from thing like prosperity (we really don't need Foxconn?) 
- (Slashdot) - Isnt that like one of the **oldest military plays in the books**?

## Potential False Positives

If there is any grammatical context in which the error may lie (for example, a word may always be invalid if preceded by a preposition), let us know. Provide a few examples, if possible.

hippietrail:
# Issues 
Resolves #821 

# Description

Fixing the wrong variant of the idiom the **oldest** trick **in the books** to use the correct singular **book**.

# How Has This Been Tested?

All of the distinct examples from the feature request were turned into unit tests.

# Checklist

### DIFF
--- a/harper-core/src/linting/oldest_in_the_book.rs
+++ b/harper-core/src/linting/oldest_in_the_book.rs
@@ -65,6 +65,20 @@ impl ExprLinter for OldestInTheBook {
         src: &[char],
         _ctx: Option<(&[Token], &[Token])>,
     ) -> Option<Lint> {
+        let np = &toks[2..toks.len() - 4];
+        let tricky = np.iter().any(|n| {
+            n.span
+                .get_content(src)
+                .eq_any_ignore_ascii_case_str(&["trick", "tricks"])
+        });
+
+        let message = if tricky {
+            "This idiom should use singular `book` instead of plural `books`."
+        } else {
+            "If this is a play on the idiom `oldest trick in the book`, it should use singular `book` instead of plural `books`."
+        }
+        .to_string();
+
         Some(Lint {
             span: toks.last()?.span,
             lint_kind: LintKind::Usage,
@@ -72,7 +86,7 @@ impl ExprLinter for OldestInTheBook {
                 "book",
                 toks.last()?.span.get_content(src),
             )],
-            message: "This idiom should use singular `book` instead of plural `books`.".to_string(),
+            message,
             ..Default::default()
         })
     }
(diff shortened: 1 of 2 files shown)


harper-core/src/linting/oldest_in_the_book.rs
```diff
@@ -0,0 +1,152 @@
...
+            "Ok, I realized now that I was making the oldest mistake in the books with my code, dividing my v by 2 instead of dividing it by 5.",
+            OldestInTheBook::default(),
+            "Ok, I realized now that I was making the oldest mistake in the book with my code, dividing my v by 2 instead of dividing it by 5.",
+        );
+    }
+
+    #[test]
+    fn fix_chromatic_alterations() {
+        assert_suggestion_result(
+            "One of the oldest chromatic alterations in the books is the raising of the leading tone",
```
elijah-potter: Are there any cases that should be ignored? Where the rule does _not_ apply?

hippietrail: Good point. This sentence may be a literal use rather than a play on the idiom. I'm looking at hits not on GitHub to see broader patterns. I expect not to be able to computationally detect literal vs idiom except when the noun is "trick". In other cases the `message` should be something like`If this is a play on "the oldest trick in the book" ...`

PR #2348, opened 2025-12-16
feat(chrome-ext): show Harper version string in the popup

elijah-potter:
# Issues 

This was a problem reported in [Discord](https://discord.com/channels/1335035237213671495/1359600621979832520/1449962305621200966)

# Description

Apparently it isn't obvious which version of the Harper Chrome extension is installed. I've added it to the main popup menu, so it's easy to find.

# Demo

image

# How Has this Been Tested?

Manually.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/popup/Popup.svelte
+++ b/packages/chrome-plugin/src/popup/Popup.svelte
@@ -56,10 +56,10 @@ function openSettings() {
        }}><Fa icon={faArrowLeft}/></Button>
     {:else}
       <div>
-        <span class="text-sm font-mono">{version}</span>
         {#if versionMismatch}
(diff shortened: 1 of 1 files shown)


hippietrail:
As a tiny bonus feature you could query `https://writewithharper.com/latestversion` and put a little dot or emoji next to it if it's not the latest.

elijah-potter:
Great idea! I've added a little warning emoji if they're not equal, plus a little explanation in the title field.

hippietrail:
> Great idea! I've added a little warning emoji if they're not equal, plus a little explanation in the title field.

Was in my head since I was doing similar things when I was playing with `harper-core` in Tauri and some TUIs a few weeks ago.

elijah-potter:
Thanks for the suggestion and for taking a look!

PR #2363, opened 2025-12-19
feat: try to detect space in wrong spot between words

hippietrail:
# Issues 
N/A

# Description

While typing I noticed my fingers were sometimes getting out of sync and I was hitting the space bar a character too early or too late by one character.

This linter has two constructors:
- `new()` only matches if neither of the resulting words is in the dictionary.
- `sensitive()` matches if either of the resulting words is missing from the dictionary, but this results in a lot more false positives.

# How Has This Been Tested?

Unit tests for the space in both positions, for both the regular and "sensitive" patterns.

# Checklist

### DIFF
--- a/.github/workflows/just_checks.yml
+++ b/.github/workflows/just_checks.yml
@@ -39,7 +39,6 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: ".node-version"
-          package-manager-cache: false
       - name: Enable Corepack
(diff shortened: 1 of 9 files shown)


hippietrail:
> Looks great! It passes my testing. I can imagine using this extensively myself. Thanks!

As a bonus when it finds false positives for me they most often reveal oddities in the dictionary that should be fixed.

PR #2374, opened 2025-12-24
fix: warn about clashing linter names

### LINKED ISSUE #2343: There are no errors or warnings if two Linters share the same name.
**Describe the bug**

This could be considered a feature request but it's also kind of a bug.

Linters can be handwritten, in which case you see them in the `src/linting` directory when you start working on your new one.

But they can also be generated by macros, such as the `phrase_corrections` ones. In this case you don't see them when you start working on your new one and might inadvertently reuse a name already used.

If so there are no error or warning messages while compiling or testing.

Maybe your new linter will silently block the existing one from working.
Maybe the old linter will silently block your new one and you'll waste time in frustration until you figure out why.

**Expected behavior**

Stopping the build process would be best but I'm not sure if it would be possible?
Aborting during running with an error message output would be acceptable during development. Not sure if such code should also be present in the release build or if Rust has a way to include code only in dev builds?
Minimum would be to just emit an error warning and continue to run.

hippietrail:
# Issues 
Resolves #2343 

# Description

In fact this finds 3 pairs of clashing linters already!
- `Intact` - one in `ClosedCompounds`, the other in `phrase_corrections`.
- `Nobody` - one in `ClosedCompounds`, the other standalone.
- `OfCourse` - one standalone, the other in `phrase_corrections`.

I have intentionally not tried to resolve any of these as part of this PR.

I included another warning in this PR related to linters:

- In `harper-cli lint` if after processing any `--only` and/or `--ignore` commandline arguments there are no lints configured, a warning will be issued. This is most commonly the result of making a typo when testing a new linter you're working on.

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/harper-core/src/linting/its_contraction/general.rs
+++ b/harper-core/src/linting/its_contraction/general.rs
@@ -124,21 +124,15 @@ impl General {
                 && next_kind
                     .is_some_and(|kind| kind.is_upos(UPOS::SCONJ) || kind.is_upos(UPOS::PART))
         } else if modifier.kind.is_upos(UPOS::VERB) || modifier.kind.is_upos(UPOS::AUX) {
-            if next_non_whitespace_word(source, modifier.span.end).is_some_and(|word| {
-                (matches!(
-                    word.as_str(),
-                    "is" | "was" | "were" | "be" | "been" | "being" | "to"
-                )) && !strong_predicative_verbs.contains(&modifier_lower.as_str())
-            }) {
-                false
-            } else if next_kind.is_some_and(|kind| {
-                (kind.is_noun() || kind.is_proper_noun())
-                    && !strong_predicative_verbs.contains(&modifier_lower.as_str())
-            }) {
-                false
-            } else {
-                true
-            }
+            let blocks_contraction = !strong_predicative_verbs.contains(&modifier_lower.as_str())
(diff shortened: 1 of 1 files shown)


elijah-potter (changes requested):
I like the idea, but I don't love having these methods do IO of any kind. Would it be possible to refactor this into a test case on the curated `LintGroup`?

hippietrail:
> I like the idea, but I don't love having these methods do IO of any kind. Would it be possible to refactor this into a test case on the curated `LintGroup`?

Putting this straight in my inbox for today!

hippietrail:
> I like the idea, but I don't love having these methods do IO of any kind. Would it be possible to refactor this into a test case on the curated `LintGroup`?

I made it a test by adding an `Option<Vec>` to `LintGroup` for linter names that clash. This way I was able to keep references to `LintGroup` as they were.

I took the opportunity to refactor the curated linter constructor too by adding new macros to add linters that use either a dialect or dictionary. But not for `SpellCheck` which, for now, is the only one that takes both.

I **did not** remove any of the three clashing `Linter`s. Up to you if you want to roll that into this PR when you test it, or do it as a separate PR after this one.

Until then this PR will show up as failing of course!

PR #2390, opened 2025-12-28
fix: don't change 'you' to 'your' if the previous word is a verb

### LINKED ISSUE #1919: "You" flagged to change to "your" when it's the legit indirect object of the verb
**What got flagged?**

Image

**Why is this incorrect?**
I'm not sure if this happens for all verbs that take indirect objects. There may be a subcategory of those where the indirect object is a kind of "beneficiary".

hippietrail:
# Issues 
Resolves #1919

# Description

The `PossessiveYour` linter was flagging "you" to change to "your" even when it was the object of a verb.

# How Has This Been Tested?

The examples from the bug report are used as unit tests.

As a great bonus a bunch of false positives from the snapshots go away and no new false positives or negatives are introduced there.

# Checklist

### DIFF
--- a/harper-core/src/linting/possessive_your.rs
+++ b/harper-core/src/linting/possessive_your.rs
@@ -33,7 +33,20 @@ impl ExprLinter for PossessiveYour {
         self.expr.as_ref()
     }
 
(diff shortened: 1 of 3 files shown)


elijah-potter (approved):
Looks perfect. Thanks!

PR #2393, opened 2025-12-29
refactor(core): simplify code

86xsk:
# Description

Replaces what seems to be reimplementations of `split_inclusive` with their standard library equivalent.

# How Has This Been Tested?

- `cargo test`

# Checklist


elijah-potter (approved):
Good catch! TIL about the `split_inclusive` method. I wouldn't be surprised if this had performance benefits as well. Thanks!

PR #2400, opened 2025-12-31
feat: number agreement in this/these type(s) of thing(s)

hippietrail:
# Issues 
N/A

# Description

I noticed there's a lot of variation in which of the three words which have grammatical number agree in constructions like:
- this kind of thing
- those types of things

Ideally it should work for arbitrary nouns.
Since noun phrases are pretty tricky to detect and since noun phrases don't usually start with a plural noun, it's currently restricted to plural nouns and the word "thing" in its singular or plural form. For now.

# How Has This Been Tested?

Unit tests for many variations, mostly from GitHub.

# Checklist

### DIFF
--- a/harper-core/src/linting/this_type_of_thing.rs
+++ b/harper-core/src/linting/this_type_of_thing.rs
@@ -21,7 +21,7 @@ impl Default for ThisTypeOfThing {
                         ])
                         .t_ws(),
                     )
-                    .then(SequenceExpr::word_set(&["of", "of"]).t_ws())
+                    .then(SequenceExpr::aco("of").t_ws())
                     .then_any_of(vec![
                         // "thing" is common in this construction and won't be part of a compound noun.
                         Box::new(WordSet::new(&["thing", "things"])),


harper-core/src/linting/this_type_of_thing.rs
```diff
@@ -0,0 +1,257 @@
...
+            expr: Box::new(
+                SequenceExpr::word_set(&["this", "these", "that", "those"])
+                    .t_ws()
+                    .then(
+                        SequenceExpr::word_set(&[
+                            "kind", "kinds", "sort", "sorts", "type", "types",
+                        ])
+                        .t_ws(),
+                    )
+                    .then(SequenceExpr::word_set(&["of", "of"]).t_ws())
```
elijah-potter: Why are there two "of" entries?

hippietrail: Probably a brain fart on my part by not noticing something silly the code completion AI did when hitting tab. Grr.

PR #2401, opened 2025-12-31
fix: `Handful` was in `linting/mod.rs` but missing from `linting/lint_group.rs`

### LINKED ISSUE #2366: `Handful` seems not to be integrated
**Describe the bug**

While starting work on a new linter I noticed that `Handful` is in `harper-core/src/linting/mod.rs` but absent from `harper-core/src/linting/lint_group.rs`

When I run the tests in `handful.rs` by clicking on "Run Tests" in VS Code, all the tests pass.

Image

But if I paste all the tests into `https://writewithharper.com/` none of them are flagged (except the spelling mistake in the last one, which is flagged by `SpellCheck` and not by `Handful`.

Maybe it's disabled intentionally? I made this issue rather than a PR just in case.

Perhaps we need some kind of test to check for this kind of thing? I think I found one a couple of weeks ago but didn't get around to reporting it, though I think I brought it up on Discord where nobody noticed it.

hippietrail:
# Issues 
Closes #2366 

# Description

Last week I discovered an orphaned `Linter` that was only half-integrated and filed an issue about it.
With no feedback I'm assuming that was an oversight and it should be fully integrated.

# How Has This Been Tested?

The linter already had unit tests in it.

# Checklist

### DIFF
--- a/harper-core/src/linting/lint_group.rs
+++ b/harper-core/src/linting/lint_group.rs
@@ -75,6 +75,7 @@ use super::free_predicate::FreePredicate;
 use super::friend_of_me::FriendOfMe;
 use super::go_so_far_as_to::GoSoFarAsTo;
 use super::good_at::GoodAt;
+use super::handful::Handful;
 use super::have_pronoun::HavePronoun;
 use super::have_take_a_look::HaveTakeALook;
 use super::hedging::Hedging;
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
Yup. That was an oversight. Thanks!

PR #2404, opened 2026-01-01
feat(cli): recognize *.markdown as Markdown

scop:
# Issues 

N/A

# Description

.markdown is the other official extension for Markdown.

https://www.iana.org/assignments/media-types/text/markdown

# Demo

N/A

# How Has This Been Tested?

Ran the CLI on a few *.markdown files, noted that it now does the right thing with Markdown links wrt AnA.

# Checklist

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -937,7 +937,7 @@ fn load_file(
         .extension()
         .map(|v| v.to_str().unwrap())
     {
-        Some("md") => Box::new(Markdown::default()),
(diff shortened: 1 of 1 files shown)


elijah-potter (changes requested):
You'll also need to make the necessary changes to `harper-ls`, specifically in `harper-ls/src/backend.rs`.

This is a good catch. I wasn't aware that was official. Thanks!

elijah-potter (approved):
I just remembered that the relevant section of `harper-ls` was refactored a while ago. It doesn't need to be modified after all. I apologize for my blunder. I'll merge this right away. Thanks!

PR #2437, opened 2026-01-07
fix: don't flag "need to run into"

### LINKED ISSUE #2433: Incorrectly flagged: "So they don't `need to run` into this problem"
**What got flagged?**
> So that they don't need `to` run into this problem in the future.

Harper reports: "need to should be followed by a verb, not a noun or pronoun like run."

**Why is this incorrect?**
I'm pretty sure this is a grammatically correct sentence, using `need to run into`. So I think Harper is incorrect here.

From the error message, it seems like Harper thinks `run` is a noun, but I'm pretty sure `run` is an intransitive verb.

**Example of correct usage:**
Harper should allow `need to run into`.

## Harper version
This happens with the Harper Firefox extension 1.3.0

---

Oh and thanks for developing this awesome tool! This is super promising, and I hope more people get involved to make it even better in the coming years!

hippietrail:
# Issues 

Fixes #2433

# Description

Fixes false positive "need to run into" from being flagged.

# How Has This Been Tested?

A unit test using the example in the issue.

# Checklist

### DIFF
--- a/harper-core/src/linting/need_to_noun.rs
+++ b/harper-core/src/linting/need_to_noun.rs
@@ -24,7 +24,7 @@ impl Default for NeedToNoun {
                     || tok.kind.is_unlintable()
                     || tok.kind.is_pronoun()
             }),
-            Box::new(WordSet::new(&["about", "it"])),
+            Box::new(WordSet::new(&["about", "into", "it"])),
         ]);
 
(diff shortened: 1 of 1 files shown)


Greenheart (approved):
Thanks for the quick fix! :)

Works well for me when testing this patched version locally.

PR #2443, opened 2026-01-08
feat: Make the dialect indicator in the VS Code status bar clickable to change dialect

### LINKED ISSUE #1095: Make the dialect indicator in the VS Code status bar clickable to change dialect
## Description

The status bar item for Harper currently is read-only displaying the country code and flag of the country whose dialect is active: 

Image

VS Code allows these status bar items to be clickable. It would be great if clicking it allowed you to change the active dialect.

## Component

Which component of Harper would this feature affect? For example:
- VS Code extension

## Use Cases

For now mostly for people working on `dictionary.dict` dialect annotations probably?

hippietrail:
# Issues 

Closes #1095

# Description

Simply click the image indicator in the VS Code status bar and a popup will appear where you can choose a different dialect. You no longer have to open the settings.

It took me longer to figure out how to build and test this than it did for the Amp Free agent to implement it.

# Demo

image

# How Has this Been Tested?

Manually in both VS Code and Windsurf

# Checklist

### DIFF
--- a/packages/vscode-plugin/src/extension.ts
+++ b/packages/vscode-plugin/src/extension.ts
@@ -1,5 +1,5 @@
-import type { ExtensionContext, QuickPickItem } from 'vscode';
-import { ConfigurationTarget, commands, StatusBarAlignment, type StatusBarItem, Uri, window, workspace } from 'vscode';
+import type { ExtensionContext, QuickPickItem, StatusBarItem } from 'vscode';
+import { ConfigurationTarget, commands, StatusBarAlignment, Uri, window, workspace } from 'vscode';
 import type { Executable, LanguageClientOptions } from 'vscode-languageclient/node';
 import { LanguageClient, ResponseError, TransportKind } from 'vscode-languageclient/node';
 
@@ -100,9 +100,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
 		commands.registerCommand('harper.languageserver.restart', startLanguageServer),
 	);
 
-	context.subscriptions.push(
-		commands.registerCommand('harper.changeDialect', changeDialect),
-	);
+	context.subscriptions.push(commands.registerCommand('harper.changeDialect', changeDialect));
 
 	await startLanguageServer();
 


mcecode (changes requested):
LGTM.

Just one thing, can you add something like the following to the `contributes.commands` array in `package.json` so the change dialect command can also be run directly through the command palette:

```json
{
	"command": "harper.changeDialect",
	"title": "Harper: Change Dialect",
	"description": "Change the dialect Harper uses for spellchecking"
}
```

I think it can be useful for more keyboard-oriented users.

packages/vscode-plugin/src/extension.ts
```diff
@@ -195,6 +198,23 @@ function updateDialectStatusBar(): void {
...
+	const dialectNames = ['American', 'British', 'Australian', 'Canadian', 'Indian'];
+	const dialects: QuickPickItem[] = dialectNames.map((name) => ({
+		label: name,
+	}));
+
+	const selected = await window.showQuickPick(dialects, {
+		placeHolder: 'Select Harper dialect',
+	});
+
+	if (selected && typeof selected !== 'string') {
```
mcecode, line 211: Oh, one question/possible change. Why do you need to check the type of `selected` here? As far as I checked, selected would either be an `object` in the shape of `{ label: "Dialect" }` or `undefined`, so shouldn't `if (selected) { ... }` suffice?

hippietrail: > Oh, one question/possible change. Why do you need to check the type of `selected` here? As far as I checked, selected would either be an `object` in the shape of `{ label: "Dialect" }` or `undefined`, so shouldn't `if (selected) { ... }` suffice?

That's probably me trusting the AI assistant a bit because I don't do enough TypeScript to be really confident with the type wrangling stuff. I appreciate this kind of feedback!

hippietrail:
> LGTM.
> 
> Just one thing, can you add something like the following to the `contributes.commands` array in `package.json` so the change dialect command can also be run directly through the command palette:
> 
> ```json
> {
> 	"command": "harper.changeDialect",
> 	"title": "Harper: Change Dialect",
> 	"description": "Change the dialect Harper uses for spellchecking"
> }
> ```
> 
> I think it can be useful for more keyboard-oriented users.

What a great idea! I'm not as keyboard-oriented as I used to be.

mcecode:
One last thing, it's kind of out of scope for this PR, but it's a simple change, and since we're updating the extension anyway, I thought I'd bring it up. I noticed while testing that we missed to add "Indian" as a valid value for the `harper.dialect` config in `package.json`. This means that it doesn't appear in the settings UI and that it's considered an invalid value in `settings.json` files. Can you add it like so in `package.json` to address that?

```jsonc
"harper.dialect": {
	"scope": "resource",
	"type": "string",
	"enum": [
		"British",
		"American",
		"Canadian",
		"Australian",
		"Indian" // This one
	],
	"default": "American",
	"description": "Set which dialect of English Harper should expect."
},
```

hippietrail:
> One last thing, it's kind of out of scope for this PR, but it's a simple change, and since we're updating the extension anyway, I thought I'd bring it up. I noticed while testing that we missed to add "Indian" as a valid value for the `harper.dialect` config in `package.json`. This means that it doesn't appear in the settings UI and that it's considered an invalid value in `settings.json` files. Can you add it like so in `package.json` to address that?

Dammit! I thought I was so very thorough looking at all the places the original dialects were used when I made the Indian English PR. But apparently this is the second spot I missed! - You're right that it should be its own PR imho.

hippietrail:
@mcecode I tried. I really, really tried. But I ended up frustrated and hating programming. See this discord thread. It's probably trivially easy for anyone else to see what I'm doing wrong: https://discord.com/channels/1335035237213671495/1335036505181257739/1459150365667627174

elijah-potter:
> Dammit! I thought I was so very thorough looking at all the places the original dialects were used when I made the Indian English PR. But apparently this is the second spot I missed! - You're right that it should be its own PR imho.

I missed it too. Might need a better solution if we keep adding dialects. 

Is this PR ready to merge?

hippietrail:
> One last thing, it's kind of out of scope for this PR, but it's a simple change, and since we're updating the extension anyway, I thought I'd bring it up. I noticed while testing that we missed to add "Indian" as a valid value for the `harper.dialect` config in `package.json`. This means that it doesn't appear in the settings UI and that it's considered an invalid value in `settings.json` files. Can you add it like so in `package.json` to address that?
> 
> ```json-comments
> "harper.dialect": {
> 	"scope": "resource",
> 	"type": "string",
> 	"enum": [
> 		"British",
> 		"American",
> 		"Canadian",
> 		"Australian",
> 		"Indian" // This one
> 	],
> 	"default": "American",
> 	"description": "Set which dialect of English Harper should expect."
> },
> ```

Hey would you be able to make a dedicated PR for this fix? I got frustrated and stalled out trying to test your other suggestion and haven't come back to it yet.

hippietrail:
> > Dammit! I thought I was so very thorough looking at all the places the original dialects were used when I made the Indian English PR. But apparently this is the second spot I missed! - You're right that it should be its own PR imho.
> 
> I missed it too. Might need a better solution if we keep adding dialects.

That's worthy of filing an issue to encapsulate the Dialect better. I remember experimenting with something like that back when Dialect was first added.

> Is this PR ready to merge?

My last session on this was such a mood killer that I didn't come back to it yet and didn't want to include @mcecode 's idea without being able to test it.

It is ready to merge without that though and one of us can add that in a separate PR with appropriate testing.

mcecode:
Hi, sorry it took me a while to get back to this.

---

> Is this PR ready to merge?

I think this PR is okay to merge since my review only requested for improvements which can be added later.

---

@hippietrail, I'm not really on social media these days so I haven't been able to check the Discord link you sent. Will try to make an account soon so I can view it and do comms there too. Hopefully I can help you or I can make the PR myself at some point.

hippietrail:
> Hi, sorry it took me a while to get back to this.

No problem at all. Sorry about my frustration affecting this issue!

> > Is this PR ready to merge?
> 
> I think this PR is okay to merge since my review only requested for improvements which can be added later.
> 
> @hippietrail, I'm not really on social media these days so I haven't been able to check the Discord link you sent. Will try to make an account soon so I can view it and do comms there too. Hopefully I can help you or I can make the PR myself at some point.

Very understandable and wise! I plan to try again myself but I want to allow a time and space where if it leads to frustration again it won't stop me coding in a huff (-:

PR #2452, opened 2026-01-09
chore: +vibe coding terms; add `/` annotation for phrasal verbs

hippietrail:
# Issues 
N/A

# Description

This is a step toward mixing the phrasal verbs in with the other multi-word entries, which will be, in-turn, a step toward moving those in with the general entries - after more tweaks to the annotation flags.

⚠️ The use of `/` as a flag is tentative - because it's the same as the separator character between the word and the annotations it might break external dictionary parsing tools.

I can always change this to another character but in the meantime think of it as fuzz-testing.

# How Has This Been Tested?

All current tests in `cargo test` pass.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -37288,7 +37288,7 @@ overwrite/VGSN
 overwritten/VT
 overwrote/Vt
 overwrought/VJ
-overzealous/J
+overzealous/JY
 oviduct/NSg
 oviparous/J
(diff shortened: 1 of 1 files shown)


hippietrail:
> I like this! In an odd way, I feel that using the `/` tag is thematically appropriate for the usage here.

I try possibly too hard to always choose something mnemonic for new annotation flags.

PR #2458, opened 2026-01-11
refactor(core): merge semantically identical funcs

86xsk:
# Description

Merges two functions which seem to have an identical purpose.

# How Has This Been Tested?

- `cargo test`

# Checklist


elijah-potter (approved):
Makes sense to me. Good catch!

PR #2461, opened 2026-01-11
feat: Add get_rel() to get a token relative to the end of a slice

hippietrail:
# Issues 
N/A

# Description

- Add TokenStringExt::get_rel() for clean access to tokens near the end of the slice
- Update several linters to use get_rel()

The doctest was generated by the Windsurf LLM because I don't have a feel for how to write them yet - please critique if it's no good!

# How Has This Been Tested?

Rather than adding any new unit tests, I found places where existing linters look for tokens relative to the end of the token slice, then checking that all the unit tests in those linters still pass.

# Checklist

### DIFF
--- a/harper-core/src/token_string_ext.rs
+++ b/harper-core/src/token_string_ext.rs
@@ -94,12 +94,14 @@ pub trait TokenStringExt: private::Sealed {
     ///
     /// # Examples
     /// ```
-    /// # use harper_core::{Token, TokenStringExt};
+    /// # use harper_core::{Token, TokenStringExt, parsers::{Parser, PlainEnglish}};
     /// # fn main() {
-    /// let tokens = []; // In a real test, this would be actual tokens
-    /// let _last = tokens.get_rel(-1);  // Gets the last token
-    /// let _third_last = tokens.get_rel(-3);  // Gets the third last token
-    /// let _first = tokens.get_rel(0);  // Gets the first token
+    /// let source = "The cat sat on the mat.".chars().collect::<Vec<_>>();
+    /// let tokens = PlainEnglish.parse(&source);
+    /// assert_eq!(tokens.get_rel(0).unwrap().span.get_content_string(&source), "The");
+    /// assert_eq!(tokens.get_rel(1).unwrap().kind.is_whitespace(), true);
+    /// assert_eq!(tokens.get_rel(-1).unwrap().kind.is_punctuation(), true);
+    /// assert_eq!(tokens.get_rel(-2).unwrap().span.get_content_string(&source), "mat");
     /// # }
     /// ```
     ///
@@ -118,11 +120,7 @@ pub trait TokenStringExt: private::Sealed {
             return None;
         }
 
-        let idx = if index >= 0 {
-            index as usize
-        } else {
-            (len + index) as usize
-        };
+        let idx = if index >= 0 { index } else { len + index } as usize;
 
         slice.get(idx)
     }


harper-core/src/token_string_ext.rs
```diff
@@ -90,6 +90,43 @@ pub trait TokenStringExt: private::Sealed {
...
+    /// ```
+    /// # use harper_core::{Token, TokenStringExt};
+    /// # fn main() {
+    /// let tokens = []; // In a real test, this would be actual tokens
+    /// let _last = tokens.get_rel(-1);  // Gets the last token
+    /// let _third_last = tokens.get_rel(-3);  // Gets the third last token
+    /// let _first = tokens.get_rel(0);  // Gets the first token
+    /// # }
+    /// ```
+    ///
```
86xsk, line 107: I would consider using some asserts here, so that the example code would function as a test and ensure correct behavior of the function.

Perhaps something like:
```rust
    /// # Examples
    /// ```
    /// # use harper_core::{Token, TokenStringExt, parsers::{PlainEnglish, Parser}};
    /// # fn main() {
    /// let source = "Insert sentence here".chars().collect::<Vec<_>>();
    /// let tokens = PlainEnglish.parse(&source);
    /// assert_eq!(tokens.get_rel(-1).unwrap().span.get_content_string(&source), "here");  // Gets the last token
    /// assert_eq!(tokens.get_rel(-3).unwrap().span.get_content_string(&source), "sentence");  // Gets the third last token
    /// assert_eq!(tokens.get_rel(0).unwrap().span.get_content_string(&source), "Insert");  // Gets the first token
    /// # }
    /// ```
```

(Just a rough draft; It could probably be cleaned up a bit.)

harper-core/src/token_string_ext.rs
```diff
@@ -90,6 +90,43 @@ pub trait TokenStringExt: private::Sealed {
...
+    fn get_rel(&self, index: isize) -> Option<&Token>
+    where
+        Self: AsRef<[Token]>,
+    {
+        let slice = self.as_ref();
+        let len = slice.len() as isize;
+
+        if index >= len || -index > len {
+            return None;
+        }
```
86xsk, line 121: I'm not 100% sure, but I think this check might be redundant since `get()` will just return `None` if the index is invalid anyway.

harper-core/src/token_string_ext.rs
```diff
@@ -90,6 +90,43 @@ pub trait TokenStringExt: private::Sealed {
...
+
+        let idx = if index >= 0 {
+            index as usize
+        } else {
+            (len + index) as usize
+        };
```
86xsk: Not sure if it would help readability or not, but you could slightly reduce the size of the code by doing the `as usize` cast on the result of the `if` expression, just before semicolon :p

harper-core/src/token_string_ext.rs
hippietrail: > I'm not 100% sure, but I think this check might be redundant since `get()` will just return `None` if the index is invalid anyway.

`get()` will return `None` if the index is too high but since it doesn't/can't when the index it too low it's nice to fail early and avoid indentation and no extra effort to check low and high together.

That's how I analyzed it when the AI rewrote my original version. I thought it looked more "API-ish" than what I wrote (-:

harper-core/src/token_string_ext.rs
hippietrail: > I would consider using some asserts here, so that the example code would function as a test and ensure correct behavior of the function.
> 
> Perhaps something like:
> 
> ...
> 
> (Just a rough draft; It could probably be cleaned up a bit.)

Great idea! It's about time I learn how to do these well.

harper-core/src/token_string_ext.rs
hippietrail: > Not sure if it would help readability or not, but you could slightly reduce the size of the code by doing the `as usize` cast on the result of the `if` expression, just before semicolon :p

That's the kind of thing I like to do in my code too. I think when I wrote it myself if was more dense and I just liked the clean look of the AI's rewrite...

elijah-potter (approved):
Looks good to me! This is already proving useful. Thanks!

PR #2473, opened 2026-01-13
perf(test): improve test performance

86xsk:
# Description

Reduces the amount of time it takes to run `cargo test`.

Before rebasing on master and with only the first two commits, this reduced the amount of time it takes to run `cargo test` from ~78s to ~57s on my system. This is primarily achieved by not running the chunker/tagger when creating the `Document` in `FixedPhrase::from_phrase()`.

After rebasing, the `does_not_oom_with_repeated_lints()` test seems to have regressed significantly in performance, and running `cargo test` took about ~180s. I took a quick crack at trying to resolve this in 0659fab16affd9b7d84b4fc7aa0c5da842f98bfc. This reduced the `cargo test` time to ~53s, albeit I'm not 100% certain that the test will still work as intended (I'm not super familiar with Wasm).

(The regressions in `does_not_oom_with_repeated_lints()` seemed to have been caused by `lex_email_address()` and `lex_url()` and the way they're used in some manner related to Weir. It seemed to be caused by the iterators at the beginning of those functions trying to find the ':' for the URL, or the '@' for the email. Replacing those iterators with a loop and adding early exits seemed to significantly reduce the impact of this. Without any changes that test took >100s, with those early exits it took ~37s, and with 0659fab16affd9b7d84b4fc7aa0c5da842f98bfc it takes about ~1.5s. I didn't spend too much time investigating why this became so much of an issue after Weir was added, but I think it could be an interesting direction for further optimization.)

# How Has This Been Tested?
- `cargo test`

# Checklist

### DIFF
--- a/harper-wasm/src/lib.rs
+++ b/harper-wasm/src/lib.rs
@@ -647,6 +647,19 @@ mod tests {
     use super::*;
     use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate, System};
 
+    /// Get memory usage for the process with the given PID, in bytes.
+    ///
+    /// This will fail if the requested process does not exist.
+    #[must_use]
+    fn get_mem_usage_of_process(sys: &mut System, pid: Pid) -> Option<u64> {
+        sys.refresh_processes_specifics(
+            ProcessesToUpdate::Some(&[pid]),
+            false,
+            ProcessRefreshKind::nothing().with_memory(),
+        );
+        sys.process(pid).map(|process| process.memory())
+    }
+
     /// If a word from another dialect is added to the user dictionary, it should be considered
     /// part of the user's dialect as well.
     #[test]
@@ -666,18 +679,7 @@ mod tests {
         let pid = Pid::from_u32(std::process::id());
         let mut sys = System::new();
 
-        macro_rules! refresh_sys {
-            () => {
-                sys.refresh_processes_specifics(
-                    ProcessesToUpdate::Some(&[pid]),
-                    false,
-                    ProcessRefreshKind::nothing().with_memory(),
-                );
-            };
-        }
-
-        refresh_sys!();
-        let mut prev_memory_usage = sys.process(pid).unwrap().memory();
+        let mut prev_memory_usage = get_mem_usage_of_process(&mut sys, pid).unwrap();
 
         if (0..10).all(|_| {
             // Run a few times.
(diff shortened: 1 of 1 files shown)


harper-core/src/document.rs
```diff
@@ -88,6 +88,19 @@ impl Document {
         Self::new(text, &PlainEnglish, &FstDictionary::curated())
     }
 
+    /// Create a new document simply by tokenizing the provided input and applying fix-ups. The
+    /// contained words will not contain any metadata.
+    ///
+    /// This avoids running potentially expensive metadata generation code, so this is more
+    /// efficient if you don't need that information.
+    pub fn new_basic_tokenize(text: &str, parser: &impl Parser) -> Self {
```
elijah-potter: I don't want to expose this function to the broader ecosystem right now. Could we mark it as `pub(crate)`?

harper-wasm/src/lib.rs
```diff
@@ -661,17 +662,44 @@ mod tests {
...
 
     #[test]
-    fn does_not_oom_with_repeated_lints() {
-        for _ in 0..3000 {
-            let mut linter = Linter::new(Dialect::American);
+    fn no_memory_leak_with_repeated_lints() {
+        let pid = Pid::from_u32(std::process::id());
+        let mut sys = System::new();
+
+        macro_rules! refresh_sys {
```
elijah-potter: Is there a particular reason this is a macro instead of a function?

harper-wasm/src/lib.rs
```diff
@@ -661,17 +662,44 @@ mod tests {
     }
 
     #[test]
-    fn does_not_oom_with_repeated_lints() {
-        for _ in 0..3000 {
-            let mut linter = Linter::new(Dialect::American);
+    fn no_memory_leak_with_repeated_lints() {
```
elijah-potter, line 678: I really like this refactor. It appears to be way more reliable and obviously much faster.

elijah-potter (changes requested):
Overall, this is great work. There are a few questions I need answered before I can merge this. Thanks!

harper-wasm/src/lib.rs
86xsk: I refactored it into a function inside that `test` module. If there's a better place for it, please let me know.

(iirc I initially tried making it a closure that would capture `pid` and `sys` to avoid having to pass them as arguments. Capturing `&mut sys` made the borrow checker unhappy though, and the closest alternative that came to mind at the time was a macro.)

PR #2485, opened 2026-01-15
feat: detect browser dialect

hippietrail:
# Issues 
N/A

# Description

The Harper browser extension always starts in US English mode on Chrome and Safari. I made sure I had my OS and browser settings all on Australian English and no difference.

This is a 99% agent-coded fix to detect the browser language region (dialect) setting.
I've looked at the code [but so far I haven't been able to run the tests](https://discord.com/channels/1335035237213671495/1461079667367739472/1461346412905369732). Hence just a draft PR for now.

# Demo

# How Has This Been Tested?

Manually only so far.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/background/index.ts
+++ b/packages/chrome-plugin/src/background/index.ts
@@ -1,5 +1,4 @@
-import { BinaryModule, Dialect, type LintConfig, LocalLinter } from 'harper.js';
-import { detectBrowserDialect } from './detectDialect';
+import { BinaryModule, type Dialect, type LintConfig, LocalLinter } from 'harper.js';
 import { type UnpackedLintGroups, unpackLint } from 'lint-framework';
 import type { PopupState } from '../PopupState';
 import {
@@ -43,6 +42,7 @@ import {
 	type SetUserDictionaryRequest,
 	type UnitResponse,
 } from '../protocol';
+import { detectBrowserDialect } from './detectDialect';
 
 console.log('background is running');
 


packages/chrome-plugin/src/background/index.ts
```diff
@@ -380,10 +396,53 @@ async function getIgnoredLints(): Promise<string> {
...
+	// If user hasn't set a dialect, try to detect from browser language
+	if (resp.dialect === undefined) {
+		return detectBrowserDialect();
+	}
+
 	return resp.dialect;
 }
 
+/** Detect English dialect from browser language settings */
+function detectBrowserDialect(): Dialect {
```
elijah-potter: This logic seems sound. I'd prefer it in its own file with some tests attached.

packages/chrome-plugin/src/background/index.ts
```diff
@@ -48,15 +48,31 @@ chrome.runtime.onInstalled.addListener((details) => {
 	}
 });
 
+let linter: LocalLinter;
+let linterReady = false;
```
elijah-potter: This is redundant. If the linter is `null`, we can assume it is not ready. This variable doesn't provide additional information and can become inconsistent with the type system if used incorrectly.

packages/chrome-plugin/src/background/index.ts
```diff
@@ -393,11 +452,13 @@ async function setActivationKey(key: ActivationKey) {
 	await chrome.storage.local.set({ activationKey: key });
 }
 
-function initializeLinter(dialect: Dialect) {
+async function initializeLinter(dialect: Dialect) {
```
elijah-potter: This function does not need to be modified for this PR.

packages/chrome-plugin/src/background/index.ts
```diff
@@ -380,10 +396,53 @@ async function getIgnoredLints(): Promise<string> {
 }
 
 async function getDialect(): Promise<Dialect> {
-	const resp = await chrome.storage.local.get({ dialect: Dialect.American });
+	const resp = await chrome.storage.local.get({ dialect: undefined });
+
+	// If user hasn't set a dialect, try to detect from browser language
+	if (resp.dialect === undefined) {
+		return detectBrowserDialect();
```
elijah-potter, line 413: Perfect.

elijah-potter (changes requested):
This PR is close. There are some irrelevant changes that seem to have crept in without explanation. Overall, I'm a big fan.

packages/chrome-plugin/src/background/index.ts
hippietrail: > This function does not need to be modified for this PR.

I think the bits you're questioning are from the second commit where we (agent and I) were trying to track down the build problems. So they might be "clutching at straws" before we figured out that the problems were in the WASM/thesaurus stuff and not in the new feature we were working on.

Thanks for casting a more expert eye over it!

PR #2488, opened 2026-01-15
fix: WASM/thesaurus `just test` failure

hippietrail:
# Issues 
N/A

# Description

I was getting failures in `just test` on my system. This is outside my fields of expertise and the changes were found with an AI agent that's been impressing me lately. They definitely need to be analysed by somebody knowledgeable.

@elijah-potter this pertains to the thread in discord where I mentioned the problems I was having testing.

The agent's summary upon completion of the session:
```
Perfect. Summary of all changes:

Summary of Fixes

1. WASM Build Errors (Resolved ✅)

Root cause: zstd-sys can't compile to wasm32-unknown-unknown on macOS
Solution:
Removed thesaurus from harper-core default features
Added explicit features = ["thesaurus"] to all non-WASM crates
Added default-features = false to WASM-building crates
```

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -10,7 +10,7 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = "0.4"
 
-harper-core = { path = "../harper-core" }
+harper-core = { path = "../harper-core", features = ["thesaurus"] }
 harper-typst = { path = "../harper-typst" }
 harper-literate-haskell = { path = "../harper-literate-haskell" }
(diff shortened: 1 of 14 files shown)


elijah-potter:
It looks like there is a documented workaround for `zstd-sys` on MacOS: https://github.com/gyscos/zstd-rs/wiki/Compile-for-WASM

That said, I feel that the real solution is to use a compression library that doesn't have as many build requirements. 

I've tried it out over at #2490. Would you give it a whirl?

PR #2491, opened 2026-01-15
feat(core): introduce Weirpacks

elijah-potter:
# Description

See my recent blog post for the full context: https://elijahpotter.dev/articles/imagine-a-weir-marketplace

In short: With Weir making it easier to write rules and otherwise extend Harper, it would be useful to be able to share collections of rules with other users. To that end, I've defined "Weirpack" to be a shareable file containing collections of Harper rules. 

I've set up the infrastructure to read and write Weirpacks, including changes to `harper-core`, `harper.js`, and an interactive playground available at the `/weir/playground` endpoint.

I intend to follow up on this PR with improvements to the Weir interpreter and ways to load Weirpacks from disk and over the network.

# Demo

image

# How Has This Been Tested?

Mostly manually. This PR will remain not be merged until I feel confident in the robustness of the systems.

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2793,8 +2793,8 @@ dependencies = [
  "trie-rs",
  "unicode-blocks",
  "unicode-script",
- "zip 2.4.2",
  "unicode-width",
+ "zip 2.4.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


hippietrail:
Hmm this could mean I could contribute more "opinionated" rules that don't have to default to "on". Will keep it in mind.

elijah-potter:
> Hmm this could mean I could contribute more "opinionated" rules that don't have to default to "on". Will keep it in mind.

Precisely.

PR #2506, opened 2026-01-16
feat(core): create a good number of new Weir rules

elijah-potter:
# Issues 

No specific issues are addressed here.

# Description

With the help of an LLM, I generated a whole bunch of new rules to address situations I have encountered.

In doing so, I encountered some issues with Weir that need to be addressed.

# How Has This Been Tested?

A good number of unit tests contained within the files themselves.

# Checklist

### DIFF
--- a/harper-core/src/linting/weir_rules/AsFollows.weir
+++ b/harper-core/src/linting/weir_rules/AsFollows.weir
@@ -1,7 +1,7 @@
 expr main (as follow)
 
 let message "Use `as follows`."
-let description "Corrects the phrase `as follow`, which is sometimes produced by hypercorrection. While it appeared briefly in 19th-century English, it is now considered archaic; modern standard usage requires `as follows` regardless of number."
+let description "Corrects the phrase `as follow`, which is sometimes produced by overcorrection. While it appeared briefly in 19th-century English, it is now considered archaic; modern standard usage requires `as follows` regardless of number."
 let kind "Grammar"
 let becomes "as follows"
 


harper-core/src/linting/weir_rules/AsFollows.weir
```diff
@@ -0,0 +1,23 @@
+expr main as follow
+
+let message "Introduce the following details with `as follows` when presenting a list or explanation."
+let description "Transforms the outdated phrase `as follow` into the standard `as follows` when it introduces additional content."
```
hippietrail: "Outdated" is probably too weak. I'd never heard "as follow" before. Apparently it was used briefly two hundred years ago. Dictionaries use "archaic" for this. "Outdated" would be for words that have gone out of fashion in the last couple of generations, like "how do you do" perhaps.

It's likely everybody using this today is completely unaware of the 19th century usage and it's just a plain hypercorrection thinking it should agree in number with the noun.

harper-core/src/linting/weir_rules/DoubleNegative.weir
```diff
@@ -0,0 +1,24 @@
+expr main <([(didn't), (did not), (did n't)] [need, have, want, make, take, get] no), no>
+
+let message "A negated auxiliary followed by `no` creates a double negative; replace `no` with `any` to keep a single negation."
```
hippietrail: Only "be", "do", "have", and the modal verbs are auxiliaries. Not "want", "make", "take", or "get".

harper-core/src/linting/weir_rules/PedalToTheMetal.weir
```diff
@@ -0,0 +1,24 @@
+expr main pedal to the medal
+
+let message "Reserve `pedal to the metal` for the classic acceleration idiom."
+let description "Replaces the misspelled `pedal to the medal` with the intended `metal` spelling whenever the phrase refers to putting the throttle down."
```
hippietrail: "Put the throttle down" sounds very odd. Perhaps "Depress the accelerator" or just "accelerate".

But actually those are verbs whereas "pedal to the metal" is an adverbial phrase. ie it describes **how fast/crazily** someone/something is going. The wording "Put the pedal to the metal" would be a verb phrase.

PR #2538, opened 2026-01-19
fix(core): fix `AnA` false-positives

### LINKED ISSUE #2406: AnA false positives, mostly acronyms
**What got flagged?**

Trying out AnA on https://github.com/home-assistant/home-assistant.io/tree/current/source/_integrations flagged a bunch of false positives, categorized below.

All uppercase acronym cases

- a UDP connection
- a UPB device 
- a UPS or power device
- a USB 2.0 port
- an HEVC HLS stream

Other acronymish ones

- an FFmpeg-compatible input file
- an [mDNS](https://en.wikipedia.org/wiki/Multicast_DNS) tool
- an RFLink device
- a [udev](https://en.wikipedia.org/wiki/Udev) rule

Other cases

- a Honeywell alarm panel
- a OneDrive folder
- a Ubiquiti UniFi Network application

**Why is this incorrect?**

Per pronounciation.

**Example of correct usage:**

N/A, above list should be accepted (+ flagged if the _other_ a/an is incorrectly used).

### LINKED ISSUE #2525: False positive: `AnA` flags "a USB" → "an USB"
Image

"An" is used before words starting with a vowel sound. The sound of "USB" is "you ess bee", starting with a consonant. As such "A USB" is correct and "An USB" is wrong.

86xsk:
# Issues 

Fixes #2525
Fixes #2406 

# Description

Fixes some reported false-positives in `AnA`.

To simplify the handling of empty words, I've made `starts_with_vowel` return `Option<InitialSound>`. I'm not sure if that's the best idea, or if it might be better to return `InitialSound::Either` for empty words instead. The latter would help reduce code noise somewhat, since we wouldn't need all the `Some()` and the unwrapping wherever the function is called.

# How Has This Been Tested?

- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/char_string.rs
+++ b/harper-core/src/char_string.rs
@@ -4,7 +4,7 @@ use std::borrow::Cow;
 use smallvec::SmallVec;
 
 // TODO: remove this when `SmallVec` allows retrieving this value in a const context.
-pub const CHAR_STRING_INLINE_SIZE: usize = 16;
+pub(crate) const CHAR_STRING_INLINE_SIZE: usize = 16;
 
 /// A char sequence that improves cache locality.
 /// Most English words are fewer than 12 characters.


harper-core/src/char_string.rs
```diff
@@ -3,9 +3,12 @@ use std::borrow::Cow;
 
 use smallvec::SmallVec;
 
+// TODO: remove this when `SmallVec` allows retrieving this value in a const context.
```
elijah-potter, line 6: That would be sweet!

PR #2540, opened 2026-01-19
chore: Update dictionary with new entries

Inhishonor:
# Issues 
N/A

# Description

- Added preindustrial
- Added reapprove
- Added preparer
- Added phosphene
- Added balter
- Added dysania
- Added semidormant
- Added pravity

# Demo
N/A

# How Has This Been Tested?
Manually linted a file with above words

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -13912,6 +13912,7 @@ baloney/~Ng
 balsa/~NwgS
 balsam/~NwSgV
 balsamic/JN
+balter/VSGd
 baluster/NSg
 balustrade/~NgS
 bamboo/~NwSgJV
@@ -22428,6 +22429,7 @@ dynastic/~J
 dynasty/~NSg
 dyno/NSg
 # dys               # prefixes that are not also words in their own right don't belong in the dictionary
+dysania/Nmg
 dysentery/~Nmg
 dysfunction/~Nmg
 dysfunctional/~J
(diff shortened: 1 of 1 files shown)


harper-core/dictionary.dict
```diff
@@ -13912,6 +13912,7 @@ baloney/~Ng
 balsa/~NwgS
 balsam/~NwSgV
 balsamic/JN
+balter/VSGd
```
hippietrail, line 13915: I was wondering if this rare but intersting word should go into the specialized section - it's mostly for tech words but I've also put words from other fields I'm interested in that don't seem mainstream enough - such as car model names.

I'm in two minds about this - does anyone else have any thoughts?

harper-core/dictionary.dict
```diff
@@ -22428,6 +22429,7 @@ dynastic/~J
 dynasty/~NSg
 dyno/NSg
 # dys               # prefixes that are not also words in their own right don't belong in the dictionary
+dysania/Nmg
```
hippietrail, line 22432: As with 'balter' I was also wondering if this rare but interesting word should stay in the mainstream section of the dictionary here or move to the specialized section.

harper-core/dictionary.dict
elijah-potter: I think it's fine where it is. It's uncommon, but not necessarily from a specialized field.

elijah-potter (approved):

PR #2541, opened 2026-01-19
fix(core): capitalization issue in `SpellCheck`

### LINKED ISSUE #2524: `SpellCheck` is flagging `MacOS` → `MacOS`
Image

The correct spelling (stylization if you prefer) is `macOS` and I'm pretty sure this used to work a couple of months ago.

It should flag `MacOS` as a misspelling. But it should suggest `macOS` and not `MacOS` as the correction.

My guess is that it's checking if a word is a proper noun and capitalizing it, which is a false assumption for the past couple of decades.

86xsk:
# Issues 
Fixes #2524

# Description

Skips capitalization of suggestions that contain uppercase characters in any position except the first in `SpellCheck`. The intention is to avoid uppercasing words that have fixed-casing(?), like "macOS".

# How Has This Been Tested?

- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/linting/spell_check.rs
+++ b/harper-core/src/linting/spell_check.rs
@@ -87,7 +87,15 @@ impl<T: Dictionary> Linter for SpellCheck<T> {
             if let Some(mis_f) = word_chars.first()
                 && mis_f.is_uppercase()
             {
-                for sug_f in possibilities.iter_mut().filter_map(|w| w.first_mut()) {
+                for sug_f in possibilities.iter_mut().filter_map(|w| {
+                    // Skip words that have uppercase chars in any position except the first.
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/spell_check.rs
```diff
@@ -513,4 +521,14 @@ mod tests {
             SpellCheck::new(FstDictionary::curated(), Dialect::American),
         );
     }
+
+    #[test]
+    fn no_improper_suggestion_for_macos() {
```
elijah-potter, line 526: I don't care too much, but next time would you include the relevant issue in the test name? For example, `allows_improper_caps_issue_2524`. It makes it easier for people down the road to find what the test is looking for.

PR #2556, opened 2026-01-21
fix(core): allow both a/an for "LED" in `AnA`

### LINKED ISSUE #2550: False positive: AnA "an LED"
**What got flagged?**
```
    ╭─[lutron.markdown:1:1]
    │
 56 │ Each full-width button on a Lutron SeeTouch, Hybrid SeeTouch, and Tabletop SeeTouch Keypad has an LED that can be controlled by Home Assistant. Performing an action of `switch.turn_off` or `switch.turn_on` against the appropriate LED entity will control the keypad LED.
    │                                                                                                ─┬  
    │                                                                                                 ╰── [Miscellaneous::AnA] (pri 31): Incorrect indefinite article.
────╯
```
https://github.com/home-assistant/home-assistant.io/blob/f1c8b9b2b1a5368f002717d9c911876532078517/source/_integrations/lutron.markdown?plain=1#L56

**Why is this incorrect?**
I believe "LED" is pronounced letter by letter, and so "an" is correct for the leading "el".

**Example of correct usage:**
N/A

86xsk:
# Issues 

Fixes #2550

# Description

Allows both a/an to precede LED by returning `InitialSound::Either` from `starts_with_vowel()`.

[According to Wiktionary](https://en.wiktionary.org/wiki/LED), both the acronym and initialism pronunciation of the word is used.

# How Has This Been Tested?

- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/linting/an_a.rs
+++ b/harper-core/src/linting/an_a.rs
@@ -123,6 +123,10 @@ fn starts_with_vowel(word: &[char], dialect: Dialect) -> Option<InitialSound> {
         return None;
     }
 
+    if matches!(word, ['S', 'Q', 'L'] | ['L', 'E', 'D']) {
+        return Some(InitialSound::Either);
+    }
+
     // Try to get the first chunk of a word that appears to be a partial initialism.
     // For example:
     // - `RFL` from `RFLink`
@@ -145,9 +149,6 @@ fn starts_with_vowel(word: &[char], dialect: Dialect) -> Option<InitialSound> {
     let is_likely_initialism = word.iter().all(|c| !c.is_alphabetic() || c.is_uppercase());
 
     if word.len() == 1 || (is_likely_initialism && !is_likely_acronym(word)) {
-        if matches!(word, ['S', 'Q', 'L']) {
-            return Some(InitialSound::Either);
-        }
         return Some(
             if matches!(
                 word[0].to_ascii_uppercase(),


86xsk:
Sorry @hippietrail, I just saw your comment from the removed(?) PR in my email now.

Yeah, I'm not 100% certain on the `Either` for LED myself. You do bring up some good points. I think there is a balance between being pedantically correct and excessively permissive, and this case is tricky because it happens to sit right in the middle of it.

For another interesting data point, [Google Books Ngram Viewer shows that both](https://books.google.com/ngrams/graph?content=a+LED%2Can+LED&year_start=1800&year_end=2022&corpus=en&smoothing=3#) "a LED" and "an LED" are used, but "an LED" is about 4-5x more common. In comparison, the difference between "a ubiquitous, an ubiquitous" (which `AnA` both allows) is >10x.

I think I lean *slightly* towards allowing either for LED. Mostly because I feel it might be a little annoying for some to have "a LED" linted if they feel that is correct. Though it would certainly be a little detrimental if we want to encourage writing in a consistent/standard manner. Maybe in the future, we could partially counter this by automatically enforcing an/a consistency for each `InitialSound::Either` word?

hippietrail:
> I think I lean _slightly_ towards allowing either for LED. Mostly because I feel it might be a little annoying for some to have "a LED" linted if they feel that is correct.

Ah but by this logic we would lean both ways due to the people who feel it's correct when it gets linted _and_ the people who feel it's incorrect when it gets permitted.

> Though it would certainly be a little detrimental if we want to encourage writing in a consistent/standard manner. Maybe in the future, we could partially counter this by automatically enforcing an/a consistency for each `InitialSound::Either` word?

But the `InitialSound::Either` logic could be moved into the dictionary parsing logic and then become part of a word's metadata, which can be queried by new linters. Compare to the `OrthInfo` code for similar logic and the `OxfordCommas` and `NoOxfordCommas` linters for linters which enforce preferences.

PR #2569, opened 2026-01-22
fix(chrome-ext): problems with incorrect dialect selection and reversion

elijah-potter:
# Description

While testing version `1.5.0` of the Chrome extension, I observed two issues related to dialect management. I believe both were introduced in #2485.

The biggest issue was that the Chrome extension would not properly persist the dialect. If a user changed their dialect, it would be saved, then immediately overwritten by the default path `getDialect` function.
The second issue, which exacerbated the first, was that the browser would incorrectly infer the dialect from the locale. My browser had the `en-US` locale, which would be mapped to British English.

I've gone ahead and tweaked the logic to fix both problems. I've confirmed that the fix works with manual testing.

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/background/detectDialect.ts
+++ b/packages/chrome-plugin/src/background/detectDialect.ts
@@ -27,6 +27,8 @@ function localeToDialect(locale: string): Dialect {
 	if (lower.startsWith('en')) {
 		// New Zealand → Australian (closest match)
 		if (lower.includes('en-nz') || lower.includes('en_nz')) return Dialect.Australian;
+		// American
+		if (lower.includes('en-us') || lower.includes('en_us')) return Dialect.American;
 		// Other English variants → British as fallback
 		if (lower.match(/^en[-_]/)) return Dialect.British;
 		// Plain 'en' → American (default)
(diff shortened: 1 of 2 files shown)


hippietrail:
D'oh! I should've written that code myself instead of looking over the LLM's code - took me a while to figure out why it wasn't caught.

But your fix should be in the "explicit match" section rather than the "fallback" section:
```rs
	// Explicit matches
	if (lower.includes('en-us') || lower.includes('en_us')) return Dialect.American;
	if (lower.includes('en-gb') || lower.includes('en_gb')) return Dialect.British;
	if (lower.includes('en-au') || lower.includes('en_au')) return Dialect.Australian;
	if (lower.includes('en-ca') || lower.includes('en_ca')) return Dialect.Canadian;
	if (lower.includes('en-in') || lower.includes('en_in')) return Dialect.Indian;

	// Fallback for English variants
	if (lower.startsWith('en')) {
		if (lower.includes('en-nz') || lower.includes('en_nz')) return Dialect.Australian;
		// American
		// Other English variants → British as fallback
		if (lower.match(/^en[-_]/)) return Dialect.British;
		// Plain 'en' → American (default)
```

PR #2570, opened 2026-01-23
chore: add words/annotations to curated dict

86xsk:
# Description

Adds words/annotations for some terms I've encountered that weren't in the curated dictionary.

Adds duplexing, recoverability, spooler, striping, and workgroup.

# How Has This Been Tested?

- `cargo test`
- manually

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -41500,7 +41500,7 @@ recordkeeping/Ng
 recoup/VdG
 recourse/~NgV
 recoverable/JNU
-recoverability/Nm
+recoverability/Nmg
 recovery/~NwSg
 recreant/JNgS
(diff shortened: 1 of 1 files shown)


harper-core/dictionary.dict
```diff
@@ -41500,6 +41500,7 @@ recordkeeping/Ng
 recoup/VdG
 recourse/~NgV
 recoverable/JNU
+recoverability/Nm
```
hippietrail: I always include `/g` for every noun not ending in "s":
- The hybrid formalism allows the recoverability's analysis based on the controllability concept of hybrid systems.
- recoverability's role in crypto cannot be overlooked

I notice that Chrome flags the possessive version though.

PR #2581, opened 2026-01-24
chore: justfile recipe names consistent+guessable

hippietrail:
# Issues 
N/A

# Description

Some `justfile` recipe names include hyphens between words, some just glom all words together.
Some places refer to a plugin while some refer to an extension.
Some recipe names use a full word and some use an abbreviation.

Occasionally a recipe name is not intuitive and it's not immediately obvious that there's a recipe that does what you're looking for. For instance I always confuse the POS-tagging and snapshot tests.
When testing dictionary curation PRs it might be quicker to test just these two than to run `cargo test`.
But we had a recipe named `runsnapshots` that called two internal tests: `test_pos_tagger`, with an intuitive name; and `test_most_lints`, with an unintuitive name since it's the one that tests the snapshots.

Since I wanted to address this last problem I decided to just add `alias`es to the other recipes. `justfile` didn't support `alias` when we first started using it but has done so for months now.

Please feel free to critique my choices!

# How Has This Been Tested?

I manually tested some, but not all, of the recipes.

# Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -24,6 +24,7 @@ build-wasm:
   fi
 
 # Build `harper.js` with all size optimizations available.
+alias build-harper-js := build-harperjs
 build-harperjs: build-wasm 
   #!/usr/bin/env bash
   set -eo pipefail
@@ -47,6 +48,7 @@ build-lint-framework: build-harperjs
   pnpm install
   pnpm build
 
+alias test-harper-js := test-harperjs
 test-harperjs: build-harperjs
   #!/usr/bin/env bash
   set -eo pipefail
@@ -69,6 +71,7 @@ test-obsidian: build-obsidian
   pnpm playwright install
   pnpm test
 
+alias dev-wordpress := dev-wp
 dev-wp: build-harperjs
   #!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -688,11 +715,17 @@ suggestannotation input:
   }
 
 # Audit the curated dictionary for any issues.
+alias audit-dictionary := auditdictionary
 alias auditdict := auditdictionary
-
+alias audit-dict := auditdictionary
 auditdictionary DIR="harper-core":
```
elijah-potter: I think this PR looks great, and I like a lot of the aliases that you've added to make things easier to guess. I would like to standardize on chain-case for all of these names. Would you mind removing all the instances where the words have just been mushed together, like auditdictionary?

hippietrail: > I think this PR looks great, and I like a lot of the aliases that you've added to make things easier to guess. I would like to standardize on chain-case for all of these names. Would you mind removing all the instances where the words have just been mushed together, like auditdictionary?

Absolutely. I also found them ungainly but some had been around for so long that I didn't want to presume. And, for the short ones, didn't seem like a terrible idea.

Things I won't change to chain-case:
- dogfood
- precommit
- setup

Old things I will change to chain-case:
- addnoun
- getforms
- getmetadata
- getnps
- registerlinter
- runsnapshots

Others I don't really consider old enough to be too worried about changing

justfile
```diff
@@ -339,10 +343,18 @@ install:
 dogfood:
   #!/usr/bin/env bash
   cargo build --release
-  for file in `fd -e rs`
-  do
-    echo Linting $file
-    ./target/release/harper-cli lint $file
+  
+  if command -v fd &> /dev/null; then
```
elijah-potter, line 347: I appreciate this.

PR #2586, opened 2026-01-25
fix: confusing wording in `OneOfTheSingular`

hippietrail:
The construction one of the ... uses a singular noun. → ... should use a plural noun.

image

The people who write this mistake probably think they're writing correct English which would lead them to wonder why the popup is telling them what they wrote.

I'm not sure whether I thought the original wording made sense or whether it was a brainfart. But I just hit it in my regular dogfooding with Harper Glasses and saw it gives the wrong impression.

Should be clearer now.


elijah-potter:
Yeah, I can see why that would be confusing. Thanks for the patch!

PR #2588, opened 2026-01-25
fix: allow `check-rust` to complete without `cargo hack`

hippietrail:
# Issues 
N/A

# Description

I noticed lately that when I do `just precommit` that the test comes to an abrupt halt after doing all kinds of heavy testing with a dozen Chromium windows open at once etc. just to tell me I don't have `cargo hack` installed.

This PR will instead issue a warning at that point but continue the testing.

instead of failing at this point it will issue the warning
```
⚠️  cargo-hack not found. Install with 'cargo install cargo-hack' for complete feature testing.
```

I'm not 100% clear if this is acceptable or not though. Another way would be to check if it's installed at the start of the test so that you don't waste time and system resources only for the test to fail to complete.

I'm not a shell expert and the LLM's first method of checking for the presence of `cargo hack` didn't actually work - so please check carefully!

# How Has This Been Tested?

Manually.

# Checklist


elijah-potter:
I would prefer this to work similarly to how we have it set up with `wasm-opt`. Right now, if you wish to skip the `wasm-opt` step (if you do not wish to have it installed, or if you are impatient like me), you can simply export `DISABLE_WASM_OPT` in your `.bashrc`. I think if you don't want to have `cargo-hack` installed, we should have a similar environment variable, like `DISABLE_CARGO_HACK`.

hippietrail:
> I would prefer this to work similarly to how we have it set up with `wasm-opt`. Right now, if you wish to skip the `wasm-opt` step (if you do not wish to have it installed, or if you are impatient like me), you can simply export `DISABLE_WASM_OPT` in your `.bashrc`. I think if you don't want to have `cargo-hack` installed, we should have a similar environment variable, like `DISABLE_CARGO_HACK`.

Is like this OK or would you prefer without the warning since `DISABLE_WASM_OPT` doesn't have one?

elijah-potter:
My main concern is that I don't want CI to fail silently if the `cargo hack` set up breaks. Would you mind making sure that the command will fail if `cargo hack` is missing and `DISABLE_CARGO_HACK` is off?

hippietrail:
> > I would prefer this to work similarly to how we have it set up with `wasm-opt`. Right now, if you wish to skip the `wasm-opt` step (if you do not wish to have it installed, or if you are impatient like me), you can simply export `DISABLE_WASM_OPT` in your `.bashrc`. I think if you don't want to have `cargo-hack` installed, we should have a similar environment variable, like `DISABLE_CARGO_HACK`.
> 
> Is like this OK or would you prefer without the warning since `DISABLE_WASM_OPT` doesn't have one?

Like this right?

hippietrail:
Messed up this PR with an AI - no idea how to fix it - right now I can't find any other old PRs that are messed up.

PR #2602, opened 2026-01-26
feat(core): add `allows` keyword to Weir

### LINKED ISSUE #2487: Weir should have special syntax for asserting that no lints should be thrown

When we write rules in Rust, we have assertions (most commonly `assert_no_lints`) that check to make sure that the rule does not apply any changes to a given test. Weir does not currently have great syntax for this. The closest thing is to use the `test` keyword with the same string twice. This is not ideal.

**Proposed Solution**

I think we should have a special keyword (maybe `ignores`) to assert that a Weir rule does not apply a change to a given string. This is something I've run into and something others have brought up to me independently (@hippietrail).

**Examples**

`ignores "This is a grammatically correct sentence."`

**Component**
- [x] Core engine
- [ ] Plugin/Extension
- [ ] Other: _____

elijah-potter:
# Issues 

Resolves #2487

# Description

I've added the keyword in the syntax described in the issue.

I've also refactored many of the existing rules to use the new syntax.

# How Has This Been Tested?

Additional unit tests.

# Checklist

### DIFF
--- a/harper-core/src/linting/correct_number_suffix.rs
+++ b/harper-core/src/linting/correct_number_suffix.rs
@@ -27,7 +27,7 @@ impl Linter for CorrectNumberSuffix {
                     span: suffix_span,
                     lint_kind: LintKind::Miscellaneous,
                     message: "This number needs a different suffix to sound right.".to_string(),
-                    suggestions: vec![Suggestion::ReplaceWith(correct_suffix.to_chars())],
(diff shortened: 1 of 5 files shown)


elijah-potter:
I didn't see @hippietrail's comment before. I think he is right: `allows` is a better keyword. I've changed it accordingly.

PR #2606, opened 2026-01-26
fix(obsidian-plugin): allow initialization to complete asynchronously

### LINKED ISSUE #2601: Obsidian plugin delays app startup by 2,500ms
**Describe the bug**
Harper plugin takes 2,500ms to load, impacting Obsidian startup time.

Before the app fully loads, it waits for all code within the plugin onload() function to execute. If there is any code in the onload function that can be executed asynchronously, after the app finishes loading, put it in `onLayoutReady` instead. Here's a guide for optimizing plugin performance: <https://docs.obsidian.md/plugins/guides/load-time>

**To Reproduce**
1. Download and enable plugin.
2. Restart Obsidian

**Expected behavior**
The app should start up with minimal delay.

**Platform**
Obsidian

elijah-potter:
# Issues 

Fixes #2601

# Description

As it turns out, most of the work that was being done in `onload` could be deferred. Instead of waiting for the internal state of the plugin to be initialized at startup, we simply kick off the task and let it complete at its own pace. 

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/packages/obsidian-plugin/src/index.ts
+++ b/packages/obsidian-plugin/src/index.ts
@@ -26,7 +26,7 @@ export default class HarperPlugin extends Plugin {
 		}
 
 		const data = await this.loadData();
-		await this.state.initializeFromSettings(data);
+		this.state.initializeFromSettings(data);
 		this.registerEditorExtension(this.state.getCMEditorExtensions());
 		this.setupCommands();
 		this.setupStatusBar();


Vinzent03:
I already started with my work, before I saw this pr. I took a different way and fixed some other performance things as well in #2611

elijah-potter:
Superseded by #2611.

PR #2611, opened 2026-01-27
fix(obsidian-plugin): postpone initialization until layout is ready

Vinzent03:
# Issues 

#2416
#2601

# Description

The whole initialization is postponed until the layout is ready. This allows the user to see their files start editing them while harper is still loading.

In addition, I saw that the settings pane continuously reloads the current config every second. I don't think this is necessary as the settings pane is not opened often and reloading it every seconds seems a bit strange. Instead it refreshes the view now on open.

I also added false as argument to `enableEditorLinter` to prevent it to flicker at start. It is not necessary to reinit at that point, because it simply calls `initializeFromSettings` which is already done before.

# Demo

# How Has This Been Tested?

manually

# Checklist

### DIFF
--- a/packages/obsidian-plugin/src/HarperSettingTab.ts
+++ b/packages/obsidian-plugin/src/HarperSettingTab.ts
@@ -11,7 +11,6 @@ import { linesToString, stringToLines } from './textUtils';
 const LintSettingId = 'HarperLintSettings';
 
 export class HarperSettingTab extends PluginSettingTab {
-	private state: State;
 	private settings: Settings;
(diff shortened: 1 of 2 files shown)


packages/obsidian-plugin/src/HarperSettingTab.ts
```diff
@@ -55,7 +53,12 @@ export class HarperSettingTab extends PluginSettingTab {
 		});
 	}
 
-	display() {
+	display(update = true) {
```
elijah-potter, line 56: Good call.

PR #2617, opened 2026-01-27
fix(chrome-ext): preserve cursor position in generic contenteditable replacement

### LINKED ISSUE #2616: Cursor jumps to start of text block after applying suggestion in Notion
**Describe the bug**
After applying a grammar suggestion in Notion, the cursor jumps to the beginning of the text block instead of staying at the replacement position.

**To Reproduce**
1. Open a Notion page in the browser with Harper extension enabled
2. Type "This is an test" in a text block
3. Click the Harper highlight on "an"
4. Apply the suggestion to replace with "a"
5. Cursor jumps to the beginning of the block

**Expected behavior**
Cursor should remain right after the replaced word ("a"), matching behavior in other editors.

**Platform**
Chrome extension, Brave browser, Fedora

draphy:
# Issues
Fixes #2616

# Description
`replaceGenericContentEditable` used `el.textContent = value`, which destroys DOM structure and loses cursor position. This affects editors like Notion that fall through to the generic handler.

Now tries surgical replacement via `selectSpanInEditor` + `replaceTextInRange` first, falling back to the original behavior if that fails.

Also extracts `testBasicSuggestionRichText` and `testCanIgnoreRichTextSuggestion` shared test utilities (matching the existing `testMultipleSuggestionsAndUndo` pattern) and adds cursor position verification to the basic suggestion test.

# How Has This Been Tested?
The basic suggestion test now verifies cursor position by using `ArrowRight×3 + Backspace` from the expected cursor position.

# Checklist


elijah-potter:
Thank you so much!

Issue #2411, opened 2026-01-02
The order of words in the dictionary can affect spell checking


hippietrail:
In my new dictionary curation I sorted the dictionary and that resulted in the order of `Tex` vs `TeX` being reversed, which caused #1873 to show up again.

It was very hard to track this down as there were no code changes in the PR.

The dictionary is case-folded when it's compiled so that `Tex` and `TeX` are the same entry with the same metadata including orthography info.

Case folding should be commutative and not matter what the order is.

I'm not familiar with the spellcheck code. There is some concept of canonical spelling involved. Case folded entries don't have a canonical spelling. Or rather they have multiple canonical spellings but they can't necessarily be retrieved. It might depend on the heuristics though.

I suspect a bug in how the metadata for the first and second word getting case folded is done but it could be a bug in how the canonical spelling is decided.

Here's some example tests that can be added to `harper-core/src/spell/rune/mod.rs` to show the effect:
```rs
    #[test]
    fn case_folded_tex_before_TeX() {
        // Test: Tex comes before TeX in the dictionary
        let words = parse_word_list("2\nTex/O\nTeX/O").unwrap();
        let attributes = AttributeList::parse(
            &json!({
                "affixes": {},
                "properties": {
                    "O": {
                        "#": "proper noun",
                        "metadata": {
                            "noun": {
                                "is_proper": true
                            }
                        }
                    }
                }
            })
            .to_string(),
        )
        .unwrap();

        let mut expanded = WordMap::default();
        attributes.expand_annotated_words(words, &mut expanded);

        // Both should be in the map
        assert!(expanded.get_with_str("Tex").is_some());
        assert!(expanded.get_with_str("TeX").is_some());

        // The canonical spelling should be TeX (last one wins)
        let entry = expanded.get_with_str("tex").unwrap();
        let canonical: String = entry.canonical_spelling.iter().collect();
        assert_eq!(canonical, "TeX");
    }

    #[test]
    fn case_folded_TeX_before_Tex() {
        // Test: TeX comes before Tex in the dictionary
        let words = parse_word_list("2\nTeX/O\nTex/O").unwrap();
        let attributes = AttributeList::parse(
            &json!({
                "affixes": {},
                "properties": {
                    "O": {
                        "#": "proper noun",
                        "metadata": {
                            "noun": {
                                "is_proper": true
                            }
                        }
                    }
                }
            })
            .to_string(),
        )
        .unwrap();

        let mut expanded = WordMap::default();
[...]

Issue #1688, opened 2025-08-04
False positive: 'PR' and username '@xxx' flagged


jan-kubica:
## Description

Example:
`Thank you for the PR @hippietrail`

Both PR and the username are flagged. I would expect that:

- `PR` is being recognised as a valid abbreviation (both Pull Request and Public Relations are common)
- whatever comes in a single token starting with @ is not being checked for correctness 

## Screenshots

<img width="1992" height="540" alt="Image" src="https://github.com/user-attachments/assets/a5dd7894-4ab3-4533-9ed2-4cc88f57d3d3" />

hippietrail:
`PR` is definitely in the dictionary. It's one of a few words that for some reason get marked as spelling mistakes despite being in the dictionary. I suspect it's a side-effect of how the dictionary conflated as case-variants into a single entry but I've never dug into it. I'll be on the road for a week so probably won't have time to investigate it.

I'm pretty sure `@` is treated as its own token and `hippietrail` wouldn't be in the dictionary. Are you saying `@` followed by anything wordish should be exempted from spellcheck? That's certainly possible.

@elijah-potter Do you have insights on the former and opinions on the latter?

jan-kubica:
Exactly, my personal preference would be that if a word starts with `@`, it is treated as an internal reference and therefore whatever directly follows in this single word should be exempted from spellcheck.

hippietrail:
> Exactly, my personal preference would be that if a word starts with `@`, it is treated as an internal reference and therefore whatever directly follows in this single word should be exempted from spellcheck.

If somebody wants to try to code this up and make a PR they should mimic my code for `condense_filename_extensions()` in `document.rs` but this one would be a bit simpler. It should probably be called `condense_mentions()` or `condense_at_mentions()`

There are probably other things with `@` that are not springing to mind so it should be behind a preference. I don't think such prefs are easy to make as of yet?

86xsk:
> `PR` is definitely in the dictionary. It's one of a few words that for some reason get marked as spelling mistakes despite being in the dictionary. I suspect it's a side-effect of how the dictionary conflated as case-variants into a single entry but I've never dug into it.

Debugging indicates that the issue relates to this code block here, at `linting/spell_check.rs:76`:
```rust
if let Some(word_metadata) = word.kind.as_word().unwrap() {
    if word_metadata.dialects.is_dialect_enabled(self.dialect)
        && (self.dictionary.contains_exact_word(word_chars)
[...]

PR #2630, opened 2026-01-30
feat(core)!: more explicit handling of case-sensitivity in dictionaries

### LINKED ISSUE #2585: "OS" flagged even though it's in `dictionary.dict`
Image

Right here around line 7568:
```
OR/CNO              # Oregon
OS/ONgJ
OSHA/Og             # agency
```
Not sure it should be marked as a proper noun or adjective though. I'll check and probably adjust those in my upcoming dictionary curation PR.

### LINKED ISSUE #3073: False Positive: text.md suggested to be text.MD
**What got flagged?**
It reports the conical spelling of "md" in harper is "MD".

**Why is this incorrect?**
 It suggests the change even when md is a file extension. I understand that paths in markdown are generally inside \``, but not always, and filenames should always be lowercase.

**Example of correct usage:**
I made the install.md file like you asked.

Edit:

I'm now noticing that this applies to more than just md. For example with .zip. I think there should be a list of file extensions that get special treatment.

### LINKED ISSUE #3080: [False Positive] `Apr` -> `APR` when used as abbreviated month
**What got flagged?**
```md
Apr 1st
```

`Apr` was flagged with: `This word's canonical spelling is all-caps. (Harper OrthographicConsistency)`

**Why is this incorrect?**
The correct capitalization of month abbreviation is to only capitalize the first letter. Harper does not give this hint for other month abbreviations.

**Example of correct usage:**
```md
- Apr 1st: April fools day.
```

86xsk:
# Issues 

Fixes #2585
Fixes #3073
Fixes #3080
Related to #1688
Related to #2411

(Probably a few more, this has been a long-running issue.)

# Description

- Changes the dictionary API to allow more explicitly handling capitalization/orthographic word variations.
    - Replaces `WordId` with `CanonicalWordId` and `CaseFoldedWordId`.
        - `CanonicalWordId` hashes the input word as is, without lowercasing or normalization.
        - `CaseFoldedWordId` lowercases and normalizes the word before hashing it (this is the current behavior of `WordId`).
        - Adds `WordIdPair` ~and `EitherWordId`~ to make it easier to work with `CanonicalWordId`/`CaseFoldedWordId`.
    - Querying the dictionary case-insensitively now means you will get a collection of words, not an individual word.
        - However, the current behavior, where a word is queried case-insensitively and returns the merged metadata across all of its casing/orthographic variants is preserved in the ~`get_word_metadata_combined`~[^1] functions. 
    - Querying the dictionary case-sensitively will now return a *specific* entry in the curated dictionary, allowing differentiation for case/orthographic variants when desired.
- Fixes the issue where some words would be linted by `SpellCheck` despite being present in the dictionary.
- ~Makes `SpellCheck` no longer care about capitalization, fully transferring that responsibility to `OrthographicConsistency` (which already handles it anyway for the most part).~
- Re-adds "Pr" to the curated dictionary, since it no longer conflicts with "PR".
- Allows the `derived_from` field in `DictWordMetadata` to contain multiple words.
  - Adds the `DerivedFrom` struct to contain this data without exposing implementation details. The current implementation uses a Vec which allows for the insertion of unique values only.

~Running `cargo bench --bench parse_essay` showed a performance regression of about \~3% across the board with these changes. I didn't notice any performance issues when testing the modified build of `harper-ls`.~

~After some performance work these changes now show a \~7-27% performance improvement on the benchmarks over master, see https://github.com/Automattic/harper/pull/2630#discussion_r2810461338.~

Changes which produced these performance improvements have been split off into their own PRs and merged separately.

# How Has This Been Tested?

- `cargo test`
- Manually building and testing the modified harper-ls on a few Markdown files

# Checklist

[^1]: `get_word_metadata_combined` has been renamed to `get_word_metadata` in 5a5ad04bf6149c112ce9b3e0a0df179ffa05b0bb, preserving the current name.


hippietrail:
Does it handle the same word having multiple `WordId`s? For instance you might have dictionary entries like:
```
- unpack/Vd # also creates "unpacked"
- packed/VtTU # also creates "unpacked
- unpacked/J # specifically creates "unpacked"
```
This happens a lot more than expected and causes some bugs of this type to crop up. I had a PR for it for many months that just tracked them but didn't really do anything with them as I wasn't sure what was and wasn't planned etc.

86xsk:
I believe it will merge the metadata for words that have identical canonical word IDs, as seems to be the case here.

image

The code in `expand_annotated_word` adds or appends metadata based on the canonical word ID (i.e. the hash of the word exactly as stated in the dictionary or as expanded from another word, without lowercasing or normalization). "unpacked" will always have the same word ID regardless of whether it was specifically defined or came from an expansion, so all instances/manifestations of "unpacked" will have their metadata merged.

If in theory we had, say, a proper noun "Unpacked" (whether specifically defined or coming from an expansion), that would not be merged. It would remain a separate entry that can be queried with the appropriate dictionary functions.

However, tokens that originate from a `Document` still currently use the merged/combined metadata. So an "unpacked" token would have both the "unpacked" and "Unpacked" metadata merged.

```rust
// In Document::parse()
// Annotate DictWord metadata
for token in sent.iter_mut() {
    if let TokenKind::Word(meta) = &mut token.kind {
        let word_source = token.span.get_content(&self.source);
        let mut found_meta = dictionary
            .get_word_metadata_combined(word_source) // <---
            .map(|c| c.into_owned());

        if let Some(inner) = &mut found_meta {
            inner.pos_tag = token_tags[i].or_else(|| inner.infer_pos_tag());
            inner.np_member = Some(np_flags[i]);
        }

        *meta = found_meta;
        i += 1;
    } else if !token.kind.is_whitespace() {
        i += 1;
    }
}
```

hippietrail:
I don't recall everything perfectly at the moment but both I think both cases are possible and only one is handled:
1. One `WordID` can refer to multiple words - due to case folding
2. One word can have multiple `WordID`s

I believe you're talking about 1. which works but is behind a few bugs. But 2. is distinct. I found my old PR: https://github.com/Automattic/harper/pull/1035

Apologies if I've mixed up any concepts myself - it's been a while (-:

86xsk:
Oh, I see what you mean now. Yeah no, at the moment I think this PR mostly sticks with the old behavior where once a `DictWordMetadata` has its one and only`derived_from` set, it's not updated again.

I did change the field to store a `WordIdPair` (storing both a canonical and case-folded word ID for its sole "parent" word), but I don't think that helps too much in this case. If I recall correctly, I did this so I could mimic the previous case-folded lookup in the `DerivedFrom` pattern, while still storing a canonical parent word ID too.

I'll take another look though. I wasn't too certain with that change as is, and the PR you mentioned does bring up an issue I haven't even considered.

hippietrail:
> Oh, I see what you mean now. Yeah no, at the moment I think this PR mostly sticks with the old behavior where once a `DictWordMetadata` has its one and only`derived_from` set, it's not updated again.

If I recall correctly the old behaviour is for each new one to blindly stomp on the previous `derived_from`?

> I did change the field to store a `WordIdPair` (storing both a canonical and case-folded word ID for its sole "parent" word), but I don't think that helps too much in this case. If I recall correctly, I did this so I could mimic the previous case-folded lookup in the `DerivedFrom` pattern, while still storing a canonical parent word ID too.

Yes I never took on the case folding as I was worried about stepping on the toes of the main devs thinking it might've been an intentional design choice plus I didn't find out it was definitely case folding until months after I did the multi-`derived_from` patch.

> I'll take another look though. I wasn't too certain with that change as is, and the PR you mentioned does bring up an issue I haven't even considered.

It looks like you're fixing up a whole bunch of sticky issues at once - good stuff!

86xsk:
> If I recall correctly the old behaviour is for each new one to blindly stomp on the previous `derived_from`?

I could be wrong, but I believe it will just keep the first value that was set, since in `DictWordMetadata::or`...

```rust
derived_from: self.derived_from.or(other.derived_from),
```

...it sets the `derived_from` to the result of [`Option::or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.or), which keeps the value if the option is already `Some`. (I can't say for sure that the value doesn't get overwritten elsewhere though.)

---

I threw together the changes so that it now supports a word having multiple `derived_from` words. Not too certain on all the details though. For instance, there would now be two types named `DerivedFrom` in harper-core.

hippietrail:
> > If I recall correctly the old behaviour is for each new one to blindly stomp on the previous `derived_from`?
> 
> I could be wrong, but I believe it will just keep the first value that was set, since in `DictWordMetadata::or`...
> 
> ```rust
> derived_from: self.derived_from.or(other.derived_from),
> ```
> 
> ...it sets the `derived_from` to the result of [`Option::or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.or), which keeps the value if the option is already `Some`. (I can't say for sure that the value doesn't get overwritten elsewhere though.)
> 
> I threw together the changes so that it now supports a word having multiple `derived_from` words. Not too certain on all the details though. For instance, there would now be two types named `DerivedFrom` in harper-core.

Cool. I remember when I was working on that stuff that it turned out I missed some cases where a word could have multiple `derived_from`s - I found the code a bit hard to follow. I think I got some bikeshedding patches in that hopefully made it a bit easier though.

86xsk:
The test seems to be failing because it expects "need" to be the suggestion for "ned" (probably from SpellCheck). However, with the changes, it's getting a lint to replace it with "Ned" instead (which must be coming from OrthographicConsistency trying to suggest what it thinks is the canonical spelling).

SpellCheck wouldn't lint for "ned" anymore since it's been made case-insensitive and "Ned" is in the curated dictionary.

I'll look into it, but I'm not sure yet how best to approach it. The simplest solution would be to remove "Ned" from the curated dictionary, but it would be ironic since part of the motivation for this PR was to avoid having to remove valid entries from the dictionary.

hippietrail:
Have a look at #2411 and see if you can get the words I mentioned such as "Metal" and "ANN" de-commented without causing "metal" and "Ann" to get flagged as misspelled.

86xsk:
I can confirm that does work. All `cargo test` tests pass with those words added, and the only difference in the snapshots is that "metal" is considered a proper noun (since `Document` parsing still uses merged case-insensitive metadata for tokens).

harper-core/src/spell/dictionary.rs
```diff
@@ -32,26 +32,63 @@ pub trait Dictionary: Send + Sync {
...
+
+    /// Get the correct canonical capitalizations for the given word.
+    fn get_correct_capitalization_of(&self, word: &[char]) -> Vec<&'_ [char]>;
+
     /// Get the associated [`DictWordMetadata`] for any capitalization of a given word.
-    fn get_word_metadata(&self, word: &[char]) -> Option<Cow<'_, DictWordMetadata>>;
+    ///
+    /// Since the dictionary might contain words that differ only in capitalization, this may
+    /// return multiple entries.
+    fn get_word_metadata(&self, word: &[char]) -> Vec<&DictWordMetadata>;
```
elijah-potter: I'm concerned, because many of these changes would be breaking, and thus would require a major version bump. I'm prepared to do that (it's worth it in this case), but we would want to make any other changes that would require such a version bump at the same time. Is there anything pressing you think should be changed in the API?

harper-core/src/spell/dictionary.rs
```diff
@@ -32,26 +32,63 @@ pub trait Dictionary: Send + Sync {
...
+    fn get_word_metadata_exact(&self, word: &[char]) -> Option<&DictWordMetadata>;
+
     /// Get the associated [`DictWordMetadata`] for any capitalization of a given word.
     /// If the word isn't in the dictionary, the resulting metadata will be
     /// empty.
-    fn get_word_metadata_str(&self, word: &str) -> Option<Cow<'_, DictWordMetadata>>;
+    ///
+    /// Since the dictionary might contain words that differ only in capitalization, this may
+    /// return multiple entries.
+    fn get_word_metadata_str(&self, word: &str) -> Vec<&DictWordMetadata>;
```
elijah-potter: This function gets called _a lot_. Would it be possible to return a value that doesn't allocate?

harper-core/src/linting/spell_check.rs
```diff
@@ -66,6 +66,15 @@ impl<T: Dictionary> SpellCheck<T> {
...
 }
 
+/// `SpellCheck` will typically ignore words that exist in the dictionary, regardless of casing.
+/// However, these (case-sensitive) exceptions will be linted anyway.
+/// As an example, we want `SpellCheck` to suggest "need" for "ned", even though "Ned" exists in the
+/// curated dictionary.
+///
+/// This is a somewhat hacky workaround. When a more general solution can be found, this should be
+/// removed.
+static CASING_EXCEPTIONS: &[&[char]] = &[&['n', 'e', 'd']];
```
elijah-potter: I don't like this solution. I feel that this would be resolved by making the spellcheck case-sensitive again. Is that assessment correct?

harper-core/src/linting/orthographic_consistency.rs
```diff
@@ -158,20 +172,29 @@ mod tests {
...
+            "So should harper and automattic.",
+            OrthographicConsistency::default(),
+            2,
+        );
+    }
+
+    #[test]
+    fn ikea_should_be_all_caps() {
         assert_suggestion_result(
-            "Lego bricks encourage creativity.",
```
elijah-potter, line 171: Why was the Lego test removed?

harper-core/tests/text/linters/Spell.snap.yml
```diff
@@ -17,17 +17,6 @@ Suggest:
 
 
 
-Lint:    Spelling (63 priority)
-Message: |
-       7 | My favourite color is blu.
-         |                       ^~~ Did you mean to spell `blu` this way?
```
elijah-potter, line 23: This was a proper correction.

harper-core/src/spell/word_id.rs
```diff
@@ -1,22 +1,60 @@
...
 
 use foldhash::fast::FixedState;
 use serde::{Deserialize, Serialize};
 
 use crate::{CharString, CharStringExt};
 
-/// An identifier for a particular word.
+/// An identifier for a particular word with canonical casing.
+#[derive(Hash, Copy, Clone, PartialEq, Eq, PartialOrd, Debug, Serialize, Deserialize)]
+pub struct CanonicalWordId {
```
elijah-potter: I think the `CanonicalWordId` and `CaseFoldedWordId` should be defined in separate files so that they're easier to find.

harper-core/src/spell/word_id.rs
```diff
@@ -1,22 +1,60 @@
...
+
+    /// Create a word ID from a string.
+    /// Requires allocation, so use sparingly.
+    pub fn from_word_str(text: impl AsRef<str>) -> Self {
+        let chars: CharString = text.as_ref().chars().collect();
+        Self::from_word_chars(chars)
+    }
+
+    /// Reinterpret this ID as a [`CaseFoldedWordId`].
+    ///
```
elijah-potter: Great comments.

harper-core/src/dict_word_metadata.rs
```diff
@@ -760,7 +731,34 @@ impl DictWordMetadata {
 
     /// Same thing as [`Self::or`], except in-place rather than a clone.
     pub fn append(&mut self, other: &Self) -> &mut Self {
-        *self = self.or(other);
+        macro_rules! merge {
```
elijah-potter, line 787: Out of curiosity, was there a reason to move this logic from the `or` function?

elijah-potter (changes requested):
I like this PR, but I have some notes. I strongly feel that we should revert your changes to how `SpellCheck` respects capitalization. It's letting a lot of false-positives through. 

Thanks for taking on this problem. You've done a lot of great work.

harper-core/src/linting/orthographic_consistency.rs
86xsk: We have both "Lego" and "LEGO" defined in the curated dictionary, and both variants appeared to be correct within the context of the test sentence as far as I could tell.

For instance, [the Wikipedia page for Lego](https://en.wikipedia.org/wiki/Lego) uses "Lego" throughout most of its writing, but it mentions that "LEGO" is how it's often stylized.

The only other alternative I can think right now would've been to remove "Lego" from the curated dictionary.

harper-core/src/spell/dictionary.rs
86xsk: I don't have any developed ideas off the top of my head, but I'll take a look and see if I can improve the API in any way. Especially since up until now I was more concerned with keeping the PR as small/focused as possible given the scale of the changes.

harper-core/src/dict_word_metadata.rs
86xsk: Since `or` clones, `append` calling `or` would also clone even though it's not necessary.

Since a `Vec` (via `DerivedFrom`) was added in `DictWordMetadata`, cloning would no longer be just a fixed-size memcpy, so I thought it would be prudent to avoid such cloning if possible.

harper-core/src/spell/dictionary.rs
hippietrail: > This function gets called _a lot_. Would it be possible to return a value that doesn't allocate?

You might be able to set it up as an iterator that copies each `DictWordMetadata` into a blank one on the stack? (I haven't actually looked into it, and could be overthinking it as well...)

harper-core/src/linting/orthographic_consistency.rs
hippietrail: > We have both "Lego" and "LEGO" defined in the curated dictionary, and both variants appeared to be correct within the context of the test sentence as far as I could tell.
> 
> For instance, [the Wikipedia page for Lego](https://en.wikipedia.org/wiki/Lego) uses "Lego" throughout most of its writing, but it mentions that "LEGO" is how it's often stylized.
> 
> The only other alternative I can think right now would've been to remove "Lego" from the curated dictionary.

I have both removed words/capitalizations from the dictionary and commented them out, to prevent bugs like this. But I ***hate*** doing it! In the real world it's perfectly normal that words can be spelled or hyphenated or capitalized multiple ways. Ideally fixing that would be in the scope of this PR. But if it's not I would prefer words that create problems to be commented out ready to be re-enabled when that can be fixed.

harper-core/tests/text/linters/Spell.snap.yml
hippietrail: > This was a proper correction.

I bet there's a `BLU` in the dictionary... Almost: `Blu` (no idea why)

Any word that's normally lowercase can appear in titlecase or all-caps, but the reverse is not true.

In the future we could support people writing in all-lowercase and in that context it would be fair to not flag `blu` when `Blu` is in the dictionary. But not while we lack such a feature.

... I bet "Blu" is there because it's a part of "Blu-Ray".

This is something else I hate. It's similar to adding single Spanish or French words to the dictionary because one or a handful of foreign phrases are used in English. IMHO the correct way to deal with these is to enhance `SpellCheck` to handle at least two-word lexemes separated by space or hyphen. Since Harper is a full grammar checker and not just a spelling checker that should remain on the table. (Not part of this PR of course.)

86xsk: > I bet there's a BLU in the dictionary... Almost: Blu (no idea why)

Yeah. The way it appears on the diff view is a little deceptive. The lint directly above that one showed that the word was still getting linted, but by `OrthographicConsistency` (suggesting it be corrected to `Blu`). (Though of course, that's a much less valuable suggestion in this context.)

Regardless, I've made `SpellCheck` case-sensitive again, seemingly without reintroducing the multiple case variants bug.

harper-core/src/spell/dictionary.rs
86xsk: I've tried, but I haven't been able to figure this one out yet.

- We can't just trivially return an `impl Iterator<...>` because generics aren't dyn-compatible. Returning a `Box<dyn Iterator<...>>` would be even worse, since you'd be allocating and then possibly paying the cost of dynamic dispatch on each call to `next` as well.
- I tried creating a custom iterator+slice type that would hold a reference to the underlying WordMap and a slice of indices from the WordMap's `case_folded`, but this required a lot of complex code, and it all fell apart because `MergedDictionary` holds multiple WordMaps.
- We could have a thread-local `RefCell<Vec>` and then reuse it, as is done in `within_edit_distance.rs`, I haven't tried this yet.

I did make some optimizations in other areas to counteract the performance regression of this PR though.

The benchmarks now show (on my machine):
- ~10% improvement on `parse_essay`, (from a77437d2f966aa6ad670fdbb1310cd5489c5c32a, 7f93814280f8888b6b22aa2b47a2968f9ce2344f)
- ~27% improvement on `lint_essay`, (from a5196da0b18bcc89e063b4de07a1d63c53453692)
- ~7% improvement on `lint_essay_uncached`, (also from a5196da0b18bcc89e063b4de07a1d63c53453692)

compared to master.

hippietrail: Some of this is deeper than my current Rust level. @elijah-potter is better at Rust and knows the overall Harper system much better so hereby asking for ideas/guidance.

harper-core/src/spell/dictionary.rs
86xsk: I've gone ahead and made some changes while experimenting with ways to avoid the allocation in `get_word_metadata`. To do this, I've split `Dictionary` into the new dyn-incompatible trait `CommonDictFuncs`, and made it return an `impl Iterator` directly, rather than collecting to a `Vec`.

I've also included some other breaking changes for things that were occupying my mind. For instance:
- `LintGroup::merge_from` now takes an owned value, rather than taking a `&mut LintGroup` and clearing it.
- Functions that return a fixed static value, like `FstDictionary::curated` now return a `&'static` rather than an `Arc`.
- Some of the dictionary functions which used to return `DictWordMetadata` now return `WordMapEntry`. This allows callers to get the canonical spelling of the word (via `WordMapEntry::canonical_spelling`) without having to make any additional calls to the dictionary, at the cost of it taking slightly more code to get to the `DictWordMetadata`. Some of the functions were renamed to reflect this change, e.g. `get_word_metadata` -> `get_word`.

I've also gone off-track and made some completely unrelated changes, like changing some types/functions that take generics to take dyns instead (where it didn't measurably hinder performance). This reduced the harper-ls release build size by about ~1MiB.

I would appreciate any opinions or ideas for improvement, especially since I may have overdone things, or gone down a subpar trail of thought. I would have no issues with rolling back some of the unrelated changes to keep things more focused.

harper-core/src/linting/word_press_dotcom.rs
elijah-potter, line 1: The brand WordPress.com, when expressed in snake case is `wordpress_dotcom`. Unfortunately, this is in Automattic's style guide. Would you keep it as it was?

harper-brill/src/lib.rs
```diff
@@ -1,39 +1,37 @@
...
 
 fn uncached_brill_tagger() -> BrillTagger<FreqDict> {
     serde_json::from_str(BRILL_TAGGER_SOURCE).unwrap()
 }
 
 /// Get a copy of a shared, lazily-initialized [`BrillTagger`]. There will be only one instance
 /// per-process.
-pub fn brill_tagger() -> Arc<BrillTagger<FreqDict>> {
-    (*BRILL_TAGGER).clone()
+pub fn brill_tagger() -> &'static BrillTagger<FreqDict> {
```
elijah-potter: Good stuff! Thanks!

harper-core/src/document.rs
```diff
@@ -102,18 +98,14 @@ impl Document {
 
     /// Parse text to produce a document using the built-in [`PlainEnglish`]
     /// parser and a provided dictionary.
-    pub fn new_plain_english(text: &str, dictionary: &impl Dictionary) -> Self {
+    pub fn new_plain_english(text: &str, dictionary: &WordMap) -> Self {
```
elijah-potter: Would you add a small comment here explaining to future readers why a `Dictionary` is no longer sufficient?

harper-core/src/document.rs
```diff
@@ -536,7 +512,7 @@ impl Document {
     }
 
     fn condense_dotted_truncations(&mut self) {
-        self.condense_expr(&Self::DOTTED_TRUNCATION_EXPR.with(|v| v.clone()), |_| {})
+        DOTTED_TRUNCATION_EXPR.with(|expr| self.condense_expr(expr, |_| {}));
```
elijah-potter: Noice!

harper-core/src/spell/merged_dictionary.rs
```diff
@@ -5,42 +5,45 @@ use std::sync::Arc;
...
 
 /// A simple wrapper over [`Dictionary`] that allows
 /// one to merge multiple dictionaries without copying.
 ///
 /// In cases where more than one dictionary contains a word, data in the first
 /// dictionary inserted will be returned.
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct MergedDictionary {
+    merged_word_map: WordMap,
```
elijah-potter, line 23: Would be nice to see a little comment that explains the relationship between the word map and the dictionaries.

elijah-potter (changes requested):
I apologize that it has taken me so long to look at this PR. I hope you understand: It's a lot to parse! That said, this PR is important to me, so I hope it only takes one more round of revision.

I have some small notes, but there's one big think I can't seem to figure out. Why is the `MutableDictionary` being replaced with the `WordMap`, which was intended to be an implementation detail? I'm not suggesting it's wrong, but rather that it wouldn't be my first instinct. I'm certain there's a bit I'm missing.

Thanks for making all of this happen. The overall quality of Harper's code would be greatly improved should this be merged.

harper-core/src/linting/word_press_dotcom.rs
86xsk: Sure thing.

The reason I initially did this is, in another (outdated and unpublished) branch, I was experimenting with ways to refactor `lint_group.rs`, and one of the issues I encountered was that converting from camel/snake case worked in all cases but that one.

(I was trying to use `paste`'s case conversion to be able to derive the `mod` name from the linter's name, avoiding the need for the `use` statements. There is certainly quite a few alternative ways to do that which don't require this change though.)

86xsk:
> I apologize that it has taken me so long to look at this PR. I hope you understand: It's a lot to parse! That said, this PR is important to me, so I hope it only takes one more round of revision.

No worries at all! I'm not sure if I would've been able to review something like this at all, which is a big part of the reason why I've been trying to split-off some of the freestanding and non-breaking changes in this PR into their own small PRs over the past couple of days.

> I have some small notes, but there's one big think I can't seem to figure out. Why is the `MutableDictionary` being replaced with the `WordMap`, which was intended to be an implementation detail? I'm not suggesting it's wrong, but rather that it wouldn't be my first instinct. I'm certain there's a bit I'm missing.

My thinking with that is that (partways through the PR) they ended up serving more or less the same purpose, and both are `pub`. So you end up in a situation where `MutableDictionary` provides helper functions, but doesn't expose all the functionality of `WordMap`, and you have `WordMap`, which can do everything that `MutableDictionary` can do and more, but it doesn't provide the same helper functions. I guess I was thinking something along the lines of: "this is unnecessarily complicated, why not just move those helper functions into `WordMap`?"

(For a somewhat similar reason, I also added functions to `Dictionary` that return a `&WordMapEntry` rather than a `&DictWordMetadata`.  A `WordMapEntry` contains the metadata, but it also contains the canonical spelling of the word, which is something the caller might want.)[^2]

---
(The below is kind of unrelated, but I wrote it out before realizing, and it might provide a bit more context behind some of the changes.)

Another thing is that, especially earlier on in this PR, I was trying to reduce the size of the `Dictionary` trait as much as possible, and even thinking of removing it entirely (or at least, removing the requirement for dyn-compatibility). There were a few reasons for this:
1. Before adding default implementations to most `Dictionary` functions, changing the signature of any of them required jumping around 4-5 files to repeat the changes, which was tedious.
1. The `Dictionary` trait needed to be dyn-compatible, which made it difficult to avoid allocating return types. (e.g. https://github.com/Automattic/harper/pull/2630#discussion_r2810461338)
1. It led to the use of `dyn`/`impl` in places were it wasn't beneficial/necessary[^1].

[^1]: `dyn` adds dynamic dispatch, `impl` (when used as the type of a function argument) acts like a generic, which can result in a lot of monomorphization, increasing the size of the executable and compile times. For instance, last I checked, this PR reduces the release build size of harper-ls by ~1MiB, primarily by removing the use of `impl` in places. Removing the `impl Dictionary` in `Document` functions saved ~0.5MiB alone.
[^2]: I didn't get around to it, but I feel like merging `DictWordMetadata` and `WordMapEntry` is something that might be worth exploring at some point.

harper-core/src/document.rs
86xsk: I changed those `Document` functions to make them take `&dyn Dictionary` instead, and I added a comment explaining why we're avoiding the use of `&impl Dictionary` above the `impl Document {`

(Taking `&dyn Dictionary` seemed to reduce the size of harper-ls by an additional ~0.02MiB, and it doesn't seem to have negatively affected the benchmarks either.)

86xsk:
I replaced uses of `WordMap` with `MutableDictionary`/`FstDictionary` in places unrelated to dictionary implementation.

One idea that occurred to me was merging (what remains of) `MutableDictionary` into `WordMap`, and then renaming `WordMap` into something like `MapDictionary`, partly because it would remove the `MutableDictionary -> WordMap` type alias, partly because the name `MutableDictionary` seems like it might become problematic if we add another type of dictionary that is mutable in the future. I'm not sure that's a good idea though, I would appreciate any thoughts.

harper-cli/src/main.rs
```diff
@@ -1,14 +1,12 @@
...
-    CanonicalWordId, CommonDictFuncs, Dictionary, FstDictionary, MutableDictionary,
-};
+use harper_core::spell::{CanonicalWordId, Dictionary, FstDictionary, MutableDictionary};
 use hashbrown::HashMap;
 use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::{self, BufReader};
 use std::path::PathBuf;
-use std::sync::Arc;
+// use std::sync::Arc;
```
hippietrail, line 9: Is this intentionally left commented-out rather than removed?

86xsk: Sort of, I guess? That's the way it currently is on master, and I restored it as part of an effort to minimize the size of the PR as much as possible. (There's still a bit of work to be done on that end.)

harper-core/src/dict_word_metadata_orthography.rs
```diff
@@ -54,6 +54,18 @@ impl Default for OrthFlags {
 }
 
 impl OrthFlags {
+    /// An `OrthFlags` which all casing-related flags set.
```
hippietrail: Is this a typo for An `OrthFlags` **with** all casing-related flags set?

elijah-potter:
@86xsk, please let me know when you'd like me to re-review this.

harper-cli/src/main.rs
```diff
@@ -1039,11 +1039,14 @@ fn line_to_parts(line: &str) -> (String, String) {
...
 
-    let children = dictionary
-        .words_iter()
-        .filter(|e| dictionary.get_word_metadata(e).unwrap().derived_from == Some(id));
+    let children = dictionary.get_word_map().iter().filter_map(|wme| {
+        wme.metadata
+            .derived_from
+            .contains(id)
+            .then_some(&wme.canonical_spelling)
+    });
```
86xsk: This logic had to be updated since derived_from is now a Vec of word IDs, rather than a single one.

harper-core/src/dict_word_metadata/derived_from.rs
```diff
@@ -0,0 +1,73 @@
...
+};
+
+use serde::{Deserialize, Serialize};
+
+use crate::spell::CanonicalWordId;
+
+/// A container for storing word IDs that a word is considered to be derived from.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Hash)]
+pub struct DerivedFrom {
+    inner: BTreeSet<CanonicalWordId>,
```
86xsk: Have to use `BTreeSet` to satisfy `PartialOrd`, which is required by `DictWordMetadata`.

harper-core/src/linting/sentence_capitalization.rs
```diff
@@ -68,7 +68,7 @@ impl<T: Dictionary> Linter for SentenceCapitalization<T> {
                                 .take_while(|&c| !c.is_whitespace() && *c != '-' && *c != '\'')
                                 .any(|&c| c.is_uppercase())
                             {
-                                continue;
+                                continue 'sentence;
```
86xsk: This certainly isn't the ideal way to implement this (for instance, the first if check could be hoisted out of the loop), but refactoring this would make the diff harder to follow. It can easily be refactored later.

harper-core/src/linting/split_words.rs
```diff
@@ -101,21 +101,21 @@ impl ExprLinter for SplitWords {
...
-            let Some(cand_meta) = self.dict.get_word_metadata(candidate) else {
+            let Some(cand_meta) = self.dict.get_word_metadata_exact(candidate) else {
                 continue;
             };
             if !cand_meta.common {
                 continue;
             }
 
-            let Some(rem_meta) = self.dict.get_word_metadata(remainder) else {
+            let Some(rem_meta) = self.dict.get_word_metadata_exact(remainder) else {
```
86xsk: This produces a lot of false-positives without `exact`.

harper-core/src/spell/dictionary.rs
```diff
@@ -1,57 +1,293 @@
...
-    /// Check if the dictionary contains the exact capitalization of a given word.
-    fn contains_exact_word_str(&self, word: &str) -> bool;
+    // NOTE: methods with the bound `Self: Sized` bound are not dynamically dispatchable
+    // (i.e. they can't be called through a &dyn Dictionary)
+    //
+    // To get around this, you can call `get_word_map` and then call the desired method on the
+    // resulting word map.
+
+    /// Get the underlying [`WordMap`] used by the dictionary.
+    fn get_word_map(&self) -> &WordMap;
```
86xsk: This is needed to support dyn-incompatible return types (i.e. `impl Iterator<...>`) and to provide default implementations for many of `Dictionary`'s functions.

harper-core/src/spell/fst_dictionary.rs
```diff
@@ -52,22 +53,22 @@ impl FstDictionary {
         (*DICT).clone()
     }
 
-    /// Construct a new [`FstDictionary`] using a wordlist as a source.
-    /// This can be expensive, so only use this if fast fuzzy searches are worth it.
-    pub fn new(mut words: Vec<(CharString, DictWordMetadata)>) -> Self {
-        words.sort_unstable_by(|(a, _), (b, _)| a.cmp(b));
-        words.dedup_by(|(a, _), (b, _)| a == b);
```
86xsk: Don't need to dedupe since `WordMap` is already guaranteed to contain no duplicates.

harper-core/src/spell/dictionary.rs
```diff
@@ -1,57 +1,293 @@
...
+    spell::{CanonicalWordId, CaseFoldedWordId, WordMap, WordMapEntry},
+};
+
+pub(crate) static CURATED_DICT_STR: &str = include_str!("../../dictionary.dict");
+pub(crate) static ANNOTATIONS_STR: &str = include_str!("../../annotations.json");
 
 /// An in-memory database that contains everything necessary to parse and analyze English text.
 ///
 /// See also: [`super::FstDictionary`] and [`super::MutableDictionary`].
-#[blanket(derive(Arc, Ref))]
```
86xsk: blanket doesn't seem to work when some functions contain the `where Self: Sized` bound. In lieu of this, I've added manual impls below this trait.

harper-core/src/spell/fst_dictionary.rs
```diff
@@ -18,7 +19,7 @@ use crate::{CharString, CharStringExt, DictWordMetadata};
 pub struct FstDictionary {
     /// Underlying [`super::MutableDictionary`] used for everything except fuzzy finding
     mutable_dict: Arc<MutableDictionary>,
-    /// Used for fuzzy-finding the WordId of words or metadata
+    /// Used for fuzzy-finding the index of words or metadata
```
86xsk: Not a necessary change. IndexMap allows for retrieving values by index, which I believe is slightly more efficient than retrieving values by their keys.

I didn't undo this along with the other unnecessary changes since WordId would need to be changed to CanonicalWordId anyway, and I'm not aware of any benefit that using a word ID would have over using the word index here. Though if there is something, I'd be happy to revert this change as well.

Issue #293, opened 2024-11-28
feat: Ignorable `Matcher` triggers.


grantlemons:
Root Cause of: #243, #287 

The Matcher (`harper-core/src/linting/matcher.rs`) linter recommends replacements in very specific cases. Unfortunately it has some problems.
One of these is that the recommendations cannot be ignored without disabling the lint rule entirely.

I think it would be good to create a more granular configuration for Matcher, so users can enable and disable different recommendations as the choose, but the key thing in my eyes is that users should be able to ignore these, as the fact these behave a lot like spelling suggestions leads users to expect them to be able to interact with them as they would a spelling suggestion.

elijah-potter:
I believe the real solution is to refactor the various ideas presented in the `Matcher` into individual linters (which can be enabled/disabled). 

[This commit](https://github.com/Automattic/harper/commit/d8967394b3e4b88837440b4184ec3328b84fad8f#diff-453075085818b767acf7579938091f49476d15d902b229dbae0620827ff79f64) is a good example of what I mean.

elijah-potter:
This has been completed for quite a while now.

PR #2638, opened 2026-02-01
feat: worth to do → worth doing

### LINKED ISSUE #789: flag "worth to do" instead of "worth doing" etc
## Description

Non-native speakers often use "worth to \<verb>" where standard English uses "worth \<verb>ing", "worth it to \<verb>" etc.

## Resources

- [English StackExchange](https://english.stackexchange.com/questions/507711)
- [r/grammar](https://www.reddit.com/r/grammar/comments/nu8nta/worth_to_do_worth_doing/)
- [r/EnglishLearning](https://www.reddit.com/r/EnglishLearning/comments/zasgop/is_it_worth_buying_is_it_worth_to_buy_both_correct/)
- [Quora](https://www.quora.com/Which-is-the-correct-way-to-say-it-worth-to-do-or-worth-doing)
- [WordReference.com forum](https://forum.wordreference.com/threads/worth-to-do-vs-worth-doing.1318799/)
- [English Learners StackExchange](https://ell.stackexchange.com/questions/327829)
- [HiNative](https://hinative.com/questions/17164352)

## Examples

- The fundamental question is: is it **worth to have** this proposal if `#{ x: 1 } !== #{ x: 1 }`?
- I was wondering, if it would be any **worth to** PR the addition of simple/codeberg to the Repository Icon section
- [How the UITests could be important? is it **worth to integrate** ...](https://github.com/mobile-dev-inc/maestro/discussions/2066)
- So my question is will it be **worth to start** exploring with .net maui.

## Potential False Positives

- The pattern "... X is/are/was/etc. worth to Y", as in "what is it worth to you" is common. The key would be a verb vs a noun after "to". But many words are both verbs and nouns and not people using words or abbreviations etc. as verbs as in the "worth to PR" example

hippietrail:
# Issues 

Fixes #789 

# Description

Changes the verb lemma (present tense) from to gerund in phrases such as "Is it worth to do?" -> "Is it worth doing?"

# How Has This Been Tested?

There's a whole bunch of unit tests based on sentences found on GitHub that should cover all cases.

# Checklist

### DIFF
--- a/harper-core/src/linting/worth_to_do.rs
+++ b/harper-core/src/linting/worth_to_do.rs
@@ -1,5 +1,6 @@
 use crate::{
-    CharStringExt, Lint, TokenStringExt,
+    CharStringExt, Lint, Token, TokenStringExt,
+    char_ext::CharExt,
     expr::{Expr, SequenceExpr},
     linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk},
     spell::Dictionary,
@@ -41,7 +42,7 @@ where
         self.expr.as_ref()
     }
 
-    fn match_to_lint(&self, toks: &[crate::Token], src: &[char]) -> Option<super::Lint> {
+    fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {
         let tolemtoks = &toks[toks.len() - 3..];
         let lemtok = toks.last()?;
         let tolemspan = tolemtoks.span()?;
@@ -65,10 +66,7 @@ where
         }
 
         if let Some(last_letter) = lemstr.chars().last()
-            && !matches!(
-                last_letter.to_ascii_lowercase(),
-                'a' | 'e' | 'i' | 'o' | 'u'
-            )
+            && !last_letter.is_vowel()
         {
             let double_consonant = format!("{}{}ing", lemstr, last_letter);
             if self.dict.contains_word_str(&double_consonant) {


harper-core/src/linting/worth_to_do.rs
```diff
@@ -0,0 +1,323 @@
...
+            let replace_e_with_ing = format!("{}ing", &lemstr[..lemstr.len() - 1]);
+            if self.dict.contains_word_str(&replace_e_with_ing) {
+                gerunds.push(replace_e_with_ing);
+            }
+        }
+
+        if let Some(last_letter) = lemstr.chars().last()
+            && !matches!(
+                last_letter.to_ascii_lowercase(),
+                'a' | 'e' | 'i' | 'o' | 'u'
```
86xsk: If you import `CharExt`, you can do `!last_letter.is_vowel()`.

harper-core/src/linting/worth_to_do.rs
```diff
@@ -0,0 +1,323 @@
...
+where
+    D: Dictionary,
+{
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        self.expr.as_ref()
+    }
+
+    fn match_to_lint(&self, toks: &[crate::Token], src: &[char]) -> Option<super::Lint> {
```
86xsk: `super::` is redundant since `Lint` is already imported.

`crate::` isn't, but I would consider just importing `crate::Token` too.

hippietrail: > `super::` is redundant since `Lint` is already imported.
> 
> `crate::` isn't, but I would consider just importing `crate::Token` too.

D'oh I normally type those bits manually but it looks like the LLM beat me and made it messy. Thanks!

elijah-potter:
> Fixes https://github.com/Automattic/harper/issues/293

What do you mean by this?

hippietrail:
> > Fixes #293
> 
> What do you mean by this?

Hmm that's definitely wrong. Let me figure out what I did / what the right issue is... Should be the correct one now: https://github.com/Automattic/harper/issues/789

PR #2651, opened 2026-02-04
fix: don't lowercase "This"

### LINKED ISSUE #2459: `UseTitleCase` seems overzealous/too opinionated insisting "This" should be "this".
**What got flagged?**

Image

In our own GitHub template this rule wants us to change "This", which looks right to me, to "this", which looks wrong to me.

[The Wikipedia page on titlecase](https://en.wikipedia.org/wiki/Title_case) references half a dozen style guides and they all say pronouns should have their first letter capitalized.

A caveat is that if's very common to talk of pronouns while only thinking about personal pronouns. "This" is a demonstrative pronoun and a determiner but the Wikipedia article and the style guide excerpts do not mention those categories either way.

hippietrail:
# Issues 

Fixes #2459

# Description

The code has some special logic that was confusing due to using a double negative, around handling non-possessive determiners (my, our, your, his, her, its, their).

I just removed that check and nothing broke.

I did want to disable this linter by default but doing so causes `harper-core/tests/test_sources/title_case_errors.md` to fail. I think `harper-core/tests/run_tests.rs` assumes tested linters are enabled by default and a way to test linters that default off is lacking?

If I'm missing something please let me know!

# Demo

# How Has This Been Tested?

I added some new unit tests based on the GitHub template heading from the bug report and using all of the possessive determiners.

# Checklist

### DIFF
--- a/harper-core/src/linting/use_title_case.rs
+++ b/harper-core/src/linting/use_title_case.rs
@@ -44,7 +44,7 @@ impl<D: Dictionary + 'static> Linter for UseTitleCase<D> {
 
 #[cfg(test)]
 mod tests {
-    use crate::linting::tests::assert_suggestion_result;
+    use crate::linting::tests::{assert_no_lints, assert_suggestion_result};
     use crate::spell::FstDictionary;
 
     use super::UseTitleCase;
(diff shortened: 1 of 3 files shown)


elijah-potter:
The merge failed because GitHub crashed.

PR #2657, opened 2026-02-04
refactor(harper.js): we do not need named types for Weirpacks

elijah-potter:
# Description

When looking through the `harper.js` API reference, I saw that there were some named types that I would prefer to leave as-is. I've updated it to be clearer.

# How Has This Been Tested?

This is a type-only change. No logic was affected.

# Checklist


elijah-potter:
I've also added some documentation for things that needed clarity.

PR #2660, opened 2026-02-05
harper-core: flag 'few' in more time contexts

### LINKED ISSUE #1114: Flag "few" used when it should be "a few"
## Description

This is extremely common among Arabic speakers no matter how good their English is, but I'm pretty sure others make it too.

"Few" has the implication of "hardly any", insufficient, not enough, so tends negative.
"A few" on the other hand tends positive as in "better than nothing" or "not a lot but probably enough".

## Resources

- https://www.grammarly.com/blog/commonly-confused-words/few-a-few/
- https://www.britannica.com/dictionary/eb/qa/The-Difference-Between-Few-and-A-Few-

## Examples

- [A teacher is in the process of generating **few** reports based ...](https://gist.github.com/rajkushwaha0521/09cd5b807b5c84b9df36432d22c5d5e9)
- After **few** minutes all kube-system pods begin crashing.
- [Tomcat 9.0 server stop **few** minutes after start without any ...](https://github.com/redhat-developer/vscode-rsp-ui/issues/282)
- [Fix **few** typos](https://github.com/usetrmnl/firmware/actions/runs/13876611431)
  This is a pretty good example as to a native English speaker it looks like an instruction to fix some typos but not to fix all of them.
- [After **few** seconds window turns white · Issue #11099](https://github.com/podman-desktop/podman-desktop/issues/11099)
- [It's great and now I am sharing **few** scripts.](https://github.com/ggml-org/whisper.cpp/discussions/2894)
  Implies that previous more scripts were shared.
- **Few** minutes after I posted this I found that unless I set the below explicitly in settings:
- can you tell me **few** specific problems with using rust for systems programming?

## Potential False Positives

I imagine it will be quite difficult to design heuristics to detect this.

LouisLau-art:
Closes #1114.

## What changed
- Expand the existing `FewUnitsOfTimeAgo` rule from only matching `few <time unit> ago` to matching `few <time unit>` and then using surrounding context to decide whether it should be corrected to `a few`.
- This allows catching very common constructions like `After few minutes ...` and `Few minutes after ...`.

## Heuristics (intentionally conservative)
Triggers when `few <time unit>` is followed by `ago/after/before/later` or preceded by `after/before/in/within/for/since`.
Skips when it already has an article (`a/an/the`) or when `few` is clearly negative (`too/very/so/quite/how few`).

## Tests
- Re-enable sentence-start cases (previously ignored).
- Add new test cases for `after few minutes` / `few minutes after`.
- Add basic false-positive guards (`the few ...`, `too few ...`, `past few ...`).


harper-core/src/linting/few_units_of_time_ago.rs
```diff
@@ -40,32 +33,93 @@ impl ExprLinter for FewUnitsOfTimeAgo {
...
+                "this"
+                    | "that"
+                    | "these"
+                    | "those"
+                    | "my"
+                    | "our"
+                    | "your"
+                    | "his"
+                    | "her"
+                    | "their"
```
hippietrail, line 81: See if `.is_determiner()` on `TokenKind` works here.

harper-core/src/linting/few_units_of_time_ago.rs
```diff
@@ -40,32 +33,93 @@ impl ExprLinter for FewUnitsOfTimeAgo {
...
-        for tok in toks.iter().take(3) {
-            if tok.span.get_content_string(src).eq_ignore_ascii_case("few") {
-                span = Some(tok.span);
-                break;
-            }
+        // If `few` is explicitly being used in the negative/insufficient sense (e.g. `too few`),
+        // inserting an article would be incorrect.
+        if matches!(
+            prev_word.as_deref(),
+            Some("too" | "very" | "so" | "quite" | "how")
```
hippietrail, line 92: See if `.is_degree_adverb()` on `TokenKind` works here.

hippietrail (commented):
So cool to see someone working on this! I made the original `few_units_of_time_ago` linter before I added `match_to_lint_with_context()`

LouisLau-art:
Thanks for the context — really appreciate you building the original linter! I used the existing behavior as a baseline and tried to keep the new context checks minimal + well-covered by tests.

harper-core/src/linting/few_units_of_time_ago.rs
elijah-potter: I think this PR is close, but I am also interested if that method would work instead. @LouisLau-art, would you give it whirl?

elijah-potter (changes requested):
@LouisLau-art, I'd love to get this merged. Would you give that method a whirl and remove the parenthesized bit at the end of the description? Thanks!

harper-core/src/linting/few_units_of_time_ago.rs
hippietrail: It's a method on `TokenKind`, which is accessed from the `.kind` field of the `Token`. Note that my/our/your/his/her/their are considered "possessive determiners" so should be included. If not there's a method to check for those too. Did you leave out "its" intentionally?

hippietrail (commented):
I added some thoughts to my previous review.

elijah-potter:
I'm going to close this for now, since we haven't heard back from you in a while. Please feel free to open a new PR if you decide to revisit this.

PR #2664, opened 2026-02-05
feat(core): create expression references

elijah-potter:
# Issues 

N/A

# Description

This PR introduces "expression references", which allow you to define reusable expressions in Weir that can be composed together to create progressively more complex behavior. I've also included a new rule that detects missing determiners that takes advantage of the new syntax. I suggest you read the updated documentation in this PR and take a look at the example rule for an idea of what it looks like.

# How Has This Been Tested?

Extensive unit tests.

# Checklist

### DIFF
--- a/harper-core/src/weir/mod.rs
+++ b/harper-core/src/weir/mod.rs
@@ -6,12 +6,12 @@ mod error;
 mod optimize;
 mod parsing;
 
-use std::collections::{HashSet, VecDeque};
+use std::collections::VecDeque;
 use std::str::FromStr;
 use std::sync::Arc;
 
 pub use error::Error;
-use hashbrown::HashMap;
+use hashbrown::{HashMap, HashSet};
 use is_macro::Is;
 use parsing::{parse_expr_str, parse_str};
 use strum_macros::{AsRefStr, EnumString};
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/weir_rules/MissingDeterminer.weir
```diff
@@ -0,0 +1,73 @@
...
+
+expr requestMissingDet <([get, provide, give, send, share, attach, include, add, need, want, see, submit, create, report, file, reproduce] @requestBareNounPhrase), ( )>
+expr narrativeVerbObjectMissingDet <([painted, built, dropped, raised, found, hid, cheered] @narrativeBareNounPhrase), ( )>
+expr narrativePrepObjectMissingDet <([in, on] [studio, tree, floor]), ( )>
+
+expr main [@requestMissingDet, @narrativeVerbObjectMissingDet, @narrativePrepObjectMissingDet]
+
+let message "Add a determiner before this noun phrase."
+let description "Detects likely missing determiners in common request phrases and offers to insert one where necessary."
+let kind "Grammar"
```
hippietrail, line 15: These could all also be fixed by making the noun plural, equally common too, especially among people whose first language is an East Asian language.

hippietrail (commented):
Determiner is also missing when the noun phrase is supposed to be plural but is singular by mistake - very common.

PR #2671, opened 2026-02-07
Add corrections for idioms and eggcorns

JanDalhuysen:
"all of the sudden" -> "all of a sudden"

"for all intensive purposes" -> "for all intents and purposes"

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -61,6 +61,14 @@ pub fn lint_group() -> LintGroup {
             "Corrects `adieu` to `ado`.",
             LintKind::Eggcorn
         ),
+        "AllOfTheSudden" => (
+            &[
+                ("all of the sudden", "all of a sudden"),
+            ],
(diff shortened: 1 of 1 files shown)


hippietrail:
Harper already handles both of these. Maybe you have them turned off in your settings?
image
image
----
Oh sorry I didn't realize this was a pull request. I was thinking it was just a feature request.

We used to use `phrase_corrections` for these but we replaced those with `.weir` files though we did keep `phrase_set_corrections` which group a bunch of related ones under a single rule, such as different inflections of the verb or singular and plural nouns.

Let me make sure the ones you worked on are Weir files... Here:
- `harper-core/src/linting/weir_rules/AllOfASudden.weir`

I think "all intensive purposes was both a `phrase_correction` and then a `.weir` file but very recently I found more variants of it in the wild and turned into a full linter here:
- `harper-core/src/linting/all_intents_and_purposes.rs`

elijah-potter:
I appreciate that you were so eager to contribute, @JanDalhuysen, but as @hippietrail mentioned, we already support these cases. I'm going to close this PR, but feel free to contribute any others you might think of!

PR #2680, opened 2026-02-09
feat: thru→through

hippietrail:
# Issues 
N/A

# Description

I noticed that the spelling "thru", mostly used in advertising but probably also texting, doesn't get flagged at all.
I've added a Weir rule to change it to "through".

# How Has This Been Tested?

I added a sentence from GitHub as a unit test.

# Checklist


elijah-potter (approved):

elijah-potter:
The merge failed because GitHub crashed.

Issue #948, opened 2025-03-20
Feature request: add user-configurable regex Masker


holmanb:
Sections can be ignored by block and in-line quotes ('`', '```'), but this requires modifying comments which otherwise wouldn't quoting. It would be quite powerful to allow users to provide regex which allows them to describe patterns which should be ignored by default.

Consider the following examples:

Sometimes shell commands are useful in code comments, for example:
```bash
# Install dependencies with the following command
# $ some_command.sh
```

Similarly editor commands might have another common syntax:

```lua
-- View active clients with the following command
-- :lua print(vim.inspect(vim.lsp.get_active_clients()))
``` 

or
```vimscript
" Set color with the following command
" :colo darkblue
``` 

A user might want to use a regex such as `[\:\$].*$` to ignore comment content which is known to match one of these conventions, which are not necessarily a structured comment syntax.

alcroito:
Any chance this is re-considered? Is there a reason this is closed as not planned, rather than just "we didn't get to it yet"?

hippietrail:
> Any chance this is re-considered? Is there a reason this is closed as not planned, rather than just "we didn't get to it yet"?

It was closed by a bot that went a bit wild a couple of months ago. I think it's up to @elijah-potter to decide if this is practicable for reopening...

elijah-potter:
This was closed erroneously.

alcroito:
Any news if the obsidian regex masker could be brought to the lsp server as well, so all editors can benefit from it?

PR #2684, opened 2026-02-09
feat(obsidian): allow users to mask out text using regular expressions

### LINKED ISSUE #2634: Add rule to exclude checking for [[links]]
My links often to contain complete or correct grammar, since I sometimes use citation keys from Zotero, and then files have names such as Smith2020Thing which of course isn't a valid word.

**Proposed Solution**
Add a rule to stop checking inside [[square brackets]].

**Examples**

Image

**Additional Context**
I think it was discussed here https://github.com/Automattic/harper/issues/112 but as far as I can tell this hasn't been implemented.

elijah-potter:
# Issues 

Should fix #2634

# Description

In this PR, I've created the necessary logic to ignore portions of documents from Harper's gaze using regular expressions. I've exposed the functionality in the Obsidian settings menu (pictured below).
If nothing goes awry, we can expose similar functionality in the language per #948.

# Demo

image

# How Has This Been Tested?

Additional unit tests + manual testing.

# Checklist

### DIFF
--- a/harper-core/src/mask/regex_masker.rs
+++ b/harper-core/src/mask/regex_masker.rs
@@ -1,6 +1,6 @@
 use regex::Regex;
 
-use crate::Span;
+use crate::{Span, offsets::build_byte_to_char_map};
 
 use super::{Mask, Masker};
 
@@ -28,14 +28,7 @@ impl RegexMasker {
 impl Masker for RegexMasker {
     fn create_mask(&self, source: &[char]) -> Mask {
         let source_s: String = source.iter().collect();
-        let mut byte_to_char = vec![0; source_s.len() + 1];
-
-        let mut char_idx = 0;
-        for (byte_idx, c) in source_s.char_indices() {
-            byte_to_char[byte_idx] = char_idx;
-            char_idx += 1;
-            byte_to_char[byte_idx + c.len_utf8()] = char_idx;
-        }
+        let byte_to_char = build_byte_to_char_map(&source_s);
 
         let mut mask = Mask::new_blank();
 
--- a/harper-core/src/lib.rs
+++ b/harper-core/src/lib.rs
@@ -19,6 +19,7 @@ mod lexing;
 pub mod linting;
 mod mask;
 mod number;
+mod offsets;
 pub mod parsers;
 pub mod patterns;
 mod punctuation;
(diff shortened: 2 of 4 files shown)


harper-core/src/mask/regex_masker.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn create_mask(&self, source: &[char]) -> Mask {
+        let source_s: String = source.iter().collect();
+        let mut byte_to_char = vec![0; source_s.len() + 1];
+
+        let mut char_idx = 0;
+        for (byte_idx, c) in source_s.char_indices() {
+            byte_to_char[byte_idx] = char_idx;
+            char_idx += 1;
+            byte_to_char[byte_idx + c.len_utf8()] = char_idx;
+        }
```
86xsk: It might be a good idea to extract this into a function. (There also seems to be very similar code in the Markdown parser, if you grep the repo for `byte_to_char`. Changing that might be out of scope, but you could probably use the same function for both.)

harper-core/src/mask/regex_masker.rs
```diff
@@ -0,0 +1,137 @@
+use regex::Regex;
+
+use crate::Span;
+
+use super::{Mask, Masker};
+
+pub struct RegexMasker {
+    regex: Regex,
+    exclude_matches: bool,
```
86xsk, line 10: I think it would be helpful to document what `exclude_matches` means.

harper-core/src/mask/regex_masker.rs
elijah-potter: I've gone ahead and done that. I wasn't sure what to call it. Let me know what your thoughts are.

86xsk: LGTM :+1: 
The concise documentation and tests also make the code a lot easier to understand.

andras-tim:
Thank you for the quick fix.

alcroito:
If a variant of this gets merged to fix #948 it would be nice if it took a list of regexes, not just one. The regexes would then be `OR`ed.
It's nicer to specify a list, rather than having to have one long textfield with lots of `|`.

elijah-potter:
Good thinking! I'll definitely consider making it a list.

PR #2082, opened 2025-10-15
feat: add jjdescription parser

Tijs-B:
# Issues 

jjdescription was briefly mentioned in https://github.com/Automattic/harper/issues/79.

# Description

jjdescription file types are used by [jujutsu](https://jj-vcs.github.io/jj/latest/) (a VCS) to write messages in, similar to git commits. We can use tree-sitter-jjdescription and the existing markdown parser to parse these kinds of files. It is super helpful to have spell checking on these files as most of the plain English I write as a developer is in jj descriptions.

# Demo

# How Has This Been Tested?

I'm running this already for some time (in helix), found no issues.

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests to cover my changes

PR #2688, opened 2026-02-10
feat: Add Zig support

### LINKED ISSUE #218: Zig Support

hippietrail:
# Issues 

Resolves #218

# Description

This has been on my TODO list for a while. I felt that the AI coding agent I've been using is probably good enough to try this.
I specifically asked it to base its implementation on the `jjdescription` support from PR #2082 

I guess `jjdescription` doesn't support `harper-cli` which seems to make sense, so I asked the agent to implement that too.

The integration tests it made pass and I've tried testing on some Zig code I have lying around with `harper-cli` and it seems to work.

But since this is AI generated code, please would some people that know how Tree-sitter language support works give it a thorough check.

There are two Zig Tree-sitter grammars on GitHub: [tree-sitter-grammars/tree-sitter-zig](https://github.com/tree-sitter-grammars/tree-sitter-zig) and [tree-sitter/zig-tree-sitter](https://github.com/tree-sitter/zig-tree-sitter) - I asked the agent to use the latter.

Zig only uses one type of comment: `//` which makes it probably more straightforward that most programming languages.

----

**Update**

It turns out using `jjdescription` was not a good idea since programming languages with Tree-Sitter parsers should use `harper-comments`

While guiding the agent on how to rework this PR I discovered that though Zig only supports `//` and not `/* */` comments, it does actually support multiple kinds of comments. The other kinds are doc comments `///` and top-level doc comments `//!` - this version supports all three.

# How Has This Been Tested?

AI-generated integration tests + manual testing.
For the updated version the AI generated "clean" and "dirty" test files similar to what other languages use.

I found Zig files on my system that had all three kinds of comments to test manually.

# Checklist

- [x] <s>I have</s> The coding agent has added tests to cover my changes


elijah-potter:
Since this is a `TreeSitterParser`, is there a reason it couldn't be made a part of `harper-comments`?

hippietrail:
> Since this is a `TreeSitterParser`, is there a reason it couldn't be made a part of `harper-comments`?

No reason at all!

Ah I guess `jjdescription` doesn't apply there since it's not a programming language, so not the ideal template I thought it might be?

elijah-potter:
> Ah I guess jjdescription doesn't apply there since it's not a programming language, so not the ideal template I thought it might be?

Yeah, that was the reason it wasn't added to the comment package. There _are_ languages that should be implemented like the jjdescription parser, but not Zig. Would you mind reworking it?

hippietrail:
> > Ah I guess jjdescription doesn't apply there since it's not a programming language, so not the ideal template I thought it might be?
> 
> Yeah, that was the reason it wasn't added to the comment package. There _are_ languages that should be implemented like the jjdescription parser, but not Zig. Would you mind reworking it?

On it!

hippietrail:
> LGTM. Thanks for the iteration!

Thanks for the perfect feedback!

PR #2693, opened 2026-02-11
fix(lint-framework): `SuggestionKind.InsertAfter` removing text

elijah-potter:
# Description

I noticed that when the refactored lint framework (after @draphy's fixed) encountered an instance of `SuggestionKind.InsertAfter`, it would replace the text preceding the actual problem spot. I've fixed this by applying the suggestion before passing it to the new code that actually interacts with the DOM.

# How Has This Been Tested?

Manually.

# Checklist


draphy:
Good catch! Should we also add tests covering the `InsertAfter` suggestion flows so future changes don't break this? I'm Happy to add those in this PR if you'd like.

elijah-potter:
I'm not sure, since there isn't a reliable way of triggering that kind of lint right now. It's a system that is in flux. I think the best move is to merge this now. If the problem crops up again after we've solidified that system, we'll add the appropriate tests.

PR #2698, opened 2026-02-12
docs: add Zig to list of supported languages and other missing pieces

hippietrail:
# Issues 
N/A

# Description
In my recent Zig support PR I overlooked adding Zig to the list of supported languages.

Later I was unable to test it in VS Code though it works fine in `harper-cli` and I found some more files I hadn't added or modified.

But VS Code still doesn't seem to be recognizing zig in my testing - so something else must still be wrong.

# Checklist


hippietrail:
Well the agent put on a very poor show when I brought it out in public. This PR is too much of a mess to save now even after I told it not to push!

PR #2718, opened 2026-02-13
fix(core): the `NoLonger` lint will now also flag preceding an adjective

### LINKED ISSUE #2696: The `NoLonger` linter should also flag "not longer" before adjectives
It's especially common before multi-syllable adjectives for some reason:
- not longer able
- not longer available
- not longer possible
- not longer responsive

It's possible after adjectives in the comparative or superlative form, but it's not easy to find examples. So might be best to only look for positive form to avoid false positives.

It's perhaps likely that matching with adjectives will lead to more false positives than seems to be the case with verbs.

elijah-potter:
# Issues 

Fixes #2696

# Description

I've added an adjective match to the expression and added some new tests. See original issue for more details.

# How Has This Been Tested?

Additional unit tests.

# Checklist

### DIFF
--- a/harper-core/src/linting/no_longer.rs
+++ b/harper-core/src/linting/no_longer.rs
@@ -154,11 +154,29 @@ mod tests {
     }
 
     #[test]
+    fn fix_able_adjective() {
+        assert_suggestion_result(
+            "After the migration, we were not longer able to sign in.",
+            NoLonger::default(),
+            "After the migration, we were no longer able to sign in.",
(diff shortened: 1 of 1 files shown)


hippietrail:
Nice! I also did a PR for this today. The only difference is mine has a whole bunch more unit tests since I was hunting for false positives - but found none (-:

elijah-potter:
Closing in favor of #2723.

PR #2724, opened 2026-02-14
chore: Curate dictionary

anyllmarkevich:
# Issues 

# Description
Dictionary changes:
- generalize + B
- +GIS (+N) (Geographic Information System)
- USAF + g
- EU + g

# Demo

# How Has This Been Tested?
Ensured that all tests passed and checked for proper behavior using `just lint`.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -26730,7 +26730,7 @@ giro/~NSV
 girt/NgSVdGJ
 girth/NwgSV
 girths/N
-GIS/ON # Geographic Information System
+GIS/N # Geographic Information System
 gist/NgSV
 git/~NOSV
 gite/NS


harper-core/dictionary.dict
```diff
@@ -10374,9 +10374,9 @@ UPC/N
 UPI/ONg
 UPS/NOgV
 URL/NSg
-US/Og
-USA/Og
-USAF/O
+US/ONg
```
hippietrail: I think `/N` might only be documented as "noun" but in practice we use it as "common noun" and, unless I'm missing secondary senses, US, USA, and USAF are all proper nouns only. If not that would probably warrants comments being added.

harper-core/dictionary.dict
```diff
@@ -26496,7 +26496,7 @@ generalissimo/NgS
 generalist/NgS
 generality/~NSg
 generalization/~NgS
-generalize/~VGdS
+generalize/~VGdSB
```
hippietrail, line 26499: Nice!

harper-core/dictionary.dict
```diff
@@ -26730,6 +26730,7 @@ giro/~NSV
 girt/NgSVdGJ
 girth/NwgSV
 girths/N
+GIS/N # Geographic Information System
```
hippietrail: This one is definitely a proper noun so should have `/O` rather than `/N`.

harper-core/dictionary.dict
anyllmarkevich: Thank you so much for the clarification! I was confused about the redundancy of O/N and asked a question in the Discord about it, but that clears it up!

harper-core/dictionary.dict
anyllmarkevich: You're absolutely right. I missed that.

harper-core/dictionary.dict
hippietrail: > Thank you so much for the clarification! I was confused about the redundancy of O/N and asked a question in the Discord about it, but that clears it up!

Yeah the structure that it generates is a `proper` flag inside a `noun` struct, but that should be regarded as an implementation detail probably. I've definitely changed things pretty similar to that in the past.

Sorry I didn't see the question in the Discord!

harper-core/dictionary.dict
anyllmarkevich: I did some research online about the exact definition of GIS, and I think the common noun usage might be more common, but I’d love your input. Here are a couple of examples.

Here is [Wikipedia’s definition](https://en.wikipedia.org/wiki/Geographic_information_system):  
> A geographic information system (GIS) consists of integrated computer hardware and software that store, manage, analyze, edit, output, and visualize geographic data. Much of this often happens within a spatial database; however, this is not essential to meet the definition of a GIS. In a broader sense, one may consider such a system also to include human users and support staff, procedures and workflows, the body of knowledge of relevant concepts and methods, and institutional organizations.

Here is the [USGS definition](https://www.usgs.gov/faqs/what-a-geographic-information-system-gis):
> A Geographic Information System (GIS) is a computer system that analyzes and displays geographically referenced information. It uses data that is attached to a unique location.

However, I have heard people use GIS to refer to all the infrastructure that humans have built that performs GIS functions (which I think is a proper noun usage). I think this definition is technically incorrect (for instance, GNSS might be a more accurate term), but it might be worth including, given that it is somewhat common. What do you think? Maybe we should have both `/O` and `/N` to cover both usages?

harper-core/dictionary.dict
anyllmarkevich: Yeah, that makes sense!
And no worries! It's probably best to discuss this right in the pull request.

harper-core/dictionary.dict
```diff
@@ -26730,7 +26730,7 @@ giro/~NSV
 girt/NgSVdGJ
 girth/NwgSV
 girths/N
-GIS/N # Geographic Information System
+GIS/ON # Geographic Information System
```
hippietrail: D'oh! Sorry about this but I was wrong. I was going to say it would just be a proper noun and not also a common noun. But that's not true. I wrongly assumed it was an official designation/product/etc but upon looking into it I found that it's in fact just a regular abbreviation/acronym/initialization and just a common noun.

Or to be concise, it should be just:
```
GIS/N
```

anyllmarkevich: Don't be sorry! Talking it through clarified things for me! Besides, I've heard it used as a proper noun on multiple occasions, even though that's apparently not entirely correct. I appreciate your help! I'll try to make my next PR less eventful! XD

harper-core/dictionary.dict
hippietrail: I also did research and came to the same conclusion. I posted here that I made a wrong assumption and it should only be a common noun `/N` but I've been having connectivity problems today so it might've got lost?

anyllmarkevich: Thanks for confirming! I got your previous message, and we're on the same page! :)

Issue #2353, opened 2025-12-17
Correct adjective before possessive noun instead of after


hippietrail:
**The Error**
I'm not sure how common this is but I just came across an example.

**Examples (2-3):**
1. experimenting with newest Neovim's built-in features and related plugins

PR #2740, opened 2026-02-17
fix: `is_possessive_nominal()` had wrong logic

hippietrail:
# Issues 
N/A

# Description

While looking into #2353 I realized that `is_possessive_nominal()` (and its derivative functions) had the wrong logic due to confusing standard terminology.

Basically "nominal" is an umbrella term for nouns and pronouns.
But a "possessive noun" is one ending in `'s` that can be used as a qualifier "**Fred's** cat" or standalone "the cat is **Fred's**".
While a possessive pronoun is only the latter "the cat is **mine**" while the former is called a possessive determiner "**my** cat".

So this PR changes `is_possessive_nominal` from a combo of `is_possessive_noun() and is_possessive_pronoun()` to a combo of `self.is_possessive_noun` and `is_possessive_determiner`

I considered changing the name but after some research I decided this name is as good as any. But I did add an explanatory comment.

I also removed a negative equivalent, `is_non_possessive_nominal()` since it's not actually used yet I couldn't be sure of coming up with the right unit tests so it can be implemented again once there's a use case for it.

# How Has This Been Tested?

I added unit tests that ensure "possessive nominal" covers the correct subset of nouns and pronouns and does not cover the wrong kinds of pronouns.
All tests in `cargo test` are unaffected by the change.

# Checklist

### DIFF
--- a/harper-core/src/dict_word_metadata.rs
+++ b/harper-core/src/dict_word_metadata.rs
@@ -1821,6 +1821,40 @@ pub mod tests {
         }
     }
 
+    mod nominal {
+        use crate::dict_word_metadata::tests::md;
+
+        #[test]
+        fn my_is_possessive_nominal() {
+            assert!(md("my").is_possessive_nominal());
+        }
+
+        #[test]
+        fn mine_is_not_possessive_nominal() {
+            assert!(!md("mine").is_possessive_nominal());
+        }
+
+        #[test]
+        fn freds_is_possessive_nominal() {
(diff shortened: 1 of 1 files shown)


hippietrail:
> I see where this confusion arose. Thanks for clearing it up!

Yes even though this is something I'm always preaching about it turns out I'm still susceptible to it! The terminology inherently seems to lead to it even when you're prepared.

PR #2748, opened 2026-02-18
fix: adaptive `message` to address issue #2743

### LINKED ISSUE #2743: "obsess about it" shows the text was "obsess with it"
Image

Is this a larger issue, namely that rules don't echo the violating text verbatim?

hippietrail:
# Issues 

Fixes #2743

# Description

The `message` field was hardcoded but now mentions the actual preposition that it is replacing.

# Demo

# How Has This Been Tested?

I used the text from the issue to form a new unit test which tests the message rather than the result.

# Checklist


hippietrail:
> That was a crazy oversight. Thanks for addressing it!

I guess I probably found ways to expand the linter beyond what I originally envisaged but didn't think to update the message?

PR #2750, opened 2026-02-18
fix(core): ignore top-level domains

elijah-potter:
# Issues 

N/A

# Description

In the domains industry, it is common to refer to individual top-level domains as proper nouns. Since these domains are written as a period-word pair, Harper would mark them as the end (and subsequent beginning) of sentences. This resulted in erroneous corrections that couldn't be reliably ignored or disabled.

To fix the problem, I've set up a pass in Harper's parser that looks for examples of top-level domains in text and marks them as "unlintable". 

# How Has This Been Tested?

An additional integration test for use of some common top-level domains.

# Checklist

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -692,21 +692,19 @@ impl Document {
         self.tokens.remove_indices(to_remove);
     }
 
-    const COMMON_TOP_LEVEL_DOMAINS: &[&str] = &[
-        "ai", "app", "blog", "co", "com", "dev", "edu", "gov", "info", "io", "me", "mil", "net",
-        "org", "shop", "tech", "uk", "us", "xyz",
-    ];
-
     /// Condenses common top-level domains (for example: `.blog`, `.com`) down to single tokens.
     fn condense_common_top_level_domains(&mut self) {
+        const COMMON_TOP_LEVEL_DOMAINS: &[&str] = &[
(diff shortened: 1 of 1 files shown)


harper-core/src/document.rs
```diff
@@ -691,6 +692,59 @@ impl Document {
         self.tokens.remove_indices(to_remove);
     }
 
+    const COMMON_TOP_LEVEL_DOMAINS: &[&str] = &[
+        "ai", "app", "blog", "co", "com", "dev", "edu", "gov", "info", "io", "me", "mil", "net",
+        "org", "shop", "tech", "uk", "us", "xyz",
+    ];
```
86xsk: I'm not 100% sure on this, but I would consider placing this constant inside the function.

`Document::COMMON_TOP_LEVEL_DOMAINS` just feels a little out of place.

harper-core/src/document.rs
```diff
@@ -691,6 +692,59 @@ impl Document {
...
+
+            cursor += 1;
+
+            if cursor >= self.tokens.len() {
+                break;
+            }
+        }
+
+        self.tokens.remove_indices(to_remove);
+    }
```
86xsk, line 738: I'll admit that I haven't run through the entire function in my head, but it *feels* like something that could maybe be simplified with https://doc.rust-lang.org/std/primitive.slice.html#method.windows?

elijah-potter: I've simplified it, but not with that method. Thanks for the review!

PR #2755, opened 2026-02-19
chore: add "one fowl swoop" to "one foul swoop" and rename from "in one fell swoop"

hippietrail:
# Issues 
N/A

# Description

I noticed that the Weir rule to correct "one fell swoop" only expect the mistake to be "one foul swoop", but "one fowl swoop" is also pretty common.

While working on that I noticed that among the examples I was finding for potential unit tests that "with one fell" swoop was also common. So I renamed the rule to "OneFellSwoop" and removed the preposition from the rule so that it will work in more contexts.

I also noticed there were no tests for the "foul" version so I found and added a unit test for that one too.

# How Has This Been Tested?

Two new unit tests, one for each variant.

# Checklist


hippietrail:
broke it with AI

Issue #2389, opened 2025-12-28
The `Spaces` and `CommaFixes` linters show the numeral `1` as the suggested fix!


hippietrail:
In the text
> ... to see the tasks as well , here is the TUI for it
The erroneous space before the comma is flagged with the message "Don't use a space before a comma". _But_ the Chrome popup displays the numeral `1` as the fix!

<img width="490" height="309" alt="Image" src="https://github.com/user-attachments/assets/25c2487c-fe63-47b2-b5af-dd7c6221c30a" />

I'm sure this was working correctly quite recently. It's very similar to #2338 which displays `0` as a fix. Both bugs are related to `Suggestion::Remove` and show up in the Chrome plugin. My guess is a JavaScript coercion bug.

While writing up this bug I found the same problem in the `Spaces` linter:

<img width="374" height="195" alt="Image" src="https://github.com/user-attachments/assets/b24a3ecc-a965-4ddf-b54a-5ba86269b83b" />

elijah-potter:
I've confirmed that this issue is also resolved by #2757.

PR #2757, opened 2026-02-19
fix(chrome-ext): suggestions being displayed as `0`

### LINKED ISSUE #2338: `SingleBe` flags "it's what's" as a repeated forms of "be" and suggests replacing with "0" (zero)
**Describe the bug**

This bug was discovered in the text
> The base execution context is the global execution context: it's what's accessible everywhere in your code.

But I'm pretty sure all or many sequential words with `'s` will trigger it.

`'s` has a few meanings but in this case the expanded phrase would be "it is what is accessible". So while both `it's` and `what's` are derived from `is` in some sense, it doesn't mean there are two forms of "is" next to each other as contractions consist of two conjoined halves. So we in fact have to derived forms of "is" separated by "what".

That alone would be a false positive. But in addition to this the suggested fix is the bizarre `0` - which looks like a type conversion you'd see in JavaScript. I'm not sure what might cause this in Rust.

**To Reproduce**
1. Type or paste the text above, or experiment with pairs of words ending in `'s`
2. See the squigglies appear.
3. Click on the purple squiggly, the one for `what's`
4. See the `Drop the repeated verb form so only one instance of be remains.` message and the `0` suggestion.

**Expected behavior**
No error flagged.

**Screenshots**

Image

**Platform**
Chrome

**Additional context**
The `It's` is also a false positive, being suggested to change to the possessive `its` but `what's` is not something that `it` owns.

elijah-potter:
- **fix(chrome-ext): show human readable label instead of underlying enum**
- **fix(core): `SingleBe` should return correct suggestion type**

# Issues 

Fixes #2338

# Description

The actual datatype behind `SuggestionKind` is `number`. When the `SuggestionBox` UI was written, it was assumed that the underlying datatype was `string` and that it just contained the relevant enum value's name. This incorrect assumption led to integers being rendered when the suggestion kind was anything other than `Suggestion.ReplaceWith` or if the replacement string was empty. 

To fix this, I created a function to map from the original integer to a human readable string and used it when the replacement string was empty.

Additionally, we found that the `SingleBe` rule was returning an empty replacement string instead of `SuggestionKind.Remove`. This has also been resolved.

# Demo

image

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/harper-core/src/linting/single_be.rs
+++ b/harper-core/src/linting/single_be.rs
@@ -161,7 +161,7 @@ impl ExprLinter for SingleBe {
         Some(Lint {
             span: Span::new(whitespace_start, second_be_end),
             lint_kind: LintKind::Grammar,
-            suggestions: vec![Suggestion::ReplaceWith(vec![])],
+            suggestions: vec![Suggestion::Remove],
             message: "Drop the repeated verb form so only one instance of `be` remains.".to_owned(),
             priority: 31,
         })


hippietrail:
@elijah-potter : Are you going to fix the same problem in #2389

PR #2797, opened 2026-02-24
fix(core): stop flagging "Czech Republic" as outdated

### LINKED ISSUE #2692: "Czech Republic" is flagged as incorrect
**What got flagged?**
Czech Republic

**Why is this incorrect?**
Harper flags this as incorrect and says, "This place has been officially known as ‘Czechia’ since 2016,". This is technically true, but misleading at best. "Czech Republic" is the country's full formal name in English, and this has not changed, while "Czechia" is the country's official short name. Although the country adopted "Czechia" as the official short name and encouraged its use by English speakers, "Czech Republic" remains a valid way to refer to the country in English, and is not inappropriate, invalid, or recommended against in any context.

This is the same as the difference between "France" and "French Republic." Both are valid names for the country, and Harper does not flag "French Republic" as incorrect.

See for further details: https://en.wikipedia.org/wiki/Name_of_the_Czech_Republic#Adoption_of_Czechia

**Example of correct usage:**
Czech Republic

NaveenPrasanth:
## Summary

Fixes #2692

- Removed the `Czech Republic → Czechia` mapping from `UpdatePlaceNames`
- "Czech Republic" is the country's formal name; "Czechia" is the official short name and both are valid (like "French Republic" vs "France")
- Added a regression test to ensure "Czech Republic" is not flagged

## Test plan

- [x] `cargo test -p harper-core update_place_names` : all 17 tests pass
- [x] `cargo fmt` : applied
- [x] `cargo clippy -p harper-core` : no new warnings

### DIFF
--- a/harper-core/src/linting/update_place_names.rs
+++ b/harper-core/src/linting/update_place_names.rs
@@ -29,7 +29,6 @@ impl<'a> Default for UpdatePlaceNames<'a> {
             ((1945, "Taiwan"), &["Formosa"]),
             ((1991, "Ulaanbaatar"), &["Ulan Bator"]),
             // Europe (and nearby)
-            ((2016, "Czechia"), &["Czech Republic"]),
             ((1945, "Gdańsk"), &["Danzig"]), // TODO: Can we recommend Gdansk as well?
             ((1992, "Podgorica"), &["Titograd"]),
             ((1936, "Tbilisi"), &["Tiflis"]),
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
This looks good to me. Thanks for the patch! 

In the future, if you're addressing a specific issue #, reference it in the test name. For example, that could look like `dont_flag_czech_republic_issue_2692`.

It isn't a hard and fast rule, and it's definitely not going to stop me from merging this. Thanks again!

PR #2809, opened 2026-02-26
refactor: tests for spell/mod.rs and and linting/spell_check.rs were arbitrarily mixed

hippietrail:
# Issues 
N/A

# Description

I noticed the line `pub use spell_check::SpellCheck;` in `linting/mod.rs` seemed out of place since no other linters not driven by macros had anything similar.

It turned out that there is spelling infrastructure in `spell/mod.rs` and the actual spelling checker linter in `linting/spell_check.rs` but that spelling tests were arbitrarily placed in either.

I've sorted them out so that all tests using the spellcheck linter are now part of the spellcheck linter and tests directly on the underlying infrastructure are in `spell/mod.rs`

If I've missed anything or put some things in the wrong place please let me know.

# How Has This Been Tested?

All tests still pass after the refactoring.

# Checklist

### DIFF
--- a/harper-core/src/linting/mod.rs
+++ b/harper-core/src/linting/mod.rs
@@ -133,6 +133,7 @@ mod more_adjective;
 mod more_better;
 mod most_number;
 mod most_of_the_times;
+mod multiple_frequency_adverbs;
 mod multiple_sequential_pronouns;
 mod nail_on_the_head;
 mod need_to_noun;
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -23704,6 +23704,7 @@ esophagi/N9
 esophagus/~N0g
 esoteric/~JNQ
 esp/~
+esp./~              # abbreviation for "especially"
 espadrille/NgS
 espalier/NgSVdG
 especial/JY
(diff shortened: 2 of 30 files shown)


harper-core/src/linting/mod.rs
```diff
@@ -247,7 +247,6 @@ pub use lint_group::{LintGroup, LintGroupConfig};
 pub use lint_kind::LintKind;
 pub use map_phrase_linter::MapPhraseLinter;
 pub use map_phrase_set_linter::MapPhraseSetLinter;
-pub use spell_check::SpellCheck;
```
elijah-potter, line 256: Why was this removed?

hippietrail: That was the red flag that made me think something was wrong and dig deeper. No other linter has a `pub use` here except for the ones that are generated from macros: `MapPhraseLinter`, `MapPhraseSetLinter`, and `InitialismLinter`.

It was only there because some tests being in the wrong place needed it there. (Maybe "needed" is too strong a word, there might've been another way to get the tests in the wrong file working - but I couldn't find it in my brief look.)

I may have even put it there myself when I was working on the dialect + spellchecker interactions - I'm not sure. If I'm missing something I'm interested to learn!

----

Ironically even though `SpellCheck` wasn't legitimately used in any other module, the same day I did this I *did* use `SpellCheck` in #2812! I actually wondered it that was the best way to test the linting asserts at the time because it's inherently brittle. Now I think it is indeed best to make a specific linter to use in the module...

PR #2812, opened 2026-02-26
feat: womans→women, writed→wrote, etc.

### LINKED ISSUE #950: feat(core): Add an `assert_any_suggestion_result` test function

hippietrail:
# Issues 

Fixes #950

# Description

**Corrects wrong regular inflected forms of irregular verbs, plurals, and adjectives**
_and_
**Replaces `assert_suggestion_result` and `assert_top3_suggestion_result` with a single new implementation that can truly check all suggestions for the specified result**

This linter fixes using wrong regular forms of noun plurals, past verbs, or adjective comparatives or superlatives which should be irregular forms: childs→children, thinked→thought, gooder→better.

----
It also fixes orthographic mistakes in inflected forms such as: tomatos→tomatoes, wifes→wives, etc.

While working on this I realized that testing the linter when there would be more than three generated results was inconsistent and broken.

This turned out to be #950 `assert_any_suggestion_result` which I'd tackled before but always had trouble understanding the logic well enough to do a thorough job.

One problem was the three nested loops which I didn't quite grok. One to count the number of iterations, up to 100. One to iterate over the lints, and one to iterate over the suggestions. Especially when it seemed only the first iteration of the lint was ever used.

But secondly, even though I implemented `assert_any_suggestion_result` by refactoring the above loop logic without fully understanding it, even that didn't work due to the use of hashmaps/hashsets internally, which do not guarantee order, and hence when more than three, not the same 3 will always be first.

After thinking about this problem many times I slowly realized the 3 loops were:
1. a limit for pathological linters or linter pairs or groups where linter A's correction gets changed back to its previous test by linter B, or where linter A's correction still results in something linter A will flag, etc.
2. When a suggestion is chosen to fix the first lint, there will no longer be a lint triggered in that position, so the second lint will now be the one that comes up as the "first" lint.
3. Iterating over the individual suggestions for the current "first" lint.

I then asked a coding agent to verify if my thinking was correct. It said it was but of course I didn't trust it 100% so started testing it in many ways and getting the agent to refactor the looping logic several times.

The result is therefore a mix of AI generated code and hand-coded logic. Which means **please check carefully** though I also think that if there are still some edge cases in there they're more likely my fault than the AI's.

In the end I decided to name the new function `assert_suggestion_result` since it should work with any single linter or combination of linters. It therefor replaces `assert_top3_suggestion_result` as well.

Please don't hold back on critiquing this! I'm doing a regular PR even though I expect to make some corrections or adjustments since I find that draft PRs rarely garner any feedback.

(I also adjusted/added a few dictionary entries and irregular noun/verb entries.)

----

Since I was refactoring I took the opportunity to convert the assertions from creating Markdown documents to creating plain English documents. This will surely have less overhead and also makes the testing code more concise and so easier to follow. This unexpectedly broke 4 tests related to currency symbol placement and title case, which turned out to depend on testing in Markdown format. So I reintroduced Markdown versions of just the functions needed for those tests, and sharing the logic that ensures all suggestions are generated.

I also noticed that the `Document` creation functions the tests were using were converting back and forth between `char` arrays/vectors and `String`/`&str/ and since documents are created in a loop to search for all suggestions and since some core functionality works on `char` I added constructors using `char`s directly, which should also simplify some source code and speed up generated code.

# How Has This Been Tested?

All old tests pass. New tests that would not have passed before also pass.

I tried to find `ignore`d tests due to this assertion missing but failed to find any. If you find some please comment on the issue.

I had trouble thinking up realistic tests for some cases that wouldn't seem to abstract. I initially used the `SpellCheck` linter, but that resulted in second-guessing its behaviour. I later implemented a special test linter that lets you specify each error and correction you want it to handle as an array of many-to-many mappings.

Several tests are still not implemented but appear just as TODOs because I thought putting something abstract in there would just be confusing. Ideally they can be fleshed out when other linters run into concrete problems.

Note that the `assert_good_and_bad_suggestions()` has not yet been updated since I didn't have a concrete example and the code to support it will inherently be harder to read and review. It will therefore probably still have problems testing linters with more than three generated results.

# Checklist

### DIFF
--- a/harper-core/src/linting/mod.rs
+++ b/harper-core/src/linting/mod.rs
@@ -180,6 +180,7 @@ mod quantifier_needs_of;
 mod quantifier_numeral_conflict;
 mod quite_quiet;
 mod quote_spacing;
+mod reason_for_doing;
 mod redundant_acronyms;
 mod redundant_additive_adverbs;
 mod redundant_progressive_comparative;
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -1,4 +1,4 @@
-54400
+54500
 
 # Start of original dictionary import
 # combined with dialect spelling dictionary import.
@@ -3875,6 +3875,7 @@ Gawain/Og
 Gay/Og
 Gayle/Og
 Gaza/Og
+Gazan/JOgS
 Gaziantep/Og
 Gd/                 # Gadolinium
 Gdańsk/Og
@@ -4100,6 +4101,8 @@ Grail/g
 Grammy/Ng
 Grampians/Og
 Granada/Og
+Grand Prix/O0g
+Grands Prix/O9
 Grant/ONg
 Grass/Og
 Graves/Og
@@ -5221,6 +5224,7 @@ Jonathan/ONg
 Jonathon/Og
 Jones/Og
 Jonesboro/Og
+Jong-Un/Og
 Joni/Og
 Jonson/Og
 Joplin/Og
@@ -5345,9 +5349,10 @@ Kandinsky/g
 Kane/Og
 Kaneohe/Og
 Kankakee/Og
-Kannada/Ng          # language
+Kannada/Omg         # language
 Kano/Og
 Kanpur/Og
+Kansai/Og
 Kansan/JNgS
 Kansas/Og
 Kant/Og             # immanuel -
@@ -5715,6 +5720,7 @@ Lachesis/ONg
 Lactobacillus/O
 Lacy/Og
 Lada/NOSg
+Ladakh/Og
 Ladoga/Og
 Ladonna/g
 Lady/NOg
@@ -6731,6 +6737,7 @@ Mel/Og
 Melaka/Og
 Melanesia/Og
 Melanesian/JNg
+Melania/Og
 Melanie/Og
 Melba/Og
 Melbourne/Og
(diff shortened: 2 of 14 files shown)


harper-core/src/linting/mod.rs
```diff
@@ -619,13 +621,51 @@ pub mod tests {
...
+            && lint.message != expected_message
+        {
+            panic!(
+                "Expected lint message \"{expected_message}\", but got \"{}\"",
+                lint.message
+            );
+        }
+    }
+
+    fn transform_nth_str(text: &str, linter: &mut impl Linter, n: usize) -> String {
```
hippietrail: This function shouldn't be needed any longer, and in fact wasn't deterministic due to how the hashmaps/hashsets used internally work, which means it didn't really work.

Not sure if this is only showing up here as some kind of merge artefact?

----

I now realize that using the spell checker module here was not the best idea and actually clashes with my other PR that sorts out mixed up spelling tests between two modules at #2809 so I'll give this one a dedicated test linter...

harper-core/src/linting/mod.rs
```diff
@@ -199,7 +200,7 @@ mod somewhat_something;
 mod soon_to_be;
 mod sought_after;
 mod spaces;
-mod spell_check;
+pub mod spell_check; // pub so spell/mod.rs can use it until #2809 is merged
```
elijah-potter: This can be removed now, right?

hippietrail: > This can be removed now, right?

Yes the other PR was merged so its job here is done!

PR #2816, opened 2026-02-26
fix(core): recognize "esp." as an abbreviation

### LINKED ISSUE #2753: lowercase word after "esp." should not be flagged
**What got flagged?**

Image

**Why is this incorrect?**
"esp." is an abbreviation

**Example of correct usage:**

> I'll go, esp. if it's a free event.

NaveenPrasanth:
## Issues

Fixes #2753

## Description

The period after "esp." was treated as a sentence terminator, causing a false positive where the next word was flagged for not being capitalized (e.g. "I'll go, esp. if it's free").

Three changes:
- Add "esp" to the dotted truncation condensing list in `document.rs` (alongside "etc" and "vs") so the word and its trailing period are treated as a single token
- Rename `condense_latin` → `condense_dotted_truncations` (and related symbols) since the list now includes non-Latin truncations
- Add "esp." to `dictionary.dict` so it's not flagged as a spelling error

## How Has This Been Tested?

- Added `doesnt_flag_after_esp_issue_2753` unit test to `sentence_capitalization` tests
- All 13 sentence_capitalization tests pass
- Verified with `harper-cli lint` that the original sentence produces no lints

## Checklist


harper-core/src/document.rs
```diff
@@ -506,7 +506,7 @@ impl Document {
 
     fn uncached_latin_expr() -> Lrc<FirstMatchOf> {
         Lrc::new(FirstMatchOf::new(vec![
-            Box::new(SequenceExpr::word_set(&["etc", "vs"]).then_period()),
+            Box::new(SequenceExpr::word_set(&["esp", "etc", "vs"]).then_period()),
```
hippietrail, line 509: I would either rename `uncached_latin_expr` to something else, or make a twin function with a different name since "especially" is not Latin and it would make the function a misnomer.

Using the Google AI but not digging any deeper it suggested the term "truncation", which seems reasonable to me. So maybe `uncached_truncation` or `uncached_dotted_truncation` or `uncached_truncation_with_period`? Those would all apply to both Latin and non-Latin abbreviations that require a period.

If you opt for twin functions, one for Latin and one for non-Latin, truncation might not be clear enough on its own.

Check if there are other symbols using "latin" that would need to follow the name change too.

hippietrail (changes requested):

NaveenPrasanth:
Updated the naming to `dotted_truncation` as suggested — `DOTTED_TRUNCATION_EXPR`, `uncached_dotted_truncation_expr`, and `condense_dotted_truncations`.

harper-core/src/document.rs
```diff
@@ -145,7 +145,7 @@ impl Document {
         self.condense_dotted_initialisms();
         self.condense_number_suffixes();
         self.condense_ellipsis();
-        self.condense_latin();
+        self.condense_abbreviations();
```
hippietrail, line 148: This is just a comment since it's probably good either way but I avoided the term "abbreviation" in my review since I think abbreviations are a superset including also other kinds such as ones with no dot at the end (vet, lab, NASA, etc.), and ones with multiple dots (U.S.A., etc.)

NaveenPrasanth:
> All systems go!

Is there any action needed from my side @hippietrail my man?

hippietrail:
> > All systems go!
> 
> Is there any action needed from my side @hippietrail my man?

I'll request a review from @elijah-potter

NaveenPrasanth:
> > > All systems go!
> > 
> > 
> > Is there any action needed from my side @hippietrail my man?
> 
> I'll request a review from @elijah-potter

Thank you!

elijah-potter (approved):
This looks perfect to me. I wouldn't do it any other way. Thanks!

PR #2819, opened 2026-02-26
fix(obsidian): find temporary fix for escape button override

### LINKED ISSUE #2770: Obsidian bug, unable to escape insert mode to normal mode
**Describe the bug**

Vim mode in obsidian is unable to return from insert mode to normal mode when Harper plugin is installed

**To Reproduce**

Open obsidian, install harper plugin & enable vim mode. Press 'i' for insert mode, and then escape key should go to normal mode (but it doesn't)

**Expected behavior**

When I press escape, it goes from insert mode to normal mode

**Platform** → Obsidian

elijah-potter:
# Issues 

Fixes #2770

# Description

@stanley-910, this is a suitable temporary fix for the underling issue? We can revisit this later to improve how people can close the popup, but I want a quick fix now that I can release to those affected.

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/packages/obsidian-plugin/src/index.ts
+++ b/packages/obsidian-plugin/src/index.ts
@@ -208,7 +208,6 @@ export default class HarperPlugin extends Plugin {
 		this.addCommand({
 			id: 'harper-dismiss-focused-tooltip',
 			name: 'Dismiss focused suggestion tooltip',
-			hotkeys: [{ modifiers: [], key: 'Escape' }],
 			checkCallback: (checking) => {
 				const editorView = this.getActiveEditorView();
 				if (!editorView) return false;


elijah-potter:
I feel confident this will fix the issue. We can always release a hotfix if something comes up.

stanley-910:
Yep, that'll do it. As long as its not bound, it won't cause any issues. Doing some research, I found other plugins like LanguageTool don't dismiss the popup on escape either, so maybe its not even necessary to find a workaround so that we don't consume escape (or some other modifier) to dismiss it, after all, it will just dismiss itself when you start typing, or click else where. 

One thing that could be tuned up is that it should dismiss automatically when the vim cursor navigates to another word (currently it stays focused with no way to dismiss w/o a click or to start typing)

PR #2836, opened 2026-02-28
fix(harper-cli): return exit code 0 when no lints are found

### LINKED ISSUE #2832: `harper-cli lint` always exits with code 1, even when no lints are found
**Describe the bug**

`harper-cli lint` always returns a non-zero exit code, even for clean files with no violations.

The `lint()` function in `lint.rs` unconditionally calls `process::exit(1)` at its end, regardless of whether any lints were found. This appears to be a regression introduced in v0.73.0 by commit 7080884 (PR #2138).

This commit refactored the `lint` subcommand from `main.rs` into a new `lint.rs` file to support multiple inputs and directories. During that refactoring, the conditional exit logic appears to have been unintentionally removed — the original code had early `return Ok(())` (exit 0) paths for "no lints" and "count mode", with `process::exit(1)` only reachable when lints were found. In the new code, `process::exit(1)` is placed unconditionally at the end of the `lint()` function. This remains unfixed as of v1.9.0.

**To Reproduce**
1. Create a file with no grammar/spelling issues, e.g. `echo "That's" > /tmp/foo.txt`
2. Run `harper-cli lint /tmp/foo.txt`
3. Check the exit code: `echo $?`
4. Exit code is `1` even though "No lints found" is printed

```console
$ harper-cli core-version
harper-core v1.9.0
$ echo "That's" > /tmp/foo.txt
$ harper-cli lint /tmp/foo.txt
Note: Using user dictionary at /home/dell-08/.config/harper-ls/dictionary.txt
foo.txt: No lints found
$ echo $?
1 # expected: 0
```

**Expected behavior**
- Exit code `0` when no lints are found
- Exit code `1` (non-zero) only when lints are found

This is important for CI/CD usage where the exit code is used to determine pass/fail.

**Screenshots**
N/A

**Platform**
harper-cli (all platforms)

**Additional context**
- Prior behavior (v0.72.0)
  In `main.rs` before the refactoring, the logic was:
  ```rust
  if lints.is_empty() {
      println!("No lints found");
      return Ok(()); // exit 0
  }
  // ... print report ...
  process::exit(1) // exit 1 only when lints exist
  ```
- Current behavior (v1.9.0)
  In `lint.rs`, the function unconditionally ends with:
  ```rust
  process::exit(1); // always exit 1
  ```
  https://github.com/Automattic/harper/blob/v1.9.0/harper-cli/src/lint.rs#L281

I do not plan to create a pull request. Since I have no experience with Rust, I cannot judge whether the code is good even with AI assistance.

Ryujiyasu:
## Summary

- Fix `harper-cli lint` always exiting with code 1, even when no lints are found
- The `lint()` function in `lint.rs` unconditionally called `process::exit(1)` — a regression introduced in PR #2138
- Now exits with code 1 only when lints are present, and returns `Ok(())` (exit code 0) otherwise

Fixes #2832

## Test plan

- [x] `cargo test -p harper-cli` — all 3 tests pass
- [x] Manual verification:
  - `echo "This is correct English." | harper-cli lint` → exit code **0**
  - `echo "This is an test." | harper-cli lint` → exit code **1**


harper-cli/src/lint.rs
```diff
@@ -278,7 +280,11 @@ pub fn lint(
         lint_options.color,
     );
 
-    process::exit(1);
+    if has_lints {
```
hippietrail, line 364: Looks like this could just be
```
    if !all_lint_kinds.is_empty() {
```
The lines are so close together that I don't think it's less clear this way. Anyone?

Ryujiyasu:
Thanks for the review! Unfortunately `all_lint_kinds` is moved into `final_report()` (it takes `HashMap<LintKind, usize>` by value), so it's no longer available after the call. That's why I capture the check in `has_lints` beforehand.

An alternative would be to change `final_report` to take `&HashMap<LintKind, usize>` (by reference), which would let us use `if !all_lint_kinds.is_empty() {` directly. However, that would require additional changes inside `final_report` (e.g. replacing `.into_iter()` with `.iter()`). Would you prefer that approach?

harper-cli/src/lint.rs
Ryujiyasu: Thanks for the review! Unfortunately `all_lint_kinds` is moved into `final_report()` (it takes `HashMap<LintKind, usize>` by value), so it's no longer available after the call. That's why I capture the check in `has_lints` beforehand.

An alternative would be to change `final_report` to take `&HashMap<LintKind, usize>` (by reference), which would let us use `if !all_lint_kinds.is_empty() {` directly. However, that would require additional changes inside `final_report` (e.g. replacing `.into_iter()` with `.iter()`). Would you prefer that approach?

86xsk: I think this is good. The function has reason to take the hashmap by value, so taking it by reference would unnecessarily complicate things.

I think it might be better to return an `Err()` rather than directly calling `process::exit()` though. I don't think it's ideal that a method like `lint` would exit the process, and if I'm reading correctly, I believe that simply returning an `Ok()`/`Err()` back to `main` should achieve the desired result.

hippietrail: > Thanks for the review! Unfortunately `all_lint_kinds` is moved into `final_report()` (it takes `HashMap<LintKind, usize>` by value), so it's no longer available after the call. That's why I capture the check in `has_lints` beforehand.

Ah OK I only actually looked at the diff-snippet here and not at the full code. Apparently I hit the edge case of using that approach (-:

hippietrail: > I think it might be better to return an `Err()` rather than directly calling `process::exit()` though. I don't think it's ideal that a method like `lint` would exit the process, and if I'm reading correctly, I believe that simply returning an `Ok()`/`Err()` back to `main` should achieve the desired result.

I kind of wondered about that too as it looked inconsistent to me, but it was outside the areas of Rust I feel comfortable in.

hippietrail:
When you're done with this one, you might want to have a look at the other `harper-cli` commands and see if any others don't return appropriate success values or use `process::exit()` needlessly. I think many of the commands have only every had one or two pairs of eyes on them.

Ryujiyasu:
Thanks for the suggestion @hippietrail! I checked the other `harper-cli` commands and found two more places using `process::exit(1)` needlessly:

- `audit-dictionary`: calls `std::process::exit(1)` when issues are found
- `test`: calls `process::exit(1)` when Weir tests fail

Both can use `anyhow::bail!()` instead, since `main()` already returns `anyhow::Result<()>`.

Opened #2840 and PR #2841 for the fix.

harper-cli/src/lint.rs
Ryujiyasu: Great point\! Updated to use `anyhow::bail\!()` instead of `process::exit(1)`, so the error propagates through `main()` naturally. Thanks for the suggestion.

elijah-potter:
It looks like you need to run `just format` really quick. Otherwise, it sounds like we can merge this!

PR #2841, opened 2026-03-01
fix(harper-cli): replace process::exit(1) with anyhow::bail!() in AuditDictionary and Test

### LINKED ISSUE #2840: harper-cli: AuditDictionary and Test commands use process::exit() instead of returning errors
## Current behavior

Two `harper-cli` subcommands use `process::exit(1)` to signal failure instead of returning `Err(...)` through the normal `anyhow::Result<()>` return type of `main()`:

1. **`audit-dictionary`** ([`main.rs:815`](https://github.com/Automattic/harper/blob/master/harper-cli/src/main.rs#L815)): When dictionary audit finds issues (duplicate flags, unknown flags, unused flags), it calls `std::process::exit(1)` instead of returning an error.

2. **`test`** ([`main.rs:969`](https://github.com/Automattic/harper/blob/master/harper-cli/src/main.rs#L969)): When Weir tests fail, it calls `process::exit(1)` instead of returning an error.

## Problem

`process::exit()` terminates the process immediately, skipping destructors, buffered I/O flushes, and any cleanup logic. Since `main()` already returns `anyhow::Result<()>`, these commands should use `anyhow::bail!()` to propagate errors through the normal error handling path.

This was identified as a follow-up to #2832 / PR #2836, as noted [in this comment](https://github.com/Automattic/harper/pull/2836#issuecomment-3979049893).

## Proposed fix

- Replace `std::process::exit(1)` with `anyhow::bail!(...)` in both commands
- Remove the now-unused `process` import from `main.rs`

Ryujiyasu:
## Summary

- Replace `process::exit(1)` with `anyhow::bail!()` in `audit-dictionary` and `test` subcommands
- Remove the now-unused `process` import from `main.rs`

`process::exit()` terminates immediately, skipping destructors and buffered I/O flushes. Since `main()` returns `anyhow::Result<()>`, errors should be propagated through the normal error handling path.

Follow-up to #2836, as suggested in [this comment](https://github.com/Automattic/harper/pull/2836#issuecomment-3979049893).

Fixes #2840

## Test plan

- [x] `cargo build -p harper-cli` — compiles successfully
- [x] `cargo test -p harper-cli` — all 3 tests pass
- [x] Both commands still exit with code 1 on failure (now via anyhow instead of process::exit)

### DIFF
--- a/harper-cli/src/main.rs
+++ b/harper-cli/src/main.rs
@@ -7,7 +7,7 @@ use std::fs::File;
 use std::io::BufReader;
 use std::path::PathBuf;
 // use std::sync::Arc;
-use std::{fs, process};
+use std::fs;
 
 use anyhow::anyhow;
 use ariadne::{Color, Label, Report, ReportKind, Source};
@@ -812,7 +812,7 @@ fn main() -> anyhow::Result<()> {
                 if unused_flag_total > 0 {
                     eprintln!("  - {} unused flags found", unused_flag_total);
                 }
-                std::process::exit(1);
+                anyhow::bail!("Audit found issues in dictionary");
             }
 
             Ok(())
@@ -966,7 +966,7 @@ fn main() -> anyhow::Result<()> {
                 Ok(())
             } else {
                 eprintln!("{:?}", failing_tests);
-                process::exit(1);
+                anyhow::bail!("Some tests failed");
             }
         }
     }


Ryujiyasu:
Closing — jumped ahead too quickly. Will wait for maintainer feedback before submitting a PR.

PR #2843, opened 2026-03-01
feat(harper-cli): add --format flag with json and compact output modes

dobbry-vechur:
# Issues

No specific issue -- adds machine-readable output formats to `harper-cli lint`.

# Description

Adds a `--format` flag to `harper-cli lint` with three modes:

- **`default`** -- existing Ariadne rich output (unchanged)
- **`json`** -- structured JSON array to stdout for machine consumption
- **`compact`** -- one-line-per-lint GCC/grep-style (`file:line:col: Kind::Rule: message`)

Other changes:
- Informational messages (user dict notes) moved to stderr so stdout stays clean for structured formats
- JSON includes an optional `error` field for files that fail to load
- Machine-readable formats use plain paths (no ANSI), even when color is enabled

# Demo

**JSON:**
```
$ harper-cli lint --format json "This is an test."
[
  {
    "file": "<text>",
    "lint_count": 1,
    "lints": [
      {
        "rule": "AnA",
        "kind": "Miscellaneous",
        "span": { "start": 8, "end": 10 },
        "line": 1,
        "column": 9,
        "message": "Incorrect indefinite article.",
        "priority": 31,
        "suggestions": ["Replace with: ..."],
        "context": "an"
      }
    ]
  }
]
```

**Compact:**
```
$ harper-cli lint --format compact "This is an test."
<text>:1:9: Miscellaneous::AnA: Incorrect indefinite article.
```

# How Has This Been Tested?

- 6 new integration tests in `harper-cli/tests/output_format.rs` covering JSON validity, JSON schema fields, no-ANSI guarantees (including directory inputs with color enabled), compact format structure, and default format backwards compatibility
- All 9 tests pass (`cargo test -p harper-cli`)
- `cargo fmt --check` and `cargo clippy -Dwarnings` clean

# Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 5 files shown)


hippietrail:
At a quick/shallow look this seems really nice!

Do you think the code would get bigger or smaller / more complex or simpler, if the JSON became the default and the other output formats were converted from the JSON?

dobbry-vechur:
It would definitely get bigger and more complex. Currently:

- compact mode writes directly to stdout in single_input_report - zero allocation, just println! per lint
- Ariadne mode works directly with `Lint` structs and the `Document`, it needs the full `Doc` for source spans, token info, etc. that `JsonFileResult` doesn't carry
- JSON mode builds `JsonFileResult` structs, then serializes at the end

If JSON were the internal pivot format, we'd need to either:
- round-trip through serialization/deserialization (wasteful)
- make `JsonFileResult` carry enough data to reconstruct Ariadne reports (duplicating the Document/source info)

Deriving each output format directly from the linting results is the right approach. The three formats need different data and have different output patterns. Forcing them through a common intermediate would only add complexity for no real benefits

elijah-potter:
This looks great! Thank you!

PR #2858, opened 2026-03-02
feat(core): `chalkboard` is a closed compound

elijah-potter:
# Issues 

N/A

# Description

While writing a blog post, I noticed that Harper did not respect that chalkboard is pretty much always a closed compound and therefore should be condensed from "chalk board".

I've added it to the list of closed compounds.

# How Has This Been Tested?

Addtional unit test.

# Checklist


hippietrail:
Interesting. I can confirm that every dictionary lists only "chalkboard" and none even include "chalk board" as an also. Since "chalkboard" is in `dictionary.dict` I wonder why it wasn't already suggesting to merge the words?

elijah-potter:
I'm not sure. I decided to live on the side of "doing something". Maybe I'll look into this more deeply soon.

PR #2859, opened 2026-03-02
feat(core): flag redundant use of progressive verbs followed by "more"

elijah-potter:
# Issues 

N/A

# Description

While writing a blog post, I noticed that Harper doesn't seem to pay attention to redundant use of progressive verbs followed by "more".
That is, phrases like "I've been increasingly more attentive".

# Demo

# How Has This Been Tested?

Additional unit tests.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -29910,7 +29910,7 @@ incorrigible/JN
 incorrigibly/Ry
 incorruptibly/Ry
 increasing/~JYV6N
-increasingly/R%
+increasingly/R
 increment/~NSgVdG
 incremental/~JY
 incrementalism/N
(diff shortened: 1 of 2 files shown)


hippietrail:
> # Description
> While writing a blog post, I noticed that Harper doesn't seem to pay attention to redundant use of progressive verbs followed by "more". That is, phrases like "I've been increasingly more attentive".

Note that `increasingly` is no longer a verb. Adding `-ing` to `increase` is an "inflectional" change but adding `-ly` to `increasing` is a "derivational" change. This changes the part of speech to be an adverb and only an adverb, unlike `increasing` which was a verb, adjective, and noun at the same time.

I asked an AI what kind of adverb `increasingly` is as I wasn't quite sure. It told me it's an "adverb of degree". I recently added the `/%` annotation flag to mark adverbs of degree and `.is_degree_adverb()` though those are probably not needed for your linter if `increasingly` is the only such adverb that results in such redundancy, as seems likely.

PR #2860, opened 2026-03-02
feat: add a rule for past tense instead of past participle

jvoisin:


hippietrail:
I was pretty sure I wrote a Rust rule for this in the past few months. In particular I kept reading and hearing "have wrote" and/or "have went" all over the place so I decided to do something about it (-: Indeed those are both flagged right here: 
image
image

Are you hitting cases that the `SimplePastToPastParticiple` rule fails to flag? Or are you replacing that rule with a Weir rule? (None of my Weir rules are this complex yet, I'm better at writing rules in Rust except for the simplest ones in Weir.) Or maybe you had `SimplePastToPastParticiple` disabled or you're using an older version of Harper from before I contributed that rule?

jvoisin:
I'm sorry, this should have been a draft instead of a ready pull-request, and issued against my fork (to trigger the CI) and not upstream. I was simply playing with weir, apologies for the noise!

PR #2872, opened 2026-03-05
fix: don't flag "our IT"

### LINKED ISSUE #2870: False positive: "our IT" flagged by `MultipleSequentialPronouns`
**What got flagged?**

Image

> their sales derp was just having none of it when **our IT** director told him, point blank, that we're not moving anything into the cloud

All-caps "IT" is an initialism for "Information technology" and not the 3rd person singular neuter pronoun "it".

But if the sentence was in all caps it would once again be ambiguous.

> THEIR SALES DERP WAS JUST HAVING NONE OF IT WHEN **OUR IT** DIRECTOR TOLD HIM, POINT BLANK, THAT WE'RE NOT MOVING ANYTHING INTO THE CLOUD

But "our it" is a pretty unlikely mistake anyway so we probably don't need to dig so deep for ambiguous cases. Likewise, checking if "IT" is part of a compound like "IT director" is probably more work than needed for such a rare case.

hippietrail:
# Issues 

Fixes #2870

# Description

- Don't flag "our IT director" etc.
- Refactored to use newer methods.
- As a bonus it stops flagging 12 non-errors in Alice and Gatsby snapshots.

I also noticed that `.is_possessive_pronoun()` is kind of a "null" implementation as "mine", "ours", "yours", "his", "hers", "its", and "theirs" are just marked as pronouns. This may need to be addressed at some point.

# How Has This Been Tested?

A new unit test using the sentence that inspired the false positive report.

# Checklist

### DIFF
--- a/harper-core/src/linting/multiple_sequential_pronouns.rs
+++ b/harper-core/src/linting/multiple_sequential_pronouns.rs
@@ -1,73 +1,30 @@
-use super::Suggestion;
-use super::expr_linter::ExprLinter;
-use crate::expr::Expr;
-use crate::expr::SequenceExpr;
-use crate::linting::LintKind;
-use crate::linting::expr_linter::Chunk;
-use crate::patterns::WordSet;
-use crate::{CharStringExt, Lint, Lrc, Token, TokenStringExt};
+use crate::{
+    expr::{Expr, SequenceExpr},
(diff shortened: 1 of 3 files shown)


elijah-potter (approved):
LGTM. Thanks!

PR #2873, opened 2026-03-05
feat(chrome-ext): surface update instructions

### LINKED ISSUE #2862: Is it possible to make it easier to update the Chrome extension
We can tell that the Chrome extension is behind from this little version check: 

Image

But it's not obvious how to go about updating it from there. Clicking the warning icon or the version number does nothing. Same for the Harper logo and title in the upper left. Clicking `Settings` in the lower link bar takes us to "Harper core" settings rather than "Harper Chrome" settings, so there's no clue from there what to do to update.

Clicking on the Chrome `extensions` "puzzle piece" icon shows this popup menu:

Image

From there, clicking on the Harper "Private Grammar Checker" entry takes us back to the original popup with the version warning. Clicking on the vertical ellipses `⋮` on the right bring up a further popup menu, none of whose entries suggest they will definitely lead to a way to update: 

Image

- The first entry takes us to Harper's page on the chrome web store.
- The second entry takes us to the "Harper core" settings mentioned above.
- The "Manage extension" entry takes us to the ("Harper extension special page"](chrome://extensions/?id=lodbfhdipoipcjmlebjbgmmgekckhpfb)

Image

Here, finally, is the "Update" button, third after "Load unpacked" and "Pack extension" buttons. Easy to miss in my opinion.

Another way to get here is from the "manage extension" entry at the bottom of the "puzzle piece" popup menu mentioned above which takes us to the ["Chrome extensions special page"](chrome://extensions/) in which we can search for Harper using the search bar at the top: 

Image

Or manually scroll down to the Harper entry, which looks like this: 

Image

Again showing a version number, which turns out to be the installed version with no additional indication of whether it's the latest or if there's a newer version available. It has "Details" and "Remove" buttons and an on-off switch. The "Details" button takes us to the same ("Harper extension special page"](chrome://extensions/?id=lodbfhdipoipcjmlebjbgmmgekckhpfb) mentioned above, where Harper can finally be updated if your eyes have not glazed over by this point to the degree that it could miss the "Update" button.

----

- Is there a more straightforward way that I have not found?
- Is there a way we can integrate a more straightforward update method into the Harper Chrome Extension?
- Is there a more concise way that my screed above to guide the user to update if we can't add an update feature in code?

elijah-potter:
# Issues 

Resolves #2862

# Description

See the original issue for more details. In short, I:

1. Wrote update instructions into the documentation for both the Chrome and Firefox Extensions.
1. Make the update indicator clickable. Clicking now takes you to the update instructions.

# How Has This Been Tested?

Manually.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/popup/Popup.svelte
+++ b/packages/chrome-plugin/src/popup/Popup.svelte
@@ -44,7 +44,7 @@ function openSettings() {
 }
 
 function openUpdateHelpPage() {
-	let url;
+	let url: string;
 
 	if (detectBrowserEngine() == 'chromium') {
 		url = 'https://writewithharper.com/docs/integrations/chrome-extension#Updating-the-Extension';


packages/web/src/routes/docs/integrations/chrome-extension/+page.md
```diff
@@ -11,3 +11,19 @@ We help improve the prose and grammar of your emails, messages, and documents, a
...
+
+Your browser will automatically update the extension to the latest version as soon as it is available.
+If it is delayed, you can initiate a manual update yourself by:
+
+1. Navigating to `chrome://extensions`.
+1. Enabling Developer mode using the toggle in the top-right corner.
+1. Clicking the Update button that appears.
+
+Be advised: Google's review process is often faster than other integrations (for example, Obsidian).
+This means that Chrome Extension updates are usually delayed by a few days compared to everything else.
```
hippietrail, line 28: Did you mean to say "often slower than"? I'm not familiar with the speed of Obsidian updates but "faster than" seems to contradict saying Chrome updates are "delayed" in the next line.

elijah-potter: Yup. You caught me there.

PR #2874, opened 2026-03-05
fix(core): `SplitWords` allowing uncommon merges

### LINKED ISSUE #2763: Flagging "cornleaves" should offer suggestion "corn leaves"
... or corn husks, but that might be out of scope.

Image

Unrelated: Dutasteride should be added to the dictionary.

elijah-potter:
# Issues 

Fixes #2763

# Description

The fundamental problem was that the derivatives of the words we marked "common" were not marked so in turn. I believe we should do so, since most common words are also commonly inflected in various ways. The infections can be assumed to be also common.

I fixed this by marking the "common" property to be allowed to propagate. 

# How Has This Been Tested?

Additional tests from the original issue.

# Checklist


harper-core/tests/text/linters/The Constitution of the United States.snap.yml
```diff
@@ -1029,7 +1049,17 @@ Message: |
...
-  - Replace with: “casing”
+  - Replace with: “cousin”
+
+
+
+Lint:    Typo (31 priority)
+Message: |
+     383 | The Congress may determine the Time of chusing the Electors, and the Day on
+         |                                        ^~~~~~~ `chusing` should probably be written as `ch using`.
+     384 | which they shall give their Votes; which Day shall be the same throughout the
```
hippietrail: This seems like a half-step backwards. It's not a full step backwards though since the correct `choosing` seems not to have been one of the suggestions even before this change.

It's probably more due to the odd dictionary entry `ch/~NI^iWv` than to this PR too.

harper-core/tests/text/linters/The Constitution of the United States.snap.yml
```diff
@@ -1292,6 +1322,16 @@ Suggest:
 
 
 
+Lint:    Typo (31 priority)
+Message: |
+     449 | Services, a Compensation, which shall neither be encreased nor diminished
+         |                                                  ^~~~~~~~~ `encreased` should probably be written as `en creased`.
+     450 | during the Period for which he shall have been elected, and he shall not
+Suggest:
+  - Replace with: “en creased”
```
hippietrail: It looks like `SplitWords` is preventing a useful `SpellCheck` suggestion `increased`?

harper-core/tests/text/linters/Alice's Adventures in Wonderland.snap.yml
```diff
@@ -2926,6 +2926,16 @@ Suggest:
 
 
 
+Lint:    Typo (31 priority)
+Message: |
+    2303 | treading on her toes when they passed too close, and waving their forepaws to
+         |                                                                   ^~~~~~~~ `forepaws` should probably be written as `fore paws`.
+    2304 | mark the time, while the Mock Turtle sang this, very slowly and sadly:—
```
hippietrail: I'll add `forepaw` to the next dictionary curation.

harper-core/tests/text/linters/The Constitution of the United States.snap.yml
elijah-potter: I feel that `ch` shouldn't be marked as "common". Am I crazy?

hippietrail: > I feel that `ch` shouldn't be marked as "common". Am I crazy?

You sir are incredibly sane! `CH` is pretty common symbol for Switzerland. Lowercase `ch` could be "channel"? `I` means pronoun - could be dialectal or obsolete? `^` is an affix compression artefact that creates `chin` which probably has its own entry but it's always best to double-check. `i`, `W`, and `v` I don't know off the top of my head, which usually means they're for affixes too... `i` = `in-` -> "inch", `v` = `-ive` -> "chive", `W` = `con-` -> "conch"

Hey `ch` is not in my `dictionary.dict`! Maybe I discovered it and removed it in recent curation? Or maybe you did between when you wrote this and when I'm replying (-:

elijah-potter: I just asked my Swiss girlfriend and she says she would not be insulted if we marked "CH" as uncommon either. I think we're good 👍.

elijah-potter: Oh yeah, I think part of a dictionary curation removed it. :shrug:

hippietrail: > Oh yeah, I think part of a dictionary curation removed it. 🤷

I sometimes add stuff quickly when I spot it and then don't even remember a couple of days later.

PR #2876, opened 2026-03-05
refactor(core): refactor code in `MissingTo`

86xsk:
# Description

- Removes unnecessary struct member.
- Tries to shrink and simplify the `MissingTo::match_to_lint` function.

# How Has This Been Tested?

- `cargo test` passes; no tests have been modified.

# Checklist

### DIFF
--- a/harper-core/src/linting/missing_to.rs
+++ b/harper-core/src/linting/missing_to.rs
@@ -197,6 +197,34 @@ impl MissingTo {
 
         None
     }
+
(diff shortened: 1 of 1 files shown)


hippietrail:
This uses more than a few techniques I'm not familiar with from Rust or from Harper so I've requested @elijah-potter review it since he knows both better than me. I'm also interested to see what he might have to say so I can maybe learn (-:

elijah-potter (approved):
This looks good to me. Much better! Thank you!

PR #2884, opened 2026-03-07
feat: it´s→it's

hippietrail:
# Issues 
N/A

# Description

I added support for using the "grave accent" character `´` instead of the apostrophe `'`.
I was able to integrate it right into the existing `SemicolonApostrophe` linter and so I also renamed it to `WrongApostrophe`.

# How Has This Been Tested?

I added unit tests using sentences found on GitHub using the acute instead of the apostrophe.

# Checklist

### DIFF
--- a/harper-core/src/linting/wrong_apostrophe.rs
+++ b/harper-core/src/linting/wrong_apostrophe.rs
@@ -7,13 +7,13 @@ use crate::{
 const CONTRACTION_AND_POSSESSIVE_ENDINGS: [&str; 7] = ["d", "ll", "m", "re", "s", "t", "ve"];
 
 pub struct WrongApostrophe {
-    expr: Box<dyn Expr>,
+    expr: FirstMatchOf,
 }
 
 impl Default for WrongApostrophe {
     fn default() -> Self {
         Self {
-            expr: Box::new(FirstMatchOf::new(vec![
+            expr: FirstMatchOf::new(vec![
                 Box::new(
                     SequenceExpr::any_word()
                         .then_semicolon()
@@ -24,7 +24,7 @@ impl Default for WrongApostrophe {
                         .then_acute()
                         .then_word_set(&CONTRACTION_AND_POSSESSIVE_ENDINGS),
                 ),
-            ])),
+            ]),
         }
     }
 }
@@ -33,7 +33,7 @@ impl ExprLinter for WrongApostrophe {
     type Unit = Chunk;
 
     fn expr(&self) -> &dyn Expr {
-        self.expr.as_ref()
+        &self.expr
     }
 
     fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {


harper-core/src/linting/wrong_apostrophe.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+impl ExprLinter for WrongApostrophe {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        self.expr.as_ref()
+    }
```
86xsk, line 37: Boxing the `FirstMatchOf` isn't necessary.

```suggestion
pub struct WrongApostrophe {
    expr: FirstMatchOf,
}

impl Default for WrongApostrophe {
    fn default() -> Self {
        Self {
            expr: FirstMatchOf::new(vec![
                Box::new(
                    SequenceExpr::any_word()
                        .then_semicolon()
                        .then_word_set(&CONTRACTION_AND_POSSESSIVE_ENDINGS),
                ),
                Box::new(
                    SequenceExpr::any_word()
                        .then_acute()
                        .then_word_set(&CONTRACTION_AND_POSSESSIVE_ENDINGS),
                ),
            ]),
        }
    }
}

impl ExprLinter for WrongApostrophe {
    type Unit = Chunk;

    fn expr(&self) -> &dyn Expr {
        &self.expr
    }
```

harper-core/src/linting/wrong_apostrophe.rs
```diff
@@ -0,0 +1,184 @@
...
+        let ending = &toks.last()?;
+
+        let replacement_str = format!(
+            "{}'{}",
+            base.span.get_content_string(src).to_lowercase(),
+            ending.span.get_content_string(src).to_lowercase()
+        );
+
+        let mut lettercase_template = base.span.get_content(src).to_vec();
+        lettercase_template.extend_from_slice(ending.span.get_content(src));
```
86xsk: An alternative that's slightly shorter and avoids a `mut`.

```suggestion
        let lettercase_template =
            [base.span.get_content(src), ending.span.get_content(src)].concat();
```

harper-core/src/linting/wrong_apostrophe.rs
hippietrail: > Boxing the `FirstMatchOf` isn't necessary.

Interesting! It's been in my muscle memory to always type `expr: Box<dyn Expr>,` and `self.expr.as_ref()` for those lines because that's what I'd seen in existing code and I always type them verbatim as boilerplate.

I thought this must be a difference in `SequenceExpr` that most `ExprLinter`s use vs using `FirstMatchOf` here but they look the same and when I try making the same change to a random one it also works.

So maybe almost all of our linters have an unnecessary level of boxing?

86xsk: > So maybe almost all of our linters have an unnecessary level of boxing?

I took a quick glance at a few others, and that does seem to be the case.

Afaik, basically the only case where boxing into a `Box<dyn Expr>` might be useful is if you don't know what type of `Expr` it's going to be at compile-time. However, in most cases it seems that the linters' constructor will only ever use a single hard-coded type for the expr (e.g. `FirstMatchOf`, `SequenceExpr`).

PR #2885, opened 2026-03-08
refactor: most `ExprLinter`s don't need to be boxed

hippietrail:
# Issues 
N/A

# Description

In a very interesting review of a recent PR I recently made I learned that most or all of my `ExprLinter`s were unnecessarily boxed. This PR refactors all `ExprLinter`s where it's possible to unbox them.

Almost all were edited manually.

`SoonToBe`, `RollerSkated`, and `RightClick` were refactored by Windsurf's built-in model and I then applied the same refactoring pattern to some other `ExprLinter`s such as `Respond` and `OughtToBe`.

`OrthographicConsistency` has a slightly different refactoring.

# How Has This Been Tested?

All existing tests continue to pass.

# Checklist


harper-core/src/linting/a_while.rs
```diff
@@ -47,22 +46,19 @@ impl Default for AWhile {
 
         let map = Arc::new(map);
 
-        Self {
-            expr: Box::new(map.clone()),
-            map,
-        }
+        Self { expr: map }
```
86xsk: I think in cases like this, it's most likely possible remove the `Arc` as well, in a similar manner to b5acd00e938b0e679dae032c435334f1327d2386

PR #2888, opened 2026-03-09
refactor(core): refactor `TokenStringExt`

86xsk:
# Description

Refactor `TokenStringExt` to provide default implementations for most of the trait's methods. This reduces the amount boilerplate needed to implement the trait.

With the changes, `TokenStringExt` has only 2 required methods: `tokens` and `tokens_mut`. The rest of the methods now have a default implementation that uses one of those 2 methods.

I don't *think* this is a breaking change, since (IIUC) the `TokenStringExt` trait is sealed, so it can't be implemented outside the crate. Not 100% sure though.

# How Has This Been Tested?

- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -4,7 +4,6 @@ use std::fmt::Display;
 
 use harper_brill::{Chunker, Tagger, brill_tagger, burn_chunker};
 use itertools::Itertools;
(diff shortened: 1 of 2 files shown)


elijah-potter:
It looks like GH Actions had an outage. I'm going to requeue this.

PR #2900, opened 2026-03-10
feature: spellcheck variant ignores `@username` mentions

### LINKED ISSUE #2403: Ignore user names mention in GitHub comments
When you write "@asafm this looks great" it flags "asafm" as misspelled word. 
Adding all usernames I work with to ignore seems too much.

**Proposed Solution**
Have it ignore "@{word}" 

**Examples**
Gave above

**Component**
- [ ] Core engine
- [ ] Plugin/Extension
- [ ] Other: _____

**Additional Context**
Any other relevant info.

hippietrail:
# Issues 
Resolves #2403 

# Description

Adds a second `SpellCheck` linter which does not flag misspelled words immediately preceded by `@` on the assumption that they are username mentions.

**To use this the user must enable `SpellCheckWithUsernames` and disable `SpellCheck`** since we do not yet have either mutually exclusive linter pairs/sets or linters with options. This is the same way `OxfordComma`/`NoOxfordComma` and `ViciousCircle`/`ViciousCycle` work.

My original idea was to refactor the `lint()` method of `SpellCheck` to separate the checking of each `Token`/word in a helper function which would be called by the `lint()` method of `SpellCheckWithUsernames` only when the previous `Token` is not `@`.

But the LLM built into Windsurf came up with this more lightweight approach where `SpellCheckWithUsernames`'s `lint()` simply calls the `lint()` of `SpellCheck` and then finds the `Token`s for each `Lint` and checks if the `Token` before is an `@`.

The LLM initially did it in a clunky way that I cleaned up. Together we found `document.token_indices_intersecting` as the way to get the `Token` before a `Lint`. I've never used this API before and it bears this comment:
```
    /// Locate all the tokens that intersect a provided span.
    ///
    /// Desperately needs optimization.
```
So it might be too heavy to use this way?

I don't know what edge cases there might be other than the `@` in email addresses. `SpellCheck` might interact with punctuation in some ways that might cause edge cases when they occur near a username mention or near other uses of `@` - please torture test to find edge cases!

# How Has This Been Tested?

I added unit tests to both `SpellCheckWithUsernames` and `SpellCheck` to ensure the former does not flag username mentions or get confused by email addresses and still flags usernames lacking the `@` sigil, and to ensure that regular old `SpellCheck` still flags usernames whether with `@` or not.

# Checklist


86xsk:
I'm concerned about implementing a temporary linter to work around the fact that we currently don't support configuration for individual linters. If we do implement such configuration in the future, it would be difficult to remove this linter, since I'm pretty sure it would be a breaking change.

Some alternative ideas that come to mind:
- Use `RegexMasker` to mask out such username sequences. One thing that would make this difficult right now is that (afaik) we don't allow the user to specify multiple regex.
- Add a lex function to `lex_english_token` that lexes such username sequences into `TokenKind::Unlintable`. Or alternatively, add a fix-up in `Document::apply_fixups`, in a similar manner to existing functions like `condense_filename_extensions`. The downside is that this would not be optional, but I'm not sure how much of an issue that would cause in practice.
- Similar to the above, but just handle the case in `SpellCheck` instead.

hippietrail:
> I'm concerned about implementing a temporary linter to work around the fact that we currently don't support configuration for individual linters. If we do implement such configuration in the future, it would be difficult to remove this linter, since I'm pretty sure it would be a breaking change.

I don't think removing this linter would be a problem. So far Harper isn't much more than a year old, but it has passed version 1, so there is that. However, linters have been removed before and names of linters changing isn't really rare. It's of course also possible to retain a linter's name while replacing the implementation. I think for now such breaking changes don't seem to be a problem. That will not remain the case forever though.

> Some alternative ideas that come to mind:
> 
> * Use `RegexMasker` to mask out such username sequences. One thing that would make this difficult right now is that (afaik) we don't allow the user to specify multiple regex.
> * Add a lex function to `lex_english_token` that lexes such username sequences into `TokenKind::Unlintable`. Or alternatively, add a fix-up in `Document::apply_fixups`, in a similar manner to existing functions like `condense_filename_extensions`. The downside is that this would not be optional, but I'm not sure how much of an issue that would cause in practice.
> * Similar to the above, but just handle the case in `SpellCheck` instead.

To me I think this is an optional thing that ideally should be enabled by default by certain document formats and/or by the user. Having options on linters or having config entries for such things and not just for just enable/disable each linter would be better. Some way for a file format to specify some default settings would be best of all.

I think making it part of a masker or a lex function would make it harder to make it optional. I wrote `condense_filename_extensions` last year and it has edge cases that are hard to think up good solutions for, and I think having something like it for username mentions would be the same but with maybe more edge cases. It seems to be best to err on the side of caution and allow them to claim only the clearest unambiguous examples as lexical units. This leaves some cases up to individual linters to analyze anyway. So you want to be sure the number of unambiguous cases you can catch covers a good portion of that work to make the linters simpler.

I'd be interested to hear any thoughts @elijah-potter might have. Maybe he'll feel compulsed to work on a strategy to add per-linter options (-:

Issue #389, opened 2025-01-15
feat: Flag decades with apostrophe in the wrong place


hippietrail:
Currently both incorrect and correct usages are flagged with `Did you mean to spell “s” this way?`, which is wrong for other reasons: <img width="545" alt="Image" src="https://github.com/user-attachments/assets/f1ed98f9-e3ea-457a-b072-ea729754e855" />

What it should do is flag and suggest:

- 80's → '80s
- The 70's → The '70s
- The 1990's → The 1990s

hippietrail:
This is one I still often ponder for the best solution.

hippietrail:
🌲

sudanchapagain:
+1

PR #2911, opened 2026-03-11
refactor: two-digit decade support

hippietrail:
# Issues 
Further work on #389

# Description

Adds support for 2-digit decades.
It's extremely common that people always write decades followed by an apostrophe as though the numbers possess something. Correctly, when referring to a decade of someone's life there should be no apostrophe, and when referring to a decade of history, the apostrophe should be before the numbers, representing the elided first two digits:
- In my **20's** I was a computer nerd → In my **20s** I was a computer nerd
- In the **70's**, Disco was king → In the **'70s**, Disco was king

# How Has This Been Tested?

Many unit tests harvested from GitHub. As with the 4-digit support, the linter can't yet discern all the possibilities. So the sentences are still included but `ignore`d so future devs may continue iterating on the module without having to collect more example sentences.

# Checklist

### DIFF
--- a/harper-core/src/linting/plural_decades/two_digits.rs
+++ b/harper-core/src/linting/plural_decades/two_digits.rs
@@ -3,6 +3,22 @@ use crate::{
     linting::{LintKind, Suggestion},
 };
 
+#[derive(PartialEq)]
+enum UsageJudgment {
+    NotMistake,
+    IsMistakeForDecade,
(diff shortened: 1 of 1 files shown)


hippietrail:
> Awesome! Great stuff. I'm a fan of all of those tests.

I'm thinking of ways to make tests more data-driven generally.

PR #2912, opened 2026-03-11
perf(core): avoid cloning in `DictWordMetadata::append`

86xsk:
# Description

Avoids unnecessary cloning in `DictWordMetadata::append`.

Previously, `append` would call `or`, (which would produce a copy of the value) and then overwrite itself with the result. With the changes, `append` modifies itself in place.

To avoid code repetition, the `DictWordMetadata` merging logic was moved from `or` to `append`.

To fix a minor performance regression on lint_essay, `MergedDictionary::get_word_metadata` was adjusted to avoid excessive use of `DictWordMetadata::or`.

(Cherry-picked from #2630, with adjustments to make it work and perform well on master.)

# How Has This Been Tested?

- `cargo test`; no tests modified.
- `cargo bench`. This seems to show a minor improvement of ~3-5% on parse_essay on my machine.

# Checklist

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -1985,6 +1985,7 @@ Chad/ONg
 Chadian/JNgS
 Chadwick/Og
 Chagall/NOg
+Chagas disease/Nmg
 Chaitanya/g
 Chaitin/g
 Chaldea/O
@@ -17095,7 +17096,6 @@ cetacean/JNgS
 ceteris
 cf/~V
 cg/~
-ch/~NI^iWv
 chad/~NS
 chafe/NSVGd
 chaff/NgSVGd
@@ -18185,6 +18185,8 @@ coffeepot/NgS
 coffer/NSgV
 cofferdam/NgS
 coffin/~NSgVdG
+cofounder/NSg
+co-founder/NSg
 cog/~NSgV
 cogency/Ng
 cogent/JY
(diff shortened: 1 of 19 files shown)


hippietrail (changes requested):
This is probably a great time to rename `append()` - I could be missing something but it's always bugged me. Both `or` and `merge` are more well-named. `combine` would also work.

"Append" literally means "attach to the end of a collection". Maybe something like this is also done somewhere I'm not seeing it. More likely an early simpler version of this did in fact just add new elements, or the better names were used for something else at the time or the coder at the time just had the wrong word pop into their head?

elijah-potter:
> "Append" literally means "attach to the end of a collection". Maybe something like this is also done somewhere I'm not seeing it. More likely an early simpler version of this did in fact just add new elements, or the better names were used for something else at the time or the coder at the time just had the wrong word pop into their head?

It was used in a specific algorithm at the time and it's implementation has evolved. It was a good name, now it is not.

harper-core/src/dict_word_metadata.rs
```diff
@@ -758,7 +732,34 @@ impl DictWordMetadata {
 
     /// Same thing as [`Self::or`], except in-place rather than a clone.
     pub fn append(&mut self, other: &Self) -> &mut Self {
```
elijah-potter: I agree with @hippietrail — let's rename this function. Otherwise, this looks great!

PR #2922, opened 2026-03-13
feat(core): allow users to include dictionaries in Weirpacks

elijah-potter:
# Issues 

N/A

# Description

I've set it up so that Weirpacks (specifically in the Chrome extension right now, but soon in other places) can include a dictionary. This is especially useful for Weirpacks used for authors who write in a domain-specific language (like that found in the IUPAC definitions).

@anyllmarkevich, I believe you wanted to be notified about this?

I am leaving this PR as a draft for now, because I need to roll the feature out to more places.

# How Has This Been Tested?

Manually + an integration test.

# Checklist

### DIFF
--- a/packages/harper.js/src/Linter.test.ts
+++ b/packages/harper.js/src/Linter.test.ts
@@ -640,7 +640,7 @@ for (const [linterName, Linter] of Object.entries(linters)) {
 		// It should pass now that we've included the "word" in the dictionary via the Weirpack.
 		lints = await linter.lint(source);
 		expect(lints).toHaveLength(0);
-	});
+	}, 30000);
 }
 
 // Disabled because it significantly slows down CI


anyllmarkevich:
Amazing! Thank you so much for the heads up!

hippietrail:
Are they annotated dictionaries like `dictionary.dict` or simple wordlists?

elijah-potter:
They are full annotated dictionaries like the main one in the repo. For a Weirpack to provide a dictionary, it must include both a `dictionary.dict` file and an `annotations.json` file.

elijah-potter:
On second thought, I don't want this PR to get too bloated. I'll save further integrations (like the LSP, for example) for another day.

hippietrail:
> They are full annotated dictionaries like the main one in the repo. For a Weirpack to provide a dictionary, it must include both a `dictionary.dict` file and an `annotations.json` file.

I would think that having an `annotations.json` is too much for an addition to English. That's something that only new languages should have. We'd end up with clashing annotation flags between main dictionary and Weirpack addon dictionaries. My 2¢ anyway.

elijah-potter:
Part of the intention here is for the Weirpacks to be a way to slowly modularize Harper over time. New languages will hopefully be implemented as Weirpacks on top of the English support. 

From a technical standpoint, there will be no overlap between the Weirpack dictionaries and the curated one. They are parsed and expanded separately.

PR #2928, opened 2026-03-14
fix: don't flag "more orange than red" etc.

### LINKED ISSUE #2925: "the effects would be subtler!! than direct." "be + more + adjective + than" is correct.
**What got flagged?**
"the effects would be more subtle than direct."  flagged More + subtle (more + adjective)

**Why is this incorrect?**
"the effects would be subtler than direct" would of course be incorrect.
the above is correct. be + More + adjective + than is correct.

**Example of correct usage:**
[Your example here]

hippietrail:
# Issues 

Fixes #2925

# Description

We were flag "more [adj]" in a context where it's part of a legitimate English construction: "more [adj] than [adj]".

# How Has This Been Tested?

I added a unit test using the sentence in the GitHub issue.

# Checklist


harper-core/src/linting/more_adjective.rs
```diff
@@ -19,11 +19,15 @@ where
...
         Self {
-            expr: Box::new(
-                SequenceExpr::word_set(&["more", "most"])
-                    .t_ws()
-                    .then_positive_adjective(),
-            ),
+            expr: SequenceExpr::word_set(&["more", "most"])
+                .t_ws()
+                .then_positive_adjective()
+                .then_optional(
```
86xsk, line 26: I would consider adding a comment to state that the presence of the optional clause indicates a false positive. It would make it a little easier to understand when reading top to bottom.

harper-core/src/linting/more_adjective.rs
```diff
@@ -47,11 +51,11 @@ where
     type Unit = Chunk;
 
     fn expr(&self) -> &dyn Expr {
-        self.expr.as_ref()
+        &self.expr
     }
 
     fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {
-        // Check invariants just in case the Expr changes
+        // Abort when the optional clause is present, or when the `Expr`` changes
```
86xsk: The extra backtick after the 'Expr' seems to be a typo.

PR #2932, opened 2026-03-14
feature/obsidian custom suggestions

stanley-910:
# Issues

# Description
Add custom typo suggestions to the Obsidian plugin so users can define personal fixes like `adn: and` and surface them directly in lint suggestions.
This PR includes:
- a new `Custom Typo Suggestions` setting in Obsidian
- custom suggestions prepended ahead of built-in Harper suggestions
- `:`-based typo mapping syntax (for example `adn: and`)

**Bugs found during implementation:**

- Harper settings could render before async settings finished loading, causing `useWebWorker` to be `undefined` and crash the settings tab. Fixed by adding a safe loading state and waiting for async settings/defaults/descriptions before rendering.
- Editing the custom typo textarea caused repeated `Disabled` / `Enabled` linter churn and unstable reload behavior. Fixed by serializing settings updates and avoiding full editor-linter recreation on every settings edit.

# Demo

https://github.com/user-attachments/assets/e0aebc1c-c8c0-4c2d-be7b-bdecdb5586c0

# How Has This Been Tested?
- Added unit tests for:
  - custom typo parsing/serialization/normalization
  - settings persistence for custom typo suggestions
  - regression coverage ensuring non-delay settings edits do not recreate the editor linter
  - serialized settings writes applying in order
- Ran:
  - `pnpm exec vitest run src/State.test.ts src/customSuggestions.test.ts src/lint.test.ts --browser.enabled=false`
  - `pnpm build`
# Checklist


elijah-potter:
I love that you took initiative to build out this feature, but I would have preferred it to wrap our existing work with Weir. Is there a specific reason you chose this specific syntax? Could we achieve similar behavior by dynamically creating Weir rules?

For context, here is the Weir documentation: https://writewithharper.com/docs/weir

stanley-910:
Oh I actually wasn't familiar with Weir; Taking a look over it now, it looks very promising, and apologies if it looked like I was stepping on its toes. 

There are a couple typos I make consistently that don't show up as suggested fixes w/ Harper's current suggestions, and since I mostly use it on the Obsidian side, I thought a feature where people could define custom suggestions to specific string values might be handy... if only as a temporary work around before Weir does get fully integrated. I mostly designed it around being compact and quickly changeable (formatted like `dict {key: list[values]}`) , similar to how the rest of the Obsidian Harper UI works vis-a-vis the 'Personal Dictionary' and 'Mask' components.

So, while not a full Weir integration, I still think it could prove handy in the interim to users.

PR #2942, opened 2026-03-16
refactor: `once_cell`→`LazyLock`

hippietrail:
# Issues 
N/A

# Description

A few months back we standardized on `LazyLock` and refactored all places we were using `lazy_static`. But there was one place we were using `once_cell` instead, and it was not refactored.

This PR refactors this last spot, in the "rune" module, to `LazyLock`.

# How Has This Been Tested?

`cargo test`

# Checklist


hippietrail:
@elijah-potter I bet it's failing due to the nondeterministic thing that happens sometimes in the Rust check lately.

elijah-potter:
I bet you're right. I'll fix the problematic test now.

Issue #383, opened 2025-01-15
feat: Flag abstract nouns mistakenly used in place of certain adjectives


hippietrail:
It has become common among non-native and younger native English speakers to say things like:

- "You are bias" instead of "You are biased"
- "He is prejudice" instead of "He is prejudiced"

Among non-native speakers in some regions such as SIngapore and Malaysia a few others have become common.

- "I was very shock" instead of "I was very shocked"

That's easily the most common one.
----
Here's some I'm adding as I spot them in the wild:
- I'm really **hype** about this current wave of AI
- It is way **overhype** in 'the west' and China. → It is way **overhyped** in 'the west' and China.

hippietrail:
Still very much on my radar.

hippietrail:
🌲

PR #2943, opened 2026-03-16
feat: I am worry→I am worried

hippietrail:
# Issues 

May be possible to refactor into a general solution to #383

# Description

I heard a Japanese YouTuber say "I am worry" when it should be "I am worried". This kind of thing is very common from speakers of various Asian languages. Compare Singapore/Malaysia English "I am shock" and the recent phenomenon in younger generations of English native speakers "I am bias". See #383

Fixes all combinations of subject pronoun and present tense forms of "to be", except when the pronoun is "it" since "it is worry" is either used correctly or is a different mistake where it should be "it is worrying" or "it is a worry".

# How Has This Been Tested?

Unit tests for the various pronouns and forms of "to be" were harvested from GitHub and Stack Exchange.

# Checklist


hippietrail:
> I would love to see this refactored into a general solution, but I think it's good for now. Thanks!

Yes me too. I have a strong feeling there will be edge cases and false positives in a more general solution. That's part of the reason I started with this limited solution. The other reason was my head was full of trickier Harper ideas I was working on.

PR #2961, opened 2026-03-18
feat(core): rule that requires that the dash between numbers in a range is an en dash

elijah-potter:
# Issues 

No particular issue, I just read about this in a book and want it enforced in Harper.

# Description

From [Practical Typography](https://practicaltypography.com/hyphens-and-dashes.html): 

> The en dash has two uses.
>
> 1. It indicates a range of values (1880–1912, pages 330–39, Exhibits A–E). If you open with from, pair it with to instead of an en dash (from 1880 to 1912, not from 1880–1912).
> 2. It denotes a connection or contrast between pairs of words (conservative–liberal split, Arizona–Nevada reciprocity, Sarbanes–Oxley Act).

This PR exists to enforce point 1.

# How Has This Been Tested?

Additional unit tests.

# Checklist

### DIFF
--- a/harper-core/src/linting/numeric_range_en_dash.rs
+++ b/harper-core/src/linting/numeric_range_en_dash.rs
@@ -51,7 +51,7 @@ impl Default for NumericRangeEnDash {
         // multipart numeric forms that should keep their existing separators.
         let pattern = SequenceExpr::default()
             .then_number()
-            .then(is_target_dash as fn(&Token, &[char]) -> bool)
+            .then(is_target_dash)
             .then_number();
 
         Self { expr: pattern }


harper-core/src/linting/numeric_range_en_dash.rs
```diff
@@ -0,0 +1,218 @@
...
+    let followed_by_dash_numeric = matches!(
+        after,
+        [dash, number, ..] if is_numeric_range_dash(dash) && number.kind.is_number()
+    );
+
+    preceded_by_numeric_dash || followed_by_dash_numeric
+}
+
+pub struct NumericRangeEnDash {
+    expr: Box<dyn Expr>,
```
86xsk: #2885

harper-core/src/linting/numeric_range_en_dash.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+fn is_numeric_range_dash(token: &Token) -> bool {
+    matches!(
+        token.kind.as_punctuation(),
+        Some(Punctuation::Hyphen | Punctuation::EnDash | Punctuation::EmDash)
+    )
+}
+
+fn is_chained_numeric_form(context: Option<(&[Token], &[Token])>) -> bool {
```
86xsk, line 24: I think it would help readability slightly if there was a comment/example indicating what this is matching.

harper-core/src/linting/numeric_range_en_dash.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    fn match_to_lint_with_context(
+        &self,
+        matched_tokens: &[Token],
+        _source: &[char],
+        context: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
+        if is_chained_numeric_form(context) {
+            return None;
+        }
```
86xsk, line 76: Having the `is_chained_numeric_form` logic in its own function makes this very clean and easy to read 👍

harper-core/src/linting/numeric_range_en_dash.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+impl Default for NumericRangeEnDash {
+    fn default() -> Self {
+        // Match isolated numeric ranges like `12-14` or `3—5`.
+        // The context check below skips dates, version chains, and similar
+        // multi-part numeric forms that should keep their existing separators.
+        let pattern = SequenceExpr::default()
+            .then_number()
+            .then(is_target_dash as fn(&Token, &[char]) -> bool)
```
86xsk: rust-analyzer seems to be okay with this without the cast here, though I guess the explicitness doesn't hurt.

PR #2964, opened 2026-03-18
feat: add lint rule for "more that likely" typo

### LINKED ISSUE #2930: Need to flag "more that likely" when "more than likely" was intended.
Mixing up "than" and "that" is a common mistake. Almost definitely a typo almost all of the time.

- More than possible, likely. **More that likely**:
- ... so it is **more that likely** that I could have messed up something pretty basic
- It's **more that likely** that I'm missing something.
- There is **more that likely** some obvious solution
- **More that likely** this is due to a variable value being freed out from under the MCA variable system.
- ... though it is **more that likely** that the account is logged on
- it's more that likely that such an algorithm would pick UTXOs controlled by exchanges.

I didn't rush straight into making a Weir file for this because I found false positives:

- That said if you already have it set to cache 2 or **more that likely** won't help
- There's **more that likely** could be done here including looking into cabins with multiple rooms listed
- But it is an ambitious project with many items to complete, and many **more that likely** haven't even been added to the list yet.

ghost:
## Summary

Adds a Weir lint rule that flags the common typo "more that likely" and suggests "more than likely".

This is a frequent than/that mix-up found across codebases and documentation.

## Examples

| Input | Suggestion |
|-------|-----------|
| "It is more that likely a typo." | "It is more than likely a typo." |
| "More that likely this is due to a config issue." | "More than likely this is due to a config issue." |

Correct usage of "more that" in other contexts is not flagged:
- "There is more that I need to tell you." (no match)
- "He said more that day than usual." (no match)

## Test plan

- `cargo test -p harper-core --lib -- run_tests_for_more_that_likely` passes (6 true positives, 6 true negatives)

Closes #2930

### DIFF
--- a/harper-core/src/linting/weir_rules/MoreThatLikely.weir
+++ b/harper-core/src/linting/weir_rules/MoreThatLikely.weir
@@ -1,0 +1,25 @@
+expr main <(more that likely), that>
+
+let message "Did you mean `more than likely`?"
+let description "Corrects the common typo `more that likely` to `more than likely`."
+let kind "Typo"
+let becomes "than"
+let strategy "MatchCase"
+
+# True positives
+
+test "It is more that likely a typo." "It is more than likely a typo."
+test "More that likely this is due to a config issue." "More than likely this is due to a config issue."
+test "There is more that likely some obvious solution." "There is more than likely some obvious solution."
+test "It's more that likely that I messed something up." "It's more than likely that I messed something up."
+test "MORE THAT LIKELY this broke in the last release." "MORE THAN LIKELY this broke in the last release."
+test "So it is more that likely a bug." "So it is more than likely a bug."
+
+# False positives / true negatives
+
+allows "It is more than likely a typo."
+allows "More than likely this is a config issue."
+allows "There is more that I need to tell you."
+allows "I know more that could help."
+allows "He said more that day than usual."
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/weir_rules/MoreThatLikely.weir
```diff
@@ -0,0 +1,25 @@
+expr main <(more that likely), that>
```
hippietrail, line 1: Ooh I haven't learned this Weir feature yet!

harper-core/src/linting/weir_rules/MoreThatLikely.weir
```diff
@@ -0,0 +1,25 @@
...
+
+let message "Did you mean `more than likely`?"
+let description "Corrects the common typo `more that likely` to `more than likely`."
+let kind "Typo"
+let becomes "than"
```
hippietrail, line 6: Ah I see how it works now. Thanks!

harper-core/src/linting/weir_rules/MoreThatLikely.weir
elijah-potter: Yeah, I think it's pretty neat. It allows you to select subsections of a previously defined Weir expression.

Issue #2969, opened 2026-03-19
Need to flag "use to" when "used to" is intended


hippietrail:
### The correct phrase to use when talking about a past habit is "used to" with "used" in the simple past tense.

- Cline **use to** be able to doc in it's own separate window. ❌
- it's the way we **use to** do it - as it's a matter of preference ❌
- Normally I **use to** develop with objects in JS ❌
- I **use to** see my workspace flow and now I don't ❌

### The same phrase is used after forms of "to be" and "to get" to mean "accustomed to".

- If you are **use to** Ubuntu, then the way sudo works should not is not strange. ❌
- im pretty new to vim helix way of typing and im getting **use to** it slowly ❌
- I've tried to get **use to** the tabs.... but it it's bad enough that it makes me want to leave for another product ❌
- I wasn't **use to** doing a lot of work and studying in high school ❌
- so I'll give it some time to get **use to** and then see about making a feature request ❌
- I'm sure I will get **use to** it and enjoy it more when I am using the pods ❌
- Issues labeled as such are a good way to get **use to** the codebase. ❌
- I think it'll take time but people will get **use to** it. ❌
- this key binding is important and we really got **use to** it ❌
- However it is a challenge to began getting **use to** the keyboard when navigating VSCode ❌

### The phrase "use to" can occur in many other contexts. These must not be flagged:

- What compiler did you **use to** build your app? ✅
- LiteLLM, which we **use to** talk to LM Studio, requires this interpreter.llm.api_base ✅
- .xpi files that you can **use to** install it in your browser are available ✅
- Look to the creators of the terminal emulator you **use to** fix these issues if the previously mentioned fixes don't work for you. ✅
- Here are some of the hack I **use to** make my life working on the command line a bit easier. ✅
- you can add rules that the endpoint will **use to** allow or disallow user's connections ✅

### Passive constructions

While looking for the above examples I discovered that another very common mistake is to write "use to" instead of "used to" in normal passive constructions that are not either of the idioms mentioned above:

- It is **use to** mitigate issues with prompt poisoning ❌
- This is **use to** host fragment. ❌
- Which method is **use to** send messages? ❌
- This script is **use to** fetch all the branches from all the repositories in the current directory ❌
- In printing Prefix and escape predicates are **use to** perform validation and drive the logic of printing. ❌

PR #2975, opened 2026-03-19
feat: flag "use to" when "used to" is intended

ghost:
## Summary

Adds a phrase set correction that flags "use to" and suggests "used to" when meaning accustomed to (after forms of `be` or `get`).

## Patterns covered

- Be verbs + "use to" (am/is/are/was/were/be use to)
- Be contractions + "use to" (I'm/we're/you're/they're/he's/she's/it's use to)
- Getting/get/got + "use to"

Pronoun patterns (I/we/they use to) were intentionally excluded to avoid false positives with legitimate verb usage like "the editor I use to write code".

## Test plan

- `cargo test -p harper-core --lib -- phrase_set_corrections::tests` - 3 true positives + 2 true negatives pass
- Correct "used to" is not flagged
- "use to" as a verb is not flagged

Partial fix for #2969 (covers be/get patterns; pronoun patterns need POS-aware approach)

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -516,25 +516,28 @@ pub fn lint_group() -> LintGroup {
         ),
         "UseToUsedTo" => (
             &[
-                ("i use to", "i used to"),
-                ("we use to", "we used to"),
-                ("they use to", "they used to"),
-                ("you use to", "you used to"),
-                ("he use to", "he used to"),
-                ("she use to", "she used to"),
-                ("it use to", "it used to"),
-                ("getting use to", "getting used to"),
-                ("get use to", "get used to"),
-                ("got use to", "got used to"),
+                // "be" verbs + "use to" -> "used to" (accustomed to)
                 ("am use to", "am used to"),
                 ("are use to", "are used to"),
                 ("is use to", "is used to"),
                 ("was use to", "was used to"),
                 ("were use to", "were used to"),
                 ("be use to", "be used to"),
+                // contractions of "be"
+                ("i'm use to", "i'm used to"),
+                ("we're use to", "we're used to"),
+                ("you're use to", "you're used to"),
+                ("they're use to", "they're used to"),
+                ("he's use to", "he's used to"),
+                ("she's use to", "she's used to"),
+                ("it's use to", "it's used to"),
+                // "get" forms + "use to" -> "used to" (becoming accustomed)
+                ("getting use to", "getting used to"),
+                ("get use to", "get used to"),
+                ("got use to", "got used to"),
             ],
             "The correct form is `used to`, not `use to`.",
-            "Corrects `use to` to `used to` when describing past habits or meaning accustomed to.",
+            "Corrects `use to` to `used to` when meaning accustomed to (after forms of `be` or `get`).",
             LintKind::Grammar
         ),
         "WreakHavoc" => (
--- a/harper-core/tests/text/linters/The Great Gatsby.snap.yml
+++ b/harper-core/tests/text/linters/The Great Gatsby.snap.yml
@@ -5363,16 +5363,6 @@ Suggest:
 
 
 
-Lint:    Grammar (31 priority)
-Message: |
-    4066 | “That was his cousin. I knew his whole family history before he left. He gave me
-    4067 | an aluminum putter that I use to-day.”
-         |                         ^~~~~~~~ The correct form is `used to`, not `use to`.
-Suggest:
-  - Replace with: “I used to”
-
-
-
 Lint:    Spelling (31 priority)
 Message: |
     4070 | the window, followed by intermittent cries of “Yea—ea—ea!” and finally by a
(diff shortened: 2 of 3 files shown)


elijah-potter:
It looks like some snapshot tests failed. Don't worry — there is an easy fix.

Assuming you have Rust installed, just run `cargo test || cargo test` to regenerate the snapshot files, then commit the resulting changes.

Thanks for opening a PR!

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -514,6 +514,29 @@ pub fn lint_group() -> LintGroup {
             "Ensures proper use of the subjunctive mood in counterfactual conditional statements starting with `if only` or `I wish`.",
             LintKind::Grammar
         ),
+        "UseToUsedTo" => (
+            &[
+                ("i use to", "i used to"),
+                ("we use to", "we used to"),
+                ("they use to", "they used to"),
+                ("you use to", "you used to"),
```
hippietrail: These context-free ones will match things like "This is the editor * use to write code". Maybe more often that matching the actual mistake.

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -514,6 +514,29 @@ pub fn lint_group() -> LintGroup {
...
+                ("he use to", "he used to"),
+                ("she use to", "she used to"),
```
hippietrail: These will match things like "What editor does he/she use to write code?" Also likely to hit more false positives that actual mistakes.

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -514,6 +514,29 @@ pub fn lint_group() -> LintGroup {
...
+                ("it use to", "it used to"),
```
hippietrail: "What API does this app use to get that effect?" etc.

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -514,6 +514,29 @@ pub fn lint_group() -> LintGroup {
...
+                ("getting use to", "getting used to"),
+                ("get use to", "get used to"),
+                ("got use to", "got used to"),
+                ("am use to", "am used to"),
+                ("are use to", "are used to"),
+                ("is use to", "is used to"),
+                ("was use to", "was used to"),
+                ("were use to", "were used to"),
+                ("be use to", "be used to"),
```
hippietrail, line 525: For these ones false positives aren't springing to mind, but I always do tons of Google searches to check because language is full of surprises.

Contractions of be would also fit:
I'm use to, we're use to, you're use to, he's use to, she's use to, it's use to, they're use to.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -5363,6 +5363,16 @@ Suggest:
 
 
 
+Lint:    Grammar (31 priority)
+Message: |
+    4066 | “That was his cousin. I knew his whole family history before he left. He gave me
+    4067 | an aluminum putter that I use to-day.”
```
hippietrail: This is an edge case with a very old-fashioned spelling of today. It can be prevented by checking for following hyphen. I know two ways to do that in an `ExprLinter` but `phrase_set` can't do it and I'm not sure if Weir can do it but I'm assuming not.

harper-core/src/linting/phrase_set_corrections/mod.rs
ghost: You're right, dropped all the pronoun patterns. They're way too likely to hit legitimate verb usage. Kept only the be-verb and get-verb patterns where it's unambiguously wrong, and added contractions too (I'm/we're/etc). Also added a false positive test for the verb case.

harper-core/src/linting/phrase_set_corrections/mod.rs
ghost: Added contractions: I'm/we're/you're/they're/he's/she's/it's use to. These should all be safe since a contraction of be + use to is always wrong.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
ghost: Good catch on the old-fashioned spelling. Since phrase_set can't check for a following hyphen, this is a known limitation for now. If it becomes a problem in practice we could convert this specific pattern to an ExprLinter later.

ghost:
Good idea, updated the PR description. Changed it to reference #2969 without closing it so the harder pronoun cases stay open for someone to tackle with a POS-aware approach.

harper-core/src/linting/phrase_set_corrections/mod.rs
hippietrail: I also maintain a corpus of real-word coding-related text that I often run linters I'm working on over. It helps to discover edge cases where you never would've expected any false positives. Mine are from the trending page of GitHub that I use to download their READMEs. There's probably a ton of ways to do it.

ghost: That's a great idea, thanks for the tip. Running against real README text would definitely catch edge cases that unit tests miss.

PR #2987, opened 2026-03-20
Added native Neovim LSP config to docs

RobertCrummett:
# Discussion 
https://github.com/Automattic/harper/discussions/2286#discussioncomment-15140872

# Description
Added example native Neovim LSP config setup to docs

### DIFF
--- a/packages/web/src/routes/docs/integrations/neovim/+page.md
+++ b/packages/web/src/routes/docs/integrations/neovim/+page.md
@@ -81,6 +81,28 @@ require('lspconfig').harper_ls.setup {
 }
 ```
 
+## Native Neovim LSP Config
+
+Neovim supports language servers [natively](https://neovim.io/doc/user/lsp.html), and therefore, Neovim supports Harper natively. To set up, first make sure that `harper-ls` is available on your system path. Next, add the following lines to your `init.lua` file:
+
+```lua
+-- General LSP setup
+vim.lsp.config['*'] = {
+    capabilities = { textDocument = { semanticTokens = { multilineTokenSupport = true } } },
+    root_markers = { '.git' },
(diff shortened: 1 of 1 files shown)


packages/web/src/routes/docs/integrations/neovim/+page.md
```diff
@@ -81,6 +81,28 @@ require('lspconfig').harper_ls.setup {
...
+```lua
+-- General LSP setup
+vim.lsp.config['*'] = {
+    capabilities = { textDocument = { semanticTokens = { multilineTokenSupport = true } } },
+    root_markers = { '.git' },
+}
+vim.diagnostic.config({ virtual_lines = true })
+
+-- Harper specific setup
+vim.lsp.config['harper'] = {
```
elijah-potter, line 97: I've seen this more commonly written using `vim.lsp.config()` as a function call rather than an assignment. Can you confirm that these two methods are interchangeable? I don't want to be misleading anyone.

RobertCrummett: Function call is probably more usual. I chose table assignment here is because the [quickstart](https://neovim.io/doc/user/lsp/) uses the table assignment in their example.

`:help vim.lsp.config` states that the config can be table-assigned (last line)

```vimdocs
config({name}, {cfg})                                       *vim.lsp.config()*
    Update the configuration for an LSP client.

    Use name '*' to set default configuration for all clients.

    Can also be table-assigned to redefine the configuration for a client.
```

Based on what I see, the methods are interchangeable.

elijah-potter: Fantastic! Thank you!

PR #2994, opened 2026-03-21
feat(core): capture range for `SequenceExpr`

86xsk:
# Description

Adds the `start_capture` and `end_capture` functions to `SequenceExpr`. These functions allow limiting the range of tokens captured by a `SequenceExpr`. They're inspired by and intended to function similarly to vim regex's `\zs` and `\ze` atoms.

One implementation detail I'm not too sure about is the `RangeEnum` type. It's functionality feels fairly distant from a grammar checker, and it feels like something that might better be delegated to a dependency. I'm not sure if there is a dependency we can/should use for this, or if there's a simpler way to implement it entirely.

All the added types/functions are currently `pub(crate)`, with the intention of allowing otherwise breaking changes until this can become more refined through use/experience.

# How Has This Been Tested?

- `cargo test`

# Checklist


hippietrail:
Interesting approach! I consider myself fairly decent at regexes but I've never known about `\zs` and `\ze` before you mentioned them the other day. I only ever did capturing with `(` and `)` (or `\(` and `\)` in vim) around subexpressions.

When I played with adding capturing I was going to "enable" capturing for any selected `Pattern` or `Expr`, or maybe just `SequenceExpr` and internally it would create an array so implicitly 0, 1, 2, etc. That would either get passed to `match_to_lint` as an extra arg that would just be `None` if there were no capture groups.

Or try to "walk the tree" of `Pattern`s/`Expr`s - I got even less far with that idea than with the previous idea.

Since then I've also considered having a new kind of `Pattern` or `Expr` that would only capture its inner `Pattern` or `Expr`. <s>I didn't even try to implement that idea.</s> I did make that a few months ago and called it `Rig` - it's in PR branch #3422.

elijah-potter:
I apologize for my delay in reviewing this.

I agree with your assessment of the `RangeEnum`. Could the same behavior be replicated with a `Span`?

86xsk:
No worries!

I think it should be possible to rework this to work with spans (and it would almost certainly result in less code). I think the only reason I didn't do this in the first place is because there isn't/wasn't a clean way to represent an unbounded range (e.g. `2..`). Though for all practical purposes, I think using the min/max values of the Span index type should work the same. Trying to perfect it isn't worth the complexity in hindsight. (Though thinking about it now, it might also be practical to use a pair of https://doc.rust-lang.org/std/ops/enum.Bound.html)

I have been fairly busy lately, so I'm not sure when I'll be able to get back to this. I'll try to get this done sooner than later, but it might be a week or a few.

elijah-potter:
> I have been fairly busy lately, so I'm not sure when I'll be able to get back to this. I'll try to get this done sooner than later, but it might be a week or a few.

No worries at all. Thank you!

86xsk:
Rewrote it to use `Span` instead. While refamiliarizing myself with the code, I also noticed that the logic was incorrect for expressions that don't produce single tokens. I fixed that by making it so the capture range specifies the range of `Exprs` that contribute to `window`, rather than a being a simple clamp on the bounds of the window. To counteract the added complexity, I tried to simplify the code in `SequenceExpr::run`.

The overall diff is a little ugly. I tried to break the changes up into small commits which should hopefully be a little easier to follow.

elijah-potter (approved):
This looks great! I have only one comment, then we can merge it. Thank you!

PR #2998, opened 2026-03-22
fix(lint-framework): reconnect popup host after DOM replacement

k4sperski:
# Issues

N/A

# Description

This fixes a popup lifecycle bug in the web lint framework when sites replace `document.body` during runtime.

## User Impact

From a user's perspective, this bug manifests as Harper "silently dying" after a page transition. Even though the extension is active, clicking highlights or trying to open the suggestion popup does nothing because the UI is attempting to attach to a detached/stale document.body. Potential failure on web pages that swap the body or main container during navigation.

## Technical Changes

While validating the original popup reconnect fix, I found there were actually two disconnected-DOM failures:
- Harper's popup host could become disconnected from the DOM
- cached text field mirror elements in `TextFieldRange` could also remain attached to the old detached `body`, which caused highlight computation to fail after the DOM replacement

This patch fixes both parts:

- `RenderBox` now accepts a parent resolver, so it can reattach to the current `document.body` instead of a stale body reference
- `PopupHandler` now renders first, then opens/closes the popover only if the host is connected and in the expected state
- `TextFieldRange` now rebuilds a cached mirror arena if the cached mirror is no longer connected

This stale-mirror failure was most obvious in Firefox during validation, but the underlying disconnected-cache issue is not Firefox-specific.

# Demo

The regression test page reproduces the original failure by:
1. rendering a textarea and Harper highlight
2. replacing `document.body` while preserving the app node
3. opening the popup from the highlight location

Before this patch, that flow could fail because the popup host or the cached text field mirror remained attached to the old detached body.
After this patch, highlights continue to resolve correctly and the popup opens as expected.

# How Has This Been Tested?

- I built the Chrome extension locally.
- I built the Firefox extension locally.
- I ran the `popup_reconnect` Playwright regression in Chromium and Firefox and verified it passes with this patch applied.
- I verified that the stale text field mirror can be reproduced after `document.body` replacement, and that rebuilding disconnected cached mirrors resolves that failure mode.

# Checklist


packages/chrome-plugin/tests/popup_reconnect.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			await new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
+		}
+
+		return null;
+	}, timeoutMs);
+}
+
+async function triggerPopupFromHighlightPosition(page: Page): Promise<boolean> {
```
elijah-potter: I'd like to see some doc comments for these utility functions. Explain what they do and more importantly __why__ they do it.

Also: would be possible to merge these with the test utilities module?

elijah-potter (changes requested):
I really appreciate this PR, and it's _almost_ ready. I only have some code-quality comments. 

I tested it locally and it all seems to work as expected, in addition to the tests you've provided.

packages/chrome-plugin/tests/popup_reconnect.spec.ts
k4sperski: Addressed!

elijah-potter (approved):
__Much__ better. Thanks!

PR #3008, opened 2026-03-23
fix(core): skip PronounVerbAgreement for hyphenated compound words

### LINKED ISSUE #3006: False positive: "co-founded" triggers `PronounVerbAgreement`
While testing I saw "co-founded" flagged, and assumed `DisjointPrefixes` would suggest changing it to "cofounded", but it made a seemingly bizarre suggestion instead:

Image

The text "In 2014, he co-founded Social Chain" got flagged with "The form of the verb must agree in grammatical number with the pronoun." It wants to change "co" to "cos".

I assumed a quirk of the dictionary having "co" marked as a verb. But it's marked as `co/~NSI(dE` with no `/V` to signify "verb"

But `/I` does signify "pronoun". I know there's been new pronouns invented but this is not one that I've heard before. It's in Wiktionary:
>co (third-person singular, gender-neutral, reflexive coself)
>
>(nonstandard) Gender-neutral subject pronoun, coordinate with gendered pronouns he and she. 
>(nonstandard) Gender-neutral object pronoun, coordinate with gendered pronouns him and her.

So we could just remove this annotation, but the question is why does it change "co" to "cos"? You don't normally add an "s" to pluralize pronouns in English like you do for regular nouns or to make the 3rd person singular present form of a verb.

I think there's some broken logic revealed here that needs to be investigated for a proper fix.

mvanhorn:
## Summary

Fixes a false positive where "co-founded" triggers `PronounVerbAgreement` because "co" is annotated as a pronoun in the dictionary. The linter sees "he" (pronoun) + "co" (verb-like pronoun) and suggests changing "co" to "cos".

## Changes

In `match_to_lint_with_context`, added an early return that checks the `after` context: if the token immediately following the matched verb is a hyphen, the verb is actually a compound word prefix (like "co-" in "co-founded" or "re-" in "re-established"), not a standalone verb. The lint is skipped in this case.

Added a regression test for "In 2014, he co-founded Social Chain." which previously triggered the false positive.

**Files changed:** `harper-core/src/linting/pronoun_verb_agreement.rs`

## Testing

All 44 existing tests pass. The 2 pre-existing `#[ignore]` tests remain unchanged.

```
test result: ok. 44 passed; 0 failed; 2 ignored; 0 measured; 4963 filtered out
```

Fixes #3006

### DIFF
--- a/harper-core/src/linting/pronoun_verb_agreement.rs
+++ b/harper-core/src/linting/pronoun_verb_agreement.rs
@@ -190,6 +190,15 @@ where
 
         let verb_tok = toks.last()?;
 
+        // If the verb is immediately followed by a hyphen, it's a compound word prefix
+        // (e.g. "co-founded"), not a standalone verb. Skip it.
+        if let Some((_, after)) = ctx
+            && let [next_tok, ..] = after
+            && next_tok.kind.is_hyphen()
+        {
+            return None;
+        }
+
         if let Some((before, _)) = ctx
             && let [.., prev_word_tok, ws_tok] = before
             && ws_tok.kind.is_whitespace()
@@ -649,6 +658,14 @@ mod lints {
     fn false_positive_she_hung_up() {
         assert_no_lints(
             "When I reiterated the conditions I'd previously set, she hung up on me.",
+            PronounVerbAgreement::new(FstDictionary::curated()),
+        );
+    }
+
+    #[test]
+    fn false_positive_co_founded() {
+        assert_no_lints(
+            "In 2014, he co-founded Social Chain.",
             PronounVerbAgreement::new(FstDictionary::curated()),
         );
     }


harper-core/src/linting/pronoun_verb_agreement.rs
```diff
@@ -190,6 +190,15 @@ where
...
         let verb_tok = toks.last()?;
 
+        // If the verb is immediately followed by a hyphen, it's a compound word prefix
+        // (e.g. "co-founded"), not a standalone verb. Skip it.
+        if let Some((_, after)) = ctx
+            && let [next_tok, ..] = after
+            && next_tok.kind.is_hyphen()
+        {
+            return None;
+        }
```
hippietrail, line 200: There's a method for this that seems to have gotten merged recently in `expr_linter.rs`: `followed_by_hyphen()` by the way. Probably not very discoverable.

mvanhorn:
Dug into this. The root cause is that "co" in "co-founded" gets tagged as a verb_lemma, so the pattern matches "he" + "co" as pronoun + verb. The current fix catches this in `match_to_lint_with_context` by checking for a following hyphen.

Moving the check into the verb_lemma closure isn't possible since closures only see the current token, not neighbors. The other option would be preventing the tagger from marking compound prefixes as verbs, but that's a bigger change in the dictionary/tagger.

Would you prefer I explore the tagger-level fix, or is there another approach you had in mind?

hippietrail:
> Dug into this. The root cause is that "co" in "co-founded" gets tagged as a verb_lemma,

Oh you're right. This is because a) the dictionary is case-folded - but a PR to change that is being worked on and b) "CO" is marked as a verb!
```
CO/ONgV             # US state (Colorado)
```
I think we can safely get rid of that `/V`: [knitting] Abbreviation of [cast on]

I've just removed it in the next round of dictionary curation I've started. Thanks for pointing me in the right direction. I stupidly only checked the "co" entry in the dictionary.

elijah-potter:
@hippietrail, would you link your PR here so we can mark it as superseding this one?

hippietrail:
> @hippietrail, would you link your PR here so we can mark it as superseding this one?

https://github.com/Automattic/harper/pull/3008

elijah-potter:
#3023 supersedes this. Thanks for letting us know about this, @mvanhorn!

hippietrail:
> #3023 supersedes this. Thanks for letting us know about this, @mvanhorn!

I should mention that I didn't verify the fix.

PR #3009, opened 2026-03-23
Fix false positive for 'co-founded' in PronounVerbAgreement linter

nmurrell07:
## Summary

Fixes incorrect dictionary tag for "co" by removing an erroneous `I` marker from its part-of-speech annotation.

## The Bug

The word `co` was incorrectly tagged with an `I` flag in the dictionary definition:

```diff
- co/~NSI(dE
+ co/~NS(dE
```

This extra tag likely causes incorrect grammatical analysis or spell-check behavior for this word (e.g., treating it as a different part of speech than intended).

## What This Change Does

Removes the `I` character from the dictionary entry's tag string. The remaining tags (`~NS(dE`) appear to be the correct annotation for "co" in Harper's linguistic format.

**Note:** Without access to the original issue thread (API returned 404), I cannot confirm exactly what semantic meaning the `I` flag represents or which specific bug this resolves. Based on common dictionary tag conventions, it may indicate an incorrect part-of-speech classification.

## Verification

To verify this fix:

1. **Test spell-check behavior** for sentences containing "co" to ensure no false positives/negatives
2. **Check grammar analysis** if Harper provides grammatical parsing - confirm "co" is now classified correctly
3. **Run existing test suite** to ensure no regressions in dictionary-related tests

```bash
# Example verification command (adjust based on project structure)
cargo test --package harper-core
```

## Limitations

- Original issue context unavailable (404 errors from GitHub API)
- Cannot confirm the exact semantic meaning of the removed `I` flag without documentation or maintainer input
- No specific failing test case identified to demonstrate before/after behavior

If this change doesn't resolve the intended bug, please share more details about what behavior was expected vs. observed.

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -18059,7 +18059,7 @@ clxvi
 clxvii
 cm/~
 cnidarian/NgS
-co/~NSI(dE
+co/~NS(dE
 coach/~NgSVdG
 coachload/NS
 coachman/N0g


hippietrail:
The meanings of the dictionary annotation flags are in `annotations.json`. The `#` field of each entry is a brief description, and the optional `//` field can contain an additional comment of any kind.

> co/~NSI(dE
- `co` - the word
- `/` - delimits the word from the annotation flags
- `~` - denotes that a word is common
- `N` - denotes that the word is used as a noun (not necessarily exclusively)
- `S` - tells Harper to generate another word from this word by adding `-s` or `-es` following certain orthographic rules. Normally to generate a plural in the case of a noun and a third-person singular present form of a verb. But as the dictionary originated as one from a spellchecker where the only job of the annotation flags was to enable affix compression, there are quote a few uses where they just lead to the creation of extra words that coincidentally end is `-s` or `-es` but are not necessarily related to the main word itself. This is probably a factor in this issue.
- `I` - denotes that can word has at least one pronoun sense. Formerly the `8` flag was used for this as mentioned in the previous version of the comment. Either way `co` is only a pronoun in the most marginal way and this should be removed.
- `(` - denotes that the word can be used as a prefix as well as a word in its own right. (Prefixes that don't also occur as words don't belong in `dictionary.dict` at this time.)
- `d` - tells Harper to generate another word from this entry by adding `-ed`. Normally to generate the past tense and past participle of regular verbs. But as with `-S` above it is historically used to derive words unrelated to the main word. Here it results in the word `coed`, which should have its own independent entry.
- `E` - I have to look this one up ... This is also a leftover from the affix compression heritage of the file. It generates a dew word by adding the `dis-` prefix. Here it generates `disco`, which should also have its own independent entry.

Issue #2077, opened 2025-10-13
Canvas Clipping Issue


lukasmwerner:
**Describe the bug**
When trying to answer a question in Canvas' Quiz system, harper's suggestion popups seem to be rendered clipped

**To Reproduce**
1. Open canvas Quiz
2. Enter grammatical error
3. Open harper suggestion

**Platform**
Helium on MacOS (Chrome Extension v0.67.0), Canvas LMS (Oregon State)

<img width="332" height="109" alt="Image" src="https://github.com/user-attachments/assets/a80c01f9-6711-413d-afd8-c517279e093d" />

elijah-potter:
Hmm... I'm not sure if we _could_ resolve this. It looks like the text input is inside an `<iframe />`. What are its dimensions?

lukasmwerner:
Yeah it appears to be in an iframe. My current iframe is of height 214, and width 1038. I'm also not currently sure on how to resolve this either.

PR #3010, opened 2026-03-23
Improve embedded form support in iframes

k4sperski:
# Issues

Related: [Discussion #2990](https://github.com/Automattic/harper/discussions/2990)

Depends on: [PR #2998](https://github.com/Automattic/harper/pull/2998) for the full end-to-end behaviour on pages that replace the body/DOM after load.

# Description

## User Impact

This PR improves Harper's behaviour in embedded third-party forms in the chrome extension.

Before this patch:
- Harper could be enabled on the top-level page
- the embedded iframe would still not lint unless its own host was explicitly enabled
- even when the iframe was eligible, a textarea revealed after user interaction could still be missed by the initial scan

After this patch:
- an enabled parent page can allow linting in an unset embedded iframe
- popup-style `www.` keys continue to work for parent-origin fallback
- explicit iframe overrides still take precedence over inherited parent state
- a dynamically revealed iframe textarea is attached when the user focuses it

## Technical Details

This PR makes two focused changes that are needed for the end-to-end embedded-form flow:

1. Parent-origin inheritance for iframe lint requests:
If an iframe host does not have an explicit stored setting, Harper now falls back to the top-level tab host and allows linting when that parent domain is enabled.

    This remains a narrow lookup change:
    - inheritance only applies when the iframe host is unset
    - explicit iframe opt-in still wins when the parent is disabled
    - explicit iframe opt-out still wins when the parent is enabled
    - this does not introduce suffix matching, registrable-domain matching, or a broader domain inheritance model

2. Focus-based attachment for lazily revealed textareas:
Some embedded form providers only insert or reveal the actual textarea after the user interacts with the form. To cover that flow, the content script now also attaches to eligible textareas (`HTMLTextAreaElement`) when they receive focus.

I also aligned the background read path with the popup's existing `www.` behaviour. When the parent page is `www.example.com`, background lookups now also check the popup-style `example.com` key during reads.

# Demo

One concrete example is the embedded Greenhouse form on [emnify page](https://www.emnify.com/careers/openings/job-detail?gh_jid=4805742101&jobtitle=Ausbildung%20zum%20Fachinformatiker%20(m%2Fw%2Fd)).

In this example part of the manual-entry flow is revealed only after user clicks `Enter manually` causing textarea to appear after Harper's normal initial scan window, which is why the focus-based attachment fallback matters here.

# How Has This Been Tested?

I verified this locally with extension builds and Playwright coverage. The new iframe inheritance coverage includes:
- parent enabled, child unset
- parent enabled through a `www.` parent host when the stored key follows the popup's existing `www.`-stripped pattern
- parent enabled, child explicitly disabled
- parent disabled, child unset
- parent disabled, child explicitly enabled

The dynamic textarea regression verifies that Harper still attaches after the textarea is inserted or revealed after the normal initial scan window.

Firefox notes:
- manually verified on the emnify/Greenhouse flow when combined with [PR #2998](https://github.com/Automattic/harper/pull/2998)
- the new iframe Playwright specs work with Chrome only because Firefox extension-state seeding is not reliable in the current test harness

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/background/index.ts
+++ b/packages/chrome-plugin/src/background/index.ts
@@ -62,8 +62,8 @@ chrome.runtime.onInstalled.addListener((details) => {
 	}
 });
 
-chrome.runtime.onMessage.addListener((request, _sender, sendResponse) => {
-	handleRequest(request).then(sendResponse);
+chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
+	handleRequest(request, sender).then(sendResponse);
 
 	return true;
 });
@@ -137,12 +137,15 @@ async function enableDefaultDomains() {
 
 enableDefaultDomains();
 
-function handleRequest(message: Request): Promise<Response> {
+function handleRequest(
+	message: Request,
+	sender?: chrome.runtime.MessageSender,
+): Promise<Response> {
 	console.log(`Handling ${message.kind} request`);
 
 	switch (message.kind) {
 		case 'lint':
-			return handleLint(message);
+			return handleLint(message, sender);
 		case 'getConfig':
 			return handleGetConfig(message);
 		case 'setConfig':
@@ -202,8 +205,11 @@ function handleRequest(message: Request): Promise<Response> {
 }
 
 /** Handle a request for linting. */
-async function handleLint(req: LintRequest): Promise<LintResponse> {
-	if (!(await enabledForDomain(req.domain))) {
+async function handleLint(
+	req: LintRequest,
+	sender?: chrome.runtime.MessageSender,
+): Promise<LintResponse> {
+	if (!(await shouldLintForRequest(req, sender))) {
 		return { kind: 'lints', lints: {} };
 	}
 
@@ -216,6 +222,39 @@ async function handleLint(req: LintRequest): Promise<LintResponse> {
 	);
 	const unpackedBySource = Object.fromEntries(unpackedEntries) as UnpackedLintGroups;
 	return { kind: 'lints', lints: unpackedBySource };
+}
+
+async function shouldLintForRequest(
+	req: LintRequest,
+	sender?: chrome.runtime.MessageSender,
+): Promise<boolean> {
+	if (await enabledForDomain(req.domain)) {
+		return true;
+	}
+
+	if (await isDomainSet(req.domain)) {
+		return false;
+	}
+
+	const parentDomain = getParentDomain(sender);
+	if (parentDomain == null || parentDomain === req.domain) {
+		return false;
+	}
+
+	return await enabledForDomain(parentDomain);
+}
+
+function getParentDomain(sender?: chrome.runtime.MessageSender): string | null {
+	const url = sender?.tab?.url;
+	if (url == null) {
+		return null;
+	}
+
+	try {
+		return new URL(url).hostname;
+	} catch {
+		return null;
+	}
 }
 
 async function handleGetConfig(_req: GetConfigRequest): Promise<GetConfigResponse> {
--- a/packages/chrome-plugin/src/background/index.ts
+++ b/packages/chrome-plugin/src/background/index.ts
@@ -137,10 +137,7 @@ async function enableDefaultDomains() {
 
 enableDefaultDomains();
 
-function handleRequest(
-	message: Request,
-	sender?: chrome.runtime.MessageSender,
-): Promise<Response> {
+function handleRequest(message: Request, sender?: chrome.runtime.MessageSender): Promise<Response> {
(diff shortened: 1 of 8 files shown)


k4sperski:
One thing I want to flag explicitly is the scope of the new `focusin` fallback. The bug here is that some embedded forms reveal the real `HTMLTextAreaElement` only after interaction, so scan-based attachment misses it. This fixes that, but the current implementation applies to focused `HTMLTextAreaElements` generally, not only iframe specific flows. I think that is a reasonable trade-off, but it feels like a maintainer call on scope rather than a strict correctness issue. I’m happy to adjust it!

k4sperski:
Another good example found here on [applications form on careers page](https://clickup.com/careers?ashby_jid=7ac59a7d-eee4-437d-b793-fb54a3a85da2)

k4sperski:issue/handle-embedded-forms
Screenshot 2026-03-23 at 17 46 34

Automattic:master
Screenshot 2026-03-23 at 17 47 03

elijah-potter:
Awesome! I wonder if this will also address #2077?

packages/chrome-plugin/src/contentScript/index.ts
```diff
@@ -56,6 +56,37 @@ const keepAliveCallback = () => {
 
 keepAliveCallback();
 
+function getTextareaReasons(element: HTMLTextAreaElement, requireVisible: boolean): string[] {
```
elijah-potter, line 63: It seems to me that this function returns the possible reasons for Harper to skip a text area. Would you mind making that more clear via a doc-comment?

packages/chrome-plugin/src/background/index.ts
```diff
@@ -538,12 +579,33 @@ function formatDomainKey(domain: string): string {
 	return `domainStatus ${domain}`;
 }
 
+function getDomainLookupCandidates(domain: string): string[] {
+	const withoutWww = domain.replace(/^www\./, '');
+	return withoutWww === domain ? [domain] : [domain, withoutWww];
+}
+
+async function getStoredDomainStatus(domain: string): Promise<boolean | undefined> {
```
elijah-potter, line 593: I looks like you're normalizing the input to this function via the `getDomainLookupCandidates` function so that you can test multiple possible domains that a site might be registered under. I think this is smart, but I'd like to see it become more visible. Can you document the behavior for this function in a doc-comment?

I understand that much of the existing code lacks sufficient documentation. It's a problem I'm cracking down on.

packages/chrome-plugin/src/background/index.ts
```diff
@@ -218,6 +226,39 @@ async function handleLint(req: LintRequest): Promise<LintResponse> {
 	return { kind: 'lints', lints: unpackedBySource };
 }
 
+async function shouldLintForRequest(
+	req: LintRequest,
+	sender?: chrome.runtime.MessageSender,
```
elijah-potter, line 231: Good stuff. This feels explicit and self-documenting.

elijah-potter (changes requested):
This looks great! Overall, your work is quite legible, but I'd like to see some more explicit documentation. Just a few tweaks and this will be ready to merge.

As for the callback, it's a small enough change that I don't think it really matters if its in scope. It doesn't meaningfully adjust the amount of effort it takes to review the PR. Keep it.

k4sperski:
Hi @elijah-potter! Thank you for the review. By “scope,” I meant that the new fallback `focusin` behaviour applies to all `HTMLTextAreaElements`, not just those within the `iframe` context where the issue was originally discovered. That is not necessarily a bad thing, since it casts a wider net for textareas the initial scans may have missed, but I wanted to call it out in case introducing such broader behaviour is not desired.

~I'll take a closer look at https://github.com/Automattic/harper/issues/2077 and see if this is still an issue that needs addressing.~
^ never mind, if I am not mistaken, this tool is not something I can easily get my hands on outside of the edu system...

elijah-potter:
> I'll take a closer look at https://github.com/Automattic/harper/issues/2077 and see if this is still an issue that needs addressing.
> ^ never mind, if I am not mistaken, this tool is not something I can easily get my hands on outside of the edu system...

Yeah. Once we push out your fix, I'll reach out to some students who use Harper to confirm whether it fixed the issue.

> Hi @elijah-potter! Thank you for the review. By “scope,” I meant that the new fallback focusin behaviour applies to all HTMLTextAreaElements, not just those within the iframe context where the issue was originally discovered. That is not necessarily a bad thing, since it casts a wider net for textareas the initial scans may have missed, but I wanted to call it out in case introducing such broader behaviour is not desired.

Oh yes, that's fine.

elijah-potter (approved):
Thanks for making those changes. This looks great!

PR #3012, opened 2026-03-23
fix: case-insensitive same-form check in SimplePastToPastParticiple

### LINKED ISSUE #3011: Use the past participle lost instead of Lost when using compound tenses or passive voice.
Image

> Elite Universities Have Lost Their Way

The `SimplePastToPastParticiple` linter wants to change "Lost" to "lost".

In fact "lost" is both past tense *and* past participle. So it must not be checking for that, which is odd since most English verbs have the same form for both, just like this. The fact that it's suggesting "lost" in all lowercase suggests that it's using `Suggestion::Replace` instead of `Suggestion::replace_with_match_case`

saschabuehrle:
Fixes #3011

The preterite-to-participle filter was comparing the source text against the dictionary form case-sensitively. Verbs whose past participle is identical to the simple past (lose \u2192 lost/lost, teach \u2192 taught/taught, etc.) would still get flagged when capitalized because "Lost" != "lost" passed the filter.

Switched to `eq_ignore_ascii_case` so the comparison matches how `get_past_participle_for_preterite` itself does its lookup. Added regression tests for the exact case from the issue.

\u2014 saschabuehrle

### DIFF
--- a/harper-core/src/linting/simple_past_to_past_participle.rs
+++ b/harper-core/src/linting/simple_past_to_past_participle.rs
@@ -515,7 +515,7 @@ mod tests {
     }
 
     #[test]
-    fn dont_flag_have_lost() {
+    fn dont_flag_have_lost_issue_3011() {
         assert_no_lints(
             "Elite Universities Have Lost Their Way",
             SimplePastToPastParticiple::default(),


harper-core/src/linting/simple_past_to_past_participle.rs
```diff
@@ -513,4 +513,20 @@ mod tests {
             SimplePastToPastParticiple::default(),
         );
     }
+
+    #[test]
+    fn dont_flag_have_lost() {
```
hippietrail: It's a good idea to append the GitHub issue number to test functions that address specific bug reports.

harper-core/src/linting/simple_past_to_past_participle.rs
```diff
@@ -73,7 +73,7 @@ impl ExprLinter for SimplePastToPastParticiple {
 
         if let Some(past_participle) = IrregularVerbs::curated()
             .get_past_participle_for_preterite(&simple_past)
-            .filter(|pp| pp != &simple_past)
+            .filter(|pp| !pp.eq_ignore_ascii_case(&simple_past))
```
hippietrail, line 76: Perfect! I was confused thinking "didn't I make that linter" and "I created the `eq_ignore_ascii_case` methods and I know to always use them" so thought that couldn't be the problem. (And I was too tired to look.)

But I was totally wrong - I definitely made that boo-boo!

PR #3028, opened 2026-03-25
fix(chrome-ext): persist lint rule changes from settings page

### LINKED ISSUE #1822: Won't remember my Rules on Firefox
**Describe the bug**
Every time I change my a rule from "Default" and then close the tab, it doesn't save. If I open my settings again the rule will be reverted back from "On" or "Off" to "Default." It seems to remember when I change my settings though.

**To Reproduce**
1. Open settings
2. Change any rule (such as ACoupleMore) to "On" or "Off"
3. Close tab
4. Open settings again, Rule is still set to "Default."

**Expected behavior**
When change a Rule from default the program remembers it. 

**Platform**
Firefox Extension

queelius:
## Summary

- **Bug:** Toggling a lint rule (Default/On/Off) in the browser extension settings page never persisted. The UI updated visually but no `storage.set()` call was made, so changes were lost on page reload.
- **Root cause:** The rules `<Select>` used `on:change` (Svelte event directive syntax) to handle value changes. The `Select` component is a Svelte 4 legacy-mode component that does **not** forward DOM events (it has no `on:change` shorthand or `createEventDispatcher`). As a result, the handler was silently ignored.
- **Fix:** Replace `on:change` with `onchange` (Svelte 5 attribute syntax), which passes through `$$restProps` to the native `<select>` element where it fires correctly.

This is a one-line change. Other settings (dialect, activation key, etc.) already work because they use `bind:value` on the Select, which goes through the component's `export let value` prop. Only the rules section used `on:change`.

Fixes #1822

## Test plan

- [ ] Open the extension settings page
- [ ] Toggle any lint rule from "Default" to "On" or "Off"
- [ ] Close and reopen the settings page — the rule should remain as set
- [ ] Verify that `chrome.storage.onChanged` fires when toggling a rule (it did not before this fix)
- [ ] Build passes: `pnpm --filter chrome-plugin build` exits 0
- [ ] Linting passes: `biome check` reports no errors
- [ ] Rust tests pass: `cargo test` all green

### DIFF
--- a/packages/chrome-plugin/src/options/Options.svelte
+++ b/packages/chrome-plugin/src/options/Options.svelte
@@ -433,7 +433,7 @@ async function removeWeirpack(id: string) {
               <Select
                 size="md"
                 value={configValueToString(value)}
-                on:change={(e) => {
+                onchange={(e) => {
                   lintConfig[key] = configStringToValue(e.target.value);
                 }}
               >


elijah-potter:
Since this is such a simple and high-impact change, don't worry about adding integration tests or anything.

elijah-potter:
I believe the error in CI was spurious. I'm going to run it again.

queelius:
Thanks for taking the time to review and verify. Glad it was useful.

Issue #2956, opened 2026-03-18
Leaky abstraction of `CharString` in two places.


hippietrail:
Our `CharString` type is actually a type alias for `SmallVec`.

This is respected everywhere in the codebase except two places:

- `harper_core::linting::spell_check::SpellCheck::uncached_suggest_correct_spelling()` uses `.to_smallvec()` to convert a `&&[char]` into a `CharString`
- `harper_core::spell::rune::attribute_list::AttributeList::apply_replacement()` also creates `replaced_segment` and `remove` using `.to_smallvec()`

This seems like a code smell to me. I'm not much more than a year into learning Rust though, so maybe this is normal?

I found it because I'm experimenting with upgrading `CharString` form a type alias to a newtype to make the code more expressive by enabling us to use `==` in many places we currently use very long-winded `.eq_ignore_ascii_case_chars()` and `.eq_ignore_ascii_case_str()`

elijah-potter:
> I found it because I'm experimenting with upgrading CharString form a type alias to a newtype to make the code more expressive by enabling us to use == in many places we currently use very long-winded .eq_ignore_ascii_case_chars() and .eq_ignore_ascii_case_str()

Please go for it! I'm quite interested in making that happen. I chose not to make it its own type at the time because it would have taken more time than I was interested in allocating towards it. If you feel strongly about it, do it.

hippietrail:
> > I found it because I'm experimenting with upgrading CharString form a type alias to a newtype to make the code more expressive by enabling us to use == in many places we currently use very long-winded .eq_ignore_ascii_case_chars() and .eq_ignore_ascii_case_str()
> 
> Please go for it! I'm quite interested in making that happen. I chose not to make it its own type at the time because it would have taken more time than I was interested in allocating towards it. If you feel strongly about it, do it.

I ran up against limits of my Rust knowledge, especially type system. I got `PartialEq` working and then found the implicit uses in hash keys that are case sensitive with the default hash function. Adding a case insensitive hash function worked but I was worried that other implicit stuff might break due to becoming case insensitive. At that point I thought I could either investigate each point that happens carefully or splitting into two `CharString` types, one for case sensitive and one for insensitive. But I wasn't sure how enforceable that would be with my level of Rust knowledge.

If I kept the work I'll publish it as a draft PR. In the meantime I implemented the concise comparison function names and delegates that you probably saw.

hippietrail:
[...]

PR #3032, opened 2026-03-26
refactor: introduce `CharSlice` with `==`

hippietrail:
# Issues 
Supersedes #2956 

# Description

Or maybe it just complements #2956 
I misunderstood several related concepts around `CharString` and `PartialEq` and assumed the former could be leveraged to provide `==` to be used instead of the long-winded `eq_ascii_ignore_case` methods.

But it turns out we could achieve that via a newtype wrapping `&[char]` which wasn't what `CharString` does.

This PR implements this and propagates it through the low-hanging fruit in the codebase where it can be used.

There's probably more places I didn't spot, some of which might involve some other changes, hopefully simplifying.

Although I could get `==` working against `&[char]`, `&str`, and `[char]`, I couldn't get it to work against `String` or `&String` for reasons I don't fully understand.

I'm not 100% convinced this is a definite win as it didn't immediately have repercussions though as much of the codebase as I had anticipated. But it should all be zero cost.

The main change is that `get()` has been added alongside `get_ch()` and `get_str()` on `Token`, which returns a newtype wrapping `&[char]` that allows us to use `==` and other places that do comparison, such as `match`, with the goal of making linter code easier to read and reason about.

So please go ahead and critique any and every aspect of this idea!

# How Has This Been Tested?

`cargo test` passes 100%

# Checklist


harper-core/src/char_slice.rs
```diff
@@ -0,0 +1,93 @@
...
+        *self == &other[..]
+    }
+}
+
+/// Case-insensitive comparison with a string slice, assuming the right-hand side is lowercase ASCII.
+/// Only normalizes the left side to lowercase and avoids allocations.
+impl PartialEq<&str> for CharSlice<'_> {
+    fn eq(&self, other: &&str) -> bool {
+        let mut chit = self.0.iter();
+        let mut strit = other.chars();
```
86xsk: These variable names are hard to understand. I would personally rename `strit` to `other` or `other_iter`, but not too sure about `chit`. Maybe `self_iter` or `self_chars`?

Or maybe `chit` -> `a_chars`, `strit` -> `b_chars`?

harper-core/src/char_slice.rs
```diff
@@ -0,0 +1,93 @@
...
+impl<'a> CharSlice<'a> {
+    pub fn new(slice: &'a [char]) -> Self {
+        CharSlice(slice)
+    }
+
+    /// Get the underlying character slice.
+    pub fn as_slice(&self) -> &'a [char] {
+        self.0
+    }
+}
```
86xsk: I think it might be a good idea to use `pub(crate)` for new features when possible. It avoids making any long-term stability promise for the code before it has had time for refinement.

harper-core/src/char_slice.rs
```diff
@@ -0,0 +1,93 @@
...
+            match (c, s) {
+                (Some(c), Some(s)) => {
+                    if c.to_ascii_lowercase() != s {
+                        return false;
+                    }
+                }
+                (None, None) => return true,
+                _ => return false,
+            }
+        }
```
86xsk:
```suggestion
        chit.map(char::to_ascii_lowercase).eq(strit)
```

Maybe? (I haven't tested too thoroughly, but `cargo test` passes successfully.)

harper-core/src/char_slice.rs
```diff
@@ -0,0 +1,93 @@
...
+/// Case-insensitive comparison with a character slice, assuming the right-hand side is lowercase ASCII.
+/// Only normalizes the left side to lowercase and avoids allocations.
+impl PartialEq<&[char]> for CharSlice<'_> {
+    fn eq(&self, other: &&[char]) -> bool {
+        self.0.len() == other.len()
+            && self
+                .0
+                .iter()
+                .zip(other.iter())
+                .all(|(a, b)| a.to_ascii_lowercase() == *b)
```
86xsk:
```suggestion
        self.0
            .iter()
            .map(char::to_ascii_lowercase)
            .eq(other.iter().copied())
```

Slightly shorter alternative.

86xsk (commented):
I'm concerned about using case-insensitive comparison for the `==` operator since it's contrary to the behavior of `==` for all other string-like types.

It would make it impossible to tell whether code like `a == "Something"` is using case-sensitive comparison unless you knew the type of `a`. I also think rules like lowercasing the left side only are almost guaranteed to lead to mistakes and time wasted debugging, especially for those not already familiar with the codebase.

hippietrail:
> I'm concerned about using case-insensitive comparison for the `==` operator since it's contrary to the behavior of `==` for all other string-like types.

It shouldn't be thought of as a string type, but as a word type or text type which is specifically case-insensitive because that's what you want when comparing words the majority of the time. I should've put `NoCase` in its name though as I had it at one point when I was trying to get it to work via `CharString` before realizing that a new newtype was the proper way to do it.

Perhaps a name like `NoCaseText` Or `CaselessText` or just `Caseless`. Or since its full name isn't needed very often maybe even `CaseInsensitive` or `CaseInsensitiveCharSlice`?

> It would make it impossible to tell whether code like `a == "Something"` is using case-sensitive comparison unless you knew the type of `a`. I also think rules like lowercasing the left side only are almost guaranteed to lead to mistakes and time wasted debugging, especially for those not already familiar with the codebase.

Perhaps but before I started working on this line we had a lot of allocations and conversions and lowercase copies because people didn't seem to fully realize it's possible to compare different string types in-place via iterators. The idea is to make it natural to just use the methods that do it that way automagically. The right side should always be a literal or come from a literal, which is what's done a ton in the linting process. It is trivial to add lowercasing both sides but that would actually hide potential bugs of people using it for other than what it's for.

----
I'm going to update the PR with it renamed to `CaseInsensitiveCharSlice` in a few minutes. I'm not going to rename the file yet to keep the diffs easy to follow. Not sure a bulky name will be as good for the filename as for the type name though?

Thanks a lot for the thoughtful review!

harper-core/src/char_slice.rs
hippietrail: Hmm I picked those names when I did the original implementation of `.eq_ignore_ascii` case because I wasn't used to the Rust method names and behaviours and it helped me reason about the plumbing to get it working. The left site iterates over the `char`s of the slice, which the left side always is. The right side iterates over a `&str` a `char` at a time. I guess `char_chars` and `str_chars` or `char_iterator` or `chars_iterator` and `str_char_iterator` would be most descriptive but both doubling the word "char" and having it only once both seem poor in different ways?

harper-core/src/char_slice.rs
hippietrail: I had to finesse the original implementation quite a bit to get it to work last year but you have a much better understanding of functional Rust than me. I chatted with an AI about it and your version is a clear win. So I've used it in the original implementation in `char_string.rs` as well Thanks!

harper-core/src/char_slice.rs
hippietrail: I tried this but it causes a problem in `token.rs` where I provide the way to get the `CharSlice` directly via delegation instead of having to manually step through `.span.` every time.

harper-core/src/char_slice.rs
hippietrail: And more Rusty! Thanks.

hippietrail:
> I also think rules like lowercasing the left side only are almost guaranteed to lead to mistakes and time wasted debugging, especially for those not already familiar with the codebase.

I learned how to use `debug_assert!` to enforce the right side is lowercase in debug builds now too.

PR #3049, opened 2026-03-27
feat: in 1 day from now→in one day/one day from now

hippietrail:
# Issues 
N/A

# Description

I read or heard "in five years from now" and thought "that's redundant" and the Google AI agrees.
So I made a linter that suggests both:
- in five years
- five years from now

I realized it's extremely similar to the `AfterLater` linter so started by copying it. If we find a third similar pattern it will probably be worthwhile converting them to be generated by a macro.

# How Has This Been Tested?

Unit tests using sentences found on GitHub.

# Checklist

### DIFF
--- a/harper-core/src/expr/duration_expr.rs
+++ b/harper-core/src/expr/duration_expr.rs
@@ -13,8 +13,8 @@ impl Expr for DurationExpr {
         }
 
         let units = WordSet::new(&[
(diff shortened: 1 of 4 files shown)


elijah-potter (approved):
> I realized it's extremely similar to the AfterLater linter so started by copying it. If we find a third similar pattern it will probably be worthwhile converting them to be generated by a macro.

Good call. For now though, this PR looks good to me.

PR #3053, opened 2026-03-27
feat: support Quarto qmd files as markdown

### LINKED ISSUE #2492: Support .qmd suffix as way to identify Markdown files
The Quarto publishing system (https://quarto.org/) uses files ending with .qmd. They're basically Markdown, so just using normal Markdown parser for those should work fine.

Thank you!

itamarst:
# Issues 
Fixes #2492 

# Description
The Quarto scientific publishing system using `.qmd` suffix for its markdown files. Support identifying those as markdown.

# How Has This Been Tested?
Manually tested output on markdown `.qmd` file and same file with different suffix, and old harper-cli against the `.qmd` file. Only this branch on a `.qmd` correctly identified markdown headings.

# Checklist

- [x] I have added tests to cover my changes (or rather, it has the same amount of testing as current file detection, which is to say none. But I did test manually!)

### DIFF
--- a/harper-ls/src/backend.rs
+++ b/harper-ls/src/backend.rs
@@ -393,7 +393,7 @@ impl Backend {
                 }
             }
             "mail" => Some(Box::new(PlainEnglish)),
-            "markdown" => Some(Box::new(Markdown::new(markdown_options))),
+            "markdown" | "quarto" => Some(Box::new(Markdown::new(markdown_options))),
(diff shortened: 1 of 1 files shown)


itamarst:
Will do.

itamarst:
It's not clear to me if anyone ever actually sends a language ID string other than "markdown", but they might (was looking at https://panache.bz/user-guide/lsp for guidance) so added "quarto".

PR #3072, opened 2026-03-31
refactor(core)!: remove unused argument/member

86xsk:
# Description

Removes an unused generic/argument/struct-member in `ProperNounCapitalizationLinter`.

(Cherry-picked from #2630, with adjustments.)

# How Has This Been Tested?

- `cargo test`

# Checklist


harper-core/src/linting/lint_group/mod.rs
```diff
@@ -592,9 +592,7 @@ impl LintGroup {
 
         out.merge_from(weir_rules::lint_group());
         out.merge_from(phrase_set_corrections::lint_group());
-        out.merge_from(proper_noun_capitalization_linters::lint_group(
-            dictionary.clone(),
```
hippietrail, line 497: I'm guessing the dictionary is passed to this module so that it would be able to check for proper nouns that don't get capitalized, like `iPhone` and that either didn't pan out as envisaged or that part didn't end up getting implemented?

86xsk: I'm guessing it's the former since the linter seems to get the capitalizations from `proper_noun_rules.json`, with the dictionary not being involved at all.

The linter seems to only be concerned with multi-word proper nouns, so I think cases like `iPhone` would be outside of its scope (that would be more in the ballpark of `SpellCheck` or `OrthographicConsistency` afaik). Looking through some of the entries in the json, I'm not sure how the use of a dictionary would be helpful, since the terms all seem unambiguous in terms of correct capitalization.

Maybe the initial idea was to pull these multi-word entries from the dictionary rather than having them in a json?

elijah-potter:
For the record, I think @hippietrail is right. When I implemented this module, I expected to use a dictionary. Obviously, it didn't pan out that way.

PR #3074, opened 2026-04-01
perf(core): avoid conversion between string and char array

86xsk:
# Description

Avoids back-and-forth conversion between string and char array in `PhrasalVerbAsCompoundNoun`.

I don't honestly see this making any meaningful performance difference.

(Cherry-picked from #2630)

# How Has This Been Tested?

- `cargo test`

# Checklist

### DIFF
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -57,7 +57,7 @@ impl PhrasalVerbAsCompoundNoun {
         document: &Document,
         i: usize,
         token: &crate::Token,
-    ) -> Result<(String, Confidence), Why> {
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -57,7 +57,7 @@ impl PhrasalVerbAsCompoundNoun {
         document: &Document,
         i: usize,
         token: &crate::Token,
-    ) -> Result<(String, Confidence), Why> {
+    ) -> Result<(Vec<char>, Confidence), Why> {
         // It would be handy if there could be a dict flag for nouns which are compounds of phrasal verbs.
```
hippietrail, line 61: Oh it would be worth adding a comment here that such a dict flag has been added: `/` - not sure if it's propagated through the system to be usable yet though.

PR #3075, opened 2026-04-01
fix: await Google Docs bridge replaceText before closing suggestion popup

### LINKED ISSUE #2995: Cannot Click to Apply Suggestions in Google Docs
**Describe the bug**
In Google Docs, you cannot click the suggested action to replace the erroneous text with the correct version automatically, it instead does nothing.

**To Reproduce**
1. Create a new document in Google Docs.
2. Write anything with an error.
3. Try to click the suggested action.
4. Nothing happens.

**Expected behavior**
The erroneous text should be replaced with the proper text, like in other websites.

**Screenshots**

Image

**Platform**
Firefox

**Additional context**
Only tested in Firefox with a brand new document and no other changes.

abhishekDeshmukh74:
# Issues

Closes #2995

# Description

Fixes a race condition where clicking a suggestion in Google Docs did nothing — the erroneous text was never replaced.

**Root cause:** When the user clicks a suggestion button, `applySuggestion()` dispatches a bridge request to the Google Docs main-world script. Previously this was fire-and-forget (`void Promise.resolve(...)`), so `refocusClose()` ran immediately — moving focus away from Google Docs' text event iframe. By the time the bridge handler resumed (after an `await`), the synthetic `paste` event dispatched to `iframe.contentDocument.activeElement` targeted the wrong element, and Google Docs ignored it.

**Fix (3 files):**
- `Box.ts` — `applySuggestion` now returns `void | Promise<void>` to support async replacements
- `computeLintBoxes.ts` — `replaceGoogleDocsValue` is now `async` and **awaits** the bridge `replaceText` call instead of discarding the promise
- `SuggestionBox.ts` — The suggestion button click handler now **awaits** `applySuggestion()` before calling `refocusClose()`, ensuring the paste event fires while Google Docs' text event iframe still has focus

# Demo

N/A — This is a browser-extension race condition that requires a Google Docs environment. The fix ensures correct async ordering so the paste event reaches Google Docs before focus is moved.

# How Has This Been Tested?

Manually verified the async ordering logic. The fix is a straightforward race condition resolution — ensuring `await` before `refocusClose()` so the bridge paste event fires while Google Docs' text event iframe still has focus.

# Checklist


elijah-potter:
In my testing, this PR doesn't fix the problem. Is there anything you need from me to make sure we can get this across the finish line?

elijah-potter:
I'm commenting once more to hopefully put this back on @abhishekDeshmukh74's radar.

elijah-potter:
Since I have not heard from the PR author, I am going to close it.

Issue #3019, opened 2026-03-24
Suggestions for (at least) `SpellCheck` and `SplitWords` should be combined


hippietrail:
Here's a fairly typical typo for me. (My touchtyping seems to be degrading over the decades.)

<img width="505" height="297" alt="Image" src="https://github.com/user-attachments/assets/db051407-12d9-4399-b73d-493e94a1a4d0" />

In this case a `SpellCheck` suggestion would've **_surely_** suggested the correct word, but I only got a less-than-useful suggestion from `SplitWords`. Sometimes the opposite happens.

In the past I'm pretty sure we used to see suggestions from multiple linters at once, at least in the Chrome extension. I assume this was disabled due it often being confusing and unhelpful.

But in some instances it is definitely the right thing to. Splitting words not in the dictionary and suggesting similarish single words from the dictionary for words not in the dictionary, are really similar when you think about it. There are fairly likely some other linters that might belong in this set too.

I can think of a few ways this might be improved:

- Maybe just changing the code that removes overlapping lints is enough? Some tweaking of lint priorities might be needed.
- Maybe a "layer" needs to be created where these two are handled in a special way that's somewhat apart from the other linters in some way.
- Maybe they need to be combined into a single linter somehow. This is actually easy to do a quick solution for by making a new "wrapper" linter that calls these two and manually combines the suggestions. I recently put out a PR that wraps the `SpellCheck` linter with one that doesn't flag username mentions by just filtering out suggestions with `@` in the right place. This is probably only suitable for a mockup.

Other ideas?

hippietrail:
<img width="611" height="312" alt="Image" src="https://github.com/user-attachments/assets/25edde66-25b1-4206-9d74-10088677a717" />

<img width="779" height="299" alt="Image" src="https://github.com/user-attachments/assets/dbc26de2-fa5b-4c08-89a4-c8213fd8dbe5" />

<img width="838" height="300" alt="Image" src="https://github.com/user-attachments/assets/bba1b646-ccb0-4ed8-9307-98ed4b268fc1" />

hippietrail:
Besides `SpellCheck` vs `SplitWords` competing for suggestions instead of cooperating, another great pair just came up: `ExpandMinimum` (Weir) and `ExpandMemoryShorthands` (Rust)

hippietrail:
[...]

PR #3081, opened 2026-04-02
feat: unpossible etc. → impossible

hippietrail:
# Issues 
N/A

# Description

Fixes words using the wrong negative prefix:
- unsecure → insecure
- insecured → unsecured

This is invoked for all unknown (OOV) words in a document, the same words that `SpellCheck` and a few other linters are invoked on.

Ideally it would not clash with `SpellCheck` and other linters all offering suggestions, which depends on #3019 being addressed.

This PR also adds a new `SingleTokenPattern` called `CustomToken` which is a generic single token pattern that accepts a closure for custom matching logic. Pretty much like `.then()` in `SequenceExpr`.

# How Has This Been Tested?

Unit tests using sentences found on GitHub.

# Checklist


harper-core/src/linting/wrong_negative.rs
```diff
@@ -0,0 +1,126 @@
...
+
+pub struct WrongNegative<D: Dictionary + 'static> {
+    pattern: CustomToken<fn(&Token, &[char]) -> bool>,
+    dict: D,
+}
+
+impl<D: Dictionary + 'static> WrongNegative<D> {
+    pub fn new(dict: D) -> Self {
+        Self {
+            pattern: CustomToken::new(looks_negative_but_oov),
```
elijah-potter: You should be able to pass `looks_negative_but_oov` directly, without a struct to wrap it. This is because of the implementation of `SingleTokenPattern` in `./harper-core/src/patterns/mod.rs`.

hippietrail: I think I tried to achieve that initially but I still don't fully intuit Rust's function pointers very well. Fixed now.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -4664,6 +4664,15 @@ Suggest:
 
 
 
+Lint:    WordChoice (127 priority)
+Message: |
+    3349 | But the rest offended her—and inarguably, because it wasn’t a gesture but an
```
elijah-potter: It looks like there are a few false-positives, likely due to some incomplete dictionary entries. "Inarguably" *is* a word, from what I can tell.

elijah-potter (changes requested):
Would it be possible to address those false-positives?

harper-core/tests/text/linters/The Great Gatsby.snap.yml
hippietrail: You're right I overlooked that. "Uncommunicable" is also a legit word so I've added both to the dictionary. Thanks!

elijah-potter (approved):

PR #3093, opened 2026-04-02
fix: don't flag "to" before "only" in ToTwoToo

### LINKED ISSUE #3088: False Positive: 'to' -> 'too' preceeding 'only' (`ToTooTwo`)
**What got flagged?**
`to`

**Why is this incorrect?**
`to` should be used in this circumstance. 

**Example of correct usage:**
```md
The dog was restricted to only one can of food.
```

joaquinhuigomez:
# Issues

Closes #3088

# Description

Adds "only" to the exclusion list in the `ToTooAdverb`, `ToTooAdjectiveEnd`, and `ToTooAdjectivePunct` sub-linters. The word "only" (meaning "merely" / "just") is not a degree modifier, so "to only" should not be flagged as a `too` candidate.

# How Has This Been Tested?

Two new unit tests: full sentence ("The dog was restricted to only one can of food.") and bare fragment ("limited to only").

# Checklist

### DIFF
--- a/harper-core/src/linting/to_two_too/to_too_adverb.rs
+++ b/harper-core/src/linting/to_two_too/to_too_adverb.rs
@@ -17,7 +17,11 @@ impl Default for ToTooAdverb {
         let expr = SequenceExpr::default()
             .t_aco("to")
             .t_ws()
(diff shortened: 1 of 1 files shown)


elijah-potter:
It looks like you need to run `cargo fmt`.

elijah-potter (approved):
I went ahead and formatted it for ya. Thanks for the patch!

PR #3095, opened 2026-04-03
fix: skip PronounVerbAgreement for hyphenated compounds

### LINKED ISSUE #3006: False positive: "co-founded" triggers `PronounVerbAgreement`

mvanhorn:
## Summary

Fixes false positive where "co-founded" triggers PronounVerbAgreement because "co" is recognized as a pronoun.

## Why this matters

In sentences like "he co-founded Social Chain", the tokenizer splits "co-founded" into separate tokens. The rule then matches "co" (pronoun) + "founded" (verb) and suggests changing "co" to "cos". But "co" here is a prefix, not a standalone pronoun.

## Changes

`harper-core/src/linting/pronoun_verb_agreement.rs` (line 188): Added an early return in `match_to_lint_with_context` that checks whether the character immediately following the pronoun token is a hyphen. If so, the word is part of a hyphenated compound and should not be treated as a pronoun for agreement purposes.

## Testing

- "he co-founded Social Chain" -- no longer flagged
- "he walks" -- still correctly unflagged (correct agreement)
- "I walks" -- still correctly flagged (disagreement)
- "co works here" (without hyphen) -- still flagged if applicable (no hyphen = standalone pronoun)

Fixes #3006

### DIFF
--- a/harper-core/src/linting/pronoun_verb_agreement.rs
+++ b/harper-core/src/linting/pronoun_verb_agreement.rs
@@ -185,6 +185,13 @@ where
         let pron_tok = &toks[0];
         let is_3psg = pron_tok.kind.is_third_person_singular_pronoun();
 
+        // Skip when the pronoun is part of a hyphenated compound (e.g. "co-founded").
+        // The tokenizer splits "co-founded" into separate tokens, but "co" as a prefix
+        // is not acting as a pronoun.
+        if pron_tok.span.end < src.len() && src[pron_tok.span.end] == '-' {
+            return None;
+        }
+
         let verb_tok = toks.last()?;
 
         if let Some((before, _)) = ctx


harper-core/src/linting/pronoun_verb_agreement.rs
```diff
@@ -185,6 +185,13 @@ where
         let pron_tok = &toks[0];
         let is_3psg = pron_tok.kind.is_third_person_singular_pronoun();
 
+        // Skip when the pronoun is part of a hyphenated compound (e.g. "co-founded").
+        // The tokenizer splits "co-founded" into separate tokens, but "co" as a prefix
+        // is not acting as a pronoun.
```
hippietrail, line 190: There's a method to check for exactly this. It was probably added after this linter was written: `followed_by_hyphen` in `harper-core/src/linting/expr_linter.rs`

hippietrail (changes requested):

harper-core/src/linting/pronoun_verb_agreement.rs
mvanhorn: Thanks for pointing that out. `followed_by_hyphen` checks whether the token AFTER the entire matched expression is a hyphen (i.e., `context.after[0]`). Here the check needs to be on the first token within the match (the pronoun) being immediately followed by a hyphen in the source text, since the hyphen sits between the matched pronoun and the next character, not after the full expression.

Happy to refactor if you see a cleaner way to wire it in though.

hippietrail: > Thanks for pointing that out. `followed_by_hyphen` checks whether the token AFTER the entire matched expression is a hyphen (i.e., `context.after[0]`). Here the check needs to be on the first token within the match (the pronoun) being immediately followed by a hyphen in the source text, since the hyphen sits between the matched pronoun and the next character, not after the full expression.

Oh my apologies! I obviously didn't read your PR deeply enough and thought I saw a pattern I'd used several times.

Oh I see this is the weird "cos" one now. I thought we removed the pronoun sense since it seems to be an invented pronoun that never took off. Too late here for me to check though... (-:

mvanhorn: No worries at all! Yeah, the 'cos' case is an edge case for sure. Happy to remove it from the pronoun list if you think it's not worth keeping. Let me know if there's anything else you'd like changed.

hippietrail: Hmm it seems to get mentioned online. It's still in `dictionary.dict` - I think I'll just mark all the neopronouns in there for now.

mvanhorn: Sounds good. Let me know if there's anything else you'd like changed on this PR, or if it's ready for approval.

hippietrail: It turns out the only other neopronoun in our dictionary so far is "em". I have no idea if it's intentional. Anyway I've gone ahead and requested a review from the repo owner.

PR #3103, opened 2026-04-03
fix(harper-tex): mask additional LaTeX math environments

### LINKED ISSUE #3101: [LaTeX]  Environments such as align, eqnarray, gather etc. are not recognized as equations
**Describe the bug**
Environments such as `align`, `eqnarray`, `gather` etc. are not masked by the LaTeX parser, causing false positives on math content.

`align` requires the `amsmath` package but it is ubiquitous in math and physics writing.

**To Reproduce**
If one writes an equation using `align`
```
\begin{align}
f(x) = x^2
\end{align}
```
`f` is flagged as a typo.

**Expected behavior**
Harper should ignore the text inside align (and other mentioned) environments.

**Screenshots**

Image

**Platform**
Noticed in neovim, but it seems like a more general issue with `harper-ls`.

jurajkl:
Extends the masker to recognise align, align*, gather, gather*, multline, multline*, flalign, flalign*, alignat, alignat*, eqnarray, eqnarray*, math, and displaymath in addition to the existing equation environment. Previously, content inside these environments was grammar-checked, causing false positives.

# Issues 

Fixes  https://github.com/Automattic/harper/issues/3101.

# Description

Extends equation_at_cursor in harper-tex to use a shared is_math_env helper that covers all common amsmath environments.
Fix was done with assistance from an agent/LLM.

# Demo

# How Has This Been Tested?

Added unit tests masks_equation_star_env and masks_align_env covering   
  the new environments.
  
# Checklist


elijah-potter (approved):
This is perfect! I would not have implemented this any other way. Thanks!

PR #3104, opened 2026-04-03
feat/obsidian web lint styling

stanley-910:
# Issues
N/A

# Description
This PR adds an Obsidian plugin setting to switch between the default squiggly lint underline and an experimental web-style lint treatment.

It introduces a new `Use Web-Style Lints` toggle in the Obsidian settings tab, persists that preference in plugin state, and applies the selected rendering style by attaching either `harper-squiggly-style` or `harper-web-style` classes to lint decorations.

The web-style option uses a straight `2px` underline plus a lightly transparent background tint, while the existing squiggly underline remains the default behavior.

# Demo
image

When disabled:
- Lints render with the original squiggly underline

When enabled:
- Lints render with a straight underline
- Lints also render with a lightly transparent background highlight

# How Has This Been Tested?
- Performed a self-review of the Obsidian plugin changes in:
  - `packages/obsidian-plugin/src/HarperSettingTab.ts`
  - `packages/obsidian-plugin/src/State.ts`
  - `packages/obsidian-plugin/src/lint.ts`
- Verified that the new setting is wired through persisted plugin settings and used when generating lint decoration classes
- Verified that both lint rendering styles remain registered in the CodeMirror theme and are selected via CSS class
- Ran `pnpm --filter obsidian-plugin run build`
- Build reached Vite and module resolution, but local workspace/package setup still has existing environment issues around monorepo dependency/build configuration

# Checklist


stanley-910:
Just backward compatibility mostly, I didn't want to push a change which enabled a change globally for all users unless everything checked out. If you'd like I can push out a change with it enabled by default. Should I keep the setting to toggle between the two styles?

Issue #2901, opened 2026-03-10
Spaces flagged for indents and comments in LaTeX files


gregosimo:
**What got flagged?**
\begin{itemize}
    \item Hello world.         % This is a test
\end{itemize}

**Why is this incorrect?**
The indentation before \item is flagged as [NoFrenchSpaces]
Additionally, the spacing between "world." and "% This" is flagged as [Spaces]

**Example of correct usage:**
The above {itemize} block is normal usage and shouldn't be flagged. LaTeX is pretty good at figuring out spaces, so I think a lot of flagged issues in the source code end up just fine in the compiled code.

PR #3106, opened 2026-04-03
Fix indentation handling in harper-tex

jurajkl:
# Issues 

Partly fixes https://github.com/Automattic/harper/issues/2901
the spacing before/after comment symbol `%` issue is not addressed yet.

# Description

Fixes false-positives for indentation such as with
```
\begin{itemize}
    \item very long item that spills over
    into multiple lines
    \item other item
\end{itemize}
```
practically ignores all spaces after newline.
Prepared with assistance of an agent/LLM.

# Demo

# How Has This Been Tested?

Verified that the indentations no longer get flagged on an active tex file.
Added unit tests.

# Checklist

### DIFF
--- a/harper-tex/src/masker.rs
+++ b/harper-tex/src/masker.rs
@@ -22,6 +22,8 @@ impl harper_core::Masker for Masker {
 
             if matches!(c, '%') {
                 actions.push_back(CursorAction::PushMaskAndIncBy(1));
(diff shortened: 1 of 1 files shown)


jurajkl:
I've added a test showing that FrenchSpaces are still flagged in regular text, seems to work without any hiccups now.
Thanks!

elijah-potter (approved):
Thanks for making that change. It looks great!

PR #3113, opened 2026-04-04
feat: convenient store→convenience store

hippietrail:
# Issues 
N/A

# Description

Tried to correct the common hypercorrection "convenient store" in Asian English to "convenience store".
It turns out that "convenient store" is quite commonly used as a legitimate phrase in its own right in tech.

The linter examines the context around the phrase against lists of words that might sway the interpretation one way or the other.

# How Has This Been Tested?

There's a bunch of unit tests gathered from GitHub and the rest of the internet to cover both the common mistake and the legitimate non-mistake uses, plus a few that are too hard to judge at least by the linter.

# Checklist


harper-core/src/linting/convenient_store.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        eprintln!(
+            "🚨🚨 '{}' {} clues hint at it being a mistake, {} clues hint at it being legit.",
+            toks[0].get_str(src),
+            noun_clues,
+            adj_clues
+        );
+
+        if adj_clues > noun_clues {
+            eprintln!(
```
elijah-potter: Please remove these.

harper-core/src/linting/convenient_store.rs
```diff
@@ -0,0 +1,381 @@
...
+            "Mini glass bottle collection of convenient store.",
+            ConvenientStore::default(),
+        );
+    }
+
+    // Real sentences for which it's hard to determine if it's a true or false positive.
+
+    #[test]
+    fn dont_flag_frame_dump() {
+        // 🚨
```
elijah-potter: These emoji's seem superfluous.

harper-core/src/linting/convenient_store.rs
hippietrail: That's sloppy of me. I have a distinct memory of removing all this detritus. I mustn't've staged after that edit. All gone now.

harper-core/src/linting/convenient_store.rs
```diff
@@ -0,0 +1,356 @@
...
+
+pub struct ConvenientStore {
+    expr: SequenceExpr,
+}
+
+// If any of these are in the 'words' around the match, it looks more like a mistake for "convenience store".
+const NOUN_CLUES: &[&str] = &[
+    "11",
+    "24",
+    "7",
```
elijah-potter, line 15: Is it possible to make this work _without_ these wordlists?

hippietrail: > Is it possible to make this work _without_ these wordlists?

I don't think so but I'm very open to suggestions. I actually attempted this months ago thinking it would be easy when I found out it's unexpectedly (to me) common in its non-error sense in computing. I put it aside then because I'm loathe to use such lists too.

I think more and more of the less-low-hanging fruit Harper doesn't tackle yet will need lexical/semantic context blacklists/whitelists to detect.

elijah-potter (approved):

PR #3131, opened 2026-04-08
fix(core): detect "to" before degree words followed by nouns mid-sentence

### LINKED ISSUE #3130: "to many/much/few" not flagged when followed by a noun mid-sentence
**The Error**
The ToTooDegreeWords linter only catches "to many/much/few" at clause end or before punctuation. It misses the common mid-sentence pattern where a noun follows the degree word.

**Examples (2-3):**
1. "I ate to many cookies." → should suggest "too many"
2. "There is to much water in the glass." → should suggest "too much"
3. "There is to little time to finish." → should suggest "too little"

**References**
- Cambridge Dictionary: "too" as degree modifier before many/much/few/little + noun
- Merriam-Webster: "too" used as a function word to indicate excess

**Potential Edge Cases**
When "to" is a preposition: "give to many charities", "connected to many servers", "listen to many podcasts" these are valid and should not be flagged.

draphy:
# Issues
Fixes #3130

# Description
The `ToTooDegreeWords` linter only caught "to many/much/few" at clause end or before punctuation. It missed the common mid-sentence pattern where a noun follows the degree word (e.g. "I ate to many cookies").

Added a second pattern (`to + degree_word + noun`) with guards to prevent false positives when "to" is a preposition:
- `prepositional_preceder` check (e.g. "connected to many servers")
- Common to-taking verbs list (e.g. "give to many charities")

Also added "little" to the degree words list.

# How Has This Been Tested?
11 new tests: 5 true positives (to many cookies, to much water, to few options, to little time, to many bugs mid-sentence) and 6 false positive guards (give, listen, talk, donate, connected). All 70 ToTwoToo tests pass. Snapshot tests pass.

# Checklist


harper-core/src/linting/to_two_too/to_too_degree_words.rs
```diff
@@ -1,35 +1,150 @@
...
+    "give",
+    "gives",
+    "gave",
+    "given",
+    "giving",
+    "talk",
+    "talks",
+    "talked",
+    "talking",
+    "speak",
```
hippietrail: It would be helpful if these were in alphabetical order. it would make it easier to scan visually to see if a word is included or not.

hippietrail (changes requested):

harper-core/src/linting/to_two_too/to_too_degree_words.rs
draphy: Good catch, sorted. Thanks

harper-core/src/linting/to_two_too/mod.rs
```diff
@@ -476,4 +476,78 @@ mod tests {
...
+    #[test]
+    fn no_lint_donate_to_many_causes() {
+        assert_no_lints("They donate to many causes.", ToTwoToo::default());
+    }
+
+    #[test]
+    fn no_lint_connected_to_many_servers() {
+        assert_no_lints("The app is connected to many servers.", ToTwoToo::default());
+    }
 }
```
hippietrail, line 617: Here's some tests I made just by searching for "to many/much/few/little" with Google on GitHub. These are the failing ones but may not be related to your PR. I couldn't test against only the sublinter you worked on due to #3134 
```
    #[test]
    fn correct_data_points() {
        assert_suggestion_result(
            "Possibly to few (unique) data points / predictions.",
            ToTwoToo::default(),
            "Possibly too few (unique) data points / predictions.",
        );
    }

    #[test]
    fn correct_python() {
        assert_suggestion_result(
            "Question for checking python import times and warning if to much of regression.",
            ToTwoToo::default(),
            "Question for checking python import times and warning if too much of regression.",
        );
    }

    #[test]
    fn dont_flag_copy_files_to_many_hosts() {
        assert_no_lints("Copy files to many hosts in parallel", ToTwoToo::default());
    }

    #[test]
    fn ignore_convert_opcode_to_little_endian() {
        assert_no_lints("Small script to convert opcode to little endian format", ToTwoToo::default());
    }

    #[test]
    fn ignore_semver() {
        assert_no_lints("SemVer is a false promise that appeals to many developers.", ToTwoToo::default());
    }

    #[test]
    fn ignore_to_expired() {
        assert_no_lints("I didn't know it was going to expired.", ToTwoToo::default());
    }

    #[test]
    fn ignore_to_little_endian() {
        assert_no_lints("I was working on gpt package in go and I didn't realize that the author had handled conversion to little endian when writing header.", ToTwoToo::default());
    }

    #[test]
    fn ignore_too_theme_happy() {
        assert_no_lints("2018: we may have gotten a bit too theme-happy.", ToTwoToo::default());
    }
```
So yeah some may not pertain to your PR and some might be edge cases that are too specific or include too much bad grammar to interpret correctly. In other words I'm not saying they have to all pass, I'm just supplying them in case they might help make the PR even more solid.

harper-core/src/linting/to_two_too/to_too_degree_words.rs
```diff
@@ -1,35 +1,150 @@
...
+    expr::{AnchorEnd, Expr, FirstMatchOf, SequenceExpr},
+    patterns::{SingleTokenPattern, prepositional_preceder},
 };
 
 use super::{ExprLinter, Lint, LintKind, Suggestion};
 use crate::linting::expr_linter::Chunk;
 
+/// Verbs that commonly take "to" as a prepositional complement.
+/// Guards against false positives like "give to many charities".
+const TO_TAKING_VERBS: &[&str] = &[
```
elijah-potter, line 13: Including a big list like this feels, for lack of a better word, off. Would it be possible to replace it with a metadata query or a POS lookup?

You're much closer to the problem, so feel free to push back. This is a valuable change, so if there's no way to simplify things, I'm happy to merge this as-is.

elijah-potter (changes requested):

harper-core/src/linting/to_two_too/to_too_degree_words.rs
hippietrail: > Including a big list like this feels, for lack of a better word, off. Would it be possible to replace it with a metadata query or a POS lookup?

FWIW, I actually commented on the issue for this (https://github.com/Automattic/harper/issues/3130#issuecomment-4203372166) thinking it would be very hard to detect with heuristics and logic. I don't think I mentioned that I thought it would necessitate a list of individual cases but I was definitely thinking it. I wasn't expecting a PR addressing it to follow immediately after the issue (-:

> You're much closer to the problem, so feel free to push back. This is a valuable change, so if there's no way to simplify things, I'm happy to merge this as-is.

draphy: Explored with POS queries before but VerbData has no subcategorization flag. "give" and "ate" are tagged identically, so any POS-based check either suppresses the original fix or false-positives on every verb+to pattern.

I see two options:

1. Keep the word list here (current approach). Self-contained, same pattern as PrepositionalPrecederPattern. Simplest for now.
2. Add a takes_to_complement flag to VerbData. Reusable if future linters need the same check, but touches more files for a single one today.

I'd lean toward keeping the list until a second linter needs the same data. Happy to go either way though.

elijah-potter:
Unfortunately, it seems like 8% PR results in an 8% slowdown in the hot path for Harper. Would it be possible to move more logic out of the hot loop?

draphy:
> Unfortunately, it seems like 8% PR results in an 8% slowdown in the hot path for Harper. Would it be possible to move more logic out of the hot loop?

@elijah-potter Could you share your bench setup so I can reproduce it locally?

hippietrail:
I believe that in PR #2630 there's some `WordSet` optimization that @86xsk made. Or maybe that's one of the pieces that they already factored out into one of their smaller PRs that already got merged?

I don't know why but I can't request them to do a PR review through the usual system on GitHub...

86xsk:
> I believe that in PR #2630 there's some `WordSet` optimization that @86xsk made. Or maybe that's one of the pieces that they already factored out into one of their smaller PRs that already got merged?

I think that was merged in #3017. At a glance, I don't think it would make a difference here since it's solely related to caching `ModalVerb`, which doesn't seem to be used here.

(Now that I think about it, it might be possible to optimize `WordSet` by comparing word IDs instead of char slices.)

86xsk:
> > Unfortunately, it seems like 8% PR results in an 8% slowdown in the hot path for Harper. Would it be possible to move more logic out of the hot loop?
> 
> elijah-potter Could you share your bench setup so I can reproduce it locally?

Not who you're asking, but if it helps, I usually run:

`cargo bench --bench parse_essay -- --save-baseline master` on master, then

`cargo bench --bench parse_essay -- --baseline master` on the branch with changes.

(You can also prepend `RUSTFLAGS='-Cforce-frame-pointers=y'` to both commands, which might allow you to get cleaner results if you want to capture a performance profile of the benches. Just be wary that it will cause a full rebuild the first time you run it.)

draphy:
> > > Unfortunately, it seems like 8% PR results in an 8% slowdown in the hot path for Harper. Would it be possible to move more logic out of the hot loop?
> > 
> > 
> > elijah-potter Could you share your bench setup so I can reproduce it locally?
> 
> Not who you're asking, but if it helps, I usually run:
> 
> `cargo bench --bench parse_essay -- --save-baseline master` on master, then
> 
> `cargo bench --bench parse_essay -- --baseline master` on the branch with changes.
> 
> (You can also prepend `RUSTFLAGS='-Cforce-frame-pointers=y'` to both commands, which might allow you to get cleaner results if you want to capture a performance profile of the benches. Just be wary that it will cause a full rebuild the first time you run it.)

Thanks, On lint_essay I'm seeing around +2.5% on my machine, couldn't reproduce the full 8%. Tried a larger corpus (Gatsby, ~1144 "to" tokens) and the regression was more visible at ~+3.7%. Could be a hardware or setup difference on @elijah-potter end, not sure.

draphy:
Tried a few approaches to bring it down:
1. Merged the two patterns into a shared prefix with `then_any_of` branching
2. Captured the preceding word in the pattern via `SequenceExpr::optional` (like `ToTooAdjectiveEnd`), switched to plain `match_to_lint`
3. Replaced `WordSet` char comparison with `WordId` hashing per @86xsk's suggestion

All three land at the same ~+2.5% on essay.md. On Gatsby, option 2 eliminates the regression and option 3 actually makes it 6% faster than master since it benefits all 102 `WordSet` usages across the codebase.

On essay.md I've run out of ideas to move the ~2.5%. If anyone sees an angle I haven't tried, happy to dig in.

On the `WordSet` → `WordId` change, I can include it in this PR but it touches core infrastructure (`word_set.rs`, used by 102 patterns). All 5083 tests pass with it, but wanted your take on whether we should go with it. @elijah-potter @86xsk

elijah-potter (dismissed):
Coming back to this for a second look now. I do believe that the slowdown you've demonstrated is worth it in this case. 2.5% is fine. I've replicated that number on my machine. Perhaps something weird was going on elsewhere on my machine? Either way, I apologize for the delay.

harper-core/src/linting/to_two_too/to_too_degree_words.rs
```diff
@@ -28,8 +129,19 @@ impl Default for ToTooDegreeWords {
...
+        let before_noun = SequenceExpr::default()
+            .t_aco("to")
+            .t_ws()
+            .then_word_set(&["many", "much", "few", "little"])
+            .t_ws()
+            .then_noun();
+
         Self {
-            expr: Box::new(expr),
+            expr: Box::new(FirstMatchOf::new(vec![
```
hippietrail: Oh we don't need `vec!` any more. I converted these to work with iterators a month or two ago to save allocation and conversion.

harper-core/src/linting/to_two_too/to_too_degree_words.rs
```diff
@@ -41,11 +153,27 @@ impl ExprLinter for ToTooDegreeWords {
...
+        source: &[char],
+        context: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
         let to_tok = tokens
             .iter()
             .find(|t| t.get_ch(source).eq_ch(&['t', 'o']))?;
 
+        // Suppress when "to" is a preposition, not a typo for "too".
+        if let Some((before, _)) = context
+            && let Some(prev) = before.iter().rfind(|t| !t.kind.is_whitespace())
```
hippietrail: Is this looking for a non-whitespace anywhere in the 'before' context? What if there's a hyphen or some unexpected punctuation in there? Or am I not reading it right?

See if `preceded_by_word()` in `harper-core/src/linting/expr_linter.rs` does what you need.

hippietrail (changes requested):
I want to complete merging this since I see Elijah started it. I've got a question about how the context is being checked first though, and a small optimization.

harper-core/src/linting/to_two_too/mod.rs
draphy: Checked all 8. One was a real false positive and I fixed it:

* "appeals to many developers" - added `appeal*` to TO_TAKING_VERBS and a regression test.

The rest are known limitations or handled by other rules, so I left them as is.

Anything you'd like me to take further?

harper-core/src/linting/to_two_too/to_too_degree_words.rs
draphy: Done, dropped vec!, Thanks.

harper-core/src/linting/to_two_too/to_too_degree_words.rs
draphy: Good catch, you're right. Switched to preceded_by_word(). Also fixed the helper itself, mixed whitespace (like give \nto) made it read a whitespace token as the preceding word. Tests pass.

hippietrail:
Have you included all in the PR including false positives and ones the improved linter still can't fix?

In linters with subtle heuristics and logic I like to include these as tests at the bottom with `#[ignore]` including a descriptive reason. This way, future devs won't have to struggle to look for exceptions and edge cases, and when a number accrue a pattern may even emerge to help a future dev improve the linter.

(Apologies for not checking myself to see if you did this as my computer or connection or GitHub itself is a bit slow today and there's a ton of issues and PRs to look at.)

draphy:
> Have you included all in the PR including false positives and ones the improved linter still can't fix?
> 
> In linters with subtle heuristics and logic I like to include these as tests at the bottom with `#[ignore]` including a descriptive reason. This way, future devs won't have to struggle to look for exceptions and edge cases, and when a number accrue a pattern may even emerge to help a future dev improve the linter.
> 
> (Apologies for not checking myself to see if you did this as my computer or connection or GitHub itself is a bit slow today and there's a ton of issues and PRs to look at.)

Good call. Added them now. Each known false positive and unfixable case is an ignored test with its reason at the bottom of the ToTwoToo tests.

elijah-potter (approved):

PR #3146, opened 2026-04-09
compound_noun_after_det_adj: skip matches followed by a preposition

### LINKED ISSUE #3139: False positive: "get away with" is wrongly flagged to change to "getaway"
Image

> They might even get away with it.

Both "get away" and "get away with" are legitimate phrasal verbs and are correct in this context.

This is in the `CompoundNouns` linter.

joaquinhuigomez:
The `CompoundNounAfterDetAdj` linter flagged "get away" in sentences like *"they might even get away with it"* as a candidate closed compound noun ("getaway"), but the trailing "with" makes "get away with" a phrasal verb and the merge is wrong.

This patch overrides `match_to_lint_with_context` and uses the existing `followed_by_word` helper to drop matches where the next word token is a preposition, which is a strong indicator that the content-word pair is part of a phrasal verb.

Fixes #3139

### DIFF
--- a/harper-core/src/linting/compound_nouns/compound_noun_after_det_adj.rs
+++ b/harper-core/src/linting/compound_nouns/compound_noun_after_det_adj.rs
@@ -7,7 +7,7 @@ use crate::patterns::InflectionOfBe;
 use crate::{CharStringExt, TokenStringExt, linting::ExprLinter};
 
 use super::{Lint, LintKind, Suggestion, is_content_word, predicate};
-use crate::linting::expr_linter::Chunk;
+use crate::linting::expr_linter::{Chunk, followed_by_word};
 
 use crate::{Lrc, Token};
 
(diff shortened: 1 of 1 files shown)


elijah-potter:
It looks like your changes cause the linter to fail to detect some cases we need the rule to continue to work for. Is there a solution that works for all the test cases?

harper-core/src/linting/compound_nouns/compound_noun_after_det_adj.rs
```diff
@@ -65,6 +70,13 @@ impl ExprLinter for CompoundNounAfterDetAdj {
             return None;
         }
 
+        // If the two content words are followed by a preposition, the pair is
+        // almost certainly part of a phrasal verb (e.g. "get away with",
+        // "run out of") and merging them into a compound noun is wrong.
+        if followed_by_word(context, |tok| tok.kind.is_preposition()) {
```
hippietrail, line 76: This is too general. I would instead check that the two content words are "get" and "away" and check that the following word is specifically "with". Welcome to the fuzzy world of grammar checker edge cases (-:

hippietrail (changes requested):

joaquinhuigomez:
Closing — would need more thought to cover all the test cases cleanly.

PR #3147, opened 2026-04-09
Copyright linter: stop flagging "copywriting" as a misspelling

### LINKED ISSUE #3138: False positive: `Copyright` linter flags legit use of "copywriting"
Image

> One price for finance, one price for cybersecurity, one price for copywriting.

In this sentence the writer really is referring to "writing copy". I'm not sure if there's enough clues in the sentence for the linter to be able to detect that it's legit though...

joaquinhuigomez:
"Copywriting" is a legitimate English word for the profession of writing advertising or marketing copy. The `Copyright` phrase-set rule rewrote it to "copyrighting", producing false positives on sentences like *"one price for copywriting"* where the writer really meant the copy-writing trade.

This patch drops the `copywriting` -> `copyrighting` pair from the rule while keeping the other entries (`copywrite`, `copywrites`, `copywritten`, `copywrited`, `copywrote`), which remain valid corrections.

Fixes #3138

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -676,7 +676,6 @@ pub fn lint_group() -> LintGroup {
             &[
                 (&["copywrite"], &["copyright"]),
                 (&["copywrites"], &["copyrights"]),
-                (&["copywriting"], &["copyrighting"]),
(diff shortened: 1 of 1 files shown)


elijah-potter (changes requested):
Would you mind including a test using text from the original issue so that we can make sure this problem does not come up again? Thanks!

joaquinhuigomez:
Closing this out. Happy to reopen if there's interest.

PR #3152, opened 2026-04-10
PronounVerbAgreement: skip plural nouns to fix 'you guys' false positive

### LINKED ISSUE #3132: False positive: `PronounVerbAgreement` is triggered by "you guys"
Image

> Keep going please because I love you guys and want to keep watching

Apparently "guy" is a verb as well as a noun. Maybe the rule needs to be modified to check if the verb is also a noun, or maybe it needs a special case for "you" + "guys"

joaquinhuigomez:
"Guys" is tagged as both a verb (to guide/steer) and a plural noun, so `PronounVerbAgreement` matched "you" (non-3p-singular pronoun) + "guys" (3p-singular verb form) and flagged agreement.

Skip the lint when the matched verb token is also a plural noun and the pronoun is non-third-person-singular, since the noun reading dominates in those constructions.

Fixes #3132

### DIFF
--- a/harper-core/src/linting/pronoun_verb_agreement.rs
+++ b/harper-core/src/linting/pronoun_verb_agreement.rs
@@ -217,6 +217,14 @@ where
             }
         }
 
+        // When a non-3p-singular pronoun is followed by a word that is both a
+        // plural noun and a verb form (e.g. "you guys", "they colors"), the noun
+        // reading is far more likely than the verb reading.  Skip the lint to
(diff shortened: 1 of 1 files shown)


hippietrail (changes requested):
This looks good to me. Thanks.

Oh you should add at least one new unit test, since it's to fix an issue, include the issue number in the name of the test.

joaquinhuigomez:
Closing — no worries if the rule is handled differently.

PR #3025, opened 2026-03-24
perf(spell): skip redundant lowercase DFA search in FstDictionary

k4sperski:
# Issues

N/A

# Description

This change makes `FstDictionary::fuzzy_match` more efficient by skipping the redundant lowercase DFA build and FST search when the query is already lowercase.

Before this change, Harper always ran two searches:
- one for the original query
- one for the lowercased query

For lowercase input, those searches are the same, so the second one was unnecessary.

I also moved the repeated distance-merging logic into a small helper so the matching code is easier to read and maintain.

To make the change easier to measure, this PR adds a `spellcheck` benchmark in `harper-core` that covers:
- realistic prose from the existing essay sample
- typo lists split into lowercase, capitalised, and mixed-case inputs
- `fuzzy_match`
- `suggest_correct_spelling`
- `MergedDictionary` fuzzy matching with a single child dictionary

This should be a low-risk performance change. There is no user-facing behaviour change intended, and regression tests were added to confirm lowercase, capitalised, and uppercase queries still return the same matches.

# Benchmark Results

Optimisation target: `harper-core/src/spell/fst_dictionary.rs`
Methodology: 3 runs vs pre-optimisation baseline · Criterion `--sample-size 10`
 
### Time Reduction vs Baseline (%)
 
> `fm` = `fuzzy_match`, `scs` = `suggest_correct_spelling`, `mg` = `fuzzy_match_merged_dict_single_child`
> Subcategories: `essay`, `mixed` = misspelled_mixed, `lower` = misspelled_lowercase, `cap` = misspelled_capitalized
 
```mermaid
---
config:
  xyChart:
    width: 800
    height: 400
---
xychart-beta
    title "Average Time Reduction vs Baseline (%)"
    x-axis ["fm/essay", "fm/mixed", "fm/lower", "fm/cap", "scs/mixed", "scs/lower", "scs/cap", "mg/mixed", "mg/lower", "mg/cap"]
    y-axis "Time reduction (%)" 0 --> 55
    bar [43.4, 42.6, 49.3, 0.0, 42.6, 48.8, 0.9, 44.0, 48.1, 1.8]
```
 
### Throughput Gain vs Baseline (%)
 
```mermaid
---
config:
  xyChart:
    width: 800
    height: 400
---
xychart-beta
    title "Average Throughput Gain vs Baseline (%)"
[...]

PR #3153, opened 2026-04-11
feat(bench): add WASM benchmark harness and native allocation profiler

k4sperski:
# Issues
Follows up on #3025.

# Description

Adds benchmarking infrastructure to measure spell-check performance in WASM and track allocation overhead; two dimensions native Criterion benchmarks don't cover. `fuzzy_match` is cheap on modern machines, but on lower-end devices WASM is slower and allocation costlier; these tools make that overhead visible.

Validated against #3025 (`is_already_lower`):

| Metric | Lowercase | Mixed | Capitalised |
|--------|-----------|-------|-------------|
| WASM time reduction | 49.2% | 43.4% | ~0% |
| Allocation churn reduction | 44.7% | 38.9% | ~0% |

WASM time reductions track #3025's native Criterion numbers, so the optimisation carries through the WASM pipeline. Capitalised is flat because `is_already_lower` doesn't trigger on uppercase starts.

### What's added

- **Native allocation profiler** (`harper-core/examples/alloc_profile.rs`): counts allocs per word in `fuzzy_match` via a `#[global_allocator]`. No extra deps. Run with `just alloc-profile`.
- **WASM bench wrapper** (`harper-wasm/src/bench.rs`): feature-gated `PreparedWords` struct exposing `fuzzy_match` to JS. Constructor parses the word list and caches an `Arc<FstDictionary>`; the method just runs the spell-check loop. `cfg(feature = "bench")` keeps it out of production builds.
- **JS bench harness** (`harper-wasm/benches/wasm_bench.js`): Node script timing `fuzzy_match` under WASM to verify native perf wins carry through. Run with `just bench-wasm`.

### Measurement hygiene

- `PreparedWords` parses inputs and caches the dictionary once, outside `bench()`, so the timed region only contains `fuzzy_match` work, matching the native criterion pattern.
- The alloc profiler warms the `AUTOMATON_BUILDERS` thread_local before measuring, so the first case doesn't absorb its one-time lazy-init cost.
- `MAX_EDIT_DISTANCE = 3 and MAX_RESULTS = 200 are duplicated across the alloc profiler, the criterion bench, and the JS harness; a comment flags the invariant.
- `just bench-wasm` builds into `pkg-bench/`, so the shipping `pkg/` stays untouched

### WASM Time: baseline vs optimised

```mermaid
xychart-beta
    title "WASM fuzzy_match: ms/iter min (lower is better)"
    x-axis ["mixed", "lowercase", "capitalized"]
    y-axis "ms/iter" 0 --> 100
    bar [96, 84, 80]
    bar [54, 43, 80]
```

### Methodology

- baseline = early-exit guard removed, optimised = guard restored. Numbers are the min of 50 iterations (after 5 warmup iterations) measured on the default just bench-wasm build (wasm-opt on). Same word lists and parameters as the native Criterion bench

### Trade-offs

- JS harness isn't in CI, so value depends on contributors running `just bench-wasm` before merging perf-sensitive changes. CI integration would be the natural next step if this becomes a recurring gap.
- Running `bench-wasm` with the default `wasm-opt` build adds ~30s per rebuild. `DISABLE_WASM_OPT=1 just bench-wasm` skips it for faster dev iteration; the numbers that ship in this PR were measured with `wasm-opt` on.

### Follow-ups

- If the harness graduates to CI, pass `MAX_EDIT_DISTANCE` / `MAX_RESULTS` as args so `just bench-wasm` becomes the single source of truth for the duplicated constants.
- Extend `PreparedWords` to cover `suggest_correct_spelling` and merged-dict lookup when optimisation work warrants it.

# Demo
N/a

# How Has This Been Tested?

- `cargo build -p harper-wasm`: production build unaffected (no `bench` feature).
- `cargo clippy -p harper-wasm --features bench -- -Dwarnings`: clean.
- `just alloc-profile`: first case no longer inflated by thread-local init.
- `just bench-wasm`: runs end-to-end; `PreparedWords.free()` called so no wasm-bindgen leaks.

# Checklist


k4sperski:
Moving to draft, needs a bit more refinement!

justfile
```diff
@@ -818,6 +818,26 @@ grep-config query:
...
 
+# Run the native allocation profiler for spell-check operations.
+alias alloc-prof := alloc-profile
+alloc-profile:
+  cargo run --example alloc_profile -p harper-core --release
+
+# Build harper-wasm with bench support and run the WASM benchmark harness.
+# Runs wasm-opt by default to match the shipping build; set DISABLE_WASM_OPT=1
+# to skip it (faster rebuilds, non-shipping numbers).
+bench-wasm:
```
elijah-potter, line 922: I'd love to see a native bench command here (it wouldn't be complicated, just call `cargo bench`). That might make discoverability easier for new contributors.

elijah-potter (changes requested):
Beside that one small change, this looks great! Thanks!

k4sperski:
Hi @elijah-potter (nudge)! Requested changes are done, conflicts resolved and branch is up-to-date. Let me know if there is anything else you'd like to see here. Otherwise it would be nice to see this work merged :)

elijah-potter:
Thanks for the nudge @k4sperski. Sorry it took me so long to get back to this! It looks good. Thank you!

elijah-potter (approved):

PR #3154, opened 2026-04-11
fix(dict): remove stale ElasticSearch entry in favour of Elasticsearch

unlimitedsola:
# Issues 
N/A

# Description
 The dictionary contained two entries for the same word:
 - `Elasticsearch/g` (line 3153) — correct modern spelling
 - `ElasticSearch/Sg` (line 53845) — old branding

 Since `WordId` is hashed from the lowercase form, both map to the
 same key. The second entry was inserted after the first and overwrote
 the canonical spelling, causing Harper to suggest `ElasticSearch`
 instead of the correct `Elasticsearch`.

 Removes the stale `ElasticSearch/Sg` entry so the canonical spelling
 reflects Elastic's current branding.

# Demo
image

# How Has This Been Tested?
`just check`

# Checklist

- [] I have added tests to cover my changes

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -53842,7 +53842,6 @@ ESP32/Og            # microcontroller family
 ESP8266/Og          # microcontroller family
 EV/NSg              # electric vehicle
 EXIF                # metadata
-ElasticSearch/Sg
 Electron/Og         # web-tech framework for desktop apps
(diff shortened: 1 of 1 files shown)


hippietrail (approved):
Thanks!

elijah-potter (approved):

PR #3156, opened 2026-04-12
config: return defaults when LSP settings is null or non-object

### LINKED ISSUE #3151: "harper_ls::backend: Settings must be an object." when lsp server is used in kate
**Describe the bug**
`harper-ls --stdio` fails when ran inside kate

**To Reproduce**
1. Open Kate
2. Go to Settings -> Configure Kate -> LSP Client -> User Server Settings
3. Paste in ```{
    "servers": {
        "markdown": {
            "command": ["harper-ls","--stdio"],
            "rootIndicatorFileNames": [".git",".marksman.toml"],
            "url": "https://writewithharper.com",
            "highlightingModeRegex": "^Markdown$"
        }
     }
}```
4. Open a markdown file and accept the prompt

**Expected behavior**
Harper should load as the default language server for markdown with this configuration, it errors out instead

**Screenshots**
Image
Image

**Platform**
Kate LSP Client Support and likely KDevelop as well

joaquinhuigomez:
Some editors (e.g. Kate) send `null` or a non-object value as `params.settings` on `workspace/didChangeConfiguration`. The LSP spec allows this — it just means the client has no settings to push. Previously `from_lsp_config` bailed with *"Settings must be an object"*, which caused `harper-ls` to fail on startup in Kate.

Return the default config instead of erroring when the value is not a JSON object.

Fixes #3151

### DIFF
--- a/harper-ls/src/config.rs
+++ b/harper-ls/src/config.rs
@@ -89,7 +89,10 @@ impl Config {
         let workspace_root = workspace_root.as_path();
 
         let Value::Object(value) = value else {
-            bail!("Settings must be an object.");
+            // Some editors (e.g. Kate) send null or a non-object value on
+            // workspace/didChangeConfiguration.  Treat this the same as
+            // an empty settings object — return defaults rather than erroring.
+            return Ok(base);
         };
 
         let Some(Value::Object(value)) = value.get("harper-ls") else {


harper-ls/src/config.rs
```diff
@@ -89,7 +89,10 @@ impl Config {
         let workspace_root = workspace_root.as_path();
 
         let Value::Object(value) = value else {
-            bail!("Settings must be an object.");
+            // Some editors (e.g. Kate) send null or a non-object value on
```
elijah-potter, line 92: Does `Kate` send `null` specifically? If so, could we handle the `null` case directly with a `match` statement? I don't love the idea of this failing silently, which is what might happen with this implementation.

joaquinhuigomez:
Closing for now.

PR #3166, opened 2026-04-13
feat(chrome-ext): add delay option

elijah-potter:
# Issues 

This doesn't address any particular tracked issue that I know of. Rather, @JasonTheAdams suggested it.

# Description

This PR adds a "delay" option to the Chrome extension, not unlike the one that exists in our Obsidian plugin. In essence, it allows the user to configure a delay (from 0 to an infinite number of milliseconds) between when they stop typing and when Harper updates the highlights on the screen. This comes from feedback that the highlights can be distracting if they are applied _while_ a user types. When the delay is set to 0, there is no change from before this PR. The delay is set to 0 by default. 

# Demo

image

# How Has This Been Tested?

Manually + a mildly unhelpful integration test.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/ProtocolClient.ts
+++ b/packages/chrome-plugin/src/ProtocolClient.ts
@@ -60,6 +60,14 @@ export default class ProtocolClient {
 		await chrome.runtime.sendMessage({ kind: 'setDialect', dialect });
 	}
 
+	public static async getDelay(): Promise<number> {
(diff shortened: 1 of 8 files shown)


elijah-potter:
After some further discussion, we decided a small delay (about 300ms) would be better as a default.

JasonTheAdams (approved):
Thanks for this! I think this will be a nice little UX bump. 😄

PR #3169, opened 2026-04-13
fix(core): address several outstanding issues

### LINKED ISSUE #2917: Flag typos mixing up "were" and "where"
I'm sure we've all seen typos like this before and probably made them ourselves:

> where you able to make forward progress here?

The opposite probably happens too, typing `were` when it should be `where`, but that might be less common.

It might be possible to disambiguate in this context by checking the POS after the "where [pronoun]" since when it's not a mistake the next word is probably a present or past verb or an adverb and when it should be "were" the third word would be an adjective like in this example or a progressive verb form.

More real-world examples would help.

### LINKED ISSUE #2952: Flag using "find out" where "find" should be used on its own.
This is one I've been noticing maybe over the last year or two. Just now I heard a native speaker in his 50s use it in a YouTube video, which surprised me. Maybe it has a regional factor and isn't just a new thing. Or maybe it's one of the new things that older generations are also absorbing, like "invest in" → "invest into":

This is the sentence from the video:
> So I've been trying to **find out** ways to do the work that I want to do while walking.

Examples found by Googling that sound right vs wrong to me:
- ✅ And I just wonder what I can do to **find out** WHY this file got the evidence it has
- ❌ RESET reason **found out**
- ✅ **Finding out** when the bug for CameraRoll.js was introduced.
- ✅ EGit/JGit: how to **find out** if a file is tracked within a git repo
- ❌ **Find out** latest supported hugo version?
- ❌ Java application to **find out** plain text of given cipher text.
- ❌ A dhrystone benchmark script I made for **finding out** dhrystones per frequency, which can be used to find CPU capacities.
- ❌ I would like to know, how could I **find out** the number of bits read from a file using constbitstraem or BitArray?
- ✅ Find out what tooling can be used for branch coverage

It seems that "find out" when used correctly is very often followed by words like: about, how, if, what, when, whether, who, why.

There are definitely correct usages that are not followed by any of these words. They may be incorrect usages that are followed by one of them, but I haven't found any yet.

I find it hard to describe the difference between the two in my own words. This is how one AI put it that seems accurate to me:

### Find
**Definition**: To locate or discover a person or thing, often after searching.
**Usage**: Used with physical objects, people, or sometimes abstract concepts (like finding a solution).
**Examples**:
- "I can't find my keys."
- "She found a new job."
- "I found a great restaurant."

### Find Out
**Definition**: To discover, learn, or become aware of information, usually through investigation, research, or hearing it from someone else.
**Usage**: Used with information, secrets, or "that/why/how" clauses.
**Examples**:
- "I found out that my boss is quitting."
- "How did you find out about the surprise party?"
- "I need to find out what time the meeting starts."

### LINKED ISSUE #3129: Emdash interrupting dialog tags false positive
**What got flagged?**

Harper says no matching end quotes grammar error.

“It’ll be our”—she leaned to his ear—“shared secret.”
Image

Valid emdash usage for mid-dialog interruption according to Chicago manual of style for "dashes out"

**Why is this incorrect?**
The open quotes all have matching closing quotes and the text is following the Chicago manual of style.

**Example of correct usage:**
Posted above in what got flagged.

### LINKED ISSUE #3155: False negative: The misspelling "payed" is not flagged.
I was reading some YouTube comment and saw this mistake and wondered how Harper would handle it:

> He over payed in part to have the specification that the monument should be capable of "withstanding catastrophic events" met, ...

Harper doesn't flag "payed" as any error. I checked and it is actually in `dictionary.dict`
```
payed/V
```
And this probably came from an ancient Wiktionary import:
> (nautical) simple past and past participle of pay (to cover with a waterproof substance)

We have two options:
- Simply remove `payed` from the dictionary and users needing this specialized form can add it to their user dictionary.
- Write a tiny linter for it that warns something like "Unless you mean the nautical "covered with a waterproof substance", this is likely a spelling mistake for "paid".

----

Yes it should actually be "overpaid" in this specific snippet. Harper does flag "over paid" to change to "overpaid" but it only flags "overpayed" is a misspelling and does not suggest "overpaid" as a correction.

elijah-potter:
# Issues
- Fixes #2917
- Fixes #2952
- Fixes #3129
- Fixes #3140
- Fixes #3155

# Description
This batch fixes five recently reported lint issues in `harper-core`.

Changes made:
- Tightened `WereWhere` so it runs at sentence scope and only reports `where you ...` at sentence start.
- Rewrote `FindOut` as a Weir rule with tests and enabled it in the curated default config.
- Refined `MissingTo` to avoid a reduced-relative-clause false positive.
- Updated quote handling so em dash and en dash dialogue interruptions do not trigger `UnclosedQuotes`.
- Removed the dictionary entry for `payed` and added phrase-set corrections for `payed` and `overpayed`.

I also updated the snapshot fixtures affected by the rule changes.

# Demo
N/A

# How Has This Been Tested?
- `cargo fmt --all`
- `cargo test -p harper-core --lib were_where -- --nocapture`
- `cargo test -p harper-core --lib run_tests_for_find_out -- --nocapture`
- `target/debug/harper-cli test harper-core/src/linting/weir_rules/FindOut.weir`
- `cargo test -p harper-core --test linters`
- `cargo build -p harper-cli`
- `PATH="/tmp/fakebin:$PATH" bash -lc 'cd ~/Projects/harper-llm-fuzz && ./check.sh WereWhere'`
- `just check-rust`
- `just test-rust`

# Checklist

### DIFF
--- a/harper-core/default_config.json
+++ b/harper-core/default_config.json
@@ -1221,6 +1221,13 @@
 						},
 						{
 							"Bool": {
+								"name": "Payed",
+								"state": true,
+								"label": "Payed"
+							}
+						},
+						{
+							"Bool": {
 								"name": "RiseTheQuestion",
 								"state": true,
 								"label": "Rise The Question"
--- a/harper-core/default_config.json
+++ b/harper-core/default_config.json
@@ -4978,6 +4978,13 @@
 						},
 						{
 							"Bool": {
+								"name": "FindOut",
+								"state": true,
+								"label": "Find Out"
+							}
+						},
+						{
+							"Bool": {
 								"name": "GoSoFarAsTo",
 								"state": true,
 								"label": "Go So Far As To"
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -38465,7 +38465,6 @@ pay's
 payback/NmSg
 paycheck/~NgS
 payday/NgS
-payed/V
 payee/NSg
 payer/NSg
 payload/~NSg
--- a/harper-core/src/document.rs
+++ b/harper-core/src/document.rs
@@ -418,6 +418,8 @@ impl Document {
                             | Some(Punctuation::OpenRound)
                             | Some(Punctuation::OpenSquare)
                             | Some(Punctuation::OpenCurly)
+                            | Some(Punctuation::EmDash)
+                            | Some(Punctuation::EnDash)
                             | Some(Punctuation::Apostrophe)
                     );
 
--- a/harper-core/src/linting/lint_group/mod.rs
+++ b/harper-core/src/linting/lint_group/mod.rs
@@ -728,7 +728,6 @@ impl LintGroup {
         insert_expr_rule!(WasAloud, true);
         insert_expr_rule!(WayTooAdjective, true);
         insert_expr_rule!(WellEducated, true);
-        insert_expr_rule!(WereWhere, true);
         insert_expr_rule!(Whereas, true);
         insert_expr_rule!(WhomSubjectOfVerb, true);
         insert_expr_rule!(WidelyAccepted, true);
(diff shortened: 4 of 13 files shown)


hippietrail:
Is the Obsidian fail due to missing a new entry in `default_config.json`? Or just another new CI quirk?

elijah-potter:
Yup. I'm adding a more descriptive test to `harper-core` now.

PR #3170, opened 2026-04-14
fix: Add `Payed` rule entry to default config JSON

hippietrail:
# Issues 

#3169

# Description

Adds a default config entry for `Payed` to the JSON file

Let me know if this is the right / wrong way to submit a PR to a PR - or if this is something that shouldn't be done

# How Has This Been Tested?

`just test-obsidian`

# Checklist

### DIFF
--- a/harper-core/default_config.json
+++ b/harper-core/default_config.json
@@ -1221,6 +1221,13 @@
 						},
 						{
 							"Bool": {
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
Yes, this is right. Thank you!

PR #3172, opened 2026-04-14
feat: code on rust→code in rust etc.

hippietrail:
# Issues 
N/A

# Description

Nonnative English speakers mix up prepositions a lot. One noticeable one for programmers is when they use
- coded on C
- programming on Java
- written on Rust

This linter checks against all inflections of those three verbs, except a couple which cause too many false positives, and a surely incomplete list of programming languages.

It could be that the `WordSet`s used for this might cause a performance impact. If that's the case, and especially if it might be useful for other linters, we could flag the names of programming languages in `dictionary.dict` - I'm thinking of `⚙` or `⌘` as the symbol to mark them if that turns out to be the case.

# How Has This Been Tested?

A whole bunch of unit tests fished out of GitHub using Google.

# Checklist

### DIFF
--- a/harper-core/default_config.json
+++ b/harper-core/default_config.json
@@ -3559,6 +3559,13 @@
 								"state": true,
 								"label": "Aspire To"
 							}
+						},
+						{
+							"Bool": {
+								"name": "CodeInWriteIn",
+								"state": true,
+								"label": "Code In Write In"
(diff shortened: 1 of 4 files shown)


elijah-potter (approved):
Awesome! Thank you.

PR #3179, opened 2026-04-15
style: clean up some imports

86xsk:
# Description

Clean up some imports, mostly by merging and reordering.

Recreated/extracted from #2630, with the intent to (slightly) reduce its size. I initially expected this to be larger, but I realized some of the import cleanups in #2630 are questionable, and are probably best undone instead.

# How Has This Been Tested?

- `cargo check`

# Checklist


hippietrail (approved):
I like this clean style.

elijah-potter (dismissed):
I agree, this is much better. It looks like the failing check is just an example of the issue I'm trying to fix with #3328.

elijah-potter (approved):

PR #3181, opened 2026-04-15
fix(vscode): pre-select current dialect in the dialect picker

### LINKED ISSUE #2667: The dialect picker that pops up in VS Code when clicking the indicator in the status bar always starts with "US" selected
I recently added support for making the dialect indicator in the VS Code status bar clickable.

When you click it, a picker pops up that lets you choose a different dialect.

That all works but there is an "unpolished" quirk: The picker always starts with "US" selected instead of the current dialect of the document.

I tried to fix this yesterday but for some reason I can get very confused working on the VS Code extension in ways that my usual coding LLMs and agents are unable to help with.

It's possibly due to me using VS Code fork for my editor to work on it and I get the editor I code in and the editor that pops up for testing confused, especially when looking for debug info or trying to figure out why a change didn't work.

Another factor might be should I open the full Harper repo as the project to work on or only the VS Code directory of the repo?

Ai-chan-0411:
I noticed issue #2667 while exploring the extension — when you click the dialect indicator in the status bar, the quick-pick list always had **American** highlighted at the top, even if the workspace was already set to, say, British or Australian. That makes it easy to accidentally revert your dialect if you just press Enter.

The root cause is simple: `window.showQuickPick` was called without an `activeItems` option, so VS Code defaults to highlighting the first item. The fix reads the current `harper.dialect` setting before opening the picker, finds the matching `QuickPickItem`, and passes it as `activeItems` so the picker scrolls to and highlights the active dialect right away.

**Changes:** `packages/vscode-plugin/src/extension.ts` — 4 lines added to `changeDialect()`.

Fixes #2667

### DIFF
--- a/packages/vscode-plugin/src/extension.ts
+++ b/packages/vscode-plugin/src/extension.ts
@@ -204,8 +204,12 @@ async function changeDialect(): Promise<void> {
 		label: name,
 	}));
 
+	const currentDialect = workspace.getConfiguration('harper').get<string>('dialect', 'American');
+	const currentItem = dialects.find((d) => d.label === currentDialect);
+
 	const selected = await window.showQuickPick(dialects, {
 		placeHolder: 'Select Harper dialect',
+		activeItems: currentItem ? [currentItem] : [],
 	});
 
 	if (selected && typeof selected !== 'string') {


hippietrail:
Thanks so much! I wish I didn't get frustrated so easily almost every time I work on the VS Code extension. It would be awesome if you could include a little screengrab video of your modded version working.

Ai-chan-0411:
Thanks for the feedback! I appreciate it. I'll work on creating a quick screengrab showing the dialect picker pre-selecting the current dialect in action. In the meantime, the fix ensures that when you open the dialect picker, it pre-selects your currently active dialect instead of always defaulting to American. This should make the VS Code extension workflow smoother for you.

Ai-chan-0411:
Thanks for the kind words! Unfortunately I can't easily record a screengrab since I'm running on a headless Raspberry Pi without a display. But the change is straightforward — it reads the current dialect from the configuration and passes it as the activeItems parameter to showQuickPick, so the picker opens with the current dialect already highlighted instead of always defaulting to American. Happy to add any other tests or adjustments if needed!

hippietrail:
See also #3198

elijah-potter:
Superseded by #3198.

PR #3193, opened 2026-04-16
missing_preposition: skip matches where AUX is main-verb 'have'

### LINKED ISSUE #3187: False positive: `MissingPreposition` wrongly invoked for "and one of my children has dual citizenship"
Image

> ..., and one of my children has dual citizenship

joaquinhuigomez:
The rule flagged *"one of my children has dual citizenship"* as missing a preposition between "has" and "dual citizenship". The pattern matches `noun + AUX + adj + noun` regardless of whether the AUX is auxiliary `have` or main-verb `have` taking a direct object.

Since the POS tagger can't reliably distinguish the two, gate the lint lexically: if the AUX slot is a form of `have` (`has`/`have`/`had`/`having` and contractions), treat it as a main verb and skip.

Fixes #3187

### DIFF
--- a/harper-core/src/linting/missing_preposition.rs
+++ b/harper-core/src/linting/missing_preposition.rs
@@ -45,8 +45,25 @@ impl ExprLinter for MissingPreposition {
         &self.expr
     }
 
-    fn match_to_lint(&self, matched_tokens: &[Token], _source: &[char]) -> Option<Lint> {
+    fn match_to_lint(&self, matched_tokens: &[Token], source: &[char]) -> Option<Lint> {
         if matched_tokens.last()?.kind.is_upos(UPOS::ADP) {
+            return None;
+        }
+
+        // Skip when the AUX slot is a form of "have" acting as a main verb
+        // taking a direct object (e.g. "children has dual citizenship",
+        // "she has big dreams"). The AUX tagger can't disambiguate main
+        // "have" from auxiliary "have", so we do it here lexically.
+        let aux_tok = matched_tokens.get(matched_tokens.len().checked_sub(5)?)?;
+        let aux_text: String = aux_tok
+            .get_ch(source)
+            .iter()
+            .collect::<String>()
+            .to_ascii_lowercase();
+        if matches!(
+            aux_text.as_str(),
+            "has" | "have" | "had" | "having" | "hasn't" | "haven't" | "hadn't"
+        ) {
             return None;
         }
 
--- a/harper-core/src/linting/missing_preposition.rs
+++ b/harper-core/src/linting/missing_preposition.rs
@@ -6,7 +6,7 @@ use crate::expr::OwnedExprExt;
 use crate::expr::SequenceExpr;
 use crate::patterns::AnyPattern;
 use crate::patterns::UPOSSet;
-use crate::{Token, TokenStringExt};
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/missing_preposition.rs
```diff
@@ -45,11 +45,28 @@ impl ExprLinter for MissingPreposition {
...
+        let aux_tok = matched_tokens.get(matched_tokens.len().checked_sub(5)?)?;
+        let aux_text: String = aux_tok
+            .get_ch(source)
+            .iter()
+            .collect::<String>()
+            .to_ascii_lowercase();
+        if matches!(
+            aux_text.as_str(),
+            "has" | "have" | "had" | "having" | "hasn't" | "haven't" | "hadn't"
+        ) {
```
hippietrail, line 66: There's a helper for this in `char_string.rs`: `eq_any_ignore_ascii_case_str()`

You might want to include some `#[test]`s for hasn‘t/haven‘t too to see if Harper's curly-apostrophe logic covers this or if you'll have to add them explicitly.

You should always add at least one `#[test]` anyway in these kinds of linting bugfixes.

hippietrail (changes requested):
I'm not able to dig into this right now properly but I can point out a helper function you can use and please add some tests, including for the typographical apostrophe. Thanks!

elijah-potter (changes requested):
Would you mind addressing the issues highlighted by the linter in `just check-rust`? Thanks!

joaquinhuigomez:
@hippietrail @elijah-potter pushed:

- Switched to `CharStringExt::eq_any_ignore_ascii_case_str` per the inline suggestion.
- Added typographical-apostrophe variants for `hasn\u{2019}t` / `haven\u{2019}t` / `hadn\u{2019}t` (macOS smart-quote substitution emits these).
- While adding tests I hit a bug: the AUX position was indexed from `matched_tokens.len() - 5`, but the expression's trailing `AnyPattern` optionals shift that when the match captures a sentence terminator. Switched to finding the AUX by UPOS — verified with the regression test below.
- Added three tests: the issue's exact "one of my children has dual citizenship" phrasing, plus `haven't` with both straight and curly apostrophes.

`cargo fmt` and `cargo clippy` are clean for this file. `just check-rust` does surface two `Arc<SequenceExpr>` warnings, but those are pre-existing on `main` and unrelated to this PR.

harper-core/src/linting/missing_preposition.rs
```diff
@@ -54,16 +54,23 @@ impl ExprLinter for MissingPreposition {
...
+        let aux_tok = matched_tokens.iter().find(|t| t.kind.is_upos(UPOS::AUX))?;
+        if aux_tok.get_ch(source).eq_any_ignore_ascii_case_str(&[
+            "has",
+            "have",
+            "had",
+            "having",
+            "hasn't",
+            "haven't",
+            "hadn't",
+            // Typographical / curly apostrophe variants.
```
hippietrail, line 69: Smart/curly quotes are automatically handled in some code paths. So try unit tests with curls quotes without this logic and if they still pass remove this logic. If this logic is needed for the tests to pass, keep the logic in that case.

hippietrail (changes requested):
Thanks for the changes. I added a not to check if the smart-quote tests pass without specific logic or not.

PR #3194, opened 2026-04-16
spell: case-insensitive apostrophe diff for contraction suggestions

### LINKED ISSUE #3184: `I'm` is not among the spellcheck suggestions for `im`
Image

Image

People often type in all lowercase and without punctuation for efficiency or chatty style, etc. If you want to use Harper to correct these into standard orthography it misses that "im" means "I'm", the contraction for "I am".

joaquinhuigomez:
`score_suggestion`'s apostrophe-diff bonus used case-sensitive `starts_with`/`ends_with`, so lowercase "im" never matched "I'm" as a single-character apostrophe difference and missed the `-8` score bonus that surfaced it for the capitalized "Im".

Compare the slices with a case-insensitive helper and fix the off-by-one around the apostrophe position so "im" -> "I'm" works the same way "Im" -> "I'm" already did. Added a regression test.

Fixes #3184

### DIFF
--- a/harper-core/src/spell/mod.rs
+++ b/harper-core/src/spell/mod.rs
@@ -313,12 +313,19 @@ fn score_suggestion(misspelled_word: &[char], sug: &FuzzyMatchResult) -> i32 {
         score -= 5;
     }
 
-    // Promote suggestions that differ only by an apostrophe
+    // Promote suggestions that differ only by an apostrophe.
+    // Case-insensitive so lowercase forms like "im" still map to "I'm".
+    let eq_ignore_case = |a: &[char], b: &[char]| -> bool {
+        a.len() == b.len()
+            && a.iter()
+                .zip(b.iter())
+                .all(|(x, y)| x.eq_ignore_ascii_case(y))
+    };
     let check_apostrophe_diff = |longer: &[char], shorter: &[char]| -> bool {
         if let Some(pos) = longer.iter().position(|&c| c == '\'' || c == '’') {
             longer.len() - 1 == shorter.len()
-                && longer.starts_with(&shorter[..pos])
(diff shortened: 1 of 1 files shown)


harper-core/src/spell/mod.rs
```diff
@@ -313,12 +313,19 @@ fn score_suggestion(misspelled_word: &[char], sug: &FuzzyMatchResult) -> i32 {
...
+                .zip(b.iter())
+                .all(|(x, y)| x.eq_ignore_ascii_case(y))
+    };
     let check_apostrophe_diff = |longer: &[char], shorter: &[char]| -> bool {
         if let Some(pos) = longer.iter().position(|&c| c == '\'' || c == '’') {
             longer.len() - 1 == shorter.len()
-                && longer.starts_with(&shorter[..pos])
-                && longer.ends_with(&shorter[pos..])
+                && eq_ignore_case(&longer[..pos], &shorter[..pos])
+                && eq_ignore_case(&longer[pos + 1..], &shorter[pos..])
```
hippietrail, line 328: We have helpers for exactly this in `char_string.rs` such as:
- `starts_with_ignore_ascii_case_str()`
- `ends_with_ignore_ascii_case_str()`

hippietrail (changes requested):
This looks like the source of the problem, thanks!
We already have starts_with and ends_with helpers though they might be newer than the problematic logic you found. They avoid allocations and string conversions by using iterators and require the LHS to be `[char]` and the RHS to be specified in lowercase.

There's a new clippy warning too that's hitting all our PR's, needing three loops in markdown.rs changed to use `while let` - it seems CI is flagging this in your PR too.

PR #3195, opened 2026-04-16
one_of_the_singular: skip when followed by a pronoun clause

### LINKED ISSUE #3186: False positive in `OneOfTheSingular`
Image

> Yours is one of the best I’ve found.

It must be due to "best" being considered a noun in the dictionary. Better look into how true that is. Maybe just special-casing "best" will be the right fix?

joaquinhuigomez:
The rule flagged *"one of the best I've found"* as missing plural agreement on "best", because it treated the matched adj as a singular noun even though "I've found" starts a new clause modifying the preceding substantive.

When the next word-like token after the match is a pronoun (starting a relative or independent clause), the matched adj is being used as a substantive, not an out-of-agreement singular noun. Skip the lint in that case.

Fixes #3186

### DIFF
--- a/harper-core/src/linting/one_of_the_singular.rs
+++ b/harper-core/src/linting/one_of_the_singular.rs
@@ -80,6 +80,18 @@ impl<D: Dictionary + 'static> ExprLinter for OneOfTheSingular<D> {
         {
             return None;
         }
+
+        // Skip if the next word-like token is a pronoun, which starts a
+        // new clause modifying the matched tail — e.g. "one of the best
+        // I've found", "one of the best that he saw". In those cases the
+        // matched adj (here "best") is used as a substantive adjective,
+        // not an out-of-agreement singular noun.
+        let after = ctx?.1;
+        if let Some(next_word) = after.iter().find(|t| t.kind.is_word_like())
+            && next_word.kind.is_pronoun()
+        {
+            return None;
+        }
         let nounspan = nountok.span;
         let singular = nounspan.get_content(src);
         let mut plural_s = singular.to_vec();
--- a/harper-core/src/linting/one_of_the_singular.rs
+++ b/harper-core/src/linting/one_of_the_singular.rs
@@ -1,7 +1,9 @@
 use crate::{
     CharStringExt, Lint, Token,
     expr::{Expr, SequenceExpr},
-    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk, expr_linter::followed_by_word},
+    linting::{
+        ExprLinter, LintKind, Suggestion, expr_linter::Chunk, expr_linter::followed_by_word,
+    },
     spell::Dictionary,
 };
 
--- a/harper-core/src/linting/one_of_the_singular.rs
+++ b/harper-core/src/linting/one_of_the_singular.rs
@@ -1,7 +1,7 @@
 use crate::{
     CharStringExt, Lint, Token,
     expr::{Expr, SequenceExpr},
-    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk},
+    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk, expr_linter::followed_by_word},
     spell::Dictionary,
 };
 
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/one_of_the_singular.rs
```diff
@@ -80,6 +80,18 @@ impl<D: Dictionary + 'static> ExprLinter for OneOfTheSingular<D> {
...
             return None;
         }
+
+        // Skip if the next word-like token is a pronoun, which starts a
+        // new clause modifying the matched tail — e.g. "one of the best
+        // I've found", "one of the best that he saw". In those cases the
+        // matched adj (here "best") is used as a substantive adjective,
+        // not an out-of-agreement singular noun.
+        let after = ctx?.1;
+        if let Some(next_word) = after.iter().find(|t| t.kind.is_word_like())
```
hippietrail, line 90: Thanks! Here you could use `followed_by_word()` from `harper-core/src/linting/expr_linter.rs`

hippietrail (changes requested):

harper-core/src/linting/one_of_the_singular.rs
elijah-potter: @joaquinhuigomez, would you give this a shot?

joaquinhuigomez:
@hippietrail @elijah-potter pushed:

- Switched the pronoun check to `followed_by_word` from `expr_linter.rs`.
- Added regression tests for the relative-pronoun and personal-pronoun substantive-adjective cases ("one of the best that he saw" / "one of the best he saw").

Worth flagging: the original report's "I've found" case slips through both versions because the POS tagger emits `I've` / `I'm` / `I'd` as a single titlecase Word token with `pronoun: None` metadata, so neither `is_pronoun()` nor the helper catches it. I'd rather fix that on the tagger side (these contractions arguably should carry pronoun metadata) than special-case the apostrophe in this rule — happy to file a follow-up issue for that if you agree.

harper-core/src/linting/one_of_the_singular.rs
```diff
@@ -81,15 +81,12 @@ impl<D: Dictionary + 'static> ExprLinter for OneOfTheSingular<D> {
...
-        let after = ctx?.1;
-        if let Some(next_word) = after.iter().find(|t| t.kind.is_word_like())
-            && next_word.kind.is_pronoun()
-        {
+        // Skip when followed by a pronoun-led clause. The matched
+        // singular form is being used as a substantive (noun phrase
+        // head) and the pronoun starts a relative or independent
+        // clause — e.g. "one of the best he saw", "one of the best
+        // that they tried".
+        if followed_by_word(ctx, |t| t.kind.is_pronoun()) {
```
hippietrail, line 91: While I was running this over my test corpus to see if it stopped flagging anything the old version did, or started flagging anything that's not an error, I was reading the comment and came to a realization.

This is not detecting the real problem in the bug report: Yours is one of the best I’ve found.

It's nothing to do with the following pronoun as can be seen from:
- ❌ This is one of the pizza **I** told you about.
- ✅ This is one of the pizzas **I** told you about.

The real problem is that there is something grammatically/linguistically special about "best". In the dictionary it's marked as a noun because it can act as a noun but lexically it's an adjective.

There are various conditions under which adjectives can be used as nouns and I think we should deal only with the one we have run into for now, which is the **_superlative_** form.

Instead of checking the following word in `match_to_lint` it should modify the pattern/`Expr`:
```
            (t.kind.is_non_possessive_noun() || t.kind.is_adjective())
                && !t.kind.is_preposition() // "in" etc.
                && !t.kind.is_pronoun() // "who" etc.
                && !t.kind.is_superlative_adjective() // "best", "worst", etc.
                && !t.get_ch(s)
                    .eq_any_ignore_ascii_case_str(&["ah", "few", "first", "said", "uh"])
```
Sorry I didn't spot this earlier!

One of the patterns I've noticed with AI coding at, least for a grammar checker, is that they don't seem to have the 'imagination' to find good tests or to extrapolate and generalize from one known false-positive pattern to more than extremely close variants. When I'm working on rules manually I use Google with `OR` operators to try to find examples of mistakes, and I test new versions of linters against a corpus of English I've collected. Mine is a collection of readmes from GitHub repos.

hippietrail (changes requested):
I realized a bit late that the PR mischaracterized the problem and therefore came up with an invalid fix that just coincidentally passed the specific test.

PR #3198, opened 2026-04-17
fix(vscode): pre-select current dialect in the dialect picker

LanceYing2004:
https://github.com/user-attachments/assets/5b3aed1a-df5d-45f6-a3e3-76490be7f184

…(#2667)

# Issues 
Previously, clicking the dialect indicator in the status bar would always show the picker with `"American"` selected by default, even when the active document was already using another dialect such as British or Canadian. This made the UI inconsistent with the actual document state.

# Description
This PR fixes the VS Code dialect picker so that it opens with the **current document dialect** highlighted, instead of always starting with **US/American**.

- added a helper to read the current dialect from the active document configuration
- replaced the simple quick pick call with a `QuickPick` instance so the current dialect can be preselected/highlighted when the menu opens

# Demo
Video demo has been attached.

# How Has This Been Tested?
I tested this change manually in the VS Code Extension Development Host.

**I specifically verified:**

- {Austrilian, British, Canadian, Indian} is highlighted when the active document dialect is {Austrilian, British, Canadian, Indian}
- Changing the dialect still works as expected
- Canceling the picker leaves the current dialect unchanged

# Checklist


hippietrail:
See also: #3181

elijah-potter:
This PR supersedes #3181.

elijah-potter (approved):
I think this looks great! Thank you so much for contributing to Harper for your class. I hope to hear from you all again!

elijah-potter:
I've requested the view of a second pair of eyes. @mcecode, if you don't time to take a look, no worries. Just let me know.

mcecode (commented):
LGTM, just one question for OP.

packages/vscode-plugin/src/extension.ts
```diff
@@ -204,9 +204,8 @@ async function changeDialect(): Promise<void> {
 		label: name,
 	}));
 
-	const selected = await window.showQuickPick(dialects, {
-		placeHolder: 'Select Harper dialect',
-	});
+	const currentDialect = getCurrentDialect();
```
mcecode, line 207: Wouldn't doing

```ts
const currentDialect = workspace.getConfiguration('harper').get<string>('dialect', '');
```

work the same here since it's what's used in `updateDialectStatusBar` and we just want what's pre-selected to be the same as what's in the status bar?

LanceYing2004: That's honestly a great point, so basically I was aiming to preselect the dialect that is effectively applied to the current document, rather than only the global/user setting.

My thinking is that, in the possible edge case where the dialect is overridden at the workspace or resource level, the picker reflects what the active file is actually using, or the setting is just directly set to this specific file only, by URI, instead of using the global setting.

But still, nevertheless, I also understand the point for keeping it aligned with `updateDialectStatusBar()` for consistency. I’m happy to switch to the simpler global lookup if you like.

mcecode: I understand. I think proceeding with the approach you've taken is fine for now.

---

Just a note for the future @elijah-potter @hippietrail, while reviewing this PR, I realized that some users might get confused due to these differences:

- Status bar shows workspace set setting (if not set, falls back to global then default setting)
- Quick pick shows resource set setting (if not set, falls back to workspace then global then default setting)
- Selecting a dialect from the quick pick changes the global setting

I think a disconnect between the status bar and the quick pick is unlikely since not many users would set a dialect just for a specific file (hence my question to OP). However, I think setting a dialect via the quick pick might cause confusion for some. For example, a user sets a dialect (British) via workspace setting. This is reflected on the status bar and quick pick, but when they select a new dialect (Australian) via the quick pick, it will update not the workspace setting but the global setting. So, even if the user changes their dialect (to Australian) via quick pick, what's applied to the current file and what will be shown on the status bar will still be the original (British).

This is just a heads up that it might come up in the future since I'm not sure how many people will bump into this edge case.

mcecode (approved):

packages/vscode-plugin/src/extension.ts
hippietrail: > I understand. I think proceeding with the approach you've taken is fine for now.
> 
> Just a note for the future @elijah-potter @hippietrail, while reviewing this PR, I realized that some users might get confused due to these differences:
> 
> * Status bar shows workspace set setting (if not set, falls back to global then default setting)
> * Quick pick shows resource set setting (if not set, falls back to workspace then global then default setting)
> * Selecting a dialect from the quick pick changes the global setting

Eww. This triggers some partial memories from playing with some VS Code extension or other. Maybe the Harper one. I'm not sure how well I understood these subtleties when I added the original language indicator. Probably very poorly.

What's the difference between "global" and "default"? Is there not a "document" setting? Only a workspace one? Or is that what "resource" means?

> I think a disconnect between the status bar and the quick pick is unlikely since not many users would set a dialect just for a specific file (hence my question to OP).

I can imagine somebody working on UK and US versions of documentation at the same time and having each set to the respective dialect.

> However, I think setting a dialect via the quick pick might cause confusion for some. For example, a user sets a dialect (British) via workspace setting. This is reflected on the status bar and quick pick, but when they select a new dialect (Australian) via the quick pick, it will update not the workspace setting but the global setting. So, even if the user changes their dialect (to Australian) via quick pick, what's applied to the current file and what will be shown on the status bar will still be the original (British).
> 
> This is just a heads up that it might come up in the future since I'm not sure how many people will bump into this edge case.

This is all very confusing to me. I expect it would be full of surprises to VS Code users who try to work with it.

mcecode: > What's the difference between "global" and "default"? Is there not a "document" setting? Only a workspace one? Or is that what "resource" means?

To be honest it's also quite confusing to me sometimes, but for our purposes, from most to least specific:

- Resource = file type level setting, so yes, it's more or less document setting (I'm least confident in this definition as I rarely set settings this specific)
- Workspace = folder or workspace-level settings set in `.vscode/settings.json` or `.code-workspace` file
- Global = user or global-level settings set in `Code/User/settings.json`
- Default = the default value of a setting set by the extension in `pacakge.json`, in this case, it's `American`

elijah-potter: I'll keep this all in mind. I haven't heard of anyone getting meaningfully confused __yet__, but I'll keep my head on a swivel nonetheless.

PR #3213, opened 2026-04-20
Hereby: require verb that is not also a noun

### LINKED ISSUE #3211: False positive: Flags to join "here by" in "I got here by skill" where they should not be joined
Image

> I got here by skill.

Though this doesn't sound totally idiomatic to me, it is grammatical and I just spotted it in the wild. Maybe not by a native speaker. To me "I got here due to skill" or "I got here thanks to skill" or "I got here through skill" sound more natural, but those would be a different linter to the one that suggests "here by" should be joined.

Perhaps just looking for an NP after "by" and aborting the suggestion would be enough?

joaquinhuigomez:
The rule flagged *"I got here by skill"* because "skill" is tagged as both verb and noun, and `then_verb()` accepts either sense. "I got here by NP" is grammatical "by" plus a noun phrase, not a mis-spaced "hereby".

Replace `then_verb()` with `then_kind_is_but_is_not(is_verb, is_noun)` so only unambiguous verb usages trigger the suggestion. Added a regression test.

Fixes #3211

### DIFF
--- a/harper-core/src/linting/hereby.rs
+++ b/harper-core/src/linting/hereby.rs
@@ -1,6 +1,6 @@
 use crate::expr::Expr;
 use crate::expr::SequenceExpr;
-use crate::{Token, TokenStringExt};
+use crate::{Token, TokenKind, TokenStringExt};
 
 use super::{ExprLinter, Lint, LintKind, Suggestion};
 use crate::linting::expr_linter::Chunk;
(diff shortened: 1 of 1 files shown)


hippietrail (approved):
Looks good to me! Thanks.

elijah-potter (approved):

PR #3227, opened 2026-04-23
fix(chrome-plugin): auto-close report page on tab switch or navigation #3225

### LINKED ISSUE #3225: Close Extension Report Page When Done
The "Report a Problematic Lint" page should automatically close when the user switches to a new site or is otherwise no longer interested in reporting anything.

muhammed-mukthar:
# Issues
Closes #3225

# Description
The "Report a Problematic Lint" popup page was persisting in
`chrome.storage.local` even after the user navigated away or switched
tabs. The next time the popup opened (on any page), it would still
show the report form.

Fix: store the triggering tab ID (`reportTabId`) when the report page
is opened, then add two listeners in the background service worker:
- `chrome.tabs.onActivated` — resets popup state when the user switches
  to a different tab
- `chrome.tabs.onUpdated` — resets popup state when the originating tab
  navigates to a new URL

# Demo
N/A (bug fix — report page no longer re-appears after tab switching or navigation)

# How Has This Been Tested?
Manually: triggered the report page, switched tabs, reopened the popup — 
confirmed it shows the main page instead of the report form.

# Checklist

### DIFF
--- a/packages/chrome-plugin/src/background/index.ts
+++ b/packages/chrome-plugin/src/background/index.ts
@@ -74,6 +74,21 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
 	return true;
 });
 
+chrome.tabs.onActivated.addListener(async ({ tabId }) => {
+	const result = await chrome.storage.local.get(['popupState', 'reportTabId']);
+	if (result.popupState?.page === 'report-error' && result.reportTabId !== tabId) {
+		await chrome.storage.local.set({ popupState: { page: 'main' } });
+	}
+});
+
+chrome.tabs.onUpdated.addListener(async (tabId, changeInfo) => {
+	if (!changeInfo.url) return;
+	const result = await chrome.storage.local.get(['popupState', 'reportTabId']);
+	if (result.popupState?.page === 'report-error' && result.reportTabId === tabId) {
+		await chrome.storage.local.set({ popupState: { page: 'main' } });
+	}
+});
+
 let linter: LocalLinter;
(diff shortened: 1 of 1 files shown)


muhammed-mukthar:
@elijah-potter

packages/chrome-plugin/src/background/index.ts
```diff
@@ -74,6 +74,21 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
 	return true;
 });
 
+chrome.tabs.onActivated.addListener(async ({ tabId }) => {
+	const result = await chrome.storage.local.get(['popupState', 'reportTabId']);
+	if (result.popupState?.page === 'report-error' && result.reportTabId !== tabId) {
+		await chrome.storage.local.set({ popupState: { page: 'main' } });
```
elijah-potter: Would you mind moving the get and set logic to two distinct functions. Since there's a bit of repetition, it might be better to isolate the prop drilling into something a bit better documented and reusable.

elijah-potter (changes requested):
I've tested it and it works! This PR is almost ready to go, but needs just one quality-of-life tweak before it can be merged.

Thanks!

packages/chrome-plugin/src/background/index.ts
muhammed-mukthar: hey @elijah-potter  i have made changes as you have said

elijah-potter (approved):
This looks good to me! Thanks for making those changes.

PR #3232, opened 2026-04-24
feat: GOP party → GOP / Grand Old Party

hippietrail:
# Issues 
N/A

# Description

I noticed "GOP party" in a YouTube comment and thought "we have a linter that can fix that"!

It turned out to be a little complicated by the fact that it's the first one that's a proper noun so when fully-spelled-out each word should begin with a capital.

Since this involved computationally creating the right-side of a string test that requires the right side to be all-lowercase, I've also included debug-time-only asserts to ensure right sides of such tests are always all-lowercase. And a pair of unit tests to make sure that they work as intended.

# How Has This Been Tested?

A unit test for the redundancy and two unit tests for the lowercase-right-side invariant.
Thence `just test-rust`

# Checklist


harper-core/src/char_string.rs
```diff
@@ -101,6 +102,15 @@ impl CharStringExt for [char] {
     }
 
     fn eq_str(&self, other: &str) -> bool {
+        // Assert that the right-hand side is all-lowercase as required
+        debug_assert!(
```
elijah-potter, line 106: I'm a little confused here. The name of the function (and the way __I__ would use it) does _not_ imply that the right side is lowercase. Why is that necessary?

harper-core/src/char_string.rs
```diff
@@ -119,6 +129,15 @@ impl CharStringExt for [char] {
     }
 
     fn eq_ch(&self, other: &[char]) -> bool {
+        // Assert that the right-hand side is all-lowercase as required
+        debug_assert!(
```
elijah-potter, line 133: Same as my other comment.

harper-core/src/char_string.rs
hippietrail: It's always been documented. It was an optimization because in practice the right side was always a literal:

`harper-core/src/char_string.rs`

image

Should I add a to-lowercase into the loop on the right-hand side and remove this requirement?

elijah-potter (dismissed):

elijah-potter (approved):

Issue #3134, opened 2026-04-08
Add ability to test sub-linters of a merged linter


hippietrail:
While testing out PR #3131 I wanted to try adding various extra tests.

The PR modifies only the `ToTooDegreeWords` sub-linter. The merged linter is called `ToTwoToo`. I tried running `harper-cli lint` with just the linter being worked on but ran into this: `Warning: Cannot enable unknown rule 'ToTooDegreeWords'.`

Oddly, while trying various things I found out that the `TooTo` sub-linter does not throw such a warning, but every other sub-linter does.

```
% cargo run \
  --manifest-path harper-the-fourth/harper/Cargo.toml \
  -p harper-cli \
  -- lint --format default --only ToTooAdjectiveEnd,ToTooAdjectivePunct,ToTooAdverb,ToTooAdjVerbEdPunct,ToTooChunkStartComma,ToTooDegreeWords,ToTooPronounEnd,ToTooEos,TooTo harper/harper-core/tests/text harper-test readmine/readmes |& less -R

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.89s
     Running `.hippietrail/cargo-harper-consumers/debug/harper-cli lint --format default --only ToTooAdjectiveEnd,ToTooAdjectivePunct,ToTooAdverb,ToTooAdjVerbEdPunct,ToTooChunkStartComma,ToTooDegreeWords,ToTooPronounEnd,ToTooEos,TooTo harper/harper-core/tests/text harper-test readmine/readmes`
Warning: Cannot enable unknown rule 'ToTooAdjectiveEnd'.
Warning: Cannot enable unknown rule 'ToTooAdjectivePunct'.
Warning: Cannot enable unknown rule 'ToTooAdverb'.
Warning: Cannot enable unknown rule 'ToTooAdjVerbEdPunct'.
Warning: Cannot enable unknown rule 'ToTooChunkStartComma'.
Warning: Cannot enable unknown rule 'ToTooDegreeWords'.
Warning: Cannot enable unknown rule 'ToTooPronounEnd'.
Warning: Cannot enable unknown rule 'ToTooEos'.
```

It should be possible to test individual sublinters of a merged linter using the `harper-cli` developers' debug tool.

PR #3237, opened 2026-04-25
feat(harper-cli): Add ability to run sublinter tests separately in harper-cli

eth1cal-ai:
# Issues 
Partially addresses #3134 

# Description
(Disclaimer about AI use, [as requested by the contributor notes](https://elijahpotter.dev/articles/harper's-policy-on-agent-PRs): I used an LLM to help me reason through parts of this PR. I tried to make sure this PR is brief and stayed grounded by typing out the parts that were useful myself.)
This PR adds the ability for devs to use the sublinters of ToTwoToo on the command line. For example, running the following command:
```
cargo run \                                                                        
-p harper-cli \
-- lint --format default --only ToTooAdjectiveEnd,ToTooAdjectivePunct,ToTooAdverb,ToTooAdjVerbEdPunct,ToTooChunkStartComma,ToTooDegreeWords,ToTooPronounEnd,ToTooEos,TooTo random.txt
```

Gives the same output as:
```
cargo run \                                                                        
-p harper-cli \
-- lint --format default --only ToTwoToo random.txt
```

When random.txt contains something like “She wants ice cream, to.”

To be honest, I still need to:
1. Add this ability for other merged linters
2. It seems like the reason TooTo might not have given a warning (and doesn’t properly lint words in this new commit) is because there’s another rule in phrase_set_corrections named TooTo. I can definitely look into renaming that to make sure it works. 
3. I can also add more in-depth cli tests to isolate the expected behavior of calling every single sublinter individually on the command line.

But I figured I might submit a PR first to see if all the checks have passed/I’m on the right track for code style, to maybe see if this is an appropriate way to handle allowing sublinters to be accessed on the command line, and because I saw it’s also encouraged for users to submit a PR ASAP. I can definitely comment again - or make another PR, to make sure that PRs are kept short - when I have more commits for the other merged linters’ sublinters, if that’s preferable, to make sure that you don’t have to look at my code twice.

# Demo
Here's a video of how this test works when run on a random .txt file that fails on merged linter ToTwoToo and on sublinter ToTooEos.

https://github.com/user-attachments/assets/e7168425-21bc-4596-935c-17d151acce06

- First, I ran it on all available sublinters of ToTwoToo.
- Then, I ran it on only ToTwoToo.
- Lastly, I ran it on ToTooEos.

# How Has This Been Tested?
I ran `cargo test -p harper-core to_two_too` and it passes all the tests still. I also ran `cargo test -p harper-cli` just in case, even though I didn't change any code corresponding to those files (although I figured it would be helpful since of course these code changes aim to affect those files as well). I also created a random txt file (as mentioned above, containing something like "She wants ice cream, to.") to try and isolate the behavior when I was initially testing it. I can definitely add more tests if needed. 

# Checklist
- [X] I have considered splitting this into smaller pull requests.


hippietrail:
> 2. It seems like the reason TooTo might not have given a warning (and doesn’t properly lint words in this new commit) is because there’s another rule in phrase_set_corrections named TooTo. I can definitely look into renaming that to make sure it works.

Oh this could be a separate issue to split out then. I wrote a detector for clashing linter names that got merged a couple of months ago <s>but I forget where it is (-:</s> It must not detect this pair, and that would no doubt be due to me not being aware that sublinters were not "just linters" in some way when I wrote it.

`harper-core/src/linting/lint_group/mod.rs` `no_linter_names_clash()`

Thanks for this work by the way!

hippietrail:
I found time to sit down and look into this more closely. I think this might not be the best way to fix it. _Unless there's not another way?_

It only handles the `ToToo` family but there are a bunch of other merged linters with sublinters too. Maybe spelling them all out piece-by-piece is the only way it can be addressed. A general solution would be preferred if such a thing can actually be done. I haven't dug that deep yet.

eth1cal-ai:
- Wait, that's totally my bad about the no_linter_names_clash - it seems like the Git Actions workflow caught that just now in the just test-rust check, so I will definitely fix that and add it to this PR! 
- That would make sense if there was a better way, rather than manually updating the rest. Would it be better to have a general solution that's geared towards fixing some of the code in harper-cli and the way that input is parsed, like if there was a function to specifically handle sublinters? With the current code I was mainly just modifying harper-core and the lint group file. I can definitely try to think on it as well, although I'm definitely still a novice at understanding and navigating the codebase

hippietrail:
I'm actually looking at enhancing `no_linter_names_clash` right now. It's pushing me into Rust territory I don't know at all around casting and traits.

I added #3241 and #3242 - For both the tree of linters needs to be walkable which it currently is not because we can't cast a Linter to a LintGroup yet.

In the case of your work here I'm not sure if that will help or not. It's still well within one of my Rust knowledge gaps. The way it's done now is very brittle because it relies in maintaining a separate list of the sublinters of linters, apart from the implementations of those linters. That's in some way similar to how/why we needed a test to find clashing linter names, because that's also an area of brittleness.

eth1cal-ai:
I see, that makes sense! 

- In that case, would it be helpful I closed this PR? I saw that you marked #3242 as a good first issue, so I can try to look at that instead if that would be better. 
- Also, in regards to #3241, I think just test-rust did end up catching the issue in the integration checks after I wrote the code for this PR (and just now locally, so I apologize for not running it on my computer first), but are you saying that `no_linter_names_clash` also should've identified the issue when you first implemented `no_linter_names_clash` a few months ago?

Thank you!

hippietrail:
> I see, that makes sense!
> 
> * In that case, would it be helpful I closed this PR? I saw that you marked [Add a new `ls-linters` command to `harper-cli` and `justfile` #3242](https://github.com/Automattic/harper/issues/3242) as a good first issue, so I can try to look at that instead if that would be better.
> * Also, in regards to [Clashing linters named `TooTo` are not detected by `no_linter_names_clash()` #3241](https://github.com/Automattic/harper/issues/3241), I think just test-rust did end up catching the issue in the integration checks after I wrote the code for this PR (and just now locally, so I apologize for not running it on my computer first), but are you saying that `no_linter_names_clash` also should've identified the issue when you first implemented `no_linter_names_clash` a few months ago?
> 
> Thank you!

No problem. It's up to you. If you think you can and want to keep working on this issue just rework it. If not, then close it.

I spent a few hours on listing the linters as a tree and discovered I was labouring under the false assumption that merging linters created `LintGroup`s but in fact those are two distinct types of collections of `Linter`s. I made a more detailed comment on one of the other issues.

PR #3263, opened 2026-04-28
feat: `Word` ctor for standard case combos; fave → favourite

hippietrail:
# Issues 
N/A

# Description

While dogfooding with Harper Tauri on YouTube comments I noticed that "fav" and "favs" are flagged so I thought they'd make good candidates for an "expand" linter.

- Also offers to expand "fave" and "faves" even though those are in the dictionary.
- Expands to either US or non-US spelling, depending on the Dialect.

While looking for examples in GitHub to use for unit tests I noticed a few hits for things like `FaVe` and realized it would be worth adding a new constructor to the `Word` `Pattern` that would avoid matching non-standard case patterns since they're often trademarks or project names etc.

So `Word::new_standard_case` is like `Word::new` and `Word::new_exact` but will match only all-lowercase, all-uppercase, and title case (first letter uppercase, remaining letters lowercase), since those three patterns can be used for arbitrary words depending just on whether they're at the start of a sentence or part of an all-caps headline, etc.

I got the Windsurf built-in coding AI to make the initial `new_standard_case` implementation but I guided it to use an internal enum and I threw away its routine to check if a word is one of the 3 standard cases and wrote my own efficient one. I kept the tests it made but added one to check all 8 combinations of case for 3-letter words.

The linter itself was written by me.

I could add a standard-cases-only version of `WordSet` to this PR if asked. If not I'm sure I'll add one later.

# How Has This Been Tested?

Unit tests using sentences mined from GitHub to cover the "fav" vs "fave" spellings, singular and plural, various case combinations, all the dialects, and test against false positives.

`just test-rust`

# Checklist


elijah-potter (approved):
This looks great! I actually found that your hand-rolled routine improved overall speed in the benches by ~6%. Thanks!

PR #3265, opened 2026-04-29
feat: less confident `SplitWords` `message` when either word is only 1 char

hippietrail:
# Issues 
N/A

# Description

While dogfooding my Harper Tauri on YouTube video comments I noticed `SplitWords` often suggests unlikely pairs where one side is a one-letter "word" yet the message said "... should **probably** be written as ... ...".

I changed the logic to use the adverb "**possibly**" instead in such cases.

# How Has This Been Tested?

I picked two examples from my test files as unit tests for the different `message`s.

# Checklist


elijah-potter (approved):
Looks good to me! I think this communicates our intent much better. Thanks!

PR #3266, opened 2026-04-29
feat: add convenience methods to `Token` for direct equality checks

hippietrail:
# Issues 
N/A

# Description

I noticed it's a very common pattern to get the content of a token and then check if it's the same as a string or char array literal. In fact there are multiple functions to get the token content, as a string or as a char array reference. It's almost always better to use the latter since it can do comparisons against other string types using only iterators and avoiding allocations and full type conversions.

In some places the change only saves a couple of characters, but in other places it saves one or two lines. I think it makes the code more expressive and easier to read and makes accidentally introducing unnecessary type conversions and allocations a little harder for new contributors copying existing code.

- Add `eq_ch()` and `eq_str()` methods to `Token` to eliminate intermediate `get_ch()`/`get_str()` calls throughout the codebase.
- Update all usage sites and remove unused `CharStringExt` imports.

# How Has This Been Tested?

`cargo test`

# Checklist


elijah-potter (commented):
I am not convinced that this is actually an improvement. I feel that there should be fewer methods for accomplishing a task, not more. This increases the number of methods one may consider when trying to do this.

On top of that, I am convinced that this change will compile to essentially the same code.

Please feel free to push back on this if you disagree.

PR #3267, opened 2026-04-29
feat: 🇸🇬 Singapore English & 🇲🇾 Malaysia English

hippietrail:
# Issues 
N/A

# Description

I'm in Malaysia for the next three months so I'm making a start on support for Singapore English 🇸🇬 and Malaysia English 🇲🇾. English is an official language in Singapore and in Sarawak state in Malaysia but it's widely used in all aspects of life everywhere in Malaysia.

It would be amazing if any Singaporeans or Malaysians wanted to help out!

# How Has This Been Tested?

- `cargo test` passes
- `just test` will probably fail but then I'll fix the issues it uncovers in non-Rust code

# Checklist


elijah-potter (commented):
Are "Singapore" and "Malaysia" English well-documented dialects? Could it make more sense to collapse the two into a single dialect? I don't know much about the regionalisms, so I'm naturally curious.

hippietrail:
> Are "Singapore" and "Malaysia" English well-documented dialects? Could it make more sense to collapse the two into a single dialect? I don't know much about the regionalisms, so I'm naturally curious.

I think they're about as well documented as Indian and even Canadian English, both of which lack anything professional online. (Australian English does have a professional dictionary online but it's subscription only plus I'm a native speaker and language nerd.) OED covers them and all other varieties of English, but only allows tiny peeks without a login. Apparently many libraries have subscriptions to their website, but none of my local libraries where I've lived have )-:

So basically I do the same for Canadian, Indian, Malaysian, and Singaporean, which is Google and ask AIs. There's also always Wiktionary, which isn't at professional level but is very handy.

English has been the language of education in Singapore for 40 years and has been a national language for over 60 years.

They do have different quirks despite being largely the same. Malaysian English has more words from Malay used in English. A couple of food words are spelled differently. They use different terms for limited liability company: "Pte Ltd" in Singapore vs "Sdn Bhd" in Malaysia. I think they're about as different as US vs Canadian or Australian vs New Zealand. That is, enough for locals to know almost instantly, but not obvious to outsiders.

PR #3275, opened 2026-04-30
feat: feet/inches and minutes/seconds from '/" to ′/″

### LINKED ISSUE #3273: Add a linter that corrects using apostrophe and quote symbols for feet and inches instead of prime and double prime
Just like we have a linter to correct three ASCII periods `...` to a Unicode ellipsis `…` we should have to that corrects `6'4"` to `6′4″`

I think this will be very difficult to do when only feet or only inches are used on their own as the ASCII apostrophe and double-quotes have various uses and will be ambiguous, require matching, etc.

But if feet and inches are used together it might avoid hitting ambiguous cases.

hippietrail:
# Issues 

Closes #3273

# Description

- Changes x'y" to x′y″ for feet and inches or minutes and seconds. (Prime and double-prime characters.)
- Changes x°y' to x°y′ and x°y'z" to x°y′z″ for degree and minutes or degrees, minutes, and seconds.

To avoid ambiguity with other uses of the ASCII apostrophe and quote characters, it needs enough context, such as the presence of the degree symbol, or having both ' and " together, but not just one alone.

Works with the ASCII chars and smart/curly/typographic apostrophes and quotes (‘/’ and “/”), and mixed combinations.

It works with or without spaces on either side of the symbols and does not change the spacing.

# How Has This Been Tested?

I managed to find a file on GitHub with two feet and inches variants to use as unit tests and contrived others that should cover all possible combinations.

I added `then_optional_whitespace()` and a shorthand `t_ows()` to `SequenceExpr` and converted all other linters that used optional whitespace to use it.

I added `is_xxx` and `then_xxx` methods to `TokenKind` and `SequenceExpr` for the degree, single prime, and double prime symbols, plus the open single quote symbol. The last change affected snapshots, which previously treat those symbols and unlintable but now treat them as punctuation. **They probably don't take part in quote-matching though.**

AI disclosure: I got the coding AI to make my original hand-coded version more Rusty, and then iterated between manual and AI fixes. The final version is just about all my manual coding now.

# Checklist


hippietrail:
Since the same symbols are used for degrees and minutes, I'm changing this to a draft while I add support and rename the linter...

elijah-potter (approved):

PR #3281, opened 2026-05-01
feat: fix wrong variants of "if worst comes to worst"

hippietrail:
# Issues 
N/A

# Description

I think I heard something like "if worse comes to worse" and looked into it. It turns out there are several variants that are accepted and several that are universally considered wrong. Preference is influenced by dialect.

This linter detects all known variants, rules out those considered correct, flags all others, and suggests the correct variants.

I haven't made a final decision of how the dialect should affect what's flagged and what's suggested. That's the only reason it's a draft for now.

AI disclaimer: I decided how the overall logic would work then used to coding AIs to help convert loops to functional style, avoid allocations and conversions, and maximize Rustiness. (It turns out the built-in "SWE-1.5" in Windsurf is not as good as the free coding AI on `you.com`)

# How Has This Been Tested?

Several unit tests to check various wrong forms will be corrected to various right forms and ensure already right forms do not get flagged.

Half a dozen unit tests from sentences found on GitHub.

`cargo test`

# Checklist


harper-core/src/linting/worst_comes_to_worst.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+}
+
+impl ExprLinter for WorstComesToWorst {
+    type Unit = Chunk;
+
+    fn match_to_lint(&self, toks: &[Token], source: &[char]) -> Option<Lint> {
+        // Return None if we matched any of the valid variants
+        let is_valid_variant = VALID_VARIANTS.iter().any(|variant| {
```
elijah-potter: Could we inline this check into the expression itself?

elijah-potter (changes requested):
I have one nit. Feel free to push back on it.

harper-core/src/linting/worst_comes_to_worst.rs
hippietrail: @elijah-potter It's turning out to have some subtle issue I can't figure out. I moved the logic into a `.but_not` clause on the `Expr` and it works for the first three variants but fails to exclude the last one, "the worst comes to the worst".

My guesses are that it might be due to that subexpression being longer than the positive parts of the `Expr` or maybe some inner part of it is hitting the 100 limit? It's hard to debug.

Maybe it's clearer to use with your deeper knowledge of how you implemented it, or you can sic an agent onto it?

I'll commit the changed version with some commented-out experiments and debug code.

hippietrail: @elijah-potter I couldn't get that working with `.but_not()` due to a lurking bug in `UnlessStep`: https://discord.com/channels/1335035237213671495/1461079667367739472/1549105711311491145

PR #3293, opened 2026-05-02
feat: in favour doing→in favour of doing

hippietrail:
# Issues 
N/A

# Description

I heard or read "in favour doing" with the missing "of" where it should be "in favour of doing". I did some Googling and it's actually common with various verbs.

This linter handles UK and US spelling and all verbs in the gerund/progressive form.

# How Has This Been Tested?

Half a dozen examples were found on GitHub and used to make unit tests.
Then `cargo test`

# Checklist

### DIFF
--- a/harper-core/default_config.json
+++ b/harper-core/default_config.json
@@ -4822,6 +4822,13 @@
 								"state": true,
 								"label": "More Than Meets The Eye"
 							}
+						},
+						{
+							"Bool": {
+								"name": "InFavourOfDoing",
+								"state": true,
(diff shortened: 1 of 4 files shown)


elijah-potter (approved):
I wonder if we could generalize this beyond the specific case?

Either way, I'm going to merge this. Thanks!

hippietrail:
> I wonder if we could generalize this beyond the specific case?
> 
> Either way, I'm going to merge this. Thanks!

I thought about this too. We probably can but I thought we could build upon it when we find real examples.

PR #3066, opened 2026-03-30
feat: Work in progress: "There is" agreement

hippietrail:
# Issues 

Addresses #1752 and #2355

# Description

Corrects disagreement in grammatical number between copula and single noun in existential constructions:
- there is [things]
- there are [thing]
- there was [things]
- there were [thing]
- is there [things]
- are there [thing]
- was there [things]
- were there [thing]
- there's [things]

So far it only handles the noun being a single word:
- No articles or determiners (there are a thing, there is many things, etc.)
- No adjectives (there is red things, etc.)
- No compound nouns (there are train station, etc.)

# How Has This Been Tested?

There are contrived tests for the basic cases and a bunch of real-word tests from GitHub and the internet generally.

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests to cover my changes

Issue #2355, opened 2025-12-17
"there are", "there is", "there was", and "there were" need to agree in number with their noun


hippietrail:
**The Error**

"There is" and "there was" are used with singular nouns (or mass) nouns.
"There are" and "there were" are used with plural nouns.

People often produce writing where these do not actually agree.

**Examples (2-3):**
1. the industry has been pushing for this kind of non-sense because **there were money** in rapid growth and no money in quality
2. decouple the editor from the viewer, so the editor won't crash if **there is errors** in the code people use
3. **There was errors** in docker compose runing.
4. **there were problem** about version but this error is solved by correcting version in these files

**Potential Edge Cases**
If there are any I'm guessing it will be in compound nouns with some words with ambiguous POS.

hippietrail:
Closely related to #1752

mauropereiira:
Closing this as a duplicate of #1752. #1752 is written in inverted question order and this issue in declarative order, but both track the same number-agreement rule for existential `there + be`. The implementation history confirms that: #3066 and its merged replacement #3303 explicitly address both issues in one linter. The declarative and mass-noun examples, POS edge case, and remaining implementation limits have been preserved on #1752.

Issue #1752, opened 2025-08-14
Should flag "is there" and "was there" with plural nouns (and "are there" and "were there" with singular)


hippietrail:
<img width="596" height="28" alt="Image" src="https://github.com/user-attachments/assets/b72690fb-794a-47dd-8b8e-03cb8428136e" />

In this example from our own GitHub issues "is there any plans" should be either "are there any plans" or "is there any plan".

More examples:
1. **Was there errors**/warning or not?
2. **Are there message** related to nim-mode in *Messages* buffer? No.
3. **Were there problem** with the upgrade?

hippietrail:
I actually started working on this about a week ago but put it aside. If anyone is interested I can see if what I have is worth posting as a draft PR.

mauropereiira:
Closed #2355 into this issue. It adds declarative order to the same existential `there + be` agreement rule: `there is`/`there was` take singular or mass nouns, while `there are`/`there were` take plural nouns.

Verbatim examples from #2355:

- `the industry has been pushing for this kind of non-sense because there were money in rapid growth and no money in quality`
- `decouple the editor from the viewer, so the editor won't crash if there is errors in the code people use`
- `There was errors in docker compose runing.`
- `there were problem about version but this error is solved by correcting version in these files`

The reported edge case is determining the head and number of compound nouns when words have ambiguous POS.

PR status: #3066 addressed both issues but closed unmerged after branch problems. #3303 superseded it and merged on May 8, 2026. It handles declarative and question order, plus `there's`, but is limited to a single noun without articles/determiners, adjectives, or compound nouns. Current `master` still has the exact `there were problem about version...` regression ignored because `problem` is also tagged as an adjective. The mass-noun and Docker examples are not named regression tests in the merged PR, so they still need verification. Neither issue has a closing PR reference.

PR #3303, opened 2026-05-05
feat: "there is" agreement

hippietrail:
# Issues 

- Addresses #1752 and #2355
- Supersedes #3066

# Description

Corrects disagreement in grammatical number between copula and single noun in existential constructions:
- there is [things]
- there are [thing]
- there was [things]
- there were [thing]
- is there [things]
- are there [thing]
- was there [things]
- were there [thing]
- there's [things]

So far it only handles the noun being a single word:
- No articles or determiners (there are a thing, there is many things, etc.)
- No adjectives (there is red things, etc.)
- No compound nouns (there are train station, etc.)

- Introduces `regular_nouns` module to handle ending changes when converting between singular and plural.
- removes rare/marginal noun sense of "none"

# How Has This Been Tested?

There are contrived tests for the basic cases and a bunch of real-word tests from GitHub and the internet generally.

# Checklist


harper-core/src/regular_nouns.rs
```diff
@@ -0,0 +1,281 @@
...
+            dict.get_word_metadata(candidate)
+                .is_some_and(|md| md.is_plural_noun())
+        })
+        .collect()
+}
+
+/// Get all valid singular forms using regular patterns, validated against dictionary
+/// Returns vector of suggestions that are confirmed to be singular nouns
+pub fn get_singulars(plural: &[char]) -> Vec<Vec<char>> {
+    let dict = FstDictionary::curated();
```
elijah-potter: I think it would be reasonable to have the dictionary as an argument to the method.

harper-core/src/linting/there_is_agreement.rs
```diff
@@ -0,0 +1,1459 @@
...
+
+#[derive(PartialEq)]
+enum WordOrder {
+    Statement, // normal: there is, etc.
+    Question,  // inverted: is there?, etc.
+}
+use WordOrder::*;
+
+#[derive(PartialEq)]
+enum Tense {
```
elijah-potter, line 38: While it __might__ be acceptable in this case to introduce these types, since they are contained within the function, I worry that there will be duplicate definitions for enums that fulfil similar roles. Is there a way we could elevate these types to a module higher in the hierarchy or eliminate them altogether?

elijah-potter (commented):
This looks good and there is obvious utility. I have a few notes though before we can merge it.

harper-core/src/linting/there_is_agreement.rs
hippietrail: It's really a simplistic/throw-away type since "tense" is a very misunderstood term with different uses by linguists/grammarians and regular people, so definitely shouldn't be elevated. I used it here because it's a concise way to mark is+are on the one hand vs was+were on the other hand.

By eliminate them do you mean just this `Tense` one or also the other enums? As in replace them with bools or such. Or I can just rename this enum to `Forms` and the variants to `IsAre` vs `WasWere`?

harper-core/src/regular_nouns.rs
hippietrail: > I think it would be reasonable to have the dictionary as an argument to the method.

Interesting. Let me look into that right now...

harper-core/src/linting/there_is_agreement.rs
elijah-potter: On second look, I think it's fine. Don't worry about it.

elijah-potter (dismissed):
We're good to go on this. Thanks!

elijah-potter (approved):

PR #3315, opened 2026-05-06
feat(harper.js): make it easier to ignore all lints

elijah-potter:
# Description

While working on the Harper editor for Harper Desktop, @jasontheadams and I noticing that the API for ignoring all the lints in a document was pretty clunky.

To remedy this, I've changed two parts of `harper.js`. 

1. The linting functions now accept a `dedup` option. This allows the consumer to change whether overlapping lints are removed (which is often desired for aesthetic reasons). The default behavior remains the same: overlapping lints are removed. This change just allows you to set `dedup = false` so you can get _all_ the problems at once.
2. I've added a batched `ignoreLints` method to the `Linter` interface. This allows you to ignore as many lints as you want, while reusing the same source text. This is helpful for code cleanliness and for performance, since you don't need to shuffle as much data around.

# How Has This Been Tested?

Additional unit tests.

# Checklist

### DIFF
--- a/harper-wasm/src/lib.rs
+++ b/harper-wasm/src/lib.rs
@@ -276,18 +276,25 @@ impl Linter {
         Ok(())
     }
 
-    pub fn ignore_lint(&mut self, source_text: String, lint: Lint) {
+    pub fn ignore_lints(&mut self, source_text: String, lints: Vec<Lint>) {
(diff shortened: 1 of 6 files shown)


hippietrail:
Being able to turn off suppressing overlapping lints is something I've been wanting to request!

PR #3323, opened 2026-05-07
feat: add `word-provenance` command to `harper-cli`

### LINKED ISSUE #3287: Add a `word-provenance` command to `harper-cli` (and maybe a `just` command too)
A `harper-cli word-provenance` command would track whether a word is
- directly in the dictionary
- generated from a base word in the dictionary by its `/` annotation flags directly
- generated by the interaction of annotation flags and the `cross-product` property
- generated by the interaction of annotation flags on two entries that have the same letters but different letter cases (Harper's dictionary is case-folded internally).
- generated by any combination of the above, or generated via multiple routes.

For instance a word like "unboxing" might be generated from `unbox/G` and also from `boxing/U` and also from `unboxing/` directly.

See #3284 for a story of how unexpected spurious words can end up in Harper's dictionary and how they can trigger spurious lints - and how the causes can be unobvious and hard to discover.

furmak331:
# Issues
Resolves #3287

# Description
This PR adds the `word-provenance` subcommand to `harper-cli` (along with a `just word-provenance` shortcut) to trace how words end up in Harper's dictionary.

To keep the hot dictionary loading path (`expand_annotated_word`) untouched and performant, I implemented this using a replay-based approach. The new `expand_with_provenance` methods in `AttributeList` mirror the expansion logic but add instrumentation to record the generation route.

It successfully tracks all 5 categories from the issue:
- **Direct entries** (base words in `dictionary.dict`)
- **Affix-generated** words (single flag expansion)
- **Cross-product interactions** (recursive expansions)
- **Case-folding collisions** (words sharing the same case-folded form)
- **Multi-route** (words generated via multiple of the above)

It also includes a `--json` flag for machine-readable output.

# Demo
image

# How Has This Been Tested?
Added 5 unit tests in attribute_list.rs verifying the provenance logic against test fixtures (direct, affix, cross-product, multi-route).
Added 5 integration tests in harper-cli/tests/word_provenance.rs testing the CLI against the real curated dictionary.
Executed cargo test and ran just word-provenance locally to verify behavior.
# Checklist


furmak331:
To let harper-cli access AttributeList and parse_word_list, I made spell::rune fully public. Would you prefer selective re-exports through spell/mod.rs instead, like pub use rune::{AttributeList, parse_word_list, ProvenanceKind, ProvenanceRecord}?

hippietrail:
Cool! I'll start looking at this. Out of interest, did you use AI in part or in full to make this. I haven't looked yet. A rust test is failing - that's often a minor formatting change or Clippy issue. I'll look at that too.
Yep. I usually try to remember to do `cargo fmt` and then `cargo clippy` and then `cargo test` before making or pushing to a PR.

harper-cli/tests/word_provenance.rs
```diff
@@ -0,0 +1,89 @@
...
+        .output()
+        .unwrap();
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(
+        stdout.contains("AFFIX"),
+        "expected AFFIX provenance for 'quickly', got: {stdout}"
+    );
+    assert!(
+        stdout.contains("quick"),
```
hippietrail: Isn't `.contains()` too fuzzy? It would be true for any string with `quickly` in it.

harper-cli/tests/word_provenance.rs
```diff
@@ -0,0 +1,89 @@
...
+    let output = harper_cli()
+        .args(["word-provenance", "hello", "quickly"])
+        .output()
+        .unwrap();
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(
+        stdout.contains("hello") && stdout.contains("quickly"),
+        "expected output for both words, got: {stdout}"
+    );
```
hippietrail, line 88: I think we definitely want some tests for the other kinds of provenance to know that they all work as expected too.
You can do this two ways:
1. either try a bunch of words, or write a tool to pick words randomly from the dictionary, and take note of interesting cases.
2. you can make `#[test]`s using tiny custom dictionaries and annotation files. Look for examples in other modules to see how.

Oh sorry, it looks like you are doing more complete testing in `harper-core/src/spell/rune/attribute_list.rs`

furmak331:
Hi @hippietrail , thanks for taking a look
> Out of interest, did you use AI in part or in full to make this.

I generally do use AI to iterate things faster but I try to make sure to review the logic and drive the architecture myself. Spent time to see if cross product provenance properly mirrors the original logic

>Yep. I usually try to remember to do cargo fmt and then cargo clippy and then cargo test before making or pushing to a PR.

Will make sure from next time . This is my first time :)

hippietrail (commented):
I haven't tried it out yet (nearly 3am here) but looking at it my impression is it's very clean. I made a couple of comments.

hippietrail:
> Hi @hippietrail , thanks for taking a look
> 
> > Out of interest, did you use AI in part or in full to make this.
> 
> I generally do use AI to iterate things faster but I try to make sure to review the logic and drive the architecture myself. Spent time to see if cross product provenance properly mirrors the original logic

Sounds similar to my current MO. It's good to know because I would keep a more careful on eye primarily AI code to spot the kinds of things that look good on the surface but have those AI-style quirks in the shadows.

> > Yep. I usually try to remember to do cargo fmt and then cargo clippy and then cargo test before making or pushing to a PR.
> 
> Will make sure from next time . This is my first time :)

🚀

hippietrail:
> To let harper-cli access AttributeList and parse_word_list, I made spell::rune fully public. Would you prefer selective re-exports through spell/mod.rs instead, like pub use rune::{AttributeList, parse_word_list, ProvenanceKind, ProvenanceRecord}?

@elijah-potter

furmak331:
> I haven't tried it out yet (nearly 3am here) but looking at it my impression is it's very clean. I made a couple of comments.


I think we may still need a few more code changes. I'll take a closer look in the morning (1am here as well). Sorry for the delay

harper-cli/tests/word_provenance.rs
furmak331: Yeah . I did it the second way. The unit tests in attribute_list.rs use the mini-dictionaries to verify every specific route (Direct, Affix, Cross-Product, Multi-route).

The tests here in harper-cli just verify the CLI output formatting and real dictionary parsing.

furmak331:
Hi @hippietrail , quick follow up on this PR. 
The requested changes and additional test assertions have been added, and CI is passing now.
Let me know if there’s anything else that needs to be adjusted before merge. Thanks!

hippietrail:
> Hi @hippietrail , quick follow up on this PR. The requested changes and additional test assertions have been added, and CI is passing now. Let me know if there’s anything else that needs to be adjusted before merge. Thanks!

Thanks for pinging me. I'll try to take a look at this today.

I've been playing with it for a few hours and it seems to work just like I was hoping. I've requested a review from @elijah-potter though to check the code, especially the changes to Harper's core:

- ‎`/spell/rune/attribute_list.rs‎`
- ‎`/spell/rune/expansion.rs‎`
- ‎`/spell/rune/mod.rs‎`
- `‎/spell/mod.rs`

harper-cli/tests/word_provenance.rs
```diff
@@ -0,0 +1,89 @@
...
+        .args(["word-provenance", "hello"])
+        .output()
+        .unwrap();
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(
+        stdout.contains("DIRECT"),
+        "expected DIRECT provenance for 'hello', got: {stdout}"
+    );
+}
```
elijah-potter, line 19: These tests are unnecessary and honestly kind of distracting. They look like the work of an AI agent. Would you mind deleting them?

elijah-potter: If you believe they are truly necessary, would you refactor them to be actual unit tests instead of these weird integration test things?

elijah-potter (changes requested):
I've had a little time to think about this.

Would it be possible to accomplish this same task with fewer changes to `harper-core` by instead constructing small `MutableDictionary`s with the expansions of individual lines in the dictionary?

For example, we would create a dictionary with just the single line `laptop/S`. When the expander creates both `laptop` and `laptops` we  can know that the latter comes from the former. Rinse and repeat with each word in the dictionary.

hippietrail:
> I've had a little time to think about this.
> 
> Would it be possible to accomplish this same task with fewer changes to `harper-core` by instead constructing small `MutableDictionary`s with the expansions of individual lines in the dictionary?
> 
> For example, we would create a dictionary with just the single line `laptop/S`. When the expander creates both `laptop` and `laptops` we can know that the latter comes from the former. Rinse and repeat with each word in the dictionary.

Wow that's a really intriguing idea I have to say!

furmak331:
Thanks, that worked out much better. `harper-core` is untouched now; the whole thing is ~150 lines in `harper-cli` plus the justfile recipe.

It expands one `dictionary.dict` line at a time into its own `MutableDictionary`, the way `forms` already does, and reports whichever entries produce the word. To name the flag responsible it re-expands the entry one flag at a time: a flag that produces the word on its own is reported as that flag, and a pair that only produces it together is a cross-product. So nothing duplicates the expansion logic, it just asks the expander. Case-folded entries like `Polish`/`polish` both turn up for either query without special handling, since `WordId` lowercases.

One thing I added beyond your sketch: expanding all 54k entries one at a time re-parses `annotations.json` for each, which puts a single query at ~42s through `just word-provenance`. So entries get ruled out 256 at a time first, and if a batch doesn't produce the word, no entry in it did. Same results, and it brings a query to ~7s, most of which is the curated dictionary load every subcommand pays anyway.

```
$ just word-provenance laptops disallows Polish boxing
'laptops' comes from:
 - laptop/~NSgJ             flag S
'disallows' comes from:
 - allow/~VGdSE             flags S and E combined
'Polish' comes from:
 - Polish/JNg               the entry itself
 - polish/~NgSVGd>Z         the entry itself
'boxing' comes from:
 - box/~NgSVGd>Zn           flag G
 - boxing/~VNg              the entry itself
```

Deleted those tests as well. Happy to add real unit tests for the flag attribution if you want coverage there, I just didn't want to re-add noise. I also dropped the `--json` flag, since it wasn't in the issue and was surface area nobody asked for.

hippietrail:
I have to say this is now looking even better than I envisaged it. Thanks!

PR #3326, opened 2026-05-08
feat: Redundant self-harm onself, self-teach oneself, do self-harm to oneself, self host it oneself

hippietrail:
# Issues 
N/A

# Description

Fixes various sentence patterns where "self" is used twice, first prefixed to a verb or noun, and then as part of a reflexive pronoun.

- self-censor oneself, self-correct oneself, self-describe oneself, self-diagnose oneself, self-govern oneself, self-harm oneself, self-identify oneself, self-medicate oneself, self-regulate oneself, self-sabotage oneself, self-teach.
- (do) self-harm to oneself
- self-host (it) oneself

Handles all inflections of the verb and all pronouns, including ones like singular "them".

Avoids flagging things like "self harm itself is not the root cause".

**BONUS** I've included an extra bit of debug output in `assert_good_and_bad_suggestions` to print out suggestions not in either the good list of bad list. Really helps in cases where you have an out-by-one index or a missing full stop in one variant etc.

AI disclosure: I used Windsurf's coding AI to refactor some of my initial versions but beyond a certain point early on, I identified all the patterns to simplify it so the final version is basically all my code now.

# How Has This Been Tested?

Examples from GitHub for the technical patterns, plus some from Stack Overflow/Exchange, and the rest from Medium and Reddit for the self-harm patterns.

# Checklist


hippietrail:
This might happen in more verbs that I thought:
- How can she **self**-humiliate her**self** so much?

PR #3330, opened 2026-05-09
feat: allow a linter to return multiple `Lint`s with different `span`s

### LINKED ISSUE #3329: Would be handy to have a way to return multiple `Lint`s from a `Linter`/`ExprLinter`
Sometimes, and especially when dealing with grammatical agreement, there are two fixes in different areas, such as a verb form that only works with singular pronouns and nouns mismatched with a plural pronoun or noun.

When there are other words in between, you can only handle that by returning multiple `Suggestion`s on the same `Lint` with the same `span` and therefore the `span` must cover the full extent of text that encompasses both parts that can be fixed.

It would be easier to code, and the code would be easier to read, if you could return two `Lint`s, each with a `span` just for the part it needs to fix.

hippietrail:
# Issues 

Resolves #3329

# Description

Before this an `ExprLinter` could return multiple `Suggestions`, but all on a single `Lint`, which is what you need most of the time.
Buy sometimes there's two ways to fix a mistake the modify different parts of the sentence not next to each.

So far the way to do this was to replace the whole section encompassing both, which could get a bit confusing and also leader to bigger code that took some effort to read. Now you can return a `Lint` for each place in the sentence that fixes just that localized spot, and another in the other place. Which should make the linter code easier to follow.

For now it only adds `match_to_lints_with_context()` but I might also add `match_to_lints()`.

As before, an `ExprLinter` should choose exactly one of the three `match_to_lint` functions to implement. Implementing none of them or multiple of them will probably blow the stack.

If any Rust experts can think of a way to detect/avoid this at compile time or runtime, a fix would be welcome!

**⚠️ NOTE:** One difference that might be surprising in some way even though totally logic in another, is that `assert_lint_count` will probably now be able to return >1 for a single operation. This does not affect any current modules or test but might if an old linter is refactored to use the new method.

The semantic change would be that the number of lints would not be the number of mistakes identified but the number of places mistakes can be fixed. In `#[test]`s this probably doesn't matter. in UIs and stats it might.

# How Has This Been Tested?

A test linter is included that fixes a pronoun and verb that don't agree by a pair of lints, each fixing one of the two words.

# Checklist


hippietrail:
@elijah-potter I don't think this one is ready to go in as-is but I'd like to hear your thoughts about the idea.

Issue #3276, opened 2026-04-30
False negative: "there own" is not flagged when it should be "their own"


hippietrail:
We have one or more linters that try to correct mixing up "there" vs "their" vs "they're" but even though "there own" is quite common we do not correct it.

Examples from GitHub:
- Split data on **there own** topics
- Provide user the option in setting to customise **there own** default Effort lvl
- auto-completion windows within emacs pop out as **there own** window in tilling mode
- Allow for modules to provide **there own** c files
- I have a number of scripts that all create **there own** connection

It's not always an error though:
- People **there own** expensive cars. (Made up example)

I wasn't expecting to find many but it looks like "they're own" is also pretty common:
- Everybody has **they're own** preferences?
- but because they are **they're own** command they have their own MshCommandRuntime
- AI does **they're own** directions of sharing their information.
- highly configurable listener (and config element) to allow users define **they're own** metrics
- components I'm working on that will eventually become **they're own** library

If we handle "they're own" it's no extra effort to also handle the missing apostrophe version:
- allowing 3rd party appstores to silently update apps installed by **theyre own**
- used for security-researchers to test LLM-BotNet capabilities on **theyre own** Local-Machines
- it would be helpful for people building **theyre own** rockets
- very useful for ppls who would like to store **theyre own** data

mauropereiira:
This looks like it was already fixed by #3336, which added the `ThereOwn` linter back in May.

I ran the examples from this issue against current `master` with `harper-cli`:

```
Split data on there own topics                                   -> ThereOwn
Allow for modules to provide there own c files                   -> ThereOwn
I have a number of scripts that all create there own connection  -> ThereOwn
Everybody has they're own preferences?                           -> ThereOwn
it would be helpful for people building theyre own rockets       -> ThereOwn
```

All five flag with "Did you mean the possessive `their`?", covering the `there own`, `they're own` and `theyre own` variants.

The false-positive example from the report, "People there own expensive cars.", stays clean, and `there_own.rs` has a regression test covering exactly that case.

Looks safe to close.

PR #3336, opened 2026-05-09
feat: add ThereOwn linter for "there own" → "their own"

kimjune01:
# Issues
N/A

# Description

Adds a `ThereOwn` linter that catches "there own", "they're own", and "theyre own" and suggests the possessive "their own". Common mistake in informal writing.

# How Has This Been Tested?

- Corrects "there own" → "their own"
- Corrects "they're own" → "their own"
- Corrects "theyre own" → "their own"
- Does not flag correct "their own"
- Does not flag "there" without "own"

# Checklist

### DIFF
--- a/harper-core/src/linting/there_own.rs
+++ b/harper-core/src/linting/there_own.rs
@@ -1,4 +1,5 @@
-use crate::linting::expr_linter::Chunk;
+use crate::char_string::CharStringExt;
+use crate::linting::expr_linter::{Chunk, followed_by_word, preceded_by_word};
 use crate::{
     Token,
     expr::{Expr, SequenceExpr},
@@ -27,9 +28,26 @@ impl ExprLinter for ThereOwn {
         &self.expr
     }
 
-    fn match_to_lint(&self, matched_tokens: &[Token], source: &[char]) -> Option<Lint> {
+    fn match_to_lint_with_context(
+        &self,
+        matched_tokens: &[Token],
+        source: &[char],
+        ctx: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
         let offender = matched_tokens.first()?;
         let template = offender.span.get_content(source);
+
+        // "there" preceded by a noun/pronoun with "own" followed by a noun/adjective
+        // is likely valid grammar: "people there own nice cars" (people who are there
+        // own nice cars). This does not apply to "they're" or "theyre" variants.
+        let is_there = template.eq_str("there");
+
+        if is_there
+            && preceded_by_word(ctx, |pw| pw.kind.is_nominal())
+            && followed_by_word(ctx, |nw| nw.kind.is_noun() || nw.kind.is_adjective())
+        {
+            return None;
+        }
 
         Some(Lint {
             span: offender.span,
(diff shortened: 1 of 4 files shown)


hippietrail:
Cool. This is one that I'd been looking at recently but had taken the deep dive into. In particular I was worried about a class of false positives I expect to be common: "people there own fancy cars".

I see the current code here does not attempt to avoid flagging these. There's a couple more in my feature request for this that you may or may not have seen but didn't reference: https://github.com/Automattic/harper/issues/3276

I'm guessing you made this with an AI. The AI tools I've used never include decent false positive checks until you give them a list of examples. They usually only do the kinds of negative checks I see at the bottom of the `mod tests` section here.

harper-core/src/linting/there_own.rs
```diff
@@ -0,0 +1,106 @@
...
+impl ExprLinter for ThereOwn {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        &self.expr
+    }
+
+    fn match_to_lint(&self, matched_tokens: &[Token], source: &[char]) -> Option<Lint> {
+        let offender = matched_tokens.first()?;
+        let template = offender.span.get_content(source);
```
hippietrail, line 38: See [my comment](https://github.com/Automattic/harper/pull/3336#issuecomment-4414169453) on how the PR in its current state does not try to avoid flagging false positives that I believe were pointed out in the original issue.

hippietrail (changes requested):
Sorry this needs to avoid some common false positives such as "people **there own** nice cars".

kimjune01:
Thanks for the review and for linking #3276 — I'd missed the issue, that example list is gold. Two commits after your review (8e799890, plus the new 70c46501) add a context guard that skips the lint when "there" is preceded by a noun or pronoun, since that pattern marks "there" as the adverb and "own" as the verb.

70c46501 adds 8 tests covering your example verbatim ("People there own expensive cars."), the four "there own" positive cases from #3276 (split data, customise effort lvl, modules c files, scripts create connection, silent appstore updates), two "they're own" cases (they're own command, they're own library), and a sentence-initial guard. `cargo test there_own` passes 16/16; `cargo run --bin harper-cli -- lint` reports "No lints found" on your example and flags the positive cases. Total harper-core lib suite: 5309 pass, 0 fail.

One thing I left as-is: "they're own" / "theyre own" still always lint, since you described those as "always a possessive error before own" and I couldn't find a counter-example. Happy to add a guard there too if you have a case in mind.

kimjune01:
Thanks for the concrete example, @hippietrail. The branch already covers that class.

In [70c4650](https://github.com/Automattic/harper/commit/70c465017c1db6da9cc329f4de73f8ad1d82591d) I added [`issue_3276_false_positive_expensive_cars`](https://github.com/Automattic/harper/blob/fix/there-own/harper-core/src/linting/there_own.rs#L187-L190) in `harper-core/src/linting/there_own.rs`:

```rust
#[test]
fn issue_3276_false_positive_expensive_cars() {
    assert_lint_count("People there own expensive cars.", ThereOwn::default(), 0);
}
```

CLI confirmation on the branch tip:

```
$ harper-cli lint /tmp/expensive.txt   # "People there own expensive cars."
No lints found

$ harper-cli lint /tmp/fancy.txt       # "people there own fancy cars"
No lints found

$ harper-cli lint /tmp/split.txt       # "Users can split data on there own topics."
[Word Choice::ThereOwn] (pri 31): Did you mean the possessive `their`?
```

Heuristic: if the token before `there` is nominal (noun/pronoun), it parses as the adverb "in that place" and `own` as the verb, so we suppress. In "split data on there own topics" the preceding token is "on" (preposition), which doesn't license the adverbial reading, so the lint fires.

Happy to add `"people there own fancy cars"` verbatim as a regression test if you want the exact string pinned.

hippietrail (approved):
Thanks for addressing everything!

elijah-potter (approved):
Looks good. Thanks!

PR #3349, opened 2026-05-11
feat: add backtick as another wrong-apostrophe mistake

hippietrail:
# Issues 
N/A

# Description

I was pretty sure people sometimes wrote `` ` `` instead of `'` just as they do with `;` and `´` but didn't want to implement it until I was sure in case it had some hidden side-effects.

Well yesterday I found such a comment on YouTube or somewhere:
> And It Won\`T Make One Bit Of Difference If I Answer Right Or Wrong. When You\`Re Rich, They Think You Really Know

# How Has This Been Tested?

A unit test using the sentence above.

# Checklist


elijah-potter (approved):
Looks great! Thanks!

PR #3352, opened 2026-05-11
feat(lint): wrong_preposition linter for common preposition errors

### LINKED ISSUE #3337: Let's have a "wrong preposition" tool in `harper-cli` and/or in the `justfile`
A common class of grammatical errors (in all languages actually) is using the correct preposition.

Sometimes there's no edge cases and we can use a Weir rule or `phrase_set_correction`.
But often that's too naïve and there are lots of false positives. So it's best to hunt for them before deciding how to write a new rule, or decide if it's even possible.

I usually use Google Search with quotes and `OR` operators but there's a better way.

`prep-pattern` (or dozens of other probably better names)

`harper-cli prep pattern install,installed,installing,installs P` (or some other format)

<s>This could invoke `harper-cli lint --only` with a `Linter` it makes on-the-fly, probably an `ExprLinter`</s>

On second thought, it wouldn't need a `Linter`, just an `Expr` on a `Document`.

If you have a corpus of files to test Harper on like I do (mine are README files from GitHub trending projects updated daily) you can pass in that directory and observe the output to see:

- All hits look like errors
- Repeated patterns of false positives
- Only some inflections lead to false positives
- etc.

----
Examples I was recently researching manually without such a tool:
- saddled to / saddled with
- inspired from / inspired by
- install in / install into / install on

quick123-666:
## Summary

This PR adds a new WrongPreposition linter for detecting common wrong preposition patterns in text.

### Patterns Detected
- different than/then ? different from
- similar with ? similar to
- married with/at/onto ? married to
- compared with ? compared to
- superior/inferior with ? superior/inferior to
- identical with ? identical to
- parallel with ? parallel to
- prefers on/at/with ? prefers to

### Implementation
- Uses SequenceExpr with WordSet for efficient pattern matching
- All patterns are unambiguous with clear corrections (low false positive risk)
- Resolves #3337

### Testing
- 25+ test cases covering all patterns
- Tests verify both flags incorrect usage and ignores correct usage


hippietrail:
Thanks. It doesn't resolve the feature request, since it's asking for a tool to help research wrong-preposition errors, not a linter which fixes wrong preposition errors.
Someone should review it soon anyway.

harper-core/src/linting/lint_group/mod.rs
hippietrail, line 511: Note that we try to keep these sorted alphabetically, and provide a shell command to help, especially if your editor of choice lets yet filter a selection of lines through a shell command.

harper-core/src/linting/lint_group/mod.rs
```diff
@@ -941,7 +946,7 @@ mod tests {
             spellcheck_lints[0]
                 .suggestions
                 .iter()
-                .any(|suggestion| suggestion.to_string() == "Replace with: “extension”")
+                .any(|suggestion| suggestion.to_string() == "Replace with: “extension�?)
```
hippietrail, line 949: Like in your other PR (if I haven't mixed them up) some tool you are using, or that your coding AI agent is using, is not set to handle UTF-8 correctly and is corrupting files. This no doubt causes tests to fail, which you should be noticing before `push`ing to the PR.

harper-core/src/linting/wrong_preposition.rs
```diff
@@ -0,0 +1,382 @@
...
+        "identical",
+        "parallel",
+    ]);
+
+    let wrong_preps = WordSet::new(&[
+        "than",   // "different than" → "different from"
+        "then",   // "different then" → "different from"
+        "with",   // "similar with" → "similar to", etc.
+        "onto",   // "married onto" → "married to"
+        "at",     // "married at" → "married to"
```
hippietrail, line 36: I believe some of these are already handled. You can check in a few ways such as pasting into [`https://writewithharper.com`](https://writewithharper.com), or using `harper-cli lint`. (I might be wrong though so worth checking.)

harper-core/src/linting/wrong_preposition.rs
```diff
@@ -0,0 +1,382 @@
...
+            ("prefers", prep) | ("prefer", prep) | ("preferring", prep) | ("preferred", prep) => {
+                if prep == "to" {
+                    return None; // already correct
+                }
+                ("to", "Use `prefers to` when indicating preference.")
+            }
+            _ => return None,
+        };
+
+        let span = tokens.last().unwrap().span;
```
hippietrail, line 126: Even though you've checked the length of `tokens` there's no reason not to use `?` rather than `.unwrap()`

harper-core/src/linting/wrong_preposition.rs
```diff
@@ -0,0 +1,382 @@
...
+    #[test]
+    fn ignores_compared_to() {
+        assert_lint_count(
+            "Life is often compared to a journey.",
+            WrongPreposition::default(),
+            0,
+        );
+    }
+
+    // Superior/inferior with → superior/inferior to
```
hippietrail, line 262: Is this an actual common error? I always research and double-check to rule out my own wrong assumptions by using Google to search for patterns on GitHub's domain to find real-world examples to use in unit tests, as well as getting a more accurate impression of how common a mistake or variant is.

harper-core/src/linting/wrong_preposition.rs
```diff
@@ -0,0 +1,382 @@
...
+        );
+    }
+
+    // Prefers on/at/with → prefers to
+    #[test]
+    fn flags_prefers_on() {
+        assert_suggestion_result(
+            "She prefers on tea over coffee.",
+            WrongPreposition::default(),
+            "She prefers to tea over coffee.",
```
hippietrail, line 334: The correction is not a grammatical English sentence.

hippietrail (changes requested):
Thanks. I've only done a partial review of this PR but please check my recommendations.

elijah-potter:
Since the PR author has not responded to @hippietrail's review, I am going to close it.

PR #3356, opened 2026-05-12
fix: no more dupe suggestions in `TransposedSpace`

### LINKED ISSUE #3355: The `TransposedSpace` linter can offer the same suggestion twice
Image

hippietrail:
# Issues 
Fixes #3355

# Description

Dupes were slipping through despite the code trying to prevent it.

It turned out to be a subtle problem between how the dictionary is case-folded, the old logic trying to add canonically-cased words, `replace_with_match_case` changing those cases anyway, and the dedupe logic using the default case-sensitive check.

So basically the logic might decide "fre dsmith" should suggest both "fred smith" and "Fred Smith", to respect both the case in the error and the case in the dictionary, correctly not dedupe them because that was intended only to remove exact duplicates, but then `replace_with_match_case` normalizing them both to "fred smith" after the dedupe check anyway.

The logic to get a canonical case spelling is flawed anyway since both "smith" and "Smith" will be in the dictionary and thus there are two canonical spellings but the API thinks there can only be one. There is a PR in progress to fix this and other lettercase bugs: #2630

So to fix this I simply removed the clever code that tried to offer multiple case versions and just use the ones in the doc and let `replace_with_match_case` do its thing.

`replace_with_match_case` has its own problems where it copies the case of each character by index rather than at the start or words or by titlecase logic or respecting proper nouns, etc.

We can always improve this again after #2630 is fully merged.

# How Has This Been Tested?

Unit test with the text that revealed the bug.

# Checklist


harper-core/src/linting/transposed_space.rs
```diff
@@ -170,12 +151,22 @@ mod tests {
...
     #[test]
-    fn test_late() {
+    fn test_3355_result() {
         assert_suggestion_result(
-            "Ands ometimes the space is a character late.",
+            "cham peng",
+            TransposedSpace::new(FstDictionary::curated()),
+            "champ eng",
+        );
+    }
```
elijah-potter, line 171: Why was this test removed or replaced?

hippietrail: > Why was this test removed or replaced?

Good question! Let me reinstate it.

relda88 (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

elijah-potter (approved):
Looks good! Thanks!

PR #3363, opened 2026-05-12
Fix FindFine false positive on 'is find <noun>'

### LINKED ISSUE #3255: False positive: `FindFine` flags "... is find someone" to change to "is fine someone"
Image

> Generally speaking what you want to try and do is find someone to replace you before you leave.

The linter is probably checking for "to be" before "find" to consider it an error. Maybe it should also look at the next word to see if it could be the object of the verb "find"?

SAY-5:
Fixes #3255

`FindFine` rewrites `is/was/... find` to `fine`, but `find` followed by a noun phrase is a legitimate verb use:

> Generally speaking what you want to try and do is find someone to replace you before you leave.

Added a postfix exception so the lint is suppressed when `find` is followed by a noun, proper noun, pronoun, or determiner (its object). Existing `fine` cases ("everything was fine for me", "I am fine not using ...") are unaffected since they are followed by ADP/PART/ADV or end of sentence.

Added a regression test; full `harper-core` test suite passes.

### DIFF
--- a/harper-core/src/linting/find_fine.rs
+++ b/harper-core/src/linting/find_fine.rs
@@ -1,7 +1,10 @@
+use harper_brill::UPOS;
+
 use crate::Token;
 use crate::expr::Expr;
 use crate::expr::SequenceExpr;
 use crate::patterns::InflectionOfBe;
+use crate::patterns::UPOSSet;
 
 use super::expr_linter::Chunk;
 use super::{ExprLinter, Lint, LintKind, Suggestion};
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/find_fine.rs
```diff
@@ -14,7 +17,15 @@ impl Default for FindFine {
     fn default() -> Self {
         let expr = SequenceExpr::with(InflectionOfBe::default())
             .t_ws()
-            .t_aco("find");
+            .t_aco("find")
+            // Don't flag when `find` is followed by something that could be its
+            // object, e.g. "what you want to do is find someone to replace you".
+            .then_unless(SequenceExpr::default().t_ws().then(UPOSSet::new(&[
```
hippietrail: Very minor but I think it helps clarity, you can use `SequenceExpr::whitespace()` instead of `SequenceExpr::default().t_ws()`

hippietrail (approved):
This looks good to me. I just have one tiny change to suggest that aids readability.

hippietrail (approved):
This looks good to me now I think.

elijah-potter (approved):
Looks good to me. Thanks!

PR #3365, opened 2026-05-12
Let me laugh

JasonTheAdams:
## Summary
- Adds a default-enabled `AllowInformalLaughter` config rule.
- Allows strictly alternating `ha`/`hah` laughter chains like `hah`, `haha`, and `hahahah`.
- Keeps malformed `hah`-style chains like `hahhah` linted, and leaves nearby interjections like `ah`, `ahah`, and `hehe` to existing behavior.

This is, admittedly, a small rule with a big emotional range.

## Tests
- `cargo test -p harper-core laughter`
- `cargo test -p harper-core default_config`
- `cargo test -p harper-core lint_descriptions_are_clean`
- `cargo run --bin harper-cli --release -- lint "hah"`
- `cargo run --bin harper-cli --release -- lint "I laughed hahahah."`
- `cargo run --bin harper-cli --release -- lint "hahhah"`


harper-core/src/linting/lint_group/mod.rs
```diff
@@ -297,6 +294,10 @@ pub struct LintGroup {
     clashing_linter_names: Option<Vec<String>>,
 }
 
+const ALLOW_INFORMAL_LAUGHTER: &str = "AllowInformalLaughter";
+const SPELL_CHECK: &str = "SpellCheck";
+const SPLIT_WORDS: &str = "SplitWords";
```
elijah-potter: I try to avoid constants like this, especially if they simply stand alone and aren't clearly connected to any particular part of the code. Would you mind inlining them?

harper-core/src/linting/lint_group/mod.rs
```diff
@@ -804,11 +796,16 @@ impl LintGroup {
...
 
         // Normal linters
         for (key, linter) in &mut self.linters {
             if self.config.is_rule_enabled(key) {
-                results.insert(key.to_owned(), linter.lint(document));
+                let mut lints = linter.lint(document);
+                if allow_informal_laughter && should_filter_informal_laughter_lints(key) {
+                    lints.retain(|lint| !is_informal_laughter(lint.get_ch(document.get_source())));
+                }
+                results.insert(key.to_owned(), lints);
```
elijah-potter: As a rule, I try to avoid letting one linter affect the output of another, especially in a composition location like the `LintGroup`. It gives the struct more responsibility than it should be given. 

Would it be possible to refactor this PR into a modification on the `RepeatedWords` rule, with some additions to the dictionary?

See:

- `harper-core/src/linting/repeated_words.rs`
- `harper-core/dictionary.dict`

elijah-potter (changes requested):
I appreciate the spirit of the rule, but I see that it's implementation is not as contained as I would prefer. Would you mind tweaking it to remove logic from `LintGroup`?

harper-core/src/linting/lint_group/mod.rs
JasonTheAdams: Hmm, so would this mean it's no longer a distinct rule that a user could disable? For things like academic papers and such they won't want this to be enabled.

elijah-potter: I agree. A separate rule, maybe called `SilenceLaughter` could be added that **stops** laughter. It would be pretty straightforward to implement.

This way, each rule's responsibility is maintained, which makes it easier to audit each one down the line.

I'd be happy to make the changes for you.

elijah-potter (approved):
This looks good. Thanks!

hippietrail:
This also reminds me of the `FillerWords` rule.

PR #3369, opened 2026-05-13
fix: allow lowercase mp3 after an

### LINKED ISSUE #3289: False positive: "an mp3" wrongly flagged to change to "a mp3"
Image

> You can also download an mp3

cfcryptotrader-rgb:
## Summary
- Treat lowercase `mp` followed by a digit (for example, `mp3`) as beginning with the vowel sound "em" for indefinite-article linting.
- Add regression coverage for both `an mp3` and `a mp3`.

Fixes #3289.

## Verification
- `cargo test -p harper-core linting::an_a -- --nocapture`
- `git diff --check`

### DIFF
--- a/harper-core/src/indefinite_article.rs
+++ b/harper-core/src/indefinite_article.rs
@@ -68,6 +68,10 @@ pub fn starts_with_vowel(word: &[char], dialect: Dialect) -> Option<InitialSound
         return Some(InitialSound::Either);
     }
 
+    if matches!(word, ['m', 'p', digit, ..] if digit.is_ascii_digit()) {
+        return Some(InitialSound::Vowel);
+    }
+
     if matches!(word, ['e', 'u', 'l', 'e', ..]) {
         return Some(InitialSound::Vowel);
     }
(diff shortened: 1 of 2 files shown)


harper-core/src/indefinite_article.rs
```diff
@@ -68,6 +68,10 @@ pub fn starts_with_vowel(word: &[char], dialect: Dialect) -> Option<InitialSound
         return Some(InitialSound::Either);
     }
 
+    if matches!(word, ['m', 'p', digit, ..] if digit.is_ascii_digit()) {
+        return Some(InitialSound::Vowel);
+    }
```
hippietrail, line 75: I think this is fine for now as uppercase `MP3` is already handled in the acronyms section and `mp-3` and `mp 3` are way less common that `mp3`. We can extend this if we find other "lowercase initialisms".

I think it's different enough from the other match arms here to warrant a comment though.

hippietrail (changes requested):
Looks good but add a comment please.

hippietrail (approved):
Looks good to me now. Thanks.

elijah-potter (approved):

hippietrail:
See also: #3531

PR #3373, opened 2026-05-13
feat(core): flag way when used for wait

### LINKED ISSUE #3345: Need to flag "way" when it's a mistake for "wait"
Another new one for me. Read it in one YouTube comment, Googled, and found it's a known thing.

I'm not sure if it also happens with "waits" or in other contexts but it definitely happens after "can't", its variants, and some very similar constructions.

Maybe this is new slang? I asked Google AI and it says it's not, but just a common mistake.

Examples:
- **can't way** for something better to be released.
- It **doesn't way** until completion though.
- Wow, just seeing this again, incredible, **can't way** to try it.
- **Can't way** to see things like this come into the project in the future .
- That's why `saveTransactionToDb` **doesn't way** for transactions to complete.
- If it's a real bother and you **can't way** for 2.10.16, go back to 2.10.12 (bug is only in 2.10.14).
- Hehe I **can't way** for a bevel node to be available in GN to avoid having to jump through hoops like that.

These patterns seem to occur in other contexts, some of which are too hard for me to figure out what the writers were trying to say, but probably not "wait".

- Reviewed this and was confused by spec with resolved value b/c line-height just takes you to css2 line ehight and doesn't **way** what resolved value is for line-height.
- Correct, the RFC does not **way** to ignore it, but also doesn't **way** what to do with it, only that the server MUST respond to it.
- I tried create+modify below file but that doesn't **way**.
-  can't **way** to proxy the strapi url to the url where nuxt is running.
- Without examining lunar in detail, I can't **way** whether this is simply a way to modify the sleep time for a particular request as per the DDC/CI spec
- the second "subscriber" probably doesn't **way** the end of the first "subscription".
- Users can't **way** anymore, so I delete from oc_share and oc_share_external the invalid link.
- Or I cannot way **how** to pass login information via some GET method or cookie.
- they are tentative (don't specify a lexical representation), eg schema:Number doesn't **way** what kind of number
- But in general the potential gain doesn't **way** up against the pain.
- `GroupByKey` which should but doesn't **way** for all elements
- but more importantly, it doesn't **way** half a ton

Some seem to be missing verbs before ***way***, such as "provide a way" or "find a way"? A couple are clearly supposed to be ***weigh*** so perhaps others too?

cfcryptotrader-rgb:
## Summary
- add a focused `WayWait` linter for `way` when it appears in clear `wait` contexts
- cover examples from #3345 such as `can't way to`, `cannot way for`, and `doesn't way until`
- keep the rule conservative by not flagging unclear `doesn't way what`-style cases

Fixes #3345.

## Verification
- `cargo fmt --check`
- `cargo test -p harper-core way_wait --lib`
- `cargo test -p harper-core --lib`
- `git diff --check`

### DIFF
--- a/harper-core/src/linting/way_wait.rs
+++ b/harper-core/src/linting/way_wait.rs
@@ -1,0 +1,117 @@
+use crate::expr::{Expr, FirstMatchOf, SequenceExpr};
+use crate::linting::expr_linter::Chunk;
+use crate::linting::{ExprLinter, Lint, LintKind, Suggestion};
+use crate::{CharStringExt, Token};
+
+/// Corrects `way` when it is likely a typo for the verb `wait`.
+pub struct WayWait {
+    expr: FirstMatchOf,
+}
+
+impl Default for WayWait {
+    fn default() -> Self {
+        let negative = SequenceExpr::word_set(&[
+            "can't", "cant", "cannot", "couldn't", "couldnt", "doesn't", "doesnt", "don't", "dont",
+            "won't", "wont",
+        ]);
+
+        let can_not = SequenceExpr::fixed_phrase("can not");
+
+        let following = SequenceExpr::word_set(&["for", "to", "until"]);
+
+        Self {
+            expr: FirstMatchOf::new(vec![
+                Box::new(
+                    SequenceExpr::with(negative)
+                        .t_ws()
+                        .t_aco("way")
+                        .t_ws()
+                        .then(following),
+                ),
+                Box::new(
+                    SequenceExpr::with(can_not)
+                        .t_ws()
+                        .t_aco("way")
+                        .t_ws()
+                        .t_set(&["for", "to", "until"]),
+                ),
+            ]),
+        }
+    }
+}
+
+impl ExprLinter for WayWait {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        &self.expr
+    }
+
+    fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {
+        let way_span = toks
+            .iter()
+            .find(|tok| tok.get_ch(src).eq_any_ignore_ascii_case_str(&["way"]))?
+            .span;
+
+        Some(Lint {
+            span: way_span,
+            lint_kind: LintKind::Typo,
+            suggestions: vec![Suggestion::replace_with_match_case(
+                "wait".chars().collect(),
+                way_span.get_content(src),
+            )],
+            message: "‘Way’ might be a typo here. Did you mean ‘wait’?".to_owned(),
+            priority: 31,
+        })
+    }
+
+    fn description(&self) -> &'static str {
+        "Corrects `way` when it appears to be a typo for `wait`."
+    }
+}
+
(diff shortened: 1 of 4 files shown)


harper-core/src/linting/way_wait.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Corrects `way` when it is likely a typo for the verb `wait`.
+pub struct WayWait {
+    expr: FirstMatchOf,
+}
+
+impl Default for WayWait {
+    fn default() -> Self {
+        let negative = SequenceExpr::word_set(&[
+            "can't", "cant", "cannot", "couldn't", "couldnt", "doesn't", "doesnt", "don't", "dont",
+            "won't", "wont",
```
hippietrail: Consider other auxiliary verbs but don't throw them all in blindly. I usually google with `site:github.com` to see if they find real mistakes and don't find non-mistakes.

But "shouldn't", "mightn't", "mustn't", "shan't".

There is actually a `ModalVerb` module but currently lacks a ctor to get only negatives. Note that the forms of "do" are auxiliaries but not modals, so if you were to integrate the use of that module you'd need to change from a `word_set` to something like `first_of` or `any_of` or `longest_of`.

harper-core/src/linting/way_wait.rs
```diff
@@ -0,0 +1,117 @@
...
+        ]);
+
+        let can_not = SequenceExpr::fixed_phrase("can not");
```
hippietrail: Oh this could also be integrated with the negative modal contractions by use of one of the first/any/longest`_of` type methods I just mentioned too.

harper-core/src/linting/way_wait.rs
```diff
@@ -0,0 +1,117 @@
...
+            "The RFC doesn't way what resolved value means.",
+            WayWait::default(),
+        );
+    }
+
+    #[test]
+    fn dont_flag_correct_wait() {
+        assert_no_lints("I can't wait to try it.", WayWait::default());
+    }
+}
```
hippietrail, line 150: Are these examples you just made up or are they from the feature request? If the feature request has a bigger set it can be a good idea to use all of them on linters that feel like they have a chance of false positives and/or false negatives. In my feature requests I try to include a range of real-world instances of the mistakes I found on GitHub.

hippietrail (commented):
Thanks. Overall it looks good. Since I don't know if you're 100% human, 100% AI or a human using an AI to some extent I haven't tailored my review for either case.

I've provided some feedback on changes to make and changes that require some consideration and may or may not be improvements.

cfcryptotrader-rgb:
Addressed the follow-up suggestions in `cc6f4a0b`:

- folded the two-token `can not` case into the same expression via `SequenceExpr::any_of`
- added the conservative negative modal contractions suggested in review (`shouldn't`, `mightn't`, `mustn't`, `shan't`, plus no-apostrophe variants)
- replaced the main positive tests with issue examples and added coverage for `can not`/negative modals

Verification:
- `rustup run stable cargo test -p harper-core way_wait --lib` (7 passed)
- `git diff --check`

elijah-potter:
Superceded by #3383.

PR #3374, opened 2026-05-13
fix(core): avoid WereWhere false positive after embedded question

### LINKED ISSUE #3214: False positive: "What do you think were the" -> "What do you think where the" (`WereWhere`)
**What got flagged?**
"What do you think **were** the…"

'were' corrected to 'where' by `WereWhere`

**Why is this incorrect?**
Here, `were` is the intended word.

In fact, even if 'where' was meant, it would be (at least in my opinion) a confusing use of grammar, and this should therefore be corrected the other way.

**Example of correct usage:**
```md
What do you think were the effects the olives had on the pasta's flavor?
```

cfcryptotrader-rgb:
## Summary
- Fixes #3214.
- Prevents `WereWhere` from rewriting `were` to `where` in the embedded-question pattern `What do you think were ...`.
- Adds a regression test using the issue sentence.

## Duplicate / spam checks
- Issue #3214 is open, unassigned, and has no comments/reservation.
- Checked open PRs for `3214`, `WereWhere`, `were where`, and `What do you think were`; no active duplicate found.
- Checked recent merged PRs; #3169 touched `WereWhere` but fixed different issues and does not cover #3214.

## Verification
- `rustup run stable cargo test -p harper-core were_where --lib` (35 passed, 3 ignored)
- `git diff --check`

### DIFF
--- a/harper-core/src/linting/were_where.rs
+++ b/harper-core/src/linting/were_where.rs
@@ -67,6 +67,33 @@ impl Default for WereWhere {
     }
 }
 
+fn looks_like_embedded_question_before_think(
+    toks: &[Token],
+    src: &[char],
+    context: Option<(&[Token], &[Token])>,
+) -> bool {
+    if !toks.first().is_some_and(|tok| {
+        matches!(tok.kind, TokenKind::Word(_)) && tok.span.get_content(src).eq_str("think")
+    }) {
+        return false;
+    }
+
+    let Some((before, _)) = context else {
+        return false;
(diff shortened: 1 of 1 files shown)


elijah-potter:
Can you confirm that you have read our agent policy?

https://elijahpotter.dev/articles/harper's-policy-on-agent-PRs

elijah-potter:
Since I have not heard from you, I am closing this PR.

PR #3376, opened 2026-05-13
fix(core): refine BetterOffPhrase contexts

### LINKED ISSUE #2863: False negative: "better of" is not flagged as a typo for "better off"
I just noticed I made this very typo in a comment on an issue thread yesterday:
> my suggestion is that for now it's better of in your User Dictionary.

I'm pretty sure we have one or multiple rules already that handle "of" vs "off" mixups as in the old issue for it: #507 

But none catch it in this context. To be fair, this one is context-sensitive and I'm not quite sure what the contexts will be, off the top of my head, that might distinguish which will be correct. But here's some I find via Googling that would be definite false positives in an overly naïve implementation:
- Apparently ECMA thought **better of** this restriction
- Curiosity got the **better of** me and I hacked up a test of this feature.
- This page describes who is the **better of** two players between and within games.
- but I think this is the **better of** the two paths
- I considered putting a default into the easybutton.css but though **better of** it.
-  phase sigmoidal fit is the **better of** the 2 sigmoidal fits
- only 0.5% compared to pngout (from **better of** /f0 and /f5 runs)

And here's some clear examples of the error that should be fixed:
- Otherwise I'm much **better of** not using the feed at all
- You may be **better of** starting from ChernSimons, which is based on KerrBL
- you are probably **better of** with other message broker solutions
  Image AHA! We already flag it when followed by "off" but as seen from the examples, we miss other examples.

Digging deeper, I see that besides the `BetterOffWith` Weir rule, we also have a `BetterOffPhrase` Weir rule. So modifying one or both of those is probably the way forward.

cfcryptotrader-rgb:
## Summary
- Fixes #2863.
- Narrows `BetterOffPhrase` so it still catches clear `better of` → `better off` adjective/adverbial contexts from the issue.
- Adds the issue's true-positive examples and false-positive examples as Weir tests.

## Duplicate / spam checks
- Issue #2863 is open, unassigned, and has no comments/reservation.
- Checked active PRs for `2863`, `better of`, and `better off`; no duplicate found.
- Checked recent merged PRs with the same terms; no duplicate found.

## Verification
- `rustup run stable cargo test -p harper-core run_tests_for_better_off_phrase --lib`
- `git diff --check`

### DIFF
--- a/harper-core/src/linting/weir_rules/BetterOffPhrase.weir
+++ b/harper-core/src/linting/weir_rules/BetterOffPhrase.weir
@@ -1,4 +1,5 @@
-expr main (better of)
+expr beBefore [am, are, is, was, were, be, been, being, feel, feels, felt]
+expr main <([(@beBefore better of), (@beBefore ADV better of), (better of [with, without, at, before, now, if, not]), (better of VERB)]), (better of)>
 
 let message "This expression usually takes `off` instead of `of`."
 let description "Rewrites `better of` to `better off` in common comparative phrasing."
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/weir_rules/BetterOffPhrase.weir
```diff
@@ -26,6 +30,12 @@ test "Of course this is better overall." "Of course this is better overall."
 test "He offered better outcomes, not quick fixes." "He offered better outcomes, not quick fixes."
```
hippietrail: Oh you can use `allows` followed by just one copy of the sentence.
It's a bit newer than using `test` followed by two copies, so there are still linters in the repo using it.

hippietrail (changes requested):
I'm not good at more advanced Weir but if all those tests are passing it's probably ready. I did point out that you can use `allows` for the negative tests.

elijah-potter (approved):

PR #3377, opened 2026-05-13
fix(core): avoid InflectedVerbAfterTo false positive after pointed

### LINKED ISSUE #2575: The base form of the verb is needed here.
> It is used to verify that the object being pointed **to outlives** the pointer ■ The base form of the verb is needed here.

Pretty sure the sentence is correct :)

cfcryptotrader-rgb:
## Summary
- Fixes #2575.
- Treats `pointed to` as a prepositional construction for `InflectedVerbAfterTo`, avoiding the false positive in `object being pointed to outlives ...`.
- Applies the existing infinitive-vs-nominal heuristic to `-s`/`-es` stems before linting, instead of unconditionally flagging them.
- Adds the issue sentence as a regression test.

## Duplicate / spam checks
- Issue #2575 is open and unassigned.
- Checked comments: no reservation/claim; only maintainer context explaining the passive-participle construction.
- Checked active PRs for `2575`, `InflectedVerbAfterTo`, `pointed to outlives`, and `outlives`; no duplicate found.
- Checked recent merged PRs with the same terms; no duplicate found.

## Verification
- `rustup run stable cargo test -p harper-core inflected_verb_after_to --lib` (15 passed, 4 ignored)
- `git diff --check`

### DIFF
--- a/harper-core/src/linting/inflected_verb_after_to.rs
+++ b/harper-core/src/linting/inflected_verb_after_to.rs
@@ -79,6 +79,10 @@ impl<T: Dictionary> Linter for InflectedVerbAfterTo<T> {
             let ed_specific_heuristics = || {
                 if let Some(prev) = document.get_next_word_from_offset(pi, -1) {
                     let prev_chars = document.get_span_content(&prev.span);
+                    if prev_chars.eq_str("pointed") {
+                        return ToVerbExpects::ExpectsNominal;
+                    }
+
                     if let Some(metadata) = self.dictionary.get_word_metadata(prev_chars) {
                         // adj: "able to" expects an infinitive verb
                         // verb: "have/had/has/having to" expect an infinitive verb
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/inflected_verb_after_to.rs
```diff
@@ -107,15 +111,13 @@ impl<T: Dictionary> Linter for InflectedVerbAfterTo<T> {
...
-                if es {
+                if es && ed_specific_heuristics() == ExpectsInfinitive {
                     lint_from_stem(&chars[..chars.len() - 2]);
                 };
             }
             if chars.ends_with(&['s']) {
                 let s = check_stem(&chars[..chars.len() - 1]);
-                // Add -s specific heuristics when needed
-                if s {
+                if s && ed_specific_heuristics() == ExpectsInfinitive {
```
hippietrail: If the same heuristics work for the -es and -s cases it should maybe be renamed so as to avoid being a misnomer that might confuse.

hippietrail (changes requested):
My brain is too fried to a full PR review but I think your helper function could have a clearer name.

elijah-potter:
@cfcryptotrader-rgb, can you confirm you've read our agent policy?

elijah-potter:
Since I haven't heard from @cfcryptotrader-rgb, I'm going to close this and other PRs.

PR #3379, opened 2026-05-13
Add verb past form only helpers

### LINKED ISSUE #3372: Add an `is_past_participle_only` and maybe an `is_simple_past_only` method to `DictWordMetadata` and `TokenKind`
Sometimes we want to check if a token is a past participle to flag something like "I seen it yesterday".
Or the opposite when a token is a simple past form like "I have went ahead".

But we can't just check for one or the other since for regular forms both are fine: "I thought about it yesterday", "I have thought about it a lot".

Currently in a linter we have to check for one and remember to also check for NOT the other.

For mass nouns we have a similar situation but it has a specialized `is_` method
```
    // Most mass nouns also have countable senses. Match those that are only mass nouns.
    pub fn is_mass_noun_only(&self) -> bool {
        if let Some(noun) = self.noun {
            matches!(
                (noun.is_countable, noun.is_mass),
                (None | Some(false), Some(true))
            )
        } else {
            false
        }
    }
```
The same can be done for verb past forms.

impetus82:
## Summary

- add `DictWordMetadata::is_verb_simple_past_only`
- add `DictWordMetadata::is_verb_past_participle_only`
- expose both helpers through `TokenKind`
- cover irregular-only forms (`ate`, `eaten`) and ambiguous/regular forms (`thought`, `walked`)

Fixes #3372.

## Testing

- `cargo fmt --package harper-core`
- `cargo test -p harper-core past_only --lib`
- `cargo test -p harper-core participle_only --lib`
- `cargo test -p harper-core neither_past_form_only --lib`
- `git diff --check`

### DIFF
--- a/harper-core/src/dict_word_metadata.rs
+++ b/harper-core/src/dict_word_metadata.rs
@@ -480,6 +480,13 @@ impl DictWordMetadata {
         })
     }
 
+    pub fn is_verb_regular_past_form(&self) -> bool {
+        self.verb.is_some_and(|v| {
+            v.verb_forms
+                .is_some_and(|vf| vf.contains(VerbFormFlags::PAST))
+        })
+    }
+
     pub fn is_verb_simple_past_form(&self) -> bool {
         self.verb.is_some_and(|v| {
             v.verb_forms
@@ -1982,6 +1989,12 @@ pub mod tests {
         }
 
         #[test]
+        fn regular_past_walked() {
+            let md = md("walked");
+            assert!(md.is_verb_regular_past_form())
+        }
+
+        #[test]
         fn simple_past_ate() {
             let md = md("ate");
             assert!(md.is_verb_simple_past_form())
@@ -2019,6 +2032,14 @@ pub mod tests {
             let md = md("walked");
             assert!(!md.is_verb_simple_past_only());
             assert!(!md.is_verb_past_participle_only());
+            assert!(md.is_verb_regular_past_form());
+        }
+
+        #[test]
+        fn irregular_past_forms_are_not_regular_past() {
+            assert!(!md("ate").is_verb_regular_past_form());
+            assert!(!md("eaten").is_verb_regular_past_form());
+            assert!(!md("thought").is_verb_regular_past_form());
         }
 
         #[test]
--- a/harper-core/src/dict_word_metadata.rs
+++ b/harper-core/src/dict_word_metadata.rs
@@ -494,6 +494,24 @@ impl DictWordMetadata {
         })
     }
 
+    pub fn is_verb_simple_past_only(&self) -> bool {
+        self.verb.is_some_and(|v| {
+            v.verb_forms.is_some_and(|vf| {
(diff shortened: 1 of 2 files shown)


harper-core/src/dict_word_metadata.rs
```diff
@@ -1975,6 +1993,34 @@ pub mod tests {
...
+
+        #[test]
+        fn eaten_is_past_participle_only() {
+            let md = md("eaten");
+            assert!(md.is_verb_past_participle_only());
+            assert!(!md.is_verb_simple_past_only());
+        }
+
+        #[test]
+        fn thought_is_neither_past_form_only() {
```
hippietrail, line 2025: Hey you know what would be even better, adding an `is_verb_regular_past_form` that does exactly this.

hippietrail (changes requested):
It looks good to me. Your implementation was quite different from the mass noun one so I asked two AIs to compare them since I didn't trust my own Rust-foo.

Your test cases gave me an idea for a third you method you could trivially add that complements these.

impetus82:
Thanks for the suggestion. I added `is_verb_regular_past_form` alongside the existing helpers, delegated it through `TokenKind`, and added coverage for `walked` plus irregular past forms (`ate`, `eaten`, `thought`).

Local checks run:
- `cargo fmt --package harper-core`
- `cargo test -p harper-core regular_past --lib`
- `cargo test -p harper-core past_only --lib`
- `cargo test -p harper-core participle_only --lib`
- `git diff --check`

harper-core/src/dict_word_metadata.rs
```diff
@@ -480,6 +480,13 @@ impl DictWordMetadata {
         })
     }
 
+    pub fn is_verb_regular_past_form(&self) -> bool {
+        self.verb.is_some_and(|v| {
+            v.verb_forms
+                .is_some_and(|vf| vf.contains(VerbFormFlags::PAST))
```
hippietrail: Actually I have a hunch that `::PAST` might be unreliable. I think a more explicit way would be more trustworthy:
```
            v.verb_forms
                .is_some_and(|vf| vf.contains(VerbFormFlags::PRETERITE)
                               && vf.contains(VerbFormFlags::PAST_PARTICIPLE))
```

hippietrail (changes requested):

harper-core/src/dict_word_metadata.rs
impetus82: Thanks, updated it to use the explicit `PRETERITE` + `PAST_PARTICIPLE` check instead of relying on `PAST`. I also adjusted the regular-past tests around `thought` and kept the distinct-form checks for `ate`, `eaten`, and `walked`.

Local checks run:
- `cargo fmt --package harper-core`
- `cargo test -p harper-core regular_past --lib`
- `cargo test -p harper-core past_only --lib`
- `cargo test -p harper-core participle_only --lib`
- `git diff --check`

hippietrail (approved):

elijah-potter (approved):
This looks good to me too. Thanks!

PR #3399, opened 2026-05-16
refactor: `OwnedExprExt::and_not` → `but_not`

hippietrail:
# Issues 
N/A

# Description

It kept gnawing at me that the `and_not` method on `OwnedExprExt` would be better expressed as `but_not`

I tried to ask an AI to adjudicate without letting it see which I was biased toward:

image

# How Has This Been Tested?

`cargo test`

# AI Disclosure

Well I didn't use an AI for the code but I did use an AI to decide if the change is warranted.

# Checklist


elijah-potter (approved):
I agree. This is technically a breaking change, but I doubt it will affect anyone.

PR #3401, opened 2026-05-16
feat: don't lint `.DS_Store` or `.json` files in dirs in `harper-cli lint` and `just lint`

hippietrail:
# Issues 

N/A

# Description

If you've used `harper-cli lint` or `just lint` on the commandline very much, you may have seen that it will try to lint files in directories that are not meaningfully lintable.

If you're on a Mac you also might've seen warnings about bad UTF-8 and not known what files it's complaining about.

This PR skips linting JSON files for the former, and `.DS_Store` files for the latter. If you do have a JSON file you want to lint you can pass it directly to the command, but if they're within a directory you pass to the command they will be ignored.

This makes testing on random directories faster and the output less cluttered.

# How Has This Been Tested?

Manually

# AI Disclosure

# Checklist


harper-cli/src/input/multi_input.rs
```diff
@@ -52,15 +52,16 @@ pub(crate) struct DirInput {
...
-                None
-            }
-        }))
+        let iter = std::fs::read_dir(&self.path)?
+            .flatten()
+            .map(|e| e.path())
+            // Skip binary files commonly in directories full of text files.
+            .filter(|p| p.file_name().is_some_and(|n| n != ".DS_Store"))
+            // Skip text-based file formats which don't deterministically contain English prose.
+            .filter(|p| !p.to_string_lossy().to_lowercase().ends_with(".json"))
```
elijah-potter: Could we refactor these lines into a constant array of `&'static str` that gets iterated through? Otherwise, this looks great!

elijah-potter (changes requested):

harper-cli/src/input/multi_input.rs
hippietrail: > Could we refactor these lines into a constant array of `&'static str` that gets iterated through? Otherwise, this looks great!

I was going to do that but I was only certain about one file type for each kind.
Did you notice that the first is for whole filenames and the second is for filename extensions?

If you know of some in each category for Linux or Windows I'll add them!

elijah-potter: I should have read closer. Does `harper-cli` currently try to read `.json` files?

hippietrail: > I should have read closer. Does `harper-cli` currently try to read `.json` files?

I know it did at one point a long time ago but I'm not sure if it still does ... the answer is YES! <s>Right I'll add them to the list to skip - great idea!</s>
Haha! - shows how long since I've dug into this code that I forgot `.json` was in there ... but I've remembered macOS has another class of files to avoid, this time by prefix `._` so I'll add that...

elijah-potter (approved):
Seems reasonable.

Thanks!

PR #3406, opened 2026-05-17
feat: `AnchorEnd` finally fixed

### LINKED ISSUE #1262: A Pattern that can match "end of chunk"
## Description

`PatternLinter`s currently run inside a chunk, which is delimited by commas as well as anything that breaks higher-level pieces such as sentence punctuation and paragraph breaks.

Sometimes a way to tell if something is a mistake or not is if anything follows it.

My current use case is an example:

"everyday" is an attributive adjective so must be followed by a noun and can't end a sentence.
"every day" is an adverb and can end a sentence.
One condition to see if "everyday" is a mistake for "every day" is if nothing follows it. Currently we can't do that.

## Component

- Core grammar checking engine

### LINKED ISSUE #1731: `AnchorStar` and `AnchorEnd` do not match start and end of **chunks** as expected.

hippietrail:
# Issues 

Fixes #1262
Fixes #1731

# Description

I've tried to get this working twice before, on my own and with LLMs.
I thought I'd give it another shot with an agent with a bigger context window that can look through more source files.

It seems to finally be working but needs a proper review.

I completed an old linter that was supposed to match at the end of a chunk but never worked.

# How Has This Been Tested?

- Implemented in a linter, `Addicting`, verified that it didn't work without this patch and does work with it.
- Added unit tests to `AnchorEnd`, `SequenceExpr`, and `Addicting`.
- `cargo test`

# AI Disclosure

- [x] <s>The sentences in the</s> unit tests were generated by an AI.

I hunted down real-world sentences for `Addicting` but the AI made 2/3 of the lower-level tests. Some were then hand-tweaked, others were not.

# Checklist

### DIFF
--- a/harper-core/src/expr/anchor_end.rs
+++ b/harper-core/src/expr/anchor_end.rs
@@ -21,10 +21,10 @@ impl Step for AnchorEnd {
         // Match if cursor is at or past the last non-whitespace token
         // This allows AnchorEnd to work in sequences where the cursor has advanced
         // past the matched content, including when cursor is past the end of the token stream
-        if let Some(last) = last_non_ws {
-            if cursor >= last {
-                return Some(0);
-            }
+        if let Some(last) = last_non_ws
+            && cursor >= last
+        {
+            return Some(0);
         }
 
         None
--- a/harper-core/src/expr/anchor_end.rs
+++ b/harper-core/src/expr/anchor_end.rs
@@ -10,26 +10,31 @@ pub struct AnchorEnd;
 
 impl Step for AnchorEnd {
     fn step(&self, tokens: &[Token], cursor: usize, _source: &[char]) -> Option<isize> {
-        if tokens
+        let last_non_ws = tokens
             .iter()
             .enumerate()
             .rev()
             .filter(|(_, t)| !t.kind.is_whitespace())
             .map(|(i, _)| i)
-            .next()
-            == Some(cursor)
-        {
-            Some(0)
-        } else {
-            None
+            .next();
+
+        // Match if cursor is at or past the last non-whitespace token
+        // This allows AnchorEnd to work in sequences where the cursor has advanced
+        // past the matched content, including when cursor is past the end of the token stream
+        if let Some(last) = last_non_ws {
+            if cursor >= last {
+                return Some(0);
+            }
         }
+
+        None
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::expr::ExprExt;
-    use crate::{Document, Span};
+    use crate::expr::{AnchorStart, ExprExt, SequenceExpr};
+    use crate::{Document, Span, TokenStringExt};
 
     use super::AnchorEnd;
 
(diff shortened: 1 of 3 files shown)


elijah-potter (approved):
Looks good to me! Thank you!

PR #3407, opened 2026-05-17
fix(core): replace Arc with Lrc in call_them and should_contract

Dronakurl:
# Issues

N/A

# Description

Replaces `std::sync::Arc` with `crate::sync::Lrc` in `call_them.rs` and `should_contract.rs`. `Lrc` is Harper's type alias that selects `Rc` or `Arc` based on the `concurrent` feature, ensuring proper `Send` + `Sync` bounds.

# Testing

- `cargo fmt --check`
- `cargo clippy --package harper-core -- -Dwarnings`

### DIFF
--- a/harper-core/src/linting/call_them.rs
+++ b/harper-core/src/linting/call_them.rs
@@ -1,7 +1,8 @@
-use std::{ops::Range, sync::Arc};
+use std::ops::Range;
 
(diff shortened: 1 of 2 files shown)


hippietrail (approved):
Well spotted. Looks good. Thanks.

elijah-potter (approved):

PR #3412, opened 2026-05-18
fix(harper-core): correct 'get access at' to 'get access to' in scoped contexts

### LINKED ISSUE #3309: Should correct "get access at" to "get access to"
I just heard this in a retrocomputing YouTube video by a native speaker:

- they really need to tabulate that spreadsheet and and they need to **get access at** some CPU time.

Even though I'd never heard this one before it's out there on GitHub:
- To **get access at** the original request within the hooks, use http middleware to inject the body and pass it around with a request context.
- Succeeded to **get access at** host system with login command ?
- ... and if someone manages to **get access at** the preferences, at least it's going to be a bit more difficult to decrypt.
- in order to **get access at models**, you typically send information to some server that runs the model and response is sent back.

There are also common false positives if this were to be implemented naïvely:
- But anyway, i don't manage to **get access at all**, so i will go another way now and try to download it all as zip from a server into the app's files folder.
- **Getting access at** the /dev/disk2 level would not normally be called a mount of /dev/disk2 that I'm aware of.
- Is it possible to **get access at least** to: file_content_type, file_size and name?
- **Get access at** getpara.com.
- **Get Access at** https://developer.bitping.com.
- Contact us to **get access at** https://tanker.io/contact.
- **get access at** https://github.com/malaysia-ai/label-studio#how-to-get-access
- Eventually we will stop using this but the quickest method **to get access at this time** would be to fill out this form.
- Today I **got access at least** to the nucleo_f767zi and I am trying the latest main branch.

It also looks like there are ambiguous cases that make sense with both "at" and "to":
- but if I set in the client "Authentication=Anonymous" still I can **got access at** server without any Username/Password!!

I also saw this one twice, which seems to be a different nonnative speaker mistake for "**can be accessed at**":
- The license text in full can be **getting access at** the file named COPYING-GPL-3.0.

favzqn:
# Issues
Closes #3309

# Description
Adds new scoped Weir rule `GetAccessTo` in:
- `harper-core/src/linting/weir_rules/GetAccessTo.weir`

Rule corrects `get access at` -> `get access to` for high-confidence resource/object contexts (e.g. features, modules, repo, API, settings, models), while avoiding naive broad matching that would create false positives in time/location/URL phrases.

Included positive and negative (`allows`) coverage in the rule tests.

# Demo
CLI check examples:
```bash
cargo run --bin harper-cli -- lint "Users can get access at premium features."
# => flags GetAccessTo
cargo run --bin harper-cli -- lint "Users can get access to premium features."
# => no lints
cargo run --bin harper-cli -- lint "Get access at https://developer.bitping.com."
# => no GetAccessTo lint
```

How Has This Been Tested?
- Ran harper-cli manually on target sentence and control sentences.
- Confirmed issue case is flagged.
- Confirmed correct phrase is not flagged.
- Confirmed URL-style phrase is not flagged.
- Confirmed get access at all is not flagged by this rule (only unrelated capitalization lint appears in sample sentence).

AI Disclosure

If Your PR Implements or Enhances a Linter

Checklist


elijah-potter (changes requested):
It looks like you need to add an entry to `default_config.json` in order for the tests to pass.

Otherwise, this looks good! Thank you!

harper-core/src/linting/weir_rules/GetAccessTo.weir
```diff
@@ -0,0 +1,23 @@
+expr main [(get access at premium), (get access at private), (get access at beta), (get access at admin), (get access at paid), (get access at dashboard), (get access at api), (get access at repo), (get access at repository), (get access at features), (get access at tools), (get access at modules), (get access at reports), (get access at settings), (get access at preferences), (get access at models), (get access at resources)]
```
hippietrail: I'm not great at the newer Weir features but is there a way to declare the list of nouns in one place and then the construction rather than repeat the whole construction over and over?

I'd definitely prefer the list in alphabetical order for humans scanning it with their eyes (-:

favzqn: Thanks for the suggestion, I tested the shared-list form locally before updating.

I tried changing:
- from explicit alternatives in expr main [...]
- to a reusable list:
  - expr access_target [...]
  - expr main (get access at @access_target)

In theory those should be equivalent for these one-word targets, but in my local run they were not.
I used this check:
```
cargo run --bin harper-cli -- lint "Users can get access at premium features."
cargo run --bin harper-cli -- lint "Users can get access to premium features."
```

Observed result:
- explicit expr main [...] form:  
  - get access at premium features is flagged (GetAccessTo)  
  - get access to premium features is not flagged
- shared-list form:  
  - get access at premium features was not flagged for me

hippietrail: Thanks for trying though - and thanks for sorting alphabetically.

PR #3418, opened 2026-05-18
feat: a/an + plural noun → a/an + singular noun

### LINKED ISSUE #3286: Need to flag indefinite article vs noun number agreement error
Some real-world grammar problems from our Discord that Harper doesn't flag:

Image

> ... but I have included **a** notic**es** on my landing ...

"**a notices**" should be either "**a notice**" or "**notices**" (or "**some notices**")

Agreement is hard because there are so many factors and nouns, verbs, and other POSes can all be phrases as well as verbs.

But I'm starting to think that we could tackle agreement problems piecemeal as we encounter the most common ones, by using merged linters of small parts that handle one focused case each.

azmifarih:
Fixes #3286

## Description
Adds `APluralNoun`, a Rust linter for the direct `a/an + plural noun` agreement case.

Examples:
- `a notices` -> `a notice`
- `a apples` -> `an apple`
- `an cars` -> `a car`

The rule uses dictionary-backed singular noun suggestions, including regular forms like `notices -> notice` and irregular forms like `children -> child`.

This first pass intentionally keeps the scope narrow. It handles direct article+noun cases only and does not attempt adjective phrases such as `a very important notices`.

## How Has This Been Tested?
- `cargo test -p harper-core a_plural_noun`
- `cargo test -p harper-core curated_default_config_lists_every_registered_rule`
- `cargo test -p harper-core --test linters`
- `cargo test -p harper-core`

## AI Disclosure

## If Your PR Implements or Enhances a Linter

## Checklist

### DIFF
--- a/harper-core/src/linting/a_plural_noun.rs
+++ b/harper-core/src/linting/a_plural_noun.rs
@@ -1,8 +1,13 @@
+use harper_brill::UPOS;
+
 use crate::{
     Dialect, IrregularNouns, Lint, Token, TokenStringExt,
     expr::{Expr, SequenceExpr},
     indefinite_article::{InitialSound, starts_with_vowel},
-    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk},
+    linting::{
+        ExprLinter, LintKind, Suggestion,
+        expr_linter::{Chunk, at_start_of_sentence},
+    },
     regular_nouns,
     spell::Dictionary,
 };
@@ -18,6 +23,7 @@ impl<D: Dictionary> APluralNoun<D> {
             expr: SequenceExpr::default()
                 .then_indefinite_article()
                 .t_ws()
+                .then_zero_or_more(SequenceExpr::default().then_non_verb_adjective().t_ws())
                 .then_plural_noun(),
             dict,
         }
@@ -27,7 +33,12 @@ impl<D: Dictionary> APluralNoun<D> {
 impl<D: Dictionary> ExprLinter for APluralNoun<D> {
     type Unit = Chunk;
 
-    fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {
+    fn match_to_lint_with_context(
+        &self,
+        toks: &[Token],
+        src: &[char],
+        ctx: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
         let span = toks.span()?;
         let noun = toks.last()?;
 
@@ -35,16 +46,36 @@ impl<D: Dictionary> ExprLinter for APluralNoun<D> {
             return None;
         }
 
+        if noun.kind.is_verb_third_person_singular_present_form()
+            && looks_like_third_person_verb_use(noun, ctx)
+        {
+            return None;
+        }
+
+        let is_start_of_sentence = at_start_of_sentence(ctx);
+
+        let article = toks.first()?;
         let plural = noun.span.get_content(src);
         let suggestions = singular_noun_suggestions(&self.dict, plural)
             .into_iter()
             .map(|singular| {
-                let mut replacement = indefinite_article_for(&singular).to_vec();
-                replacement.push(' ');
+                let article_target = article_target(toks, noun, &singular, src)?;
+                let mut replacement = indefinite_article_for(article_target).to_vec();
+                replacement.extend(&src[article.span.end..noun.span.start]);
                 replacement.extend(singular);
 
-                Suggestion::replace_with_match_case(replacement, span.get_content(src))
+                let mut case_template = span.get_content(src).to_vec();
+                if is_start_of_sentence && let Some(first) = case_template.first_mut() {
+                    first.make_ascii_uppercase();
+                }
+
+                Some(Suggestion::replace_with_match_case(
+                    replacement,
+                    case_template,
+                ))
             })
+            .collect::<Option<Vec<_>>>()?
+            .into_iter()
             .collect::<Vec<_>>();
 
         if suggestions.is_empty() {
@@ -66,6 +97,46 @@ impl<D: Dictionary> ExprLinter for APluralNoun<D> {
 
     fn description(&self) -> &str {
         "Corrects plural nouns after the indefinite article `a` or `an`."
+    }
+}
+
+trait SequenceExprExt {
+    fn then_non_verb_adjective(self) -> Self;
+}
+
+impl SequenceExprExt for SequenceExpr {
+    fn then_non_verb_adjective(self) -> Self {
+        self.then(|tok: &Token, _src: &[char]| tok.kind.is_adjective() && !tok.kind.is_verb())
+    }
+}
+
+fn looks_like_third_person_verb_use(noun: &Token, ctx: Option<(&[Token], &[Token])>) -> bool {
+    let Some((_, after)) = ctx else {
+        return false;
+    };
+
+    let mut non_ws = after.iter().filter(|tok| !tok.kind.is_whitespace());
+    let Some(next) = non_ws.next() else {
+        return noun.kind.is_upos(UPOS::VERB);
+    };
+
+    (noun.kind.is_upos(UPOS::VERB) && next.kind.is_sentence_terminator())
+        || next.kind.is_pronoun()
+        || next.kind.is_upos(UPOS::PRON)
+}
+
+fn article_target<'a>(
+    toks: &'a [Token],
+    noun: &Token,
+    singular: &'a [char],
+    src: &'a [char],
+) -> Option<&'a [char]> {
+    let first_after_article = toks.iter().skip(1).find(|tok| !tok.kind.is_whitespace())?;
+
+    if first_after_article.span == noun.span {
+        Some(singular)
+    } else {
+        Some(first_after_article.span.get_content(src))
     }
 }
 
@@ -136,6 +207,44 @@ mod tests {
     #[test]
     fn corrects_irregular_plural() {
         assert_suggestion_result("I saw a children.", linter(), "I saw a child.");
+    }
+
+    #[test]
+    fn corrects_plural_after_adjective() {
+        assert_suggestion_result(
+            "A beautiful girls is sitting in the chair now.",
+            linter(),
+            "A beautiful girl is sitting in the chair now.",
+        );
+    }
+
+    #[test]
+    fn capitalizes_plural_after_adjective_at_sentence_start() {
+        assert_suggestion_result(
+            "an beautiful girls is sitting in the chair now.",
+            linter(),
+            "A beautiful girl is sitting in the chair now.",
+        );
+    }
+
+    #[test]
+    fn corrects_article_before_adjective() {
+        assert_suggestion_result("I saw an red cars.", linter(), "I saw a red car.");
+    }
+
+    #[test]
+    fn corrects_article_before_vowel_sound_adjective() {
+        assert_suggestion_result("I saw a old errors.", linter(), "I saw an old error.");
+    }
+
+    #[test]
+    fn allows_third_person_verb_after_modified_singular_subject() {
+        crate::linting::tests::assert_no_lints("A predicate adjective follows.", linter());
+    }
+
+    #[test]
+    fn allows_third_person_verb_after_adjective_subject() {
+        crate::linting::tests::assert_no_lints("An auxiliary precedes it.", linter());
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


harper-core/src/linting/a_plural_noun.rs
```diff
@@ -0,0 +1,289 @@
...
+            "A beautiful girl is sitting in the chair now.",
+        );
+    }
+
+    #[test]
+    fn capitalizes_plural_after_adjective_at_sentence_start() {
+        assert_suggestion_result(
+            "an beautiful girls is sitting in the chair now.",
+            linter(),
+            "A beautiful girl is sitting in the chair now.",
```
hippietrail: Watch out, this one is both creating an extra correction, and accepting it!
"an" should not be changed to "A" here. There is a separate linter for that job that user may have turned off. More importantly, there are contexts and styles in which sentence fragments are OK or capitals are intentionally avoided.

harper-core/src/linting/a_plural_noun.rs
```diff
@@ -125,6 +139,40 @@ fn looks_like_third_person_verb_use(noun: &Token, ctx: Option<(&[Token], &[Token
...
+        "onwards",
+        "outwards",
+        "upwards",
+    ])
+}
+
+fn is_great_many_phrase(toks: &[Token], src: &[char]) -> bool {
+    let words = toks
+        .iter()
+        .filter(|tok| tok.kind.is_word())
```
hippietrail: This is going to ignore punctuation between words which will result in false matches. You want to check that it's strictly word/ws/word/ws/word etc. Then again lists of adjectives or adverbs can have commas between them too.

That's not to say we always go down the rabbit hole to match everything we know is possible. but it's worth knowing where the edge between what a linter actually handles and what it would ideally handle but doesn't yet.

harper-core/src/linting/a_plural_noun.rs
azmifarih: Thanks, that makes sense. I removed the sentence-start capitalization handling from this linter.

The suggestion now only fixes the article/noun agreement part, so `an beautiful girls` becomes `a beautiful girl` rather than also applying capitalization. Sentence capitalization should stay with the separate capitalization rule.

harper-core/src/linting/a_plural_noun.rs
azmifarih: Thanks, I tightened this guard.

It no longer collects only word tokens and ignores everything else. It now checks the matched token sequence directly for `a` / whitespace / `great` / whitespace / `many`, so punctuation between those words is not silently skipped.

I’m keeping the broader adjective/adverb-list cases out of scope here unless we get concrete corpus or fuzz examples.

hippietrail:
If you are feeling ambitious you could enhance this to suggest the other side of the coin where the plural was intended and the article was the mistake. I'm not sure if my is there / are there / there is / there are linter got merged or not, but it tries to handle this. Off the top of my head you suggest one with no indefinite noun and one with "some".

This could also be a separate linter but at the moment, suggestions are not combined from two linters on the same mistake/span. I've got a bug report / feature request for that but I don't think anyone is working on a strategy yet.

azmifarih:
> If you are feeling ambitious you could enhance this to suggest the other side of the coin where the plural was intended and the article was the mistake. I'm not sure if my is there / are there / there is / there are linter got merged or not, but it tries to handle this. Off the top of my head you suggest one with no indefinite noun and one with "some".
> 

Yep, I already added that in the same lint. It now offers the singular fix, the no-article form, and the some form together, so we don’t depend on cross-linter suggestion merging.

harper-core/src/linting/a_plural_noun.rs
```diff
@@ -0,0 +1,631 @@
...
+    };
+
+    if !article
+        .span
+        .get_content(src)
+        .eq_any_ignore_ascii_case_str(&["a"])
+        || !noun
+            .span
+            .get_content(src)
+            .eq_any_ignore_ascii_case_str(&["pieces"])
```
hippietrail: Since these are both single element arrays you can use the single-word methods `eq_ch()` and/or `eq_str()`

harper-core/src/linting/a_plural_noun.rs
```diff
@@ -0,0 +1,631 @@
...
+    {
+        return false;
+    }
+
+    let Some((before, _)) = ctx else {
+        return false;
+    };
+    let Some(previous_word) = before.iter().rev().find(|tok| tok.kind.is_word()) else {
```
hippietrail: Isn't this going to find the previous word even if it's before punctuation and not just before whitespace?

hippietrail (commented):
I only glanced at this quickly but noted a couple of things.

harper-core/src/linting/a_plural_noun.rs
azmifarih: Fixed for this and changed the others like this as well.

harper-core/src/linting/a_plural_noun.rs
azmifarih: Yes, you’re right. The old code walked backward until it found a word so that it could skip over punctuation. I changed it to look only at the immediate previous non-whitespace token and accept it only if that token is a word.

azmifarih:
I also fixed the plural-head handling. If the matcher consumes a trailing POS-tagged verb, the rule now backs up to the actual plural noun head and trims the lint span there. For plural-location nouns like `stairs`, it only suggests the plural-intended forms, not `A stair`.

harper-core/src/linting/a_plural_noun.rs
```diff
@@ -0,0 +1,699 @@
...
+            "She hired a senior systems engineers.",
+            linter(),
+            "She hired a senior systems engineer.",
+        );
+    }
+
+    #[test]
+    fn allows_plural_looking_modifier_before_head_noun_with_different_head() {
+        crate::linting::tests::assert_no_lints("She hired a communications director.", linter());
+    }
```
hippietrail, line 919: Here's some more tests you can add that resulted in more false positives than true positives when running this linter over my corpus:

1. Teaches you PyTorch and many machine learning concepts in **a hands**-on, code-first way.
2. It plays like 1 minute of the song and then stops, and after about 30 minutes later, the bot disconnects **an throws** DisTubeError
3. v0.9.7 turns ShadowBroker from a dashboard **a human watches** into an intelligence surface any agent can act on.
4. the cache is created when **a model loads** and cleared on unload.
5. OpenCTI is **an open source platform allowing organizations** to manage their cyber threat intelligence knowledge and observables.
6. The goal is to create **a comprehensive tool allowing users** to capitalize technical (such as TTPs and observables) and non-technical information
7. `/config tui` opens **a forms**-style config editor powered by schemaui
8. For **a quick hands**-on introduction, run one of the notebooks below.
9. When **a rollout finishes**, your code assigns a `reward` to its Trajectory, indicating the performance of the LLM.
10. information about **a web app requirements**, and it will be your job to come up
11. to use **a temporary user data** dir instead which will be cleared automatically after
12. no filesize limit, even when **a proxy limits** the request size
13. You need to be aware that this means actions like downloading and opening **an unknown file **will** make Microsoft Defender send samples** of it to the Cloud
14. Consider turning it on after clean installing **a new OS** and fully updating it.
15. If **a user signs** in as Other user, the full name of the user isn't displayed during sign-in.
16. extensions only work from **a real user data** dir
17. **a macOS binary reports** as macOS with Apple GPU, **a Linux binary reports** as Linux with NVIDIA GPU.
18. you need actual Windows fonts (Segoe UI, Calibri, Bahnschrift, etc.) from **a Windows** machine's `C:\Windows\Fonts\` directory
19. It's the best way to get **a hands-on** feel for TabPFN
20. tiny11 Core is not **a full Windows** 11 substitute but a rapid testing or development tool
21. Double-clicking **a node copies** the connections of the nearest node.
22. Ouroboros is **an Agent OS** for AI coding
23. If you use **a Windows** `cmd` development environment
24. manifest + README + 6 paper-**A inputs**/baseline + 1 paper-C inputs/baseline + Stage 3 reviewer excerpt + 6 codex-judge baseline placeholders.
25. Building **a Universal Data** Lakehouse
26. **A special thanks** to our course sponsors for making this initiative possible!
27. On top of that none of them produced PDF/**A files** (format dedicated for long time storage)
28. **An invalid prefix causes** an error.
29. Generate a timeline of **a users GitHub Repos**
30. **A BIG thanks** to all Diolinux's supporters

And my analyses:
1. The first you can handle by checking for a hyphen in the next token. This is common so there's a helper method for it. In this case "hands-on" is an idiom so the plural is correct.
2. The second is a surprise where a different mistake can resemble the mistake we're looking for. These spring up from time to time. The `an` is a typo for `and` and there's another linter for that. I would probably `#[ignore]` this one but still include it.
3. A failure to detect that "human watches" is a `noun-subject + verb` rather than a `noun + noun` compound NP - this is the kind of tricky stuff we are up against when developing a grammar checker! (-:
4. Same as 3.
5. Tricky! -ing forms of verbs can behave as both adjectives and nouns - often it's obvious to us that code groups them wrongly but difficult to figure out code to group them right
6. Same as 5. This helps up spot that the -ing verb and the "to" after the plural noun might be the way to detect this pattern.
7. Similar to 1. but this time "forms-style" is not an idiom so perhaps it should be corrected to "form-style", in which case it would be a true positive after all?
8. Almost identical to 1. which confirms this is a pattern we should handle.
9. This is a noun followed by 3rd person verb which is being misidentified as plural noun. I'm sure I saw your code has logic for this so that will need to be investigated.
10. This is either a true positive or a different mistake where "app" should be "app's"?
11. This reveals that the noun "dir" is missing from `dictionary.dict`
12. "limits" is a verb here.
13. This one will be a nightmare to handle properly. Of the several verbs that can also act as nouns in other contexts, "send" seems the most marginal to me.
14. This is probably due to the case-folding design of the dictionary and might be fixed by #2630 but maybe can be fixed fairly easily in this linter for this case?
15. Ambiguous word acting as verb but detected as noun.
16. Identical to 11.
17. Just like 15.
18. "Windows" should be marked as a proper noun in the dictionary, which should help weed this out.
19. Now we know this is common.
20. Just like 18.
21. Like 15 again.
22. Same as 14.
23. Like 18 again.
24. Weird but we can use the helper method to check if the previous token is a hyphen.
25. The way I usually handle things like this is via `.is_oov()` since most out-of-vocabulary words are nouns. Having the first letter capitalized is an even stronger hint.
26. I'd handle this as a special exception rather than marking "thanks" as also singular in the dictionary.
27. Weird! Is PDF/A a type of file? If so we can check if the previous token is a slash as well as a hyphen. To learn more we can look at the GitHub repo for `ocrmypdf/OCRmyPDF`, which is where I found this.
28. Like 18 again.
29. Hmm I didn't foresee this. Makes things tricky. Here "users" should be "user's" - ambiguity lurks everywhere in a grammar checker!
30. Confirms 26 is worth handling.

There's more but I'll stop here. This demonstrates the value of testing against a corpus of real-world English over just trying to make up typical sentences ourselves.

Let me know if you enjoy working on this type of problem - it can be very frustrating and/or very interesting!

hippietrail (changes requested):
I think you're doing great work. I've run this over my corpus of GitHub readmes and annotated the first 30 hits, most of which are false positives unfortunately.

harper-core/src/linting/a_plural_noun.rs
azmifarih: Let's do it! Yeah!

Addressed 9, 14, and 22 in a618da9c.

For 9, I added a guard for singular noun + third-person singular verb, so `a rollout finishes` is no longer treated as an `a/an + plural noun` case.

For 14 and 22, I added an APluralNoun-side guard for all-caps proper initialism heads like `OS`, covering both `a new OS` and `an Agent OS`.

I’m leaving the remaining corpus cases for follow-up commits.

azmifarih: Addressed 1,2,5,6,7,8 in [b5c93c6](https://github.com/Automattic/harper/pull/3418/commits/b5c93c6e0353cfae4a9e70e7897f4a1968cc2342) 

For 1,7,8:

Skips accepted plural hyphen modifiers using `followed_by_hyphen`, e.g. `hands-on`, `rights-based`, `standards-compliant`, and `systems-level`.

For `forms-style`, I kept it as a true positive, but limited the suggestion to the singular modifier form (`a form-style ...`) so it no longer suggests dropping the article or using `some`.

For 2, 5, 6:
added guards for the corpus false positives where:
  - `an` appears to be a typo for `and`, as in `disconnects an throws`
  - an `-ing` verb takes a plural object before `to`, as in `platform allowing organizations to manage ...`

PR #3426, opened 2026-05-19
fix: treat semicolons as chunk separators

### LINKED ISSUE #3405: Semicolons are not handled as chunk separators like commas are.
While working on getting `AnchorEnd` to work I noticed that `Expr`s that work with commas, periods, and exclamation marks break at semicolons.

The current alignment of the concept of "chunk" with commas works pretty well because commas separate clauses and other things that have some kind of "start" and "end".

But semicolons also do this, they are kind of like a higher-order comma.

But a pattern that matches at the end of a chunk will match before other important punctuation but not semicolon, which can be hard to debug. "bar⚓" matches documents like "foo bar", "foo bar. baz", "foo bar! baz", "foo bar, baz" but not "foo bar; baz"

I've wondered this for months and actually initially expected semicolon was handled just like comma. But it hasn't really come up so far because `AnchorEnd` didn't seem to work.

I haven't tested to see whether just adding the semicolon to the current logic will cause a bunch of test to fail or will "just work".

azmifarih:
# Issues
Fixes #3405

# Description
Treat semicolons as chunk terminators so end-of-chunk expression behavior is consistent with commas and similar punctuation.

This patch:
- adds `Punctuation::Semicolon` to `TokenKind::is_chunk_terminator()`
- adds chunk-splitting tests for comma and semicolon in `token_string_ext.rs`
- switches `WrongApostrophe` from `Chunk` to `Sentence` scope to preserve context after chunk boundary changes

# Demo
N/A

# How Has This Been Tested?
- `just test`
    - All relevant Rust/core tests passed.
    - One local Chrome extension options UI test failed:
      `[chromium] tests/options_structured_rules.spec.ts › structured rule settings › rule dropdown updates only the targeted flat rule`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/lint_group/mod.rs
+++ b/harper-core/src/linting/lint_group/mod.rs
@@ -809,7 +809,9 @@ impl LintGroup {
         insert_struct_rule!(WordPressDotcom, true);
         insert_expr_rule_with_dict!(WorthToDo, true);
         insert_expr_rule!(WouldNeverHave, true);
-        insert_expr_rule!(WrongApostrophe, true);
+        // Uses Sentence rather than Chunk
+        out.add("WrongApostrophe", WrongApostrophe::default());
+        out.config.set_rule_enabled("WrongApostrophe", true);
 
         // Uses Sentence rather than Chunk
         out.add("AspireTo", AspireTo::default());
--- a/harper-core/src/linting/wrong_apostrophe.rs
+++ b/harper-core/src/linting/wrong_apostrophe.rs
@@ -1,7 +1,7 @@
 use crate::{
     Token, TokenStringExt,
     expr::{Expr, FirstMatchOf, SequenceExpr},
-    linting::{ExprLinter, Lint, LintKind, Suggestion, expr_linter::Chunk},
+    linting::{ExprLinter, Lint, LintKind, Suggestion, expr_linter::Sentence},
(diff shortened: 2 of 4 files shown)


hippietrail:
Thanks! Did you try testing everything? `just test` will test all the integrations as well as just the Rust stuff. I'm a bit worried it could break something unexpectedly. Watch out though as it can bog down your system if you're low on resources. It can open ten browser windows at a time over and over, etc.

azmifarih:
For `just test` I only have one error in test-chrome-plugin:

`[chromium] tests/options_structured_rules.spec.ts › structured rule settings › rule dropdown updates only the targeted flat rule`

It seems outside of the PR.

hippietrail (approved):
Everything I see here makes sense. Since you say all tests pass even with `just test` that it's surely ready. Thanks!

Oh add that you did that to the testing section of your PR body above too so everyone knows without reading through the comments.

azmifarih:
> Everything I see here makes sense. Since you say all tests pass even with `just test` that it's surely ready. Thanks!
> 
> Oh add that you did that to the testing section of your PR body above too so everyone knows without reading through the comments.

Done for this. Thank you for your advice.

elijah-potter (approved):
This looks great! Thank you!

PR #3436, opened 2026-05-19
fix(core): proper-noun coverage for Apple i-products and product phrases

johndecker3:
# Issues 

# Description

Two small gaps in Apple-product coverage:

1. dictionary.dict: iMac, iPad, iPhone, iPod, iTunes were missing the
   `O` (proper-noun) flag that their siblings (Apple, Mac, iCloud,
   macOS, MacBook, iOS) already carry. Without `O`, `is_proper_noun()`
   returns false and downstream linters skip canonical-capitalization
   for these words.

2. proper_noun_rules.json (AppleNames): added "Apple iPod" alongside
   the existing Apple iPhone/iPad/iMac entries, plus six commonly
   typed bare multi-word phrases — "MacBook Pro/Air", "Mac Pro/Mini",
   "AirPods Pro/Max" — that users write without the "Apple" prefix.

No new linters, no behavior changes outside the affected dictionary
entries.

# Demo

# How Has This Been Tested?

- 5 new tests in orthographic_consistency.rs (iphone/ipad/ipod/imac/itunes
  in full sentences) — confirm the new `O` flag enables canonical-case
  suggestion via the orthographic consistency linter.
- 8 new tests in proper_noun_capitalization_linters.rs (apple_ipod +
  three "family" pairs for MacBook/Mac/AirPods) — modeled on the
  existing americas_lowercase / americas_allow_correct pattern.
- cargo test -p harper-core --lib: full module suites green (31/31
  orthographic_consistency, 48/48 proper_noun_capitalization_linters).

# AI Disclosure

# Checklist


harper-core/dictionary.dict
```diff
@@ -29453,12 +29453,12 @@ hysterical/~JY
...
+iMac/ONgS
 iOS/Og
-iPad/NgS
-iPhone/NgS
-iPod/NgS
-iTunes/Ng
+iPad/ONgS
+iPhone/ONgS
+iPod/ONgS
+iTunes/ONg
```
hippietrail: Those are all good. It's tricky to decide what to mark as proper noun vs common noun vs proper noun+common noun, especially when we can't rely on the capitalization test.
But actually `iTunes` should be `O` only though so was "more wrong" than the others. You can't say "the iTunes" or "an iTunes" or "two iTunes".

harper-core/proper_noun_rules.json
```diff
@@ -350,7 +351,13 @@
...
 			"Apple AirPods Pro",
 			"Apple AirPods Max",
-			"Apple Vision Pro"
+			"Apple Vision Pro",
+			"MacBook Pro",
+			"MacBook Air",
+			"Mac Pro",
+			"Mac Mini",
+			"AirPods Pro",
+			"AirPods Max"
```
hippietrail: Try to keep these alphabetically sorted so that future contributors can quickly scan the list to see if something is included or missing.

hippietrail (changes requested):
Thanks. This is definitely good. I only ask for two tweaks.

harper-core/proper_noun_rules.json
johndecker3: Thank you for the feedback.  I will make the sorting changes to the Apple product list.

harper-core/dictionary.dict
johndecker3: Quick follow-up on the iTunes entry before I make the change — I want to flag a regression concern.

Following your reasoning ("the iTunes" / "an iTunes" / "two iTunes"), I'm with you on dropping N and not adding S. But the peer entries — Disney/Og, Google/OgV, Pixar/Og, and  existing iOS/Og — all keep g, and prose containing "iTunes's library" is a usage your three examples don't seem to rule out. Did you intend iTunes/O literally, or iTunes/Og matching those peers? Happy to go either way, just want to make sure I'm not introducing a possessive false-positive against the grain of the dictionary.

Let me know your preference and I'll make the change.  Thanks!

hippietrail: Actually I always leave of `/g` from nouns ending in `s` because whether to put the apostrophe before or after is subtle and confusing and hasn't really been addressed anywhere in Harper that I've seen. I think that's mostly the status quo in the dictionary but I haven't gone through all the old entries that were there before I started contributing.

Hope this helps!

johndecker3: Thanks.  Made the suggested changes.

harper-core/proper_noun_rules.json
johndecker3: All set.  New commit that should address both requested changes.

elijah-potter (approved):

PR #3444, opened 2026-05-20
feat: analog/acoustic bike→pushbike/pedal bike

hippietrail:
# Issues 
N/A

# Description

I heard "acoustic bike" for this first time yesterday and thought it was incredibly dumb. But then I thought maybe it's playful rather than ignorant, which is the case.

But it's still worth a lint for when you want to make a document more standard English.

It handles both "acoustic" and "analog"/"analogue" spellings, and "bike" and "bicycle".
Depending on the dialect, it suggests "pushbike" and "pedal bike", or "bicycle".

Note: It doesn't yet handle the indefinite article changing from "an" to "a".

# How Has This Been Tested?

`cargo test`

# AI Disclosure

I also used Google Search to convert my version with a loop into a nice functional-style version.

# Checklist


elijah-potter:
Unfortunately, I think this might be an overcorrection. When someone uses the term "analog bike", they mean something specific and wish to convey a certain tone. Suggestions to correct "analog bike" have a real chance of suppressing an author's voice, even if we find that voice to be dumb.

hippietrail:
> Unfortunately, I think this might be an overcorrection. When someone uses the term "analog bike", they mean something specific and wish to convey a certain tone. Suggestions to correct "analog bike" have a real chance of suppressing an author's voice, even if we find that voice to be dumb.

Are people always writing for an audience of their peers? Are they always aware some words might not be understood or taken the expected way by all their readers?

My first reaction to encountering it for the first time was the same as for "vinyls player" that it was out of ignorance and I only found out it was used ironically when I dug into it. Some writers might want to avoid that reaction if made aware of it?

Also it could be off by default.

elijah-potter:
You have convinced me. Let's add it, but have it disabled by default.

hippietrail:
> You have convinced me. Let's add it, but have it disabled by default.

While doing this it occurred to me that I don't actually know the exact relationship between the boolean `state` field in the `harper-core/default_config.json` file and the boolean `default_config` field in `harper-core/src/linting/lint_group/mod.rs`

It's not really clarified in `https://writewithharper.com/docs/contributors/author-a-rule` either:
image

Should they always match? Is it an error if they don't? Do we enforce this in any way? Should we? Am I missing something?

elijah-potter:
That's been bothering me for a while. Let me draft a fix really quick.

elijah-potter:
Done. See #3755.

PR #3445, opened 2026-05-20
feat: add a new `LintKind::Preposition`

hippietrail:
# Issues 
N/A

# Description

I kept making more and more linters to fix wrong prepositions so I finally decided they should be their own `LintKind`.
I was going to call it `WrongPreposition` but there are a couple of other preposition-related lints so I ended up choosing `Preposition` over `Prepositions`... but maybe you disagree? Let me know!

I changed the `LintKind` of all the Rust and Weir rules that seemed applicable.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist


elijah-potter (commented):
This is a breaking change. How big of an improvement would this be? 

I am not opposed to merging it, but I do want to make sure it is worth it.

hippietrail:
> This is a breaking change. How big of an improvement would this be?
> 
> I am not opposed to merging it, but I do want to make sure it is worth it.

It only makes the `LintKind`s more fine-grained. So it only comes down to whether that's something desirable or not. Maybe even how fixed/finalized the set of `LintKind`s is right now.

PR #3455, opened 2026-05-21
fix(core): avoid overall false positive after over all

### LINKED ISSUE #1249: Rule "Overall" false positive: "Iterate over all"
## Description

I have this comment:
```lua
-- Iterate over all the save data in the slot.
```
Harper Suggestion: "Did you mean the closed compound `overall`?" Which is incorrect.

It flagged a non-error and suggested an incorrect fix.

I thought adding 'of' might make it more explicitly correct:

```lua
-- Iterate over all of the save data in the slot.
```

But instead it added another suggestion: "Consider simplifying to `all the`."

Provide a detailed description of the false positive flagged by Harper.

It accepts removing "all" but the comment is trying to emphasize that this function includes **all** data whereas other functions are only partial:

```lua
-- Iterate over the save data in the slot.
```

resolvicomai:
# Issues

Fixes #1249

# Description

This keeps the `Overall` closed-compound correction for cases like `over all performance`, but skips it when `over all` is followed by `the` or `of`, where `all` is acting as a quantifier in phrases such as `over all the save data` and `over all of the save data`.

The change is scoped to the `Overall` closed-compound rule. Other closed compound mappings still use the shared `MapPhraseLinter` path.

# Demo

Not visual.

# How Has This Been Tested?

- `cargo test -p harper-core linting::closed_compounds::tests::over_all -- --nocapture`
- `cargo test -p harper-core linting::closed_compounds -- --nocapture`
- `cargo check -p harper-core`
- `cargo fmt --check`
- `git diff --check`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/closed_compounds.rs
+++ b/harper-core/src/linting/closed_compounds.rs
@@ -1,6 +1,73 @@
-use crate::linting::LintGroup;
+use crate::expr::{Expr, FixedPhrase};
+use crate::linting::expr_linter::Chunk;
+use crate::linting::{ExprLinter, Lint, LintGroup, LintKind, Suggestion};
+use crate::{Token, TokenStringExt};
 
 use super::MapPhraseLinter;
+
+struct Overall {
+    expr: FixedPhrase,
+}
+
+impl Default for Overall {
+    fn default() -> Self {
+        Self {
+            expr: FixedPhrase::from_phrase("over all"),
+        }
+    }
+}
+
+impl ExprLinter for Overall {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        &self.expr
+    }
+
+    fn match_to_lint_with_context(
+        &self,
+        matched_tokens: &[Token],
+        source: &[char],
+        context: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
+        if followed_by_quantifier_complement(context, source) {
+            return None;
+        }
+
+        let span = matched_tokens.span()?;
+        let matched_text = span.get_content(source);
+
+        Some(Lint {
+            span,
+            lint_kind: LintKind::Miscellaneous,
+            suggestions: vec![Suggestion::replace_with_match_case(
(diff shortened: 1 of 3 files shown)


harper-core/src/linting/closed_compounds.rs
```diff
@@ -1,7 +1,74 @@
...
+fn followed_by_quantifier_complement(
+    context: Option<(&[Token], &[Token])>,
+    source: &[char],
+) -> bool {
+    if let Some((_, after)) = context
+        && let [ws, word, ..] = after
+        && ws.kind.is_whitespace()
+    {
+        let word = word.get_str(source);
+        return word.eq_ignore_ascii_case("the") || word.eq_ignore_ascii_case("of");
```
hippietrail: We have a helper, `followed_by_word` in `expr_linter.rs` to make this a bit more idiomatic.

hippietrail (changes requested):
Thanks. This looks reasonable but see if you can adapt it to use the helper method.

harper-core/src/linting/closed_compounds.rs
resolvicomai: Fixed in the latest push: the custom context helper is removed and the `Overall` rule now uses `followed_by_word` from `expr_linter.rs`. Re-ran `cargo test -p harper-core linting::closed_compounds -- --nocapture`, `cargo check -p harper-core`, `cargo fmt --check`, and `git diff --check`.

resolvicomai:
Updated in `8f9b161b`.

The `Overall` rule now uses the existing `followed_by_word` helper for the `over all the/of...` guard instead of open-coding the lookahead. I also refreshed the affected Alice/Gatsby snapshots.

Validation:
- `cargo test -q -p harper-core --lib closed_compounds`
- focused snapshot flow for `harper-core/tests/linting_tests.rs`
- `cargo test -q -p harper-core --lib`
- `cargo check --workspace`
- `cargo fmt --check`
- `git diff --check`

The GitHub checks are green on the latest commit.

PR #3460, opened 2026-05-21
fix: avoid BuiltIn false positive for verb phrases

### LINKED ISSUE #3288: False positive: "built in" flagged to be hyphenated when the words should remain separate
Image

> Also known as the 'New Jetty', the Mixed Clan Jetty was built in 1962.

The hyphenated form "built-in" is an **adjective** meaning something that is an integral or permanent part of something else (e.g., a built-in feature), which does not apply here.

But in this context, "built in" is a **verb phrase** indicating the year the jetty was constructed, and the two words should remain separate.

The `BuiltIn` rule is a Weir rule. It needs to be enhanced to try to determine from the words' context whether they are used properly as a verb phrase or need to be flagged to correct to an adjective. This may or may not require upgrading the rule from Weir to Rust.

jmolz:
Fixes #3288

## Summary
- narrow the BuiltIn Weir rule so built in only lints in modifier+noun contexts
- add regression coverage for verb-phrase uses like "was built in 1962" and "was built in Rust"
- keep existing in-built and modifier suggestions covered

## Tests
- cargo test -p harper-core run_tests_for_built_in
- cargo run -p harper-cli -- test harper-core/src/linting/weir_rules/BuiltIn.weir
- cargo fmt --all --check

### DIFF
--- a/harper-core/src/linting/weir_rules/BuiltIn.weir
+++ b/harper-core/src/linting/weir_rules/BuiltIn.weir
@@ -1,7 +1,19 @@
-expr main [(in built), (in-built), (built in)]
+expr safeBefore [AUX, DET, ADP, PUNCT, VERB]
+expr modifierHead [NOUN, (ADJ NOUN), (ADV ADJ NOUN), (PROPN NOUN), (PROPN PROPN NOUN)]
+expr builtInModifier <(@safeBefore built in @modifierHead), (built in)>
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):

elijah-potter:
This looks good to me. Thanks!

PR #3462, opened 2026-05-21
fix(ci): produce universal desktop binaries

elijah-potter:
# Description

Instead of producing separate x86 and ARM app bundles for macOS, this PR sets up our CI to produce universal binaries, whose updates and installers can be used on any platform.

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -18,28 +18,17 @@ jobs:
     strategy:
       matrix:
         platform:
-          - name: macOS x64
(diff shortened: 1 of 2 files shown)


hippietrail:
> # Description
> Instead of producing separate x86 and ARM app bundles for macOS, this PR sets up our CI to produce universal binaries, whose updates and installers can be used on any platform.

Universal binaries are just about double the size of normal binaries so some people avoid them if there's a choice, especially given that modern Macs don't have upgradeable storage.

PR #3494, opened 2026-05-25
feat(core): create rule to add accent to cliche

elijah-potter:
# Issues 

N/A

# Description

Adds a new Weir spelling rule, `ClicheAccent`, that suggests `cliché` when users write `cliche` without the accent.

The rule is enabled in the curated default config under Spelling and Orthography.

# Demo

N/A

# How Has This Been Tested?

Added Weir rule tests covering lowercase, uppercase, sentence-initial, punctuation-adjacent, and hyphenated uses of `cliche`, plus allowed cases for the already-accented spelling and nearby non-matches.

# Checklist


hippietrail:
There's no way in Weir at this stage to get this to also cover the plural in a single rule is there?

elijah-potter:
I've adjusted the build step for the `weir_rules` directory. You should be able to compose multiple Weir rules into a single one by placing them into a subfolder together.

Take a look at the diff.

hippietrail:
> I've adjusted the build step for the `weir_rules` directory. You should be able to compose multiple Weir rules into a single one by placing them into a subfolder together.
> 
> Take a look at the diff.

That, sir, is very awesome!

PR #3507, opened 2026-05-26
Add fastlane scaffolding for macOS code signing

mokagio:
# Issues

[AINFRA-2433](https://linear.app/a8c/issue/AINFRA-2433) — request for distribution certificate for harper-desktop.

# Description

Scaffolds the standard A8C macOS code-signing setup at the repo root using Ruby fastlane and uses it to notarize the app in CI via the new Buildkite provider.

The reason for using Buildkite instead of GitHub Actions is that we have better control over Buildkite and more security: code signing runs on machines that we own, not shared hardware or VMs.

Pick up one of the [latest Buildkite builds](https://buildkite.com/automattic/harper), and you'll find the Mac DMG in the artifacts.

<table>
<tr>
 <td>
Screenshot 2026-05-26 at 12 40 26 PM
</td>
 <td>
image
</td>
</table>

# AI Disclosure

# Checklist

- [ ] I have added tests to cover my changes — N.A.

---

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.* Further edited by Gio directly.


justfile
```diff
@@ -168,7 +168,7 @@ dev-desktop: build-harperjs build-lint-framework build-components build-harper-e
 
   cd "{{justfile_directory()}}/harper-desktop"
   pnpm install
-  cargo tauri dev
+  pnpm tauri dev
```
mokagio, line 171: Switching to `pnpm` for Tauri saved ~1m45s in CI

.buildkite/commands/build-desktop.sh
```diff
@@ -16,7 +16,19 @@ if ! command -v rustup >/dev/null 2>&1; then
 fi
 # shellcheck disable=SC1091
 . "$HOME/.cargo/env"
-rustup target add aarch64-apple-darwin x86_64-apple-darwin wasm32-unknown-unknown
+
+# Universal builds (master, tags) need both arches. PR/branch builds run
+# Apple-Silicon-only to skip the x86_64 half of the compile (~4m saved).
```
hippietrail, line 23: 🥳

mokagio:
@elijah-potter @twstokes — sorry, my latest push broke CI on this PR.
I knew shipping a change like that right before EOD wasn't a good idea.

For the record: the notarization itself actually succeeded (Apple returned `Accepted` for the `.app`), but fastlane's `notarize` action then crashed because I passed `verbose: true`, which makes `xcrun notarytool` interleave debug lines with the JSON output the action tries to parse — `expected ',' or ']' after array value at line 1 column 4`. The fix is dropping `verbose: true` from the new `notarize_macos` lane.

I'll fix it first thing tomorrow.

---

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

mokagio:
@elijah-potter @twstokes — sorry, my latest push broke CI on this PR.
I knew shipping a change like that right before EOD wasn't a good idea.

The new `notarize_macos` fastlane lane crashed the `Build, sign, notarize harper-desktop` Buildkite job. fastlane's `notarize` action runs `xcrun notarytool submit --output-format json --wait --verbose` and then does `JSON.parse(submit_response)` on the combined stdout. Passing `verbose: true` to the action makes notarytool interleave non-JSON log lines into the response, so the parse blows up with `expected ',' or ']' after array value at line 1 column 4 (JSON::ParserError)` at `notarize.rb:89`. Full log: https://buildkite.com/automattic/harper/builds/14#019e63cc-3e1d-44b4-9926-02b5b016e6ca

I'll fix it first thing tomorrow (drop `verbose: true` from the `notarize` call in `fastlane/Fastfile`).

---

*Posted by Claude Code (Opus 4.7) on behalf of @mokagio with approval.*

fastlane/Fastfile
```diff
@@ -0,0 +1,78 @@
...
+  # supply it.
+  bundle_id = JSON.parse(File.read(TAURI_CONFIG_PATH)).fetch('identifier')
+
+  notarize(
+    package: package,
+    bundle_id: bundle_id,
+    api_key: app_store_connect_api_key,
+    print_log: true,
+    verbose: true
+  )
```
twstokes, line 79:
```suggestion
  notarize(
    package: package,
    bundle_id: bundle_id,
    api_key: app_store_connect_api_key,
    print_log: true
  )
```

twstokes: [Applying this suggestion](https://github.com/Automattic/harper/pull/3507#issuecomment-4543616675) from @mokagio.

elijah-potter:
I have a few questions. Forgive me if these things are obvious somewhere. I am entirely unfamiliar with BuildKite.

- When will this build run? Does it run on every commit in every branch or PR?
- Like the existing GitHub Action workflow, does this one publish release artifacts, or does that need to happen manually?
- I see that the artifacts only include an aarch64 binary. Would it be possible produce universal binaries instead?

Thank you again for picking this up.

twstokes:
👋 @elijah-potter, @mokagio will confirm, but to go ahead and answer some of these questions:

> When will this build run? Does it run on every commit in every branch or PR?

It will run on tags, `master`, and commits that are part of a PR. This is configured on the Buildkite side.

> Like the existing GitHub Action workflow, does this one publish release artifacts, or does that need to happen manually?

This PR doesn't currently publish, but we can make that happen as well. 👍

> I see that the artifacts only include an aarch64 binary. Would it be possible produce universal binaries instead?

It looks like Gio took the route of only producing artifacts for Apple Silicon on PRs to speed things up, and `master` or tags [produce a universal binary](https://github.com/Automattic/harper/pull/3507/changes#diff-997e4b6082b944cf77be059024a3eef5337f40a79d325ccb914a1afa6b821e39R24).

elijah-potter:
@twstokes, thanks for the quick response! I understand.

The auto-publish is not mission-critical, but it could save us quite a bit of time. How difficult would it be to get that working?

twstokes:
> but it could save us quite a bit of time

💯 

> How difficult would it be to get that working?

Shouldn't be too difficult – it'll likely be a new [Fastlane call](https://docs.fastlane.tools/actions/set_github_release/).

elijah-potter:
Okay, if it's not too much trouble, let's do it.

mokagio:
The auto-publish follow-up @elijah-potter  asked about is in #3519 — currently stacked on this branch, so it'll auto-retarget to `master` once this lands.

Copilot flagged three things while reviewing #3519 that actually live in this PR's territory rather than the autopublish change. Routing them here so they can be considered against the right diff:

- `.buildkite/commands/build-desktop.sh:46` — the cargo git-deps cache is saved with key `$CARGO_DEPS_KEY-git` but the matching `restore_cache` only restores `$CARGO_DEPS_KEY`, so `~/.cargo/git/db` never restores on subsequent builds. Either restore the `-git` key too, or save `~/.cargo/git/db` under the unsuffixed key.
- `.buildkite/commands/build-desktop.sh:56` — `curl … | bash` from `raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh` is an unpinned remote installer for the signing/notarization pipeline. Worth pinning to a tag or fetching the prebuilt binary with checksum verification.
- `.nvmrc` pinned to `24.15.0` while `.node-version` stays `lts/*`. Buildkite and GitHub Actions can drift on the same commit. Either pin `.node-version` to match or move `.nvmrc` to `lts/*`.

Full Copilot text is on the corresponding threads in #3519.

---

*Posted by Claude (Opus 4.7) on behalf of @mokagio with approval.*

elijah-potter (approved):

elijah-potter:
It looks like CI is stuck due to https://github.com/microsoft/playwright/issues/40724. I'll draft a fix right away.

elijah-potter:
Fixed in #3533. Will update this branch soon.

twstokes (approved):
LGTM @mokagio. Correct me if I'm wrong, but the previous GHA (which is being removed) would run on forks, when the Buildkite pipeline won't? Just to make sure @elijah-potter is aware.

elijah-potter:
That sounds right. That's fine. Thanks, @twstokes!

PR #3515, opened 2026-05-26
fix(desktop): remove second Dock icon

### LINKED ISSUE #3499: Launching Map desktop app via "Apps", Spotlight, or Terminal shell, launches two copies
Image

----
I noticed one copy gets the full menus while the other gets only the "harper-desktop" menu:

Image

Image

The one with all the menus is the one that's associated with the Harper icon in the menu bar, which will disappear if you quit that one, but not the other one.

----
I can verify that they both have the same version number. I was thinking maybe one is the downloaded one and one is from building locally:

Image

----
**UPDATE** After building my own desktop app locally with my system local patch, I still get two copies, and can confirm that both have the new version number but now one is named **`harper-desktop`** and the other is named **Harper**: 

Image
The one named **Harper** has the minimal menus: 

Image and the one named **`harper-desktop`** has the full menus: 

Image

Note that both instances of the app have both their About and Quit menu options with the un-maccish name `harper-desktop` rather than the "Harper" or "Harper for Mac" seen elsewhere.

It's also worth noting that we don't really want any of those menu items since they're default ones that don't do anything. We should have one using `⌘`+`,` for **Settings** perhaps, as well as the **About** one, but that's about it.

elijah-potter:
# Issues 

Fixes #3499

# Description

Prevents the Harper Desktop highlighter helper from appearing as a second macOS app.

The highlighter runs in a child process with its own winit event loop. On macOS, that made the helper process show up separately in the Dock/app switcher with the default minimal app menu. This updates the highlighter event-loop setup to use macOS accessory activation policy and disables the default highlighter process menu, while leaving the main Harper Desktop app behavior unchanged.

# Demo

N/A

# How Has This Been Tested?

Manually.

# AI Disclosure

# Checklist


hippietrail (approved):
This looks right. I never used winit but in a native AppKit app it's the activation policy.

PR #3523, opened 2026-05-27
fix(core): Weir options should resolve independent of order

elijah-potter:
# Issues 

This fixes an issue mentioned to me in [Discord](https://discord.com/channels/1335035237213671495/1461079667367739472/1509217447104876684).

# Description

If two expressions existed in a Weir option (denoted with the `[]` syntax), a shorter branch might stop a longer branch from matching.

As it turnes out, this was pretty easy to fix. I just swapped out `FirstMatchof` for `LongestMatchOf`. 

# How Has This Been Tested?

Additional unit test.

# Checklist

### DIFF
--- a/harper-core/src/expr/longest_match_of.rs
+++ b/harper-core/src/expr/longest_match_of.rs
@@ -14,6 +14,10 @@ impl LongestMatchOf {
     pub fn add(&mut self, expr: impl Expr + 'static) {
         self.exprs.push(Box::new(expr));
     }
+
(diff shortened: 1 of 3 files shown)


hippietrail:
> If two expressions existed in a Weir option (denoted with the `[]` syntax), a shorter branch might stop a longer branch from matching.
> 
> As it turnes out, this was pretty easy to fix. I just swapped out `FirstMatchof` for `LongestMatchOf`.

Had a hunch it would be that (-:

Issue #3490, opened 2026-05-25
WorkerLinter: default-enabled grammar rules silently inactive until getDefaultLintConfig() is called


frankt72476-gif:
## Summary

When using `WorkerLinter` from `harper.js@2.0.0` in the browser, default-enabled grammar/style rules (e.g. `LongSentences`, `PronounVerbAgreement`, `SubjectPronoun`) do not produce lints unless `getDefaultLintConfig()` is called at least once after `setup()`. Spelling lints work regardless — only the rule-registry-driven categories silently return zero results.

`LocalLinter` does not show this behavior — defaults work without any extra calls.

## Versions

- `harper.js@2.0.0`
- Environment: Chromium-based browser via Vite-bundled SPA; should reproduce on any browser that runs Web Workers.
- WASM loaded from the `harper.js/binary` subpath.

## Reproduction

```ts
import { WorkerLinter, Dialect } from 'harper.js';
import { binary } from 'harper.js/binary';

const linter = new WorkerLinter({ binary, dialect: Dialect.American });
await linter.setup();

const LONG = "I went to the store and I bought some bread and I also bought some milk and then I walked home along the way I saw a dog and the dog was friendly so I stopped to pet it and then I continued walking home but when I got home I realized I had forgotten the eggs which was the main thing I went to the store for in the first place.";

// Returns []:
console.log((await linter.lint(LONG, { language: 'plaintext' })).length);

// Discard the return — only the side effect matters:
await linter.getDefaultLintConfig();

// Now returns 1 Readability lint ("This sentence is 72 words long."):
console.log((await linter.lint(LONG, { language: 'plaintext' })).length);
```

## Bisection

Tried each call in isolation:

- `lint()` without any config calls → 0 grammar lints
- After `getLintConfig()` → 0 grammar lints
- After `getDefaultLintConfig()` → grammar lints fire correctly
- After `setLintConfig(...)` → also arms the rules

`getDefaultLintConfig()` is documented as a read-only accessor (and the `LocalLinter` implementation behaves that way). On the worker it appears to have a side effect of materializing default rule state into the worker's active rule set.

## Workaround

Call `await linter.getDefaultLintConfig()` once after `setup()` resolves, discard the return.

## Suggestion

Either inline the default-materialization into the worker's `setup()` so its behavior matches `LocalLinter`, or rename/document the materialization call honestly (e.g. a `materializeDefaults()` method).

Happy to provide a minimal Vite reproduction repo if it'd help.

[...]

PR #3528, opened 2026-05-28
fix: initialize default grammar rules in WorkerLinter setup

Chessing234:
Bug: WorkerLinter's default-enabled grammar rules are silently inactive until getDefaultLintConfig() is called.

Root Cause: The curated rules are conditionally initialized in the Rust WASM layer via FlatConfig's curated lazy lock, but under certain conditions in a Web Worker environment, the rule set initialization isn't correctly populated when setup merely triggers linting an empty string. The getter explicitly requests materialization of the rules, breaking the dependency cycle.

Why fix is correct: Calling getDefaultLintConfigAsJSON inside LocalLinter's setup explicitly materializes the default rule state via the Wasm-bindgen exported get_default_lint_config_as_json. This ensures the rules are armed and cached before the first proper lint is dispatched, effectively fixing the bug while preserving backwards-compatibility.

### DIFF
--- a/packages/harper.js/src/LocalLinter.ts
+++ b/packages/harper.js/src/LocalLinter.ts
@@ -28,6 +28,8 @@ export default class LocalLinter implements Linter {
 
 	async setup(): Promise<void> {
 		await this.lint('', { language: 'plaintext' });
+		// Workaround for Issue 3490: ensure default rule state is materialized
+		await this.getDefaultLintConfigAsJSON();
 
 		const exported = await this.exportIgnoredLints();
 		await this.importIgnoredLints(exported);


hippietrail:
@elijah-potter Do you think a `wasm` GitHub issue label is warranted? So far I've been adding the [`harper.js`](https://github.com/Automattic/harper/issues?q=state%3Aopen%20label%3Aharper.js) label when I see WASM mentioned.

hippietrail:
Hi there. This looks interesting though it's in a part of the codebase I'm not familiar enough with to review. But you should really use our PR template and fill in the AI/bot section and you should definitely put in your PR description that this is to address #3490 - so far it's only hidden in the code comments.

Apologies for my curtness if you are in fact 100% human (-:

Issue #3525, opened 2026-05-28
False Flags: "you yourself", "out right", "fanfiction"


jupiter-august:
### "You yourself"

**What got flagged?**
The text that was incorrectly flagged.
"Ever write something you yourself hate to think about?"
"Even you yourself must admit that is pretty silly." (oddly, it didn't flag it that time?)
Rule: PossessiveYour

**Why is this incorrect?**
The "you" is flagged. It is suggesting "your", "you're a", or "you're an", all of which would be incorrect.
It's similar to using a phrase like "he himself hated that". That isn't flagged – it is correct – but in the case of "you" here, it was flagged. 

**Example of correct usage:**
"Ever write something you yourself have to think about?"

-------------------------

### "Out right" (TW: Blood/Injury due to the context being a horror fanfiction)

**What got flagged?**
(in a bullet point list of injuries)
"Deep cut down left arm, crudely stapled shut and poorly bandaged, just enough for him to not bleed out right away."
"out right" was flagged. It was just flagged again in the prior sentence saying it was flagged.
Rule: DisjointPrefixes

**Why is this incorrect?**
It suggested changing it to "outright", which would be incorrect in this case as it is followed by "right away" and is being used to indicate time.

**Example of correct usage:**
"I can't go out right away."
"The food ran out right away."

--------------------------

### Fanfiction

**What got flagged?**
Any time I write the word "fanfiction", it is flagged as a typo, with no option to just add it to my dictionary.

**Why is this incorrect?**
 It claims it should be spelled "fan fiction" with a space, but no one I know of spells it that way when speaking of fanfiction such as that on AO3. 

**Example of correct usage:**
"I have been writing fanfiction lately."
"I've read a lot of fanfiction."

hippietrail:
> ### "You yourself"
> **What got flagged?** The text that was incorrectly flagged. "Ever write something you yourself hate to think about?" "Even you yourself must admit that is pretty silly." (oddly, it didn't flag it that time?) Rule: PossessiveYour

✅ Confirmed - thanks!

> ### "Out right" (TW: Blood/Injury due to the context being a horror fanfiction)
> **What got flagged?** (in a bullet point list of injuries) "Deep cut down left arm, crudely stapled shut and poorly bandaged, just enough for him to not bleed out right away." "out right" was flagged. It was just flagged again in the prior sentence saying it was flagged. Rule: DisjointPrefixes

✅ Confirmed - thanks!

> ### Fanfiction
[...]

PR #3529, opened 2026-05-28
fix: correct false flags for yourself, out right, and fanfiction

Chessing234:
bug: PossessiveYour flagged 'you yourself', DisjointPrefixes flagged 'out right' before 'away', and fanfiction was missing from dictionary.
root cause: 'yourself' wasn't in homograph exception list for PossessiveYour, 'right' wasn't in out exceptions for DisjointPrefixes, and fanfiction wasn't in the curated dictionary.
why fix is correct: added exceptions and dictionary word to prevent these false flags.

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -24582,6 +24582,7 @@ fancywork/Ng
 fandango/~NgSV
 fandom/~N
 fanfare/~NwSgV
+fanfiction/S
 fanfic/Nmg†
 fang/~NgSVd
 fanlight/NSg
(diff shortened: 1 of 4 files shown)


harper-core/dictionary.dict
```diff
@@ -24582,6 +24582,7 @@ fancywork/Ng
 fandango/~NgSV
 fandom/~N
 fanfare/~NwSgV
+fanfiction/S
```
hippietrail, line 24585: This part is addressed thoroughly in [3469](https://github.com/Automattic/harper/pull/3469)

hippietrail (changes requested):
Please use our PR template, paying attention to the AI/bot fields.

For AI PRs I'd prefer separate PRs for each issue, even when the user has filed them as a combined GitHub issue. But I especially don't like addressing the unrelated #3490 in with the ones for #3525 - especially without being mentioned in the PR description.

That's just my opinion though, would be better to hear @elijah-potter 's opinion.

PR #3531, opened 2026-05-28
fix: false positive 'an mp3' wrongly flagged to change to 'a mp3'

### LINKED ISSUE #3289: False positive: "an mp3" wrongly flagged to change to "a mp3"

Chessing234:
Fixes #3289.

### Root Cause
Lowercase 'mp3' and 'mp4' were falling back to being treated as consonants because they aren't parsed as likely initialisms (since they contain lower-case and numbers).

### Fix
Added 'mp3' and 'mp4' as explicit exceptions in the vowel-sound checker for indefinite articles.


hippietrail:
See also: https://github.com/Automattic/harper/pull/3369

PR #3549, opened 2026-05-30
feat: the the→that the

hippietrail:
# Issues 
N/A

# Description

Saw this typo in one of @elijah-potter 's PRs and thought "well that's one we all make".
I did some research on which contexts it mostly occurs in and only the previous word is a decent hint.
The test is not foolproof but "the the" is always an error, so it suggests both "that the" and "the" as replacements.

We can delete predictor words if any are found to cause too many false positives and too few genuine mistakes.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist


harper-core/src/linting/the_the_to_that_the.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        Some(Lint {
+            span,
+            lint_kind: LintKind::Typo,
+            suggestions: ["that the", "the"]
+                .iter()
+                .map(|s| Suggestion::replace_with_match_case_str(s, ch))
+                .collect(),
+            message: "Did you mean `that the` or just `the`?".to_string(),
+            ..Default::default()
```
elijah-potter: Can we make sure this has a higher priority than the repetition rule?

elijah-potter (changes requested):
This looks great! I just want to make sure it appears over other conflicting rules.

harper-core/src/linting/the_the_to_that_the.rs
hippietrail: Good call. Before: 
image
After: 
image

elijah-potter (approved):
Looks great! Thanks!

PR #3553, opened 2026-05-31
Add version numbers next to key integrations that are often not up-to-date with the GitHub version

hippietrail:
# Issues 
N/A

# Description

Some integrations take some time to update to the new Harper version once it's published on GitHub.

This PR adds the current version numbers for the Firefox, JavaScript/npm, VS Code, and Rust (Cargo) integrations.

I'm not really satisfied with the size and placement of the version text and font. I'm not a webdev guy.

I feel that the current GitHub release version number should be somewhere too so we can see how far behind any integration is. But I'm not sure where it would fit. I know to do the code as the latest version number is available at https://writewithharper.com/latestversion it would be trivial to add using the same methods the PR already employs.

I tried to add the Chrome one, but couldn't find any JSON feed or API and CORS prevents scraping the HTML. Google AI told me Svelte can do some kind of microserver thingy to fetch, but that sounded like going too far.

The LSP-based ones don't have version lag. Obsidian uses our repo. Did I miss any?

# Demo

image

image

# How Has This Been Tested?

Manually

# AI Disclosure

I got Google to iterate on the code too, which is between an in-editor AI and an agent I think.

# Checklist


packages/web/src/routes/+page.svelte
```diff
@@ -246,8 +254,17 @@ onMount(() => {
...
 							<span class="flex min-w-0 flex-col">
-								<strong class="overflow-hidden text-ellipsis whitespace-nowrap text-[0.84rem]">{integration.name}</strong>
-								<small class="overflow-hidden text-ellipsis whitespace-nowrap text-[0.72rem] text-[#807a6e] dark:text-white/55">{integration.desc}</small>
+								<div class="flex items-center gap-1.5 overflow-hidden">
+									<strong class="overflow-hidden text-ellipsis whitespace-nowrap text-[0.84rem]">{integration.name}</strong>
+									{#if $liveVersions[integration.id]}
+										<span class="inline-flex items-center rounded-full bg-[#f4f1ea] dark:bg-white/10 px-1.5 py-0.5 text-[0.65rem] font-mono font-medium text-[#6b6455] dark:text-white/80 border border-[#e4dfd3] dark:border-white/10 select-none">
+											v{$liveVersions[integration.id]}
+										</span>
+									{/if}
```
elijah-potter, line 273: It would be nice to have a small bit of hover text (which you can simply do with the `title` attribute) that explains why we are showing a version number there. 

Something like, "This version is slightly behind the core engine due to a delay".

packages/web/src/lib/marketing/versions.ts
```diff
@@ -0,0 +1,69 @@
...
+export const liveVersions = writable<Record<string, string>>({
+	firefox: '',
+	js: '',
+	rust: '',
+	vscode: '',
+});
+
+export async function loadLiveVersions() {
+	// 1. Firefox Addon
+	try {
```
elijah-potter: I would love a brief doc-comment that explains what this function does. Also, is there a reason you don't simply return a memoized copy of the `liveVersions` object? If you do, we can reduce the number of exports from this module.

elijah-potter (changes requested):
I love the idea! I just have a few tweaks before we can merge this.

packages/web/src/routes/+page.svelte
hippietrail: Ah I was wondering about something like that, sounds like a way to handle it.
Do you think I should also add the current version? We don't already display it anywhere. I'm not sure where it would look the best either.

packages/web/src/lib/marketing/versions.ts
hippietrail: > I would love a brief doc-comment that explains what this function does.

Will do.

> Also, is there a reason you don't simply return a memoized copy of the `liveVersions` object? If you do, we can reduce the number of exports from this module.

I'm not sure I really understand.

elijah-potter: Rather than have the function update an exported object, I suggest you make that object private and return it from the function instead.

elijah-potter: Wait, scratch that. I didn't see that this was a Svelte store. Please disregard.

packages/web/src/routes/+page.svelte
elijah-potter: Sure!

packages/web/src/lib/marketing/versions.ts
hippietrail: > Wait, scratch that. I didn't see that this was a Svelte store. Please disregard.

Still grappling with the Svelte and webdev terminology and concepts (-:

elijah-potter (approved):

PR #3564, opened 2026-06-01
feat: in a better shape→in better shape

hippietrail:
# Issues 
N/A

# Description

This is a really common mistake that I think is only made by nonnative English speakers.
Saying something is "in a shape" when they mean it's fit or in good condition.

This linter is a bit conservative to avoid false positives so for now it only works after "be" and words that behave similarly to "be" in this context, and only with "better" and "worse" for now.

So I think it does pretty well at avoiding false positives but it definitely has false negatives.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist


harper-core/src/linting/in_better_shape.rs
```diff
@@ -0,0 +1,177 @@
...
+                // be
+                "am", "are", "be", "been", "being", "is", "was", "were",
+                // contractions of be
+                "i'm", "we're", "you're", "he's", "she's", "it's", "they're",
+                // contractions of be missing their apostrophes
+                "im", "were", "youre", "hes", "shes", "its",
+                // sense verbs behave somewhat similarly to be
+                "theyre", "appears", "looks", "seems", "sounds",
+                // get behaves similarly to be
+                "get", "gets", "getting", "got", "gotten",
```
elijah-potter, line 24: I get the sense that some of these could be outsourced to a more central expression. Thoughts?

elijah-potter (changes requested):

PR #3575, opened 2026-06-02
fix: prevent false positive "quite" → "quiet" (#3560)

### LINKED ISSUE #3560: False positive flagging "quite" to change to "quiet"
Image

> It's probably **quite** doable.

jdgajjar:
Fixes #3560

## Summary

Fixes a false positive where Harper incorrectly suggests changing **"quite"** to **"quiet"** in valid contexts.

Example:

> It's probably quite doable.

Previously, Harper flagged **"quite"** as a typo and suggested **"quiet"**, even though the sentence is grammatically correct.

## Root Cause

The typo rule for detecting accidental use of **"quite"** instead of **"quiet"** was too aggressive and triggered on valid uses of **"quite"** as an adverb.

### Valid uses of "quite" (should NOT be flagged)

* It's probably quite doable.
* The task is quite simple.
* The results were quite impressive.
* She was quite happy with the outcome.
* I'm quite sure this is correct.
* Quite frankly, I don't agree.
* That's quite a lot of work.

### Actual misspellings (should still suggest "quiet")

* Please be quite during the presentation.
* The library is very quite today.
* The children were surprisingly quite.

### Verification

* Confirmed that valid uses of **"quite"** are no longer reported as errors.
* Confirmed that genuine cases where **"quiet"** is intended are still detected and corrected.

### DIFF
--- a/harper-core/src/linting/quite_quiet.rs
+++ b/harper-core/src/linting/quite_quiet.rs
@@ -40,7 +40,7 @@ impl Default for QuiteQuiet {
         let adverb_quite = SequenceExpr::default()
             .then_kind_except(
                 TokenKind::is_adverb,
-                &["actually", "never", "not", "really", "generally"],
+                &["actually", "never", "not", "probably", "really", "generally"],
             )
             .t_ws()
             .t_aco("quite");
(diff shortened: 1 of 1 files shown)


elijah-potter (approved):
Looks good. Thank you!

Issue #3565, opened 2026-06-02
MacOS APP - nothing happens


ACyber-lt:
Installed latest MacOS version, all permissions are enabled etc, however now it does not mark any mistakes. Nothing. Even the test "its not alot of fun" - nothing gets flagged or suggested to fix. 

I also added Safari and few other apps to the list, it works nowhere. 
What can I do to fix it?

And second bug: checkbox to automatically add apps does not work.

ACyber-lt:
<img width="1566" height="679" alt="Image" src="https://github.com/user-attachments/assets/5370abde-262a-47f7-9532-bd8b7ced129f" />

hippietrail:
> Installed latest MacOS version, all permissions are enabled etc, however now it does not mark any mistakes. Nothing. Even the test "its not alot of fun" - nothing gets flagged or suggested to fix.

That's interesting. It works for me: 

<img width="1029" height="275" alt="Image" src="https://github.com/user-attachments/assets/ce787fc2-81b8-42de-941e-ef9c437b21e3" />

I'm on an M1 MacBook Air running Tahoe 26.5 (25F71).

> And second bug: checkbox to automatically add apps does not work.

This part I can confirm!

ACyber-lt:
I am using MacBook Pro M3 Tahoe 26.5 

I tried to fully reinstall Harper - same issue. When I reinstall and open text document, it briefly underlines the issue, however when I click on that work/underlining, it all disappears, not suggestion nothing and does not show up anymore. Weird

hippietrail:
> I am using MacBook Pro M3 Tahoe 26.5
> 
> I tried to fully reinstall Harper - same issue. When I reinstall and open text document, it briefly underlines the issue, however when I click on that work/underlining, it all disappears, not suggestion nothing and does not show up anymore. Weird

I did have trouble setting it up on my machine, but my problem was with granting accessibility permission, and you are past that...

bengabay1994:
Same here. Not working, Tried re-installing, remove from accessibility permission and re-adding. Nothing worked. Not in textEdit, Slack or any other app for that matter. I am on macbook pro m5 macOS Tahoe Version 26.5

[...]

PR #3600, opened 2026-06-05
fix(desktop): make macOS accessibility activation work when AXEnhancedUserInterface is rejected

### LINKED ISSUE #3500: Mac desktop app doesn't seem to do anything, even with Privacy & Security / Accessibility enabled
Here it is in my Notes app with System Settings visible behind it with the Harper accessibility permission on: 

Image

The updated Harper homepage doesn't seem to offer any information.

markduan:
On recent macOS 15.7.x, setting AXEnhancedUserInterface on another app's AX element fails with kAXErrorNotImplemented (-25208) for every target app (observed for AppKit apps like TextEdit and Electron apps like Slack, on macOS 15.7.7 / 24G720). Harper treated this error as retryable, so activation stayed in RetryLater forever, get_boxes() always returned no lints, and the desktop app silently did nothing.

Fix, in four parts:

- Treat kAXErrorNotImplemented as an unsupported-attribute error.
- When the attribute is unsupported, proceed to text-metrics verification instead of retrying forever: native AppKit text views expose usable AXBoundsForRange without any activation.
- Fall back to the Electron-specific AXManualAccessibility attribute when AXEnhancedUserInterface cannot be set, so Chromium-based apps (e.g. Slack) still enable their accessibility tree.
- Chromium reports degenerate zero-size AXBoundsForRange rects on the editable AXTextArea wrapper but real ones on the AXStaticText leaf nodes inside it. Add a fallback that maps lint spans onto those leaves by char-offset text matching, used by both the activation verification probe and RectCollector.

Verified on macOS 15.7.7: underlines and suggestions now appear in both TextEdit (AppKit) and Slack (Electron); previously neither worked.

Fixes #3500

# Issues 

# Description

# Demo

# How Has This Been Tested?

# AI Disclosure

# Checklist

### DIFF
--- a/harper-desktop/src-tauri/src/mac_broker.rs
+++ b/harper-desktop/src-tauri/src/mac_broker.rs
@@ -6,9 +6,9 @@ use accessibility_sys::{
     AXUIElementCopyParameterizedAttributeValue, AXUIElementGetPid, AXUIElementSetAttributeValue,
     AXValueCreate, AXValueGetType, AXValueGetValue, AXValueRef, error_string,
     kAXBoundsForRangeParameterizedAttribute, kAXErrorAttributeUnsupported, kAXErrorIllegalArgument,
-    kAXErrorNoValue, kAXErrorParameterizedAttributeUnsupported, kAXErrorSuccess,
-    kAXFocusedApplicationAttribute, kAXTrustedCheckOptionPrompt, kAXValueTypeCFRange,
-    kAXValueTypeCGRect, pid_t,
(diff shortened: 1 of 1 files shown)


markduan:
I'm trying to fix issue https://github.com/Automattic/harper/issues/3565

hippietrail:
This could explain the sometimes high CPU usage I'd noticed when not using it but switching between other apps and Activity Monitor.

elijah-potter (approved):
This looks good to me, and I can confirm that the changes have the desired effect. Thank you!

PR #3610, opened 2026-06-08
feat(harper-core): flag redundant "go back N years ago"

### LINKED ISSUE #3466: Need to flag redundant/conflicting "goes back N years ago"
I just heard this in a YouTube video:
> And this theory **goes back 40 years ago** to a deal that most people have probably never even heard of.

You "go back" a period/span of time. But "ago" indicates a point in time.

- And this theory **goes back 40 years** to a deal that most people have probably never even heard of.
- And this theory **goes back to 40 years ago** to a deal that most people have probably never even heard of.

Some research needs to be done to see if it also applies to any of the other inflections: go back, going back, gone back, went back.

There are already some `Expr` that can match units/periods of time. They can probably be used as-is or perhaps modified slightly if need be.

More examples:
- You have to **go back 20 years ago** to find a Prize awarded for an investigation of the Nation of Islam's questionable business dealings.
- In order to search for various errors, we need to be able to **go back hours or even days ago** in the logs.
- **Go back 20 years ago** and look at what the masses were using: instant cameras or some variant.

mvanhorn:
## Summary
Harper now flags the redundant phrasing "goes back N years ago", where "ago" double-marks the past that "goes back" already expresses.

## Why this matters
@hippietrail requested this in #3466: "goes back ten years ago" is redundant because "goes back" already places the reference in the past, so "ago" is conflicting. The fix follows Harper's established `ExprLinter` pattern (same shape as the neighboring rules) and suggests dropping the redundant "ago".

## Changes
- Add `go_back_units_ago` rule with the pattern matcher, suggestion, and unit tests.
- Wire it into `linting/mod.rs` and the `LintGroup` registry alphabetically.

## Testing
Unit tests in the new module cover the flagged phrasing and the corrected suggestion.

Fixes #3466

AI was used for assistance.

### DIFF
--- a/harper-core/src/linting/go_back_units_ago.rs
+++ b/harper-core/src/linting/go_back_units_ago.rs
@@ -1,0 +1,193 @@
+use crate::{
+    Lint, Token, TokenStringExt,
+    expr::{Expr, SequenceExpr, SpelledNumberExpr, TimeUnitExpr},
+    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk},
+};
+
+pub struct GoBackUnitsAgo {
+    expr: SequenceExpr,
+}
+
+impl Default for GoBackUnitsAgo {
+    fn default() -> Self {
+        let quantity = SequenceExpr::longest_of(vec![
+            Box::new(SpelledNumberExpr),
+            Box::new(SequenceExpr::default().then_number()),
+            Box::new(SequenceExpr::aco("a").t_ws().then_quantifier()),
+            Box::new(SequenceExpr::default().then_indefinite_article()),
+            Box::new(SequenceExpr::default().then_quantifier()),
+        ]);
+
+        Self {
+            expr: SequenceExpr::word_set(&["go", "goes", "going", "gone", "went"])
+                .t_ws()
+                .t_aco("back")
+                .t_ws()
+                .then_optional(quantity.t_ws())
+                .then(TimeUnitExpr::default())
+                .t_ws()
+                .t_aco("ago"),
+        }
+    }
+}
+
+impl ExprLinter for GoBackUnitsAgo {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        &self.expr
+    }
+
+    fn match_to_lint(&self, toks: &[Token], _source: &[char]) -> Option<Lint> {
+        let ago_with_space = toks[toks.len().saturating_sub(2)..].span()?;
+
+        Some(Lint {
+            span: ago_with_space,
+            lint_kind: LintKind::Redundancy,
+            suggestions: vec![Suggestion::Remove],
+            message: "`Go back` already expresses a span of time, so `ago` is redundant here."
+                .to_string(),
+            ..Default::default()
+        })
+    }
+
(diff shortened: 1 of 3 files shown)


hippietrail:
Hey thanks for this and thanks for mentioning that AI was used. Makes it easier to tweak our brains into the right mode for reviewing PRs. Have you read our policy here? https://elijahpotter.dev/articles/harper's-policy-on-agent-PRs

Also please use our actual GitHub PR template as it has more detailed tickboxes and such.

harper-core/src/linting/go_back_units_ago.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+impl Default for GoBackUnitsAgo {
+    fn default() -> Self {
+        let quantity = SequenceExpr::longest_of(vec![
+            Box::new(SpelledNumberExpr),
+            Box::new(SequenceExpr::default().then_number()),
+            Box::new(SequenceExpr::aco("a").t_ws().then_quantifier()),
+            Box::new(SequenceExpr::default().then_indefinite_article()),
+            Box::new(SequenceExpr::default().then_quantifier()),
```
hippietrail, line 18: You should be able to simplify these. "A" is a member of the indefinite article class. Look at `.then_optional()` and you should be able to make either the quantifier or the article optional in the presence of the other.

harper-core/src/linting/go_back_units_ago.rs
```diff
@@ -0,0 +1,193 @@
...
+mod tests {
+    use super::GoBackUnitsAgo;
+    use crate::linting::tests::{assert_lint_count, assert_no_lints, assert_suggestion_result};
+
+    #[test]
+    fn fixes_goes_back_40_years_ago() {
+        assert_suggestion_result(
+            "This theory goes back 40 years ago.",
+            GoBackUnitsAgo::default(),
+            "This theory goes back 40 years.",
```
hippietrail, line 69: It would be better if it could suggest both this *and* "This theory goes back to 40 years ago". So when `Suggestion` would remove "ago" while the other would insert "to".
You can use a different test assert function: `assert_good_and_bad_suggestions()` with both wanted variants as a static array slice in the first "good" arg and and empty array slice as the "bad" arg.

harper-core/src/linting/go_back_units_ago.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn doesnt_flag_later() {
+        assert_no_lints("I went back 20 years later.", GoBackUnitsAgo::default());
+    }
+
+    #[test]
+    fn doesnt_flag_unrelated_back() {
+        assert_no_lints(
+            "I went back there years ago to check the records.",
+            GoBackUnitsAgo::default(),
+        );
```
hippietrail, line 191: Are these made up test sentences or from the issue? Sometimes I put a big selection in my issues found by Googling GitHub but sometimes I only put a couple. Sorry I'm rushing this a bit and not checking (-:

hippietrail (changes requested):
Looks decent but I have some questions. Also please use the GitHub PR template.

elijah-potter:
I am closing this PR for violating the agent policy.

PR #3623, opened 2026-06-09
feat: acclimize→acclimate/acclimatize

hippietrail:
# Issues 
N/A

# Description

I heard a YouTube use the word "acclimize" and thought he must be mixing up "acclimate" and "acclimatize". Some Googling confirmed this.

`SpellCheck` does flag it and does suggest both correct words, but it seemed worth the trouble to have a specific linter anyway.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/default_config.json
+++ b/harper-core/default_config.json
@@ -5039,6 +5039,13 @@
 								"state": true,
 								"label": "Incident Report"
 							}
+						},
+						{
+							"Bool": {
+								"name": "Acclimize",
+								"state": true,
+								"label": "Acclimize"
+							}
 						}
 					]
 				}
--- a/harper-core/src/linting/acclimize.rs
+++ b/harper-core/src/linting/acclimize.rs
@@ -1,0 +1,174 @@
+use crate::{
+    Lint, Token,
+    expr::Expr,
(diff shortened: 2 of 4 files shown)


elijah-potter:
> SpellCheck does flag it and does suggest both correct words, but it seemed worth the trouble to have a specific linter anyway.

If spellcheck correctly flags it and provides reasonable suggestions, I think it might better to simply let it do its thing. I would, however, accept a test that made sure the spell check's behavior was maintained.

hippietrail:
> > SpellCheck does flag it and does suggest both correct words, but it seemed worth the trouble to have a specific linter anyway.
> 
> If spellcheck correctly flags it and provides reasonable suggestions, I think it might better to simply let it do its thing. I would, however, accept a test that made sure the spell check's behavior was maintained.

Indeed. I wondered about that quite a bit and then thought people who use it think it's a real word and are used to the tech being wrong "stupid computer is flagging perfectly normal word"...?

elijah-potter:
There's not much we can do about people thinking that, and I believe this to be a pretty insidious form of scope-creep. I think I am going to close this.

PR #3633, opened 2026-06-11
docs: add harper.js Chrome extension example

### LINKED ISSUE #3628: Docs: Add an intro for how to use `harper.js` inside a Chrome extension
I'm not a browser extension expert but I've stumbled through making a couple by following howtos or guidance of an AI.

But getting `harper.js` to work inside one seems to be next level complicated. I'm only using Google Search AI because I don't have tokens for a coding AI but every time it says we're done there's something wrong that's outside my areas of expertise.

Maybe it's obvious to a webdev or someone who's made lots of Chrome Extensions.

I was going to make a little tool similar to [harper-glasses](https://github.com/hippietrail/harper-glasses) but instead of linting the selected text, called the tokenization API and annotates the tokens in the selection with POS tags.

But a tiny example either as part of the Harper docs that just opened a popup/modal and displays the harper-core version it's built with would be nice and minimal perhaps.

I also find the [harper.js API Reference](https://writewithharper.com/docs/harperjs/ref/index.html) very opaque. It seems to list only objects and not functions. For instance how to I find the function to get the `harper-core` version? I'm sure it exists.

chatman-media:
Summary:
- document how to bundle harper.js in a Manifest V3 popup
- add the example to the harper.js README
- update the raw browser CDN imports to the current 2.4 dist entry points

Closes #3628

Checks:
- pnpm dlx @biomejs/biome@2.3.3 check packages/harper.js/README.md packages/harper.js/examples/raw-web/index.html packages/harper.js/examples/chrome-extension/README.md
- git diff --check

### DIFF
--- a/packages/harper.js/README.md
+++ b/packages/harper.js/README.md
@@ -5,4 +5,10 @@ Harper is a grammar checker designed to be easy to consume and integrate into yo
 
 [Read the documentation to learn more.](https://writewithharper.com/docs/harperjs/introduction)
 
+Examples:
+
+- [Node.js/CommonJS](./examples/commonjs-simple)
+- [Raw browser page](./examples/raw-web)
+- [Chrome extension popup](./examples/chrome-extension)
+
 Wondering how good our algorithm is? [Give it a whirl.](https://writewithharper.com)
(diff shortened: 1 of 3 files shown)


hippietrail:
Thanks for this! I'm about to check it out. Could you please use our GitHub PR template though as it includes fields regarding AI usage that we like to know.

elijah-potter:
This is definitely a bot.

PR #3653, opened 2026-06-13
feat: app picker bundle ID lookup

### LINKED ISSUE #3643: Programmatically find macOS bundle IDs so the user doesn't have to wrangle them manually in the Desktop UI
Image

Currently the user is prompted to `Use the app's macOS bundle identifier. Example: com.apple.TextEdit`

Typically Mac users are less technical than even Windows users, let alone Linux users.

A better interface would have a textbox which updates a list of matching apps by name at each edit

hippietrail:
# Issues 
Fixes #3643

# Description

Changes the `AppPickerModal` to allow the user to type the name of an app and find the bundle ID itself rather than expecting the user to find the bundle ID somehow.

My editor is letting me use the built-in coding AI for the first time in weeks.
I vibe-coded the original version in rawdog macOS API calls from Rust without using any objc crate or such.
I then took that code and vibe-coded a standalone Taui PoC app that populated a listview while the user enters app name.
Then today I showed the AI the files in Harper Desktop I thought were relevant, and the previous Tauri PoC and asked it to integrate them.

I asked it to use the objc2 crate or whatever Harper was already using for other macOS calls instead of rawdogging, but it failed to find the needed API `NSMetadataQuery` support via objc2 and decided to instead shell out to the macOS CLI tool `mdfind`.

I'll keep iterating on this but `NSMetadataQuery` requires a runloop, which the AI says will be a problem. It might be whining though since that didn't stop the previous Tauri PoC and was part of the reason I wanted to get it working in Tauri before trying to bring it into Harper.

# Demo
image

# How Has This Been Tested?

Manually.

# AI Disclosure

# Checklist

### DIFF
--- a/harper-desktop/src-tauri/src/commands.rs
+++ b/harper-desktop/src-tauri/src/commands.rs
@@ -3,7 +3,7 @@
 
 use crate::config::Config;
 use crate::highlighter_service::HighlighterService;
-use crate::os_broker::{AccessibilityPermissionStatus, OsBroker};
+use crate::os_broker::{AccessibilityPermissionStatus, AppSearchResult, OsBroker};
 use crate::{IntegrationView, accessibility_allows_highlighter_start, platform_broker};
 use harper_core::{
     Dialect, DictWordMetadata, IgnoredLints,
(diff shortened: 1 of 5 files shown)


elijah-potter (approved):
This looks excellent! I had a run at it myself last week, but I like your work better.

PR #3659, opened 2026-06-14
refactor: introduce `SequenceExpr::word_seq()`

hippietrail:
# Issues 
N/A

# Description

I thought there was a pattern of linters using sequences of words separated by whitespace and realized a new ctor and builder on `SequenceExpr` could take an array of just the words and add in the whitespace elements.

It turned out that only one linter used more than two words but there were also a few uses of `fixed_phrase` that are probably more efficient as `word_seq`.

# Demo
# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/expr/sequence_expr.rs
+++ b/harper-core/src/expr/sequence_expr.rs
@@ -303,8 +303,6 @@ impl SequenceExpr {
 
     /// Match a series of words separated by whitespace.
     pub fn then_word_seq(self, words: &'static [&'static str]) -> Self {
-        debug_assert!(!words.is_empty(), "words must not be empty");
-
         if let Some((first, rest)) = words.split_first() {
             let mut expr = self.t_aco(first);
             for word in rest {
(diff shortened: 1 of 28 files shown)


harper-core/src/expr/sequence_expr.rs
```diff
@@ -296,6 +301,21 @@ impl SequenceExpr {
         self.then(Word::new_exact(word))
     }
 
+    /// Match a series of words separated by whitespace.
+    pub fn then_word_seq(self, words: &'static [&'static str]) -> Self {
+        debug_assert!(!words.is_empty(), "words must not be empty");
```
elijah-potter: I feel like it should be OK for it to be empty. People may insert an empty array for debugging purposes.

elijah-potter (changes requested):
Looks mostly good. Just one nit.

elijah-potter (approved):

PR #3660, opened 2026-06-15
feat: depends based on→depends on

hippietrail:
# Issues 
N/A

# Description

I heard somebody say "depends based on" in a YouTube video this morning, enunciated in a way that indicated there was no intervening punctuation. I did some Googling and Google's AI insisted this is a known/common mistake though it didn't give me any links discussing it. I was able to find a good few examples on GitHub.

But to me only some of them are clearly fixed by removing the "based". Others might be fixed by adding some kind of punctuation between "depends" and "based on". I asked two AIs and they seemed mostly to think removing "based" is always the fix. I don't trust them.

So I'm putting this PR here, marked for review, with the debugging code still present. Because it's not finished but I am soliciting review and draft PRs never get any feedback.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

I used some AI for the research but not for the code or examples.

# Checklist


harper-core/src/linting/depends_based_on.rs
```diff
@@ -0,0 +1,185 @@
...
+impl ExprLinter for DependsBasedOn {
+    type Unit = Chunk;
+
+    fn match_to_lint_with_context(
+        &self,
+        matched_tokens: &[Token],
+        source: &[char],
+        context: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
+        eprintln!("🚨 {}", format_lint_match(matched_tokens, context, source));
```
elijah-potter, line 32: We cannot allow global side effects in this function.

elijah-potter (changes requested):

harper-core/src/linting/depends_based_on.rs
hippietrail: Sorry I left the debugging in because it's just here for review of what it does, not ready to go in yet. I noticed draft PRs never ever get looked at so I thought I'd try this. Sorry if the Description was too much of a wall of text to read my explanation.

elijah-potter:
> But to me only some of them are clearly fixed by removing the "based". Others might be fixed by adding some kind of punctuation between "depends" and "based on". I asked two AIs and they seemed mostly to think removing "based" is always the fix. I don't trust them.

I think I agree with the AIs. To me, the example sentences look fine. What punctuation were you thinking of? 

As an aside, I am so sorry that I did not read the description all the way.

hippietrail:
> > But to me only some of them are clearly fixed by removing the "based". Others might be fixed by adding some kind of punctuation between "depends" and "based on". I asked two AIs and they seemed mostly to think removing "based" is always the fix. I don't trust them.
> 
> I think I agree with the AIs. To me, the example sentences look fine. What punctuation were you thinking of?

Unless I'm misreading you it seems like you're **disagreeing** with the AIs. It thinks they all need fixing. The same when I paste into Google or even if I just put in google "redundant 'depends based on'". It's much more certain than I am. I can't trust my native-speaker intuition on this one any more (-:

# From pasting just the key terms:
image

# From pasting just the test section of the code in and asking which ones I should remove "based" vs which ones I should insert a comma: 
image

----
I'm inclined to put this one in the "too hard" basket.

(Sorry for pasting images but it strips all formatting, even spaces between words if I copypaste as text)

PR #3666, opened 2026-06-15
chore: dictionary curation

carlosroe:
# Issues 
N/A

# Description
This is my first time trying to commit words to Harper. Any help and criticism is highly appreciated. I am neither an expert in linguistics nor a native English speaker, just someone who would like to help by adding words I frequently use in my research!

+ACC
+CBASP
+Glu
+MRS
+rsFC
+glutamate
+pregenual
+racemate
+racemic

# How Has This Been Tested?
`cargo test`

# AI Disclosure

- [x] I used LLM to find matching annotations for my words

# Checklist


harper-core/dictionary.dict
```diff
@@ -3759,6 +3761,7 @@ GHz/
 GI/JNV
 GIF/NSg             # file format
 GIGO/
+Glu/N               # symbol for glutamate (anion of glutamic acid)
```
hippietrail: This section is for all-uppercase. It sounds silly but when I tried to re-sort in a case insensitive ordering it revealed a latent bug where order of words in the dictionary makes a difference and made passing tests fail. A PR covering all aspects of case problems in the dictionary is being worked on.

harper-core/dictionary.dict
```diff
@@ -7148,6 +7151,7 @@ Mozambique/ONg
 Mozart/NOg
 Mozilla/g           # company
 Mr/NSg
+MRS/N               # magnetic resonance spectroscopy
```
hippietrail: Same here. Easiest way to get them in the right spot is to turn on case sensitive search on your editor. Around line 6267 is where you want to look.

harper-core/dictionary.dict
```diff
@@ -7845,6 +7849,7 @@ PED/NgS             # performance-enhancing drug
 PET/NOg             # early computer
 PFC/N
 PG/JNO
+pgACC/N             # pregenual anterior cingulate cortex
```
hippietrail: Some of these look very specialized. The top/main part of the dictionary is for regular words. Specialized words are in a separate section so that they'll be easy to find if and when we decide to split them out into focussed Weir Packs. The specalized section starts around line 53,834 and 54,573 or so should be where a word starting with lowercase `pg` should go.

hippietrail (changes requested):
Some notes on the sort order of the dictionary being case sensitive and obscure/specialized words have a dedicated section.

harper-core/dictionary.dict
```diff
@@ -8510,6 +8514,7 @@ RP/NOV
 RPM/NgS
 RR/NOV
 RSA/Ng              # encryption standard
+rsFC/Nm             # resting state functional connectivity
```
hippietrail: I would put this way down around 54,604 between
```
royale/~NSgJ        # !! please check and comment !! dictionaries only list noun sense type of custard
runahead/NgS
```

hippietrail (changes requested):
Looks good. Just one more to move. Thanks.

hippietrail:
You've got failing "snapshot" tests. Your new words probably change the top 3 spelling suggestions for other unknown words in the old texts we test against. The way to fix this is to run `cargo test` again. Usually only once but occasionally twice. Then commit the changed snapshots with the PR.

PR #3678, opened 2026-06-17
perf(lint-framework): remove unconditional lint polling

### LINKED ISSUE #3299: Extension using ~8gb of ram and high CPU
**Describe the bug**
After I started using the extension (firefox) I've seen a huge spike in extensions resources used. Today it rendered my machine almost useless as the harper extension was using 8gb of memory and high CPU. I don't expect an extension checking spelling to consume that much resources. Not even with a lot of open tabs.

**To Reproduce**
Install the Firefox extension
Have many tabs open

**Expected behavior**
Extension should run wihtout using a lot of resources

**Platform**
Firefox on Arch Linux

Elias0419:
# Issues 
Fixes #3299 

# Description
This fixes the high CPU and RAM usage in the Firefox extension. It removes the unconditional polling in LintFramework and replaces it with event-driven linting.

# Demo

# How Has This Been Tested?
I've been running this patch on my machine for about a week now without issue. There is no more high CPU or RAM usage.
I checked it with  `pnpm exec biome check --write packages/lint-framework/src/lint/LintFramework.ts`

# AI Disclosure

# Checklist


elijah-potter:
If I recall, this lint polling was necessary to keep the service worker alive. It might be more prudent to find a way to make these polls less energy intense rather than remove them altogether.

Elias0419:
> If I recall, this lint polling was necessary to keep the service worker alive. It might be more prudent to find a way to make these polls less energy intense rather than remove them altogether.

Hi - thanks for having a look. I think that the keepalive timer you are thinking of is this one:

https://github.com/Automattic/harper/blob/c36bb1ca977466139827ceacad5d4e1dbc529393/packages/chrome-plugin/src/contentScript/index.ts#L51

elijah-potter:
What can be done to fix the failing tests?

Elias0419:
> What can be done to fix the failing tests?

It seems like four "Can ignore suggestion" tests are affected by the patch, in:

```
tests/draft.spec.ts 
tests/lexical.spec.ts 
tests/prosemirror.spec.ts 
tests/slate.spec.ts
```
The removed polling would delete the underline by calling update() so that it disappeared quickly without having to wait for another action to trigger an update. I think a fix could be to force refresh or explicitly remove the ignored hash after ignoreLint completes.

elijah-potter (approved):
Seems reasonable. Thank you!

PR #3691, opened 2026-06-20
Fix#3663   "even worst" is not flagged

### LINKED ISSUE #3663: "Even worst" is not flagged
We have at least one linter that handles "worse" vs "worst" but this was not flagged:

> But it's **even worst** because when you look at it, the kernel also needs to manually save a bunch of information whenever there's interrupts, right?

jdgajjar:
Issues

Fixes #3663

Description

This PR fixes a false negative where the phrase "even worst" was not flagged.

The existing "LessWorse" linter handled combinations involving "less" and "least" with "worse"/"worst", but did not account for the pattern "even worst", allowing grammatically incorrect phrases such as:

«But it's even worst because when you look at it...»

to pass without a lint.

Changes

- Extended the "LessWorse" linter to recognize "even" alongside the existing "less" and "least" patterns.
- Added a suggestion:
  - "even worst" → "even worse"
- Updated the linter description.
- Added tests covering:
  - The reported issue
  - Sentence-start capitalization
  - Question and exclamation punctuation
  - Multiline text
  - False-positive scenarios ("even worse", "the worst", etc.)
  - Regression coverage for existing "less worse" / "least worst" behavior

Demo

Before:

But it's even worst because when you look at it...

No lint generated.

After:

But it's even worst because when you look at it...

Suggestion:

But it's even worse because when you look at it...

How Has This Been Tested?

- Added unit tests covering the reported issue and related edge cases.
- Added regression tests to ensure existing "less worse" and "least worst" behavior remains unchanged.
- Ran project test suite locally.

AI Disclosure

If Your PR Implements or Enhances a Linter

Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,6 +71,7 @@ Agents should remind humans of this when possible.
 - `packages/web/src/routes/docs/contributors/author-a-rule/+page.md`: Step-by-step workflow for implementing and testing new grammar rules.
 - `packages/web/src/routes/docs/contributors/visual-studio-code/+page.md`: How to run, debug, test, and package the VS Code extension locally.
 - `packages/web/src/routes/docs/contributors/chrome-extension/+page.md`: Internal architecture and local development notes for the browser extensions.
+- `packages/web/src/routes/docs/contributors/website/+page.md`: Local development, preview, and validation for writewithharper.com (SvelteKit/SveltePress).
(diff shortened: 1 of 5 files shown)


hippietrail (changes requested):
I don't think this belongs in this existing linter. The existing linter is named and documented to deal with a quite distinct and specific mistake regarding the word "worse".

A better place would probably be in `/linting/phrase_set_corrections/mod.rs` in the `WorseOrWorst` section starting around line 985.

PR #3739, opened 2026-06-29
Fix URL lexer dropping the host and path when a port is present

chatman-media:
URLs with an explicit port lose everything after the scheme. `http://localhost:8080/api` gets tokenized with only `http://` as the URL, and `localhost:8080/api` ends up treated as ordinary text (so the host and path get spell-checked).

The cause is in `lex_hostport`: when a `:` follows the host, it looks for the end of the port by scanning for the first non-digit from the start of the host rather than from after the colon. Since the host starts with a letter (or the scan stops at the first dot), that search returns almost immediately and the port comes out zero-length, ending the URL right after the scheme.

This scans the port digits from just after the colon, and leaves the colon alone if there's no port behind it. Added a few cases to the url tests (`localhost:8080`, with a path, and an IP with a port).

### DIFF
--- a/harper-core/src/lexing/url.rs
+++ b/harper-core/src/lexing/url.rs
@@ -86,17 +86,19 @@ fn lex_hostport(source: &[char]) -> Option<usize> {
     let hostname_end = lex_hostname(source)?;
 
     if source.get(hostname_end) == Some(&':') {
-        Some(
-            source
-                .iter()
-                .enumerate()
-                .find(|(_, c)| !{
-                    let c = **c;
-                    c.is_ascii_digit()
-                })
-                .map(|(i, _)| i)
-                .unwrap_or(source.len()),
-        )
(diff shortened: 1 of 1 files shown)


hippietrail:
Yes I've made some toy projects recently annotated text tokenized by Harper, and sometimes whole URLs are interpreted as a single URL token and sometimes only the host is interpreted as a host token and all the other parts are interpreted as words and punctuation. But I haven't yet figured out the pattern or whether it's plain text input vs Markdown input.

I don't think the ones I encountered used the port, so I suspect there's more than one problem in our URL parsing.

elijah-potter:
@chatman-media, have you read our agent policy?

elijah-potter:
Given that I have not heard back from you regarding our agent policy, I am going to close this PR. Feel free to reopen if you circle back.

PR #3755, opened 2026-07-03
fix(core): use `default_config.json` as the canonical config

elijah-potter:
# Issues 

Related to #3444

# Description

For a long time, there have been two (confusingly) places where we "configure" the default config for a particular rule. One was in the `lint_config` curated list and another was in the `default_config.json`. I have removed the former from being a configuration option. The canonical version is now the `default_config.json` file. From now on, you only need to pay attention to the one file.

# AI Disclosure

# Checklist

### DIFF
--- a/packages/web/src/routes/docs/contributors/author-a-rule/+page.md
+++ b/packages/web/src/routes/docs/contributors/author-a-rule/+page.md
@@ -171,11 +171,11 @@ use super::correct_number_suffix::CorrectNumberSuffix;
 Finally, enable it in a macro invocation near the bottom:
 
 ```rust title="harper-core/src/linting/lint_group.rs"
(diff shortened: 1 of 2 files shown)


packages/web/src/routes/docs/contributors/author-a-rule/+page.md
```diff
@@ -171,11 +171,11 @@ use super::correct_number_suffix::CorrectNumberSuffix;
 Finally, enable it in a macro invocation near the bottom:
```
hippietrail: Is this wording with "enable" affected by the change? Maybe "add it with a macro invocation" would be better now?

hippietrail (commented):
Just a thought on the doc saying "enable".

PR #3757, opened 2026-07-04
feat: add YAML support

### LINKED ISSUE #2700: Feature request to support YAML
Issue #79 is getting crowded and hard to use. This issue is to make the request for **YAML** support easy to follow.

Mentioned twice: https://github.com/Automattic/harper/issues/79#issue-2266758540 and https://github.com/Automattic/harper/issues/79#issuecomment-2739083217

rodbegbie:
> **AI disclosure:** this PR was written by an AI coding agent (Claude), working with me interactively over the course of a session. I reviewed and directed the work throughout — see the "AI Disclosure" section below for specifics.

# Issues

Closes #2700

# Description

Adds YAML support to Harper: `harper-ls` and `harper-cli` now lint `.yaml`/`.yml` files, checking `#` comments and prose-like scalar values while leaving structural YAML (keys, identifiers, enum-like config values) alone.

This follows the same standalone-crate pattern as `harper-python` and `harper-html` (a new `harper-yaml` crate wired directly into `harper-ls`/`harper-cli`), rather than folding into `harper-comments`.

The three commits build on each other:

1. **`harper-yaml` crate itself** — `YamlMasker` runs two independent tree-sitter passes (comments, scalar values) over the parsed document and combines the surviving spans. Scalar-value detection has to explicitly exclude mapping keys, since unquoted YAML keys are the same tree-sitter node kind as values — this is done by comparing a candidate node's start byte against its enclosing `block_mapping_pair`/`flow_pair`'s start byte (a key is always the first token in its pair; a value always starts later, after `key:`). A small `heuristics::is_prose_scalar` function decides whether a scalar value looks like prose worth checking (3+ words, not solely a URL/path/version-shaped token) versus structural config data. A small `DedentLines` parser wrapper avoids a "double space" false positive that raw YAML block-scalar indentation would otherwise trigger.
2. **Wiring into `harper-ls`/`harper-cli`** — routes `"yaml"`/`.yaml`/`.yml` the same way `"python"`/`.py` already routes to `PythonParser`.
3. **Fuzzing, VS Code activation, and docs** — a dedicated `fuzz_harper_yaml` target (mirroring `fuzz_harper_html`), `onLanguage:yaml` activation plus a VS Code integration test, and a docs table entry.

# How Has This Been Tested?

- `cargo test --workspace` — full workspace suite passes, including 39 new unit/integration tests in `harper-yaml` covering comments, plain/quoted/block scalars, the `spellchecker:ignore`-family suppression, mapping-key exclusion, trailing same-line comments, and the URL/path/version/identifier heuristics.
- `just test-vscode` — the real VS Code extension integration suite (36 specs) passes, exercising the built `harper-ls` binary end to end.
- Manually verified via `harper-cli lint` against hand-written fixtures.

# AI Disclosure

# Checklist


harper-yaml/src/heuristics.rs
```diff
@@ -0,0 +1,196 @@
+/// Decides whether a YAML scalar value looks like prose worth
+/// grammar-checking, as opposed to structural config data
+/// (identifiers, enum values, etc).
+///
+/// Comments are never passed through this function — they're
+/// always linted, handled separately in `YamlMasker`.
+pub(crate) fn is_prose_scalar(text: &str) -> bool {
```
elijah-potter, line 7: The `(crate)` tag is almost never necessary.

harper-yaml/src/dedent.rs
```diff
@@ -0,0 +1,120 @@
...
+/// [`TokenKind::Newline`] token between lines.
+///
+/// This exists for YAML block/folded scalars: their continuation lines
+/// carry literal indentation (e.g. two leading spaces). Feeding
+/// that raw text straight to a prose parser makes the newline-plus-
+/// indentation read as a run of several space characters, which
+/// incorrectly triggers Harper's "double space" formatting rule.
+/// De-denting each line before parsing avoids that, without needing to
+/// understand YAML's specific indentation rules.
+pub(crate) struct DedentLines<P> {
```
elijah-potter: Same comment. The `(crate)` tag is superfluous.

harper-yaml/src/dedent.rs
```diff
@@ -0,0 +1,120 @@
...
+/// stitching the results back together with a single explicit
+/// [`TokenKind::Newline`] token between lines.
+///
+/// This exists for YAML block/folded scalars: their continuation lines
+/// carry literal indentation (e.g. two leading spaces). Feeding
+/// that raw text straight to a prose parser makes the newline-plus-
+/// indentation read as a run of several space characters, which
+/// incorrectly triggers Harper's "double space" formatting rule.
+/// De-denting each line before parsing avoids that, without needing to
+/// understand YAML's specific indentation rules.
```
elijah-potter: This functionality exists for other parsers (i.e. HTML). Would it be possible to centralize the logic? It would make it FAR easier to maintain.

fuzz/fuzz_targets/fuzz_harper_yaml.rs
```diff
@@ -0,0 +1,9 @@
+#![no_main]
+
+use harper_core::parsers::StrParser;
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &str| {
+    let parser = harper_yaml::YamlParser::default();
+    let _res = parser.parse_str(data);
+});
```
elijah-potter, line 9: I'm curious if you've actually run the fuzzer. If so, what were the results?

elijah-potter (changes requested):
Good stuff! I just have a few nits.

fuzz/fuzz_targets/fuzz_harper_yaml.rs
rodbegbie: Robot says:
>   Summary: Ran cargo +nightly fuzz run fuzz_harper_yaml for 120 seconds against harper_yaml::YamlParser::parse_str. 635,305 executions, no
>  crashes, no panics, coverage plateaued at 1534. Your harper-yaml parsing/masking work (48a4cc03, 101e1754) held up fine.

harper-yaml/src/dedent.rs
rodbegbie: Let me have the robot take a run at this, and see how it does.

harper-yaml/src/heuristics.rs
rodbegbie: Robot says:

>  Good instinct in general, but this one's actually required rather than superfluous: is_prose_scalar and DedentLines are declared in child
>  modules (heuristics/dedent) but called from lib.rs, the parent. Rust's privacy rule only extends downward — a private item is visible to
>  its own module and that module's descendants, not to its ancestors. So a plain fn/struct here isn't visible from lib.rs.
>
>  Confirmed by stripping the tag and running cargo check -p harper-yaml:
>
>  error[E0603]: struct `DedentLines` is private
>  error[E0603]: function `is_prose_scalar` is private
>
>  Happy to inline these into lib.rs if you'd rather avoid the separate modules, but as structured, pub(crate) needs to stay.

Let me look at Claude's suggestion about moving the dedent logic somewhere central, then come back to these and maybe move them into lib.rs if that makes sense then.

harper-yaml/src/dedent.rs
rodbegbie: >   Good catch, but I don't think we can reuse HtmlParser's approach directly. It solves a similar-looking symptom with a different mechanism:
>  it parses the whole text blob in one pass, then clamps every Space token to a max of 1, wherever it occurs
>  (harper-html/src/lib.rs:32-36). That's correct for HTML specifically, because browsers collapse all whitespace runs to a single space when
>  rendering — a source-level double space is never meaningful there. It's not safe for YAML: indentation is only structurally meaningful
>  right after a line break, and a genuine double-space typo mid-line in a scalar should still get flagged. Reusing the clamp as-is would
>  silently suppress real typos.
>
>  The closer precedent is actually Unit/Lua/Solidity in harper-comments/src/comment_parsers/ — they share DedentLines' surgical shape (split
>  by line, trim only the line-boundary whitespace/markers, parse each line independently, reinsert a single Newline token, offset spans),
>  just stripping comment initiators instead of whitespace. That pattern is duplicated three times there already, plus this one now, so
>  centralizing does seem worthwhile.
>
>  I'd rather scope that to a generic LineWise<P, S> combinator in harper-core::parsers (inner parser + a per-line "strip" function) and land
>  it as a follow-up PR against harper-comments, since Lua and Solidity have quirks (paragraph-break newlines, an SPDX special case) that
>  need care rather than a same-PR rider here. Happy to open that follow-up if useful.

I’m having it create that LineWise extraction on this branch, but not change the other parsers.  That feels like the right compromise for this PR. Let me know what you think.

rodbegbie: Landed — pulled the shared split-by-line/strip/parse/stitch mechanism into a generic `LineWise<P, S>` combinator in `harper_core::parsers` (parameterized by a per-line strip function), and `DedentLines` is now a thin wrapper supplying just the YAML whitespace-trim policy: 93075cee.

Scoped it to just this crate for now — `Unit`/`Lua`/`Solidity` in `harper-comments` share the same shape but have some quirks (paragraph-break newlines, an SPDX special case) that I'd rather migrate onto `LineWise` deliberately in a follow-up PR than as a rider here.

rodbegbie: Here’s a commit on another branch for the Unit/Lua/Solidity migration to use the LineWise combinator: https://github.com/Automattic/harper/pull/3909/changes/79f653b8473b4bdc23f3f1d70ccda84e9d130a3d

rodbegbie:
Hi @elijah-potter.  I’ve addressed your comments, and have updated the PR to cleanly land on master.  Let me know if you have more feedback. Thanks!

PR #3763, opened 2026-07-04
fix(core): catch less common that comparisons

### LINKED ISSUE #3736: False negative missing "that" as a typo for "than"
Image

We do have a linter that flags "that" when it should be "than" but I just spotted this exact typo by me right here in a comment and it's not being flagged.

> way less common **that** mp3

terminalchai:
AI/agent assistance was used to prepare this patch.

This fixes #3736 by extending the existing `ThatThan` linter to catch the reported `less common that mp3` comparison shape without enabling the broader `more/less adjective that` pattern that already has known false positives.

The matcher now supports the existing comparative-adjective pattern and a constrained `more/less common that <token containing a digit>` pattern. I also updated `match_to_lint` to locate the matched `that` token dynamically, then added true-positive tests for the reported MP3 comparison and false-positive tests for valid `less/more common that ...` clauses.

Verification:
- `cargo test -p harper-core that_than -- --nocapture`
- `cargo run -p harper-cli -- lint "This format is way less common that mp3."`
- `cargo run -p harper-cli -- lint "It is less common that people configure this manually."`
- `cargo fmt -- harper-core\src\linting\that_than.rs`
- `git diff --check`

### DIFF
--- a/harper-core/src/linting/that_than.rs
+++ b/harper-core/src/linting/that_than.rs
@@ -1,12 +1,12 @@
 use crate::linting::expr_linter::Chunk;
 use crate::{
-    Token, TokenKind,
-    expr::{Expr, SequenceExpr},
+    CharStringExt, Token, TokenKind,
+    expr::{AnchorEnd, Expr, FirstMatchOf, SequenceExpr},
     linting::{ExprLinter, Lint, LintKind, Suggestion},
 };
 
 pub struct ThatThan {
-    expr: SequenceExpr,
+    expr: FirstMatchOf,
 }
 
 impl Default for ThatThan {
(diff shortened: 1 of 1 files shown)


hippietrail:
I'm pretty sure this is going to have false positives along the lines of:
- it's less common that mp3 players break than CD players

terminalchai:
Good catch, thanks. I narrowed this so the new branch only matches the digit-bearing comparison target at the end of the chunk/sentence, with optional trailing punctuation. That keeps the reported `less common that mp3` case covered but avoids cases like `it's less common that mp3 players break than CD players`.

I added your example as a no-lint regression test and reran:
- `cargo test -p harper-core --lib that_than -- --nocapture`
- CLI lint checks for the reported typo and the false-positive example
- `git diff --check`

PR #3775, opened 2026-07-06
refactor: `assert_suggestion_result()` family of functions

hippietrail:
# Issues 
N/A

# Description

Some months back I used AI to help me convert the suggestion assertions to find any suggestion rather than just the first three, after failing a couple of attempts to do it with human brain only.

At the time I did not apply the fix also to `assert_good_and_bad_suggestions()` because it was a bit more complicated and I wanted to be sure the general fix was fine. It has now long since proven itself.

This PR goes a step further by converting these methods from using recursion to iterators, including `assert_good_and_bad_suggestions`.

I've looked through the code line by line and all tests pass, but because I used more AI than I like in the PR I'd appreciate more eyes checking it. My grasp of all the recursion vs iterator issues in Rust is probably not total at this point too.

There are some tests sprinkled through the code that don't go as far as they should due to the known limitations with the good vs bad assertion. These can be improved over time.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist


elijah-potter:
Let me check my understanding here: This PR is a refactor of the `assert_suggestion_result` family of functions to become iterative, rather than recursive. My assumption is that the goal is to simplify the code and make it easier to modify down the road. Is that correct?

hippietrail:
> Let me check my understanding here: This PR is a refactor of the `assert_suggestion_result` family of functions to become iterative, rather than recursive. My assumption is that the goal is to simplify the code and make it easier to modify down the road. Is that correct?

No it's more than that. The good/bad assert, the one you can give a list of multiple acceptable corrections and a list of bad corrections, did not use the full iterator because I wanted to be sure it worked with the simpler asserts that only checked against one correction, before adapting it to the more complicated one.

Main reason is I always found the logic hard to reason about and only solved it with AI help, so wanted another pair of eyes to check the soundness of the logic.

PR #3777, opened 2026-07-07
Fix false positive for modal 'may be'.

Tbuggee:
## Summary
- Prevent `may be` from being treated as a split compound noun for `maybe`
- Add regression coverage for reported `may be` false positives

## Testing
- `cargo test -p harper-core`
- `cargo fmt --check`

### DIFF
--- a/harper-core/src/expr/mergeable_words.rs
+++ b/harper-core/src/expr/mergeable_words.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use super::{Expr, SequenceExpr};
 use crate::spell::{Dictionary, FstDictionary};
-use crate::{CharString, DictWordMetadata, Span, Token};
+use crate::{CharString, CharStringExt, DictWordMetadata, Span, Token};
 
 type PredicateFn =
     dyn Fn(Option<&DictWordMetadata>, Option<&DictWordMetadata>) -> bool + Send + Sync;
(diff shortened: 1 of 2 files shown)


harper-core/src/expr/mergeable_words.rs
```diff
@@ -42,6 +42,12 @@ impl MergeableWords {
         let a_chars: CharString = word_a.get_ch(source).into();
         let b_chars: CharString = word_b.get_ch(source).into();
 
+        // "may be" is a modal construction, not a split compound noun.
+        // The closed word "maybe" has a noun sense, which can otherwise trigger false positives.
+        if a_chars.eq_str("may") && b_chars.eq_str("be") {
```
hippietrail, line 47: I'd soften the comment since people really do make the mistake of mixing up "maybe" and "may be" ("May be I'll see you there" etc) but not generally in a noun context ("Is that a no or a may be?")

But you're definitely right to add this check here because detecting when "maybe" or "may be" is a mistake would belong in its own linter without an error message about compounds nouns.

hippietrail (changes requested):
Looks good, I just have a minor nit about the comment.

harper-core/src/expr/mergeable_words.rs
Tbuggee: Thanks, softened the comment to avoid implying "may be" is always modal.

PR #3729, opened 2026-06-29
feat: lint the Source Control commit message box

dmytrosorokame:
# Issues 
Closes [#2988](https://github.com/Automattic/harper/issues/2988)

# Description
The VS Code extension now lints the commit message input in the Source Control panel. Those documents carry the `scminput` language id and a non-file URI, so the language client matches them by language alone, and harper-ls parses their contents with the git commit parser.

# Demo
Typing a commit message with spelling errors into the Source Control box
now surfaces Harper diagnostics and quick-fixes inline:

Screenshot 2026-06-29 at 08 07 43
Screenshot 2026-06-29 at 08 07 58
Screenshot 2026-06-29 at 08 08 32

# How Has This Been Tested?

**Automated:** Added an integration test in `packages/vscode-plugin/src/tests/suite/integration.test.ts` that opens an in-memory document, sets its language to `scminput` (the same language id VS Code gives the Source Control commit box), and asserts Harper's diagnostics. With `Errorz\n# Errorz`, the subject line is flagged while the `#` comment line is ignored — confirming the input is parsed as a git commit message. It runs in the existing `test-vscode` CI job.

**Manual:** Built `harper-ls` from this branch, ran the extension in an Extension Development Host, and typed `Fix detritis in teh code` into the Source Control commit box. Harper flags "detritis" and "teh" with suggestions, as shown in the demo.

> Note for reviewers: the linting lives in `harper-ls`, so the server must be rebuilt from this branch to see the effect — testing only the extension against a released `harper-ls` will show nothing. This mirrors the `just test-vscode` flow.

# AI Disclosure

- [ ] I am a human and didn't use any AI.
- [ ] I used LLM features of my editor, but not an agent.
- [x] I used an AI agent interactively.
- [ ] I am an agent or I got an agent to do the work autonomously.

# If Your PR Implements or Enhances a Linter

- [ ] I made up the sentences in the unit tests.
- [ ] The sentences in the unit tests were generated by an AI.
- [ ] I'm using examples from the bug report / feature request.
- [ ] I collected real-world sentences for the unit tests.

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests to cover my changes
- [x] I have considered splitting this into smaller pull requests.

PR #3782, opened 2026-07-07
feat: recognize Sublime Text's `git` language id as git commit messages

### LINKED ISSUE #3781: Support sublime-text git message
When editing commit message in sublime text, I had to do this config to make it work, in `language-ids.sublime-settings`:
```
{
  "text.git.commit": "git-commit"
}
```

It would be nice to add that language in the list for git commit: `git` (sublime text send `"languageId": "git"`)

I think here: https://github.com/Automattic/harper/blob/515a97b7f93b4afbcfe138e80e79140a3faab133/harper-ls/src/backend.rs#L364

costajohnt:
# Issues

Closes #3781

# Description

Sublime Text's LSP client reports `languageId: "git"` when you edit a git commit message, so harper-ls was not linting those buffers. This adds `git` to the git-commit parser match arm, next to the existing `git-commit` / `gitcommit` / `scminput` ids, so Sublime commit messages get the same grammar checking as the other editors.

One thing worth your call: Sublime derives the languageId from the syntax scope's second component (`basescope2languageid`), and its whole `text.git.*` family (commit, ignore, config, rebase, and so on) collapses to `git`. So this arm also catches non-commit git files edited in Sublime with harper attached. There is no way to tell them apart at the languageId layer. The blast radius is bounded: `GitCommitParser` masks through tree-sitter-gitcommit and only lints the subject and message lines, so a `.gitignore` just picks up some prose false positives, no crash. I left a comment on the arm noting this. If you would rather keep it narrower, the alternative is leaving users to add `{"text.git.commit": "git-commit"}` to their `language-ids.sublime-settings` (the reporter's current workaround), but that pushes config onto every Sublime user.

# Demo

Not applicable, this is an LSP languageId mapping. Behavior: Sublime sends `languageId: "git"` for a commit buffer, and harper-ls now routes it to the git commit parser instead of leaving it unlinted.

# How Has This Been Tested?

`cargo clippy -p harper-ls` and `cargo fmt --check -p harper-ls` are clean. The change mirrors the existing `git-commit` / `scminput` arms exactly (same target parser), differing only in the languageId string the client reports.

I did not add a Rust test. The languageId match lives inline in `update_document` with no unit-test seam, and the analogous `scminput` addition (#3729) was covered by the VS Code integration suite, which does not apply here since harper ships no Sublime client. Glad to add a test if you have a preferred harness for the languageId routing.

# AI Disclosure

Not applicable, this adds a language-id mapping rather than a linter rule.

# Checklist


elijah-potter:
I would prefer to err on the side of false-negatives. As such, I think the better course of action would be to better document the configuration workaround. If you'd like to create a PR for that, I'd be super open to that. For now, though, I think I am going to close this.

Thanks for the PR!

costajohnt:
Makes sense, thanks for the quick look. I opened https://github.com/Automattic/harper/pull/4029 documenting the configuration instead: the selector addition for `text.git.commit` plus the `language-ids.sublime-settings` mapping to `git-commit`.

PR #3792, opened 2026-07-08
fix(linting): stop flagging the adjective "leftover" as a phrasal verb

### LINKED ISSUE #3707: False positive: adjective "leftover" flagged to change to "left over"
Image

> Remnants" typically implies passive, **leftover** debris

It's not a phrasal verb or a noun in this context, it's an adjective and is marked as an adjective in `dictionary.dict`

AnayGarodia:

## What

Fixes #3707. `PhrasalVerbAsCompoundNoun` flagged the adjective **"leftover"** ("leftover debris") and suggested rewriting it to "left over".

## Why

"leftover" ends in the particle "over" and its prefix "left" is a verb, so it matches the rule's phrasal-verb shape. But used attributively it's an adjective, not a phrasal verb miswritten as a compound noun. This is the exact false-positive class the rule already handles for **"gallon"** ("gall on") and **"dragon"** ("drag on").

## How

Add `leftover` to the existing known-false-positive set in `analyse`, and update the accompanying comment. A blanket "skip adjectives" approach was rejected because "backup" is also tagged as an adjective, and skipping it would regress the rule's intended `backup` → `back up` corrections (verified locally — 2 existing tests fail with that approach).

## Testing

Added `dont_flag_leftover_adjective_3707` (reproduces the issue: 1 lint before, 0 after). Full rule suite green, no regressions:

```
cargo test -p harper-core --lib phrasal_verb_as_compound_noun   # 65 passed, 1 ignored
cargo fmt -- --check                                            # clean
cargo clippy -- -Dwarnings ...                                  # clean
```

## Compatibility

Purely narrows one false positive; no public API or output change for any other input.

### DIFF
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -84,9 +84,12 @@ impl PhrasalVerbAsCompoundNoun {
         }
 
         // * Must not be in the set of known false positives.
+        //   These end in a particle after a real verb ("gall on", "drag on",
+        //   "left over") yet aren't phrasal verbs miswritten as compound nouns.
         if noun_chars.eq_any_ignore_ascii_case_chars(&[
             &['g', 'a', 'l', 'l', 'o', 'n'],
             &['d', 'r', 'a', 'g', 'o', 'n'],
+            &['l', 'e', 'f', 't', 'o', 'v', 'e', 'r'],
         ]) {
             return Err(Why::ItsAKnownFalsePositive);
         }
@@ -826,6 +829,14 @@ mod tests {
     fn dont_flag_meltdown_3591() {
         assert_no_lints(
             "Unfortunately, Meltdown ended up being problematic.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+    }
+
+    #[test]
+    fn dont_flag_leftover_adjective_3707() {
+        assert_no_lints(
+            "\"Remnants\" typically implies passive, leftover debris.",
             PhrasalVerbAsCompoundNoun::default(),
         );
     }
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -875,6 +875,14 @@ mod tests {
         );
     }
 
+    #[test]
+    fn dont_flag_punctuation_separated_attributive_leftover_3707() {
+        assert_no_lints(
+            "The exhibition included discarded, leftover paint.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+    }
+
     // "leftover" is attributive after a number/determiner (Writer's Digest, Longman).
     // The preceding word here is adjacent, so the existing guard already covers it;
     // this pins that the redesign keeps the common attributive case quiet.
@@ -896,6 +904,15 @@ mod tests {
             "Backup your database before upgrading.",
             PhrasalVerbAsCompoundNoun::default(),
             "Back up your database before upgrading.",
+        );
+    }
+
+    #[test]
+    fn flag_backup_after_adverb_and_comma() {
+        assert_suggestion_result(
+            "Run fast, backup often.",
+            PhrasalVerbAsCompoundNoun::default(),
+            "Run fast, back up often.",
         );
     }
 }
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -145,13 +145,18 @@ impl PhrasalVerbAsCompoundNoun {
             return Err(Why::Contextual(Context::ItsInIsolation));
         }
 
-        // With no usable previous word (e.g. punctuation before it), a word that's also
-        // an adjective sitting right before a noun is attributive ("passive, leftover
-        // debris", #3707), not a phrasal verb miswritten as a compound noun. When there
-        // is a previous word the adjective/determiner/preposition guards below decide.
+        // A word that can be an adjective, sitting right before a noun and itself
+        // preceded by another adjective or determiner, is attributive rather than a
+        // phrasal verb: "passive, leftover debris" (#3707). The preceding modifier is
+        // separated by punctuation (the comma), so `maybe_prev_tok` misses it; we look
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -84,9 +84,12 @@ impl PhrasalVerbAsCompoundNoun {
         }
 
         // * Must not be in the set of known false positives.
+        //   These end in a particle after a real verb ("gall on", "drag on",
+        //   "left over") yet aren't phrasal verbs miswritten as compound nouns.
         if noun_chars.eq_any_ignore_ascii_case_chars(&[
             &['g', 'a', 'l', 'l', 'o', 'n'],
             &['d', 'r', 'a', 'g', 'o', 'n'],
+            &['l', 'e', 'f', 't', 'o', 'v', 'e', 'r'],
```
hippietrail: This is not appropriate. "Gallon" and "dragon" merely look like compound nouns that might be related to phrasal verbs but "leftover" is an actual compound noun that is related to an actual phrasal verb 'left over".

A different fix will have to be found. I highly recommend a data-based approach using as many actual mistakes and as many legit uses as you can find. AI coding agents so far seem to be terrible at this and only make up examples that fit their assumptions. I think they can't freely use Google, which is how I find examples and counterexamples when working on new linters.

hippietrail (changes requested):
This fix does not fit the pattern where it was added.

AnayGarodia:
you're right, that didn't fit. gallon/dragon are in the list because they're *coincidental* splits — "gall on"/"drag on" only look like verb+particle. "leftover" is a genuine one ("left over"), so hardcoding it was papering over the real reason.

reworked it: dropped "leftover" from the list and added a contextual guard — when there's no usable previous word (punctuation before it) and the word is itself an adjective sitting right before a noun, it's attributive ("passive, leftover debris") rather than a phrasal verb written as a compound noun. scoped to the no-previous-word case so it doesn't touch "to backup individual apps" and friends, which the existing prev-word guards already handle. generalizes past the one word, no corpus snapshot changes.

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -145,6 +145,17 @@ impl PhrasalVerbAsCompoundNoun {
             return Err(Why::Contextual(Context::ItsInIsolation));
         }
 
+        // With no usable previous word (e.g. punctuation before it), a word that's also
+        // an adjective sitting right before a noun is attributive ("passive, leftover
+        // debris", #3707), not a phrasal verb miswritten as a compound noun. When there
+        // is a previous word the adjective/determiner/preposition guards below decide.
+        if maybe_prev_tok.is_none()
+            && token.kind.is_adjective()
+            && maybe_next_tok.is_some_and(|t| t.kind.is_noun())
```
hippietrail, line 158: Another common gotcha is that many words are both nouns and adjectives, so you would want to find examples where the first such is in both categories and where the second such is in both categories.

Then you'd want to test with the first `token` being an adjective and not also being a noun and the second `maybe_next_tok` being a noun and not also an adjective.

And then you want to test whether doing so doesn't rule out too many instances of the problem you were trying to fix.

This is the kind of lint that having a big set of real-world text files or markdown files works great with, by the way.

hippietrail (changes requested):
More gotchas to look for I'm afraid. There are many subtleties in grammar checking due to words beloning in multiple POS categories.

AnayGarodia:
ok, redid this properly - dropped the earlier no-previous-word special case.

the thing #3707 was actually missing: "passive, leftover debris" already has an adjective ("passive") sitting right before the word, which is exactly what the existing guard at line 168 checks for. but the comma made get_next_word_from_offset return None, so that guard never got to run. so rather than special-casing "leftover" i made it look back past the punctuation to find that preceding adjective/determiner. same guard you already had, just no longer blocked by a comma.

kept it tight: the previous word (across punctuation) has to be an adjective or determiner, and the following word has to be a noun. so imperative errors still flag - "Backup your database before upgrading" -> "Back up ...", and "Run fast, backup often" stays flagged because "often" is an adverb, not a noun.

on the multiple-POS point: this linter only iterates noun tokens, so the target word is always a noun - there is no adjective-that-is-not-a-noun case on that side. on the following-word side i checked both noun-only (debris, dinner rolls, database) and noun+adjective.

for examples i went looking instead of inventing them. Writer's Digest and Longman for the attributive uses ("leftover paint", "three leftover dinner rolls"), and Linguix actually lists "back up" as a documented common mistake ("They backup the files" -> "back up"), plus grammar.com ("Please backup your work" -> "back up").

one thing i want to be upfront about: "Stay calm, backup files" gets suppressed now too. it is the same shape as "passive, leftover debris" - a coordinate adjective vs two imperative clauses split by a comma - and i do not think those can be told apart without a real parse. it is narrow (needs adjective, comma, the word, then a noun) and it trades against the more common attributive false positive, with no snapshot changes either way. if you would rather not make that trade i am happy to drop the approach.

AnayGarodia:
@hippietrail Added regression coverage for a second punctuation-separated attributive `leftover` use and for a comma-separated imperative error (`Run fast, backup often.`). The focused linter suite passes (69 passed, 1 pre-existing ignored). Could you please re-review?

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
AnayGarodia: Done in c8a0aa7 + 3ed422c: "leftover" is out of the gallon/dragon list — that list now keeps only the coincidental splits. The actual bug behind #3707: "passive, leftover debris" already has the preceding-modifier signal the rule checks for, but the comma stopped get_next_word_from_offset from seeing it. The fix looks back across non-terminal punctuation for the preceding adjective/determiner instead of special-casing the word.

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
AnayGarodia: Done in d236199: tests now cover the noun+adjective target before noun-only words ("leftover debris", "leftover paint", "three leftover dinner rolls") and the not-ruling-out-too-much side — "Backup your database before upgrading." and "Run fast, backup often." still get flagged. On the first-token side, an adjective-that-is-not-a-noun can't reach this code path: the linter only iterates tokens tagged noun.

Issue #3568, opened 2026-06-02
MoreAdjective should ignore compound adjectives


real-or-random:
**What got flagged?**
"There are more low-hanging fruits."

This gives me MoreAdjective that suggests replacing "more low" with "lower".

**Why is this incorrect?**
"More" refers to the "fruits" here and not to "low". The suggestion "There are lower-hanging fruits" has different meaning.

PR #3793, opened 2026-07-08
fix(linting): don't suggest an inflected form for "more <adj> <noun>"

### LINKED ISSUE #3688: False positive: "more short" flagged to be "shorted" in "more short videos".
Image

In the `MoreAdjective` rule.

AnayGarodia:

## What

Fixes #3688. `MoreAdjective` flagged **"more short videos"** and suggested the inflected form "shorter videos".

## Why

When the adjective is immediately followed by a noun it modifies, "more" quantifies the whole noun phrase ("more [short videos]") rather than intensifying the adjective, so rewriting to the comparative changes the meaning. This is the same "look one token past the adjective to spot a false positive" approach the rule already uses for the `than <adjective>` case (#2925) and the hyphen case (#3568).

## How

Add `whitespace + noun` as a third alternative in the rule's existing optional trailing clause (`FirstMatchOf`). When it matches, the span is longer than three tokens and the existing `toks.len() != 3` guard in `match_to_lint` already aborts — no new control flow. Existing true positives are unaffected because they aren't followed by a modified noun ("more fast", "more cute", "more good", "more humane than …").

## Testing

Added `dont_flag_more_short_videos_3688` (1 lint before, 0 after). Full rule suite green, no regressions:

```
cargo test -p harper-core --lib more_adjective    # 17 passed, 1 ignored
cargo fmt -- --check                              # clean
cargo clippy -- -Dwarnings ...                    # clean
```

## Compatibility

Purely narrows one Style-level false positive; no public API change.

### DIFF
--- a/harper-core/src/linting/more_adjective.rs
+++ b/harper-core/src/linting/more_adjective.rs
@@ -23,8 +23,8 @@ impl<D: Dictionary> MoreAdjective<D> {
                 // Or a following hyphen which we'll use to identify a false positive #3568
                 // Or a following noun the adjective modifies, where `more` quantifies the
                 // noun phrase ("more short videos") rather than the adjective — false positive #3688.
-                // Require noun-but-not-adjective so a following word that's also an adjective
-                // ("more common plural") doesn't wrongly suppress the comparative suggestion.
+                // Only when that noun is the phrase head: if another noun follows it, the word is
+                // itself adjectival ("more common plural noun") and we leave the suggestion alone.
                 .then_optional(FirstMatchOf::new([
                     Box::new(
                         SequenceExpr::whitespace()
@@ -33,10 +33,11 @@ impl<D: Dictionary> MoreAdjective<D> {
                             .then_positive_adjective(),
                     ) as Box<dyn Expr>,
                     Box::new(|tok: &Token, _source: &[char]| tok.kind.is_hyphen()),
-                    Box::new(SequenceExpr::whitespace().then_kind_is_but_is_not(
-                        |kind| kind.is_noun(),
-                        |kind| kind.is_adjective(),
-                    )),
+                    Box::new(
+                        SequenceExpr::whitespace()
+                            .then_noun()
+                            .then_unless(SequenceExpr::whitespace().then_noun()),
+                    ),
                 ])),
             dict,
         }
--- a/harper-core/src/linting/more_adjective.rs
+++ b/harper-core/src/linting/more_adjective.rs
@@ -357,4 +357,41 @@ mod tests {
             MoreAdjective::new(FstDictionary::curated()),
         );
     }
+
+    // Real-world sentences (all present in harper-core/tests/text/) that exercise the
+    // trailing-noun head check from #3688. The contrast between a noun head (suppress)
+    // and a noun+adjective word that modifies a following noun (still suggest) shows the
+    // check is doing real work, not passing by coincidence.
+
+    #[test]
+    fn dont_flag_more_perfect_union_3688() {
+        // US Constitution: "a more perfect Union". "Union" is the noun head, so "more"
+        // quantifies the phrase and "perfecter" must not be suggested.
+        assert_no_lints(
+            "in Order to form a more perfect Union,",
+            MoreAdjective::new(FstDictionary::curated()),
+        );
+    }
+
+    #[test]
+    fn dont_flag_more_even_then_3688() {
+        // The Great Gatsby: "loved me more even then". "then" is not a noun, so "even" is
+        // the head; "evener" must not be suggested.
+        assert_no_lints(
+            "and loved me more even then, do you see?",
+            MoreAdjective::new(FstDictionary::curated()),
+        );
+    }
+
+    #[test]
+    fn flag_more_common_before_plural_noun_3688() {
+        // A Wikipedia part-of-speech article: "the more common plural noun". "plural" is
+        // followed by the noun "noun", so it's adjectival, not the head - the comparative
+        // suggestion for "common" should still surface.
+        assert_suggestion_result(
+            "as the more common plural noun.",
+            MoreAdjective::new(FstDictionary::curated()),
+            "as the commoner plural noun.",
+        );
+    }
 }
--- a/harper-core/src/linting/more_adjective.rs
+++ b/harper-core/src/linting/more_adjective.rs
@@ -341,12 +341,20 @@ mod tests {
 
     #[test]
     fn dont_flag_more_short_videos_3688() {
-        // "more" quantifies the noun phrase ("more [short videos]"), so "shorter
-        // videos" is not the intended meaning.
         assert_no_lints(
             "I want to watch more short videos.",
             MoreAdjective::new(FstDictionary::curated()),
         );
+    }
+
+    #[test]
+    fn dont_flag_real_world_quantified_adjective_noun_phrases_3688() {
+        for text in [
+            "Young people are living in more diverse arrangements than at any point in the last 40 years.",
+            "It is among the more common ways to form a plural noun.",
+        ] {
+            assert_no_lints(text, MoreAdjective::new(FstDictionary::curated()));
+        }
     }
 
     #[test]
--- a/harper-core/src/linting/more_adjective.rs
+++ b/harper-core/src/linting/more_adjective.rs
@@ -22,7 +22,9 @@ impl<D: Dictionary> MoreAdjective<D> {
                 // Include a following "than adjective" which we'll use to identify a false positive #2925
                 // Or a following hyphen which we'll use to identify a false positive #3568
                 // Or a following noun the adjective modifies, where `more` quantifies the
-                // noun phrase ("more short videos") rather than the adjective — false positive #3688
+                // noun phrase ("more short videos") rather than the adjective — false positive #3688.
+                // Require noun-but-not-adjective so a following word that's also an adjective
+                // ("more common plural") doesn't wrongly suppress the comparative suggestion.
                 .then_optional(FirstMatchOf::new([
                     Box::new(
                         SequenceExpr::whitespace()
@@ -31,7 +33,10 @@ impl<D: Dictionary> MoreAdjective<D> {
                             .then_positive_adjective(),
                     ) as Box<dyn Expr>,
                     Box::new(|tok: &Token, _source: &[char]| tok.kind.is_hyphen()),
-                    Box::new(SequenceExpr::whitespace().then_noun()),
+                    Box::new(SequenceExpr::whitespace().then_kind_is_but_is_not(
+                        |kind| kind.is_noun(),
+                        |kind| kind.is_adjective(),
+                    )),
                 ])),
             dict,
         }
--- a/harper-core/src/linting/more_adjective.rs
+++ b/harper-core/src/linting/more_adjective.rs
@@ -21,6 +21,8 @@ impl<D: Dictionary> MoreAdjective<D> {
                 .then_positive_adjective()
                 // Include a following "than adjective" which we'll use to identify a false positive #2925
                 // Or a following hyphen which we'll use to identify a false positive #3568
+                // Or a following noun the adjective modifies, where `more` quantifies the
+                // noun phrase ("more short videos") rather than the adjective — false positive #3688
                 .then_optional(FirstMatchOf::new([
                     Box::new(
                         SequenceExpr::whitespace()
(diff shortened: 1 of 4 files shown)


AnayGarodia:
Updated the `test_most_lints` corpus snapshots: the fix removes 6 awkward `MoreAdjective` suggestions where `more/most` is followed by an adjective+noun — e.g. "more common plural noun" → "commoner", **"a more perfect Union" → "perfecter"**, "more cheerful carriages" → "cheerfuller". All removals are this one rule; no other lints changed. (The `test-chrome-plugin`/`test-firefox-plugin` failures look like unrelated browser-test flakiness — they also fail on my Rust-only sibling PRs.)

harper-core/src/linting/more_adjective.rs
```diff
@@ -29,6 +31,7 @@ impl<D: Dictionary> MoreAdjective<D> {
                             .then_positive_adjective(),
                     ) as Box<dyn Expr>,
                     Box::new(|tok: &Token, _source: &[char]| tok.kind.is_hyphen()),
+                    Box::new(SequenceExpr::whitespace().then_noun()),
```
hippietrail: This new branch's prefix overlaps with the branch's one since they both start with whitespace, which isn't necessarily wrong but I always avoid it since things like overlapping POSes are a common source of unexpected false positives and false negatives.

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
```diff
@@ -65,16 +65,6 @@ Suggest:
 
 
 
-Lint:    Style (127 priority)
-Message: |
-      32 | Correct grammatical tagging will reflect that "dogs" is here used as a verb, not
-      33 | as the more common plural noun. Grammatical context is one way to determine
-         |        ^~~~~~~~~~~ This is not an error, but an inflected form of this adjective also exists
-Suggest:
-  - Replace with: “commoner”
```
hippietrail: This is a problem even though it looks like a bonus. You and I might avoid using "commoner" as an adjective, but it's a legitimate adjective and noun. Hiding the suggestion this way is an unexpected side-effect that should probably be considered by using something other than `.then_noun()` that takes into account that it might be both a noun and an adjective.

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
```diff
@@ -393,16 +383,6 @@ Suggest:
 
 
 
-Lint:    Style (127 priority)
-Message: |
-     142 | parsing (1997) that merely assigning the most common tag to each known word and
-         |                                          ^~~~~~~~~~~ This is not an error, but an inflected form of this adjective also exists
```
hippietrail: Same as previous.

harper-core/tests/text/linters/The Constitution of the United States.snap.yml
```diff
@@ -22,16 +22,6 @@ Message: |
 
 
 
-Lint:    Style (127 priority)
-Message: |
-       5 | **We the People** of the United States, in Order to form a more perfect Union,
-         |                                                            ^~~~~~~~~~~~ This is not an error, but an inflected form of this adjective also exists
```
hippietrail: Same again. I don't like the adjective "perfecter" but it's still a legit word some users might want.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -3140,16 +3140,6 @@ Suggest:
 
 
 
-Lint:    Style (127 priority)
-Message: |
-    2073 | A dead man passed us in a hearse heaped with blooms, followed by two carriages
-    2074 | with drawn blinds, and by more cheerful carriages for friends. The friends
-         |                           ^~~~~~~~~~~~~ This is not an error, but an inflected form of this adjective also exists
```
hippietrail: Same.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -6437,16 +6427,6 @@ Suggest:
 
 
 
-Lint:    Style (127 priority)
-Message: |
-    4883 | “Of course she might have loved him just for a minute, when they were first
-    4884 | married—and loved me more even then, do you see?”
-         |                      ^~~~~~~~~ This is not an error, but an inflected form of this adjective also exists
```
hippietrail: Same.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
```diff
@@ -7650,15 +7630,6 @@ Suggest:
 
 
 
-Lint:    Style (127 priority)
-Message: |
-    5659 | One of my most vivid memories is of coming back West from prep school and later
-         |           ^~~~~~~~~~ This is not an error, but an inflected form of this adjective also exists
-Suggest:
-  - Replace with: “vividest”
```
hippietrail: Same again but unlike the others where the suggested adjectives all seemed ugly, "vividest" is a suggestion I would definitely consider.

There may be even better adjective suggestions this change prevents that we haven't thought of.

hippietrail (changes requested):
More care needs to be taken to deal with words that are both noun and adjective so that we don't block suggestions. Look at the methods starting with `.is_kind` - there should be one for "is but is not".

AnayGarodia:
switched to `then_kind_is_but_is_not(is_noun, is_adjective)` per your pointer (same shape as the one in `cant.rs`). so a genuine noun head still suppresses — "more short videos" stays quiet — but a trailing word tagged both noun *and* adjective no longer blocks the suggestion.

heads up on the corpus fallout: three style suggestions come back that `then_noun()` was silently swallowing — "more common plural" → commoner, "a more perfect Union" → perfecter, "more even then" → evener. that's the direct consequence of not blocking on the ambiguous words. perfecter/evener are debatable, but they're Style hints (dismissable), and filtering them back out would just be the hardcoded-list thing again. lmk if you'd rather draw the line tighter.

AnayGarodia:
follow-up — the `is_noun && !is_adjective` version was too blunt. it un-suppressed *every* noun+adjective word, which pulled back a couple of bad ones on non-gradable adjectives ("a more perfect Union" → perfecter, "more even then" → evener). that's not what you wanted either.

refined it to head-detection: keep `then_noun()` but require the trailing noun to be the phrase *head* — `then_unless(whitespace + noun)`. if another noun follows it, the word is functioning adjectivally ("more common plural noun") so the suggestion stays; if it's the head ("more short videos", "more perfect Union,") it suppresses. net: commoner still surfaces (the genuine noun+adjective case), perfecter and evener are gone. only two corpus lines change, both removals.

harper-core/src/linting/more_adjective.rs
```diff
@@ -29,6 +33,11 @@ impl<D: Dictionary> MoreAdjective<D> {
                             .then_positive_adjective(),
                     ) as Box<dyn Expr>,
                     Box::new(|tok: &Token, _source: &[char]| tok.kind.is_hyphen()),
+                    Box::new(
+                        SequenceExpr::whitespace()
+                            .then_noun()
+                            .then_unless(SequenceExpr::whitespace().then_noun()),
```
hippietrail, line 39: This seems confusing but it looks like it now matches "more common noun" but not "more common noun noun"? However, "more common noun" followed by something other than whitespace or "more common noun <whitespace>" followed by something other than a noun will still actually be matched?

When rules start to get confusing with exceptions to exceptions etc. you definitely will want more than a single test case so that you can chase down each one of these loose ends to make sure we're not getting the desired results by fluke.

Unfortunately `UnlessStep`, `.unless()` and `.then_unless()` all lack any tests to clarify what they are and are not intended to do.

hippietrail (changes requested):
I think we need more tests to be sure this new logic does what it's intended to do rather than just seeming to work by coincidence. With a very strong preference for real-world examples in the tests rather than contrived ones made up by an AI. Made up by a human would lie somewhere between.

AnayGarodia:
added the tests. all three are real sentences pulled straight from the corpus (harper-core/tests/text/), not made up: "a more perfect Union" from the Constitution, "loved me more even then" from Gatsby, and "the more common plural noun" from the part-of-speech tagging article. grep for them if you want to confirm.

the first two suppress (noun head, and "then" not being a noun), the third still suggests commoner since "plural" is modifying the following "noun" rather than being the head. that contrast is the whole point of the head check, so if it ever regresses one of these three will catch it.

hippietrail:
> added the tests. all three are real sentences pulled straight from the corpus (harper-core/tests/text/), not made up: "a more perfect Union" from the Constitution, "loved me more even then" from Gatsby, and "the more common plural noun" from the part-of-speech tagging article. grep for them if you want to confirm.

The problem is that three is still a tiny tiny sample. Ideally we'd be looking outside our tiny internal corpus into GitHub, Stack Overflow, Hacker News, Slashdot, Medium, Reddit where bad English abounds as well as good English. I do this with Google to varying levels of success. I understand AIs can't do this due to `robots.txt` and Google bot checks.

AnayGarodia:
@hippietrail Expanded #3688 coverage with externally sourced real-world quantified adjective-noun examples, while retaining the existing real-world positive contrast that must still lint. Focused `more_adjective` tests pass (21 passed, 1 pre-existing ignored). Could you please re-review?

harper-core/src/linting/more_adjective.rs
AnayGarodia: The two branches do still both start with whitespace, but they split on the very next token ("than" vs a noun), and both paths are pinned by tests as of 58337f3, so an overlap-induced behavior change would surface in the suite.

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
AnayGarodia: This suggestion came back in e8cc90b: "plural" is followed by the noun "noun", so it's treated as adjectival rather than the phrase head, and "commoner" still surfaces. Pinned by flag_more_common_before_plural_noun_3688.

harper-core/tests/text/linters/Part-of-speech tagging.snap.yml
AnayGarodia: Different outcome from the previous one as of e8cc90b: here "tag" is the phrase head ("the most common tag to each known word"), so this stays suppressed — the #3688 pattern the PR targets.

harper-core/tests/text/linters/The Constitution of the United States.snap.yml
AnayGarodia: As of e8cc90b this stays suppressed: "Union" is the head noun, so more/most is read as quantifying the noun phrase. Pinned by dont_flag_more_perfect_union_3688; the open trade-off for this class is summarized in the vividest thread below.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
AnayGarodia: Same shape under e8cc90b: "carriages" is the head noun, so this stays suppressed.

harper-core/tests/text/linters/The Great Gatsby.snap.yml
AnayGarodia: Stays suppressed as of e8cc90b; pinned by dont_flag_more_even_then_3688 ("and loved me more even then, do you see?").

harper-core/tests/text/linters/The Great Gatsby.snap.yml
AnayGarodia: Honest status as of e8cc90b: this one is still suppressed — "memories" is the head noun, so it falls under the same head check as "more short videos". This is the real trade-off left in the PR: the head check alone can't tell a quantifier reading from a degree reading. If you'd like contexts where the degree reading is forced (e.g. "the/my most <adj> <noun>") to keep suggesting, I can add that as a narrower carve-out — say the word.

harper-core/src/linting/more_adjective.rs
AnayGarodia: Done in 86ec915 + 58337f3: added tests chasing each loose end — head noun followed by punctuation suppresses ("a more perfect Union,"), noun followed by another noun keeps the suggestion ("the more common plural noun" -> "commoner"), a non-noun follower ("more even then"), plus externally sourced quantified examples ("more diverse arrangements than at any point..."). 21 focused tests pass.

PR #3794, opened 2026-07-08
fix(linting): don't flag sentence-initial "Backup" before a modal

### LINKED ISSUE #3661: "Backup" is flagged to become "back up" when it's the first word of a sentence and followed by a modal verb
Image

AnayGarodia:

## What

Fixes #3661. `PhrasalVerbAsCompoundNoun` flagged sentence-initial **"Backup"** and suggested "back up" in sentences like "Backup will run automatically."

## Why

The rule already treats a following present/simple-past/past verb form as evidence that the compound noun is the subject (`ItsFollowedByVerb`), e.g. "Backup is scheduled" and "Backup has finished" are correctly left alone. But **modal/auxiliary** verbs weren't covered, so "Backup **will** run", "Backup **can** be restored", and "Backup **should** complete" were still flagged.

## How

Add `next_tok.kind.is_auxiliary_verb()` to the existing "followed by a verb" check. Modals (`will`/`can`/`should`/…) are tagged as auxiliary verbs, so this closes the gap without new control flow. True positives are unaffected — they are followed by a preposition or end of clause ("I backup", "you checkout"), not by an auxiliary.

## Testing

Added `dont_flag_backup_before_modal_3661`. Full rule suite green, no regressions:

```
cargo test -p harper-core --lib phrasal_verb_as_compound_noun   # 65 passed, 1 ignored
cargo fmt -- --check                                            # clean
cargo clippy -- -Dwarnings ...                                  # clean
```

## Compatibility

Narrows one false positive; no public API or other-input behavior change.

### DIFF
--- a/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
+++ b/harper-core/src/linting/phrasal_verb_as_compound_noun.rs
@@ -835,12 +835,69 @@ mod tests {
     }
 
     #[test]
-    fn dont_flag_backup_before_modal_3661() {
-        // Sentence-initial "Backup" followed by a modal is the subject noun,
-        // not a phrasal verb miswritten as a compound noun.
+    fn dont_flag_sentence_initial_backup_before_modal_3661() {
+        // Sentence-initial "Backup" is the subject noun when a modal follows, not a
+        // phrasal verb miswritten as a compound noun. #3661 was "Backup will run
+        // automatically"; the same reasoning holds across modals.
         assert_no_lints(
             "Backup will run automatically.",
             PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "Backup can be restored from the cloud.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "Backup should complete within an hour.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+    }
+
+    #[test]
+    fn dont_flag_back_up_verb_before_modal_3661() {
+        // Real-world sentences (found by @hippietrail while reviewing #3794) where the
+        // phrasal verb "back up" is legitimately followed by a modal. They are already
+        // two words so the rule never treats them as candidates, but they document that a
+        // following modal is not on its own evidence that the word is a subject noun.
+        assert_no_lints(
+            "The data you back up should be determined by how important it is to you.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "The number of files you back up can affect both the time and the size.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "Any gift you leave to Back Up will be deducted from your estate.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "The photos you back up might capture quiet moments at home.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "Immediately swimming back up would probably be enough.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+        assert_no_lints(
+            "The data you can back up may vary depending on your operating system.",
+            PhrasalVerbAsCompoundNoun::default(),
+        );
+    }
+
+    #[test]
+    fn still_flag_backup_verb_when_no_modal_follows_3661() {
+        // The modal exception must stay narrow: a "backup" written for the verb "back up"
+        // is still flagged when a modal does not follow it.
+        assert_suggestion_result(
+            "Backup your files daily.",
+            PhrasalVerbAsCompoundNoun::default(),
+            "Back up your files daily.",
+        );
+        assert_suggestion_result(
+            "You should backup before upgrading.",
+            PhrasalVerbAsCompoundNoun::default(),
+            "You should back up before upgrading.",
         );
     }
 }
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -256,11 +256,13 @@ impl PhrasalVerbAsCompoundNoun {
             return Err(Why::Contextual(Context::ItsActuallyPartOfANounPhrase));
         }
 
-        // If the next word is a present or simple past verb form compatible with a 3rd person singular noun
+        // If the next word is a present or simple past verb form compatible with a 3rd person singular noun,
+        // or an auxiliary/modal verb ("Backup will run", "Backup can be restored"), the word is the subject noun.
```
hippietrail: There is an assumption that a modal verb can't also be used after a phrasal verb. These kinds of assumptions need to be verified. AIs are terrible at this. I use Google manually and I think they are not able to even if they hadn't already "made up their minds". Here's what I found:
- The data you **back up should** be determined by how important it is to you
- The number of files you **back up can** affect both the time required to run the backup and the size of the resulting backup.
- But if your estate is going to attract inheritance tax, any gift you leave to **Back Up will** be deducted from your estate before any tax liability is calculated.
- The photos you **back up might** capture quiet moments at home
- Swimming to the bottom, touching your elbows to a fresh fuel canister, and immediately swimming **back up would** probably be enough to kill you.
- The data you can **back up may** vary depending on your operating system or phone version.

These all read very naturally and give the impression that legit use of "back up" followed by a modal is common

hippietrail (changes requested):
Thanks for this but there are subtle things that the coding AIs are not yet good at in relation to grammar checking.

AnayGarodia:
fair caution. the subtle bug you're pointing at: I put the modal check in the general `ItsFollowedByVerb` block, so it fired mid-sentence too — but the test only justifies the sentence-initial case. mid-sentence "will/can/might" are as often nouns as modals, so that was broader than the reasoning behind it.

tightened it: the auxiliary/modal signal now only applies with nothing before the word (`maybe_prev_tok.is_none()`), which is the "Backup will run" case the test covers. the existing 3rd-person/past-form checks stay as they were. still suppression-only, no corpus snapshot changes.

if there's a specific construction you had in mind that's still off, point me at it — happy to add it as a test.

hippietrail:
> fair caution. the subtle bug you're pointing at: I put the modal check in the general `ItsFollowedByVerb` block, so it fired mid-sentence too — but the test only justifies the sentence-initial case. mid-sentence "will/can/might" are as often nouns as modals, so that was broader than the reasoning behind it.
> 
> tightened it: the auxiliary/modal signal now only applies with nothing before the word (`maybe_prev_tok.is_none()`), which is the "Backup will run" case the test covers. the existing 3rd-person/past-form checks stay as they were. still suppression-only, no corpus snapshot changes.

Yep if you've already got something like `maybe_prev_tok` that's a good approach. When you don't already have it, there's the `AnchorStart` expression.

> if there's a specific construction you had in mind that's still off, point me at it — happy to add it as a test.

I'm on the road right now so might not have the brain space to dedicate to think it through fully but I'll definitely let you know if one does pop into my head.

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
```diff
@@ -829,4 +833,14 @@ mod tests {
             PhrasalVerbAsCompoundNoun::default(),
         );
     }
+
+    #[test]
```
hippietrail, line 837: I believe it was in this thread where I posted half a dozen counterexamples earlier? Those were real-word sentences found by Googling so I would add at least one of them and in this case possibly all of them as additional tests.

hippietrail (changes requested):
More tests would help a lot.

AnayGarodia:
added the tests. pulled in all six of your back-up-before-modal sentences as real-world cases — they stay unflagged, as they should.

also widened the sentence-initial check past "will" to "can"/"should", and added the bit that actually pins down the assumption: a bare "backup" with no modal after it still gets flagged ("Backup your files daily" → "Back up your files daily", "You should backup before upgrading" → "back up"). so the exception only fires sentence-initial with a modal right after, and real errors still come through.

one honest note: your six are already two words, so the rule never considers them candidates in the first place — they pass trivially. i kept them because they document that "back up" + modal is legit, but the sentence-initial and still-flagged cases are the ones that actually exercise the change. happy to drop the trivial ones if you'd rather.

AnayGarodia:
@hippietrail Following up after the earlier revision: the modal guard remains limited to sentence-initial `backup` cases, and the reviewer-provided real-world `back up` + modal examples were added as regression tests. Could you please re-review?

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
AnayGarodia: Done in 622a87a: the modal signal no longer fires mid-sentence — it's restricted to sentence-initial position (maybe_prev_tok.is_none()), i.e. the "Backup will run" case. All six of your back-up-before-modal sentences are in the test suite as of aad85f1 and stay unflagged.

harper-core/src/linting/phrasal_verb_as_compound_noun.rs
AnayGarodia: Done in aad85f1: all six are in dont_flag_back_up_verb_before_modal_3661, alongside still-flagged guards pinning the assumption ("Backup your files daily." -> "Back up your files daily.", "You should backup before upgrading." -> "back up").

PR #3805, opened 2026-07-11
fix(core): flag "a software" with modifiers

### LINKED ISSUE #3799: False negative: `MassNouns` doesn't flag "a software" that has modifiers/qualifiers
> bibisco is a novel writing software that helps you plan, organize and write your story without getting lost.

"a software" is very common among nonnative users of English and has been adopted by some in the younger generation of native speakers. In this case I think the writer is a native Italian speaker.

Harper flags it in isolation as "a software" but not in "a novel writing software".

This is an interesting test case too because both "novel" and "writing" can be analysed as either adjectives or nouns though in this context "novel" is a noun and "writing" is a gerund, which is a verb used as a noun.

Correctly, the phrase should be hyphenated as "novel-writing software", but that's a separate job for a different linter and the linter that flags "a software" should be permissive about what it accepts.

ychampion:
AI disclosure: This change was implemented by an autonomous Codex agent and reviewed and tested in the local workspace.

# Issues

Fixes #3799

# Description

`MassNouns` now recognizes `software` as the head of a qualified noun phrase, including the reported `a novel writing software` case. Corrections preserve the original modifiers.

The matcher stops when a later noun becomes the head, so compounds such as `a software tool` and `a software engineering manager` remain unchanged. This is scoped to `software`; applying the same heuristic to every mass noun caused corpus false positives for words with valid countable senses.

# Demo

`a novel writing software` now offers `a novel writing program`, `novel writing software`, and `some novel writing software`.

# How Has This Been Tested?

- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p harper-core --quiet`
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p harper-core --no-default-features --quiet`
- `cargo clippy -p harper-core --lib -- -D warnings -A clippy::useless-borrows-in-formatting -A clippy::question-mark`
- `cargo clippy -p harper-core --lib --no-default-features -- -D warnings -A clippy::useless-borrows-in-formatting -A clippy::question-mark`
- `cargo fmt --all -- --check`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/mass_nouns/noun_countability.rs
+++ b/harper-core/src/linting/mass_nouns/noun_countability.rs
@@ -5,7 +5,7 @@ use crate::{
         ExprLinter, Lint, LintKind, Suggestion,
         expr_linter::{Chunk, followed_by_hyphen, followed_by_word},
     },
-    patterns::{IndefiniteArticle, WordSet},
+    patterns::{IndefiniteArticle, Pattern, WordSet},
 };
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/mass_nouns/noun_countability.rs
```diff
@@ -466,6 +594,68 @@ mod tests {
...
+        assert_no_lints(
+            "She is a software engineering manager.",
+            NounCountability::default(),
+        );
+        assert_no_lints(
+            "They operate a custom software development platform.",
+            NounCountability::default(),
+        );
+    }
+
```
hippietrail, line 665: The logic is quite unorthodox but you have quite a few tests. A very common with AI generated code for grammar checkers is that they lack talent for thinking up or finding edge cases. They usually just make up obviously passing tests. Humans have the same problem at first but they can learn from experience.

I'll see if I can find some examples for testing. I'm using Google with "a OR an * software"...

Should not be flagged
- A full software stack includes many components
- What makes a good software program?
- Most software developers aim to create software that bears the following principles
- CodeFusion Studio is an embedded software development platform leveraging Microsoft's Visual Studio Code.
- A third software category is that of network software, which coordinates communication between the computers linked in a network.
- we define an experiment as the Cartesian product of a given software stack and its configuration
- a novel software agent framework designed to bridge the gap between reasoning depth, efficiency, and context constraints
- A fork is the process of creating a separate and distinct piece of software by copying source code from an existing software package.
- MediaWiki is a free software open source wiki package written in PHP
- To a company software is a means to its ultimate goal of more profits. (Here "a company software" looks like a match for this mistake, but it's actually a missing comma and should be "To a company, software is a means ..." This is so tricky do detect that we might be able to put up with it if it's rare as long as we document it as something to be improved)
- Not to say that there isn't a reason software grows so fast. (Here "reason" is a noun so can look like "reason software" is a compound noun, but it's just another way to say "there isn't a reason that software grows so far". Like the previous one we might be able to put up with such a false positive if it's rare and we document it as an outstanding fault.)
- As a truism Software Engineering is the act of engineering software. (Here again "truism software" can look like a compound noun but should have a comma between, but "software engineering" is a legit compound noun, but "a ... engineering" can still look wrong. None of that is the business of this linter though.)

Should be corrected
- How do I find a quality software?
- Extensible and open source Platform as a Service software.
- GLPI is a living software.
- What's an old software that you would like to have again?
- I have not yet seen a SAP software that could not be replaced with a homemade Django/Rails web app
- My question is - have I just built a custom software for my own use-case?
- Expressive Animator is an SVG vector animation software that helps users create and export animated icons
- you would never go through chain of approvals to buy a cheap software
- We do not expect to come as a pro software like them.
- At newegg bought a virus software for $60.
- Has anyone used a software called gifcam or similar?
- But you're right, the goal is not to write test but to ensure delivery of a reliable software.
- However each software is a prototype, something that has never been made before

Should be corrected by a different linter
- This is a free software code, use fairly! (Here they're saying "a code" rather than "a software". This linter should ignore it as it's a different issue.)
- Since the WSUS configuration changed, a full software updates synchronization will occur rather than a delta synchronization. (Here the mistake is "software updates synchronization" which should use singular "update". This linter should ignore it as it's a different issue.)

Try making an individual test from each one of these sentences. They should all pass or fail as per how they're categorized, but we can tolerated one with notes in parentheses not doing so, in which case leave the tests in but with an `#[ignore]` message about why.

hippietrail (changes requested):
AI-written linters, especially for tricky mistakes that need context to disambiguate need a lot of real world test cases. I spent some human time and effort to find some to help see how close this is.

harper-core/src/linting/mass_nouns/noun_countability.rs
ychampion: Added an individual test for every sentence. All categorized cases now pass; the two ambiguous missing-comma and omitted-`that` examples remain as ignored tests with the reasons documented. The `a pro software like them` case exposed two tagger edge cases, so I scoped both fixes to `software` and reran the full default/no-default core suites and corpus snapshots.

ychampion:
Thanks for the examples. I added each sentence as its own regression case and kept the two ambiguous parser cases as documented ignores. The full default/no-default core suites and corpus snapshots pass.

PR #3817, opened 2026-07-13
feat(core): flag extraneous "so" in "very much so"

### LINKED ISSUE #1911: Flag "very much so" when the "so" doesn't refer back to something prior
**The Error**
This seems to be a fairly new error among native speakers that I've been noticing. (Apparently non-native speakers are also making this mistake.)

People now sometimes say "I very much so liked it" instead of "I very much liked it".
The version with "so" needs to refer back to something like in: "Did you like it? Yes, very much so!"

I think there may be other constructions in which this extraneous "so" is occurring.

**Examples:**
1. I very much **so** feel like this is the turning point.
2. I very much **so** have the text files with the appropriate names in the folder with the training images.
3. but I very much **so** want to make a modular software that integrates this boards
4. Hey we're also encountering this bug, and it's very much **so** a massive advantage for Bedrock players compared to Java in these cases.
5. Thanks `@kin0025` & `@primeinc` for reporting bugs, I very much **so** appreciate it.
6. And personally, I'm **very much so** enjoying it. 
7. ... who was **very much so** assaulted with questions yesterday from Congress
8. ... it looks like three of the four measurements are **very much so** out of spec

**Potential Edge Cases**
- Thank you very much so far!

ozpool:
# Issues

Closes #1911.

# Description

`very much so` only reads correctly when the `so` points back to something earlier, as in a reply — *"Did you like it? Yes, very much so."* When the `so` sits in front of the verb it is intensifying, it is redundant:

> I very much **so** appreciate it.

Harper did not flag this. This PR adds a `VeryMuchSo` Weir rule that matches `very much so` before a verb and drops the `so`, suggesting `very much`:

- `I very much so appreciate it.` → `I very much appreciate it.`
- `I'm very much so enjoying it.` → `I'm very much enjoying it.`

Legitimate uses are left untouched: sentence-final `very much so`, the `so far` / `so long` idioms, and the conjunction `so` (as in *"... very much so we left early"*). The verb trigger covers finite, progressive, and past-participle forms, which handles most of the reported examples. The noun/adjective-phrase cases (*"very much so a massive advantage"*) are deliberately left for a follow-up to keep this change small and focused.

No related PRs.

# Demo

```
$ harper-cli lint "I very much so appreciate it."
1 │ I very much so appreciate it.
  │         ╰─────── [Redundancy::VeryMuchSo]: Remove the redundant `so`.
```

# How Has This Been Tested?

Rule tests (10 correction cases + 7 negative cases, embedded in the `.weir` file):

```
$ harper-cli test harper-core/src/linting/weir_rules/VeryMuchSo.weir
All tests pass!
```

Rust suite (rule registration + curated-config drift check):

```
$ cargo test -p harper-core --lib curated_default_config_lists_every_registered_rule
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5803 filtered out

$ cargo test -p harper-core --lib run_tests_for_weir_rules
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5803 filtered out
```

Manual checks on the reported sentences confirmed the flagged cases correct and left the elliptical `Yes, very much so.` and the `Thank you very much so far.` idiom untouched.

# AI Disclosure

# Checklist


hippietrail:
Thanks for this. Can you please use our PR template and answer the AI-coding question part. It helps us with reviewing PRs.

ozpool:
Thanks! I've filled in the PR template, including the AI disclosure and linter test-source sections.

elijah-potter (changes requested):
I was able to find a false-positive with this rule:

```
The reports said they were involved, and they were very much so involved.
```

Is there a way to fix that?

hippietrail:
> I was able to find a false-positive with this rule:
> 
> ```
> The reports said they were involved, and they were very much so involved.
> ```
> 
> Is there a way to fix that?

Wow, this is a really interesting find. To me it sounded wrong intuitively, but thinking about it made it seem like it shouldn't be wrong by the rules.

So I did a deep dive into the grammar with an AI, and it turns out "so" here is acting as a "pro-form" (like a pronoun) standing in for the earlier word/phrase.

Because the sentence repeats "involved" at the end, it results in an accidental redundancy ("very much [involved] involved").

I just pulled the PR branch in and ran `harper-cli lint` against the exact sentence to see exactly how it diagnosed it, and the built-in mechanical fix actually handles it correctly:
```
[Redundancy::VeryMuchSo]: Remove the redundant so.
```
Applying this fix changes it to: `"...and they were very much involved."` This is grammatically correct. So while it's an interesting edge case, the current rule's advice is already sound! I think it's just one of those kinds of mistake that was already hard to think about and that we've become so used to hearing that it doesn't always sound wrong.

It's probably best to not just take my word and put the case to whatever AI you think is good at grammar, or see if traditional grammar/usage/style guides can help.

(I did a few rounds of an AI clarifying my explanation and me putting it back into my own words in a better way.)

ozpool:
Nice catch digging into that one. I came at it from the grammar angle and landed in the same place.

The tell is the repeated participle. In the clean proform case ("Are they involved?" / "Very much so."), the "so" stands in for "involved" and nothing follows it, so it is doing real work. But in "they were very much so involved," the participle "involved" is already spelled out right after "so." That leaves "so" pointing at a word that is also present, so you get the doubling: very much [involved] involved. Stripping "so" gives "very much involved," which is exactly what the sentence means.

So this is not a spot where "so" is load bearing. It is the redundancy the rule is meant to flag, just well camouflaged because the ear is used to hearing it. Agreed the current fix is sound, no change needed.

elijah-potter (approved):
LGTM. Thanks!

PR #3830, opened 2026-07-15
fix(core): avoid overall quantifier false positives

### LINKED ISSUE #1249: Rule "Overall" false positive: "Iterate over all"

coyaSONG:
> AI disclosure: This pull request was authored autonomously by OpenAI Codex. I reviewed Harper's contributing and agent policies, issue #1249, the related closed PR, the implementation diff, and the validation output before submission.

# Issues

Fixes #1249

# Description

The `Overall` closed-compound rule previously suggested `overall` for every occurrence of `over all`. In normal phrases such as “iterate over all the save data,” however, `over` is a preposition and `all` is a quantifier.

This change gives `Overall` a small context-aware wrapper and emits the existing `MapPhraseLinter` correction only when a preceding determiner establishes the adjectival use, as in the existing “The over all performance...” positive test. It adds the two issue examples plus quantifier cases covering a plural noun, a pronoun, and sentence-initial usage. Regenerated corpus snapshots remove four existing false positives from the U.S. Constitution, *Alice's Adventures in Wonderland*, and *The Great Gatsby*.

Related: #3455 was an earlier attempt that special-cased words following `over all`; its author closed it. This patch uses grammatical context and covers the additional false positives already present in Harper's corpus snapshots.

# Demo

Not visual.

# How Has This Been Tested?

- `just format`
- `cargo test -p harper-core -- closed_compounds::tests` (40 passed)
- `cargo test -p harper-core --test linters -- test_most_lints` (snapshot regeneration followed by a clean pass)
- `cargo test -p harper-core --lib` (5,526 passed, 284 existing ignored, 0 failed)
- `cargo clippy -p harper-core --lib -- -D warnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes`
- `cargo check -p harper-core`
- `just test-rust` (full Rust workspace, 0 failed)
- `git diff --check`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/closed_compounds.rs
+++ b/harper-core/src/linting/closed_compounds.rs
@@ -1,6 +1,52 @@
-use crate::linting::LintGroup;
+use crate::{
+    Token,
+    expr::Expr,
+    linting::{
+        ExprLinter, Lint, LintGroup,
+        expr_linter::{Chunk, preceded_by_word},
+    },
+};
 
 use super::MapPhraseLinter;
+
+struct Overall {
+    inner: MapPhraseLinter,
+}
+
+impl Default for Overall {
+    fn default() -> Self {
+        Self {
+            inner: MapPhraseLinter::new_closed_compound(["over all"], "overall"),
+        }
+    }
+}
+
+impl ExprLinter for Overall {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        self.inner.expr()
+    }
+
+    fn match_to_lint_with_context(
+        &self,
+        matched_tokens: &[Token],
+        source: &[char],
+        context: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
+        // "Over all" is normally a preposition followed by a quantifier. Only treat it as the
+        // adjective "overall" when a preceding determiner establishes that grammatical role.
+        if !preceded_by_word(context, |word| word.kind.is_determiner()) {
+            return None;
+        }
+
+        self.inner.match_to_lint(matched_tokens, source)
+    }
+
+    fn description(&self) -> &str {
+        self.inner.description()
+    }
+}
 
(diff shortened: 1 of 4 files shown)


harper-core/src/linting/closed_compounds.rs
```diff
@@ -157,6 +204,40 @@ mod tests {
...
+    #[test]
+    fn allow_over_all_plural_noun() {
+        assert_no_lints(
+            "Congress may exercise authority over all places purchased by consent.",
+            lint_group(),
+        );
+    }
+
+    #[test]
+    fn allow_over_all_pronoun() {
```
hippietrail: I would call this `allow_over_all_of_pronoun`.

hippietrail (changes requested):
I've only had a cursory glance at this and the results look good but I haven't analysed the logic yet. I would like to be explicit about the name of one test needing "of", which is very very minor.

coyaSONG:
@hippietrail The requested test rename to `allow_over_all_of_pronoun` was applied in `ab04b4e9`, and I have now resolved the outdated review thread. The current head still has all repository checks passing. When convenient, could you re-review the updated patch so the stale `CHANGES_REQUESTED` state can be refreshed?

PR #3847, opened 2026-07-18
fix: Google Docs bridge and replaceText issue for the popup suggestion

### LINKED ISSUE #2995: Cannot Click to Apply Suggestions in Google Docs

Yuvi-raj-P:
# Issues 

Hopefully should close https://github.com/Automattic/harper/issues/2995

# Description
Problem: When you click a linted word on google docs, the suggestions and the grammar check works however when you go to the popup to replace the error with harper suggestion the error does not get replaced.

Solution:
1) From @abhishekDeshmukh74, initially thought race conditions are what causing the issue. So I went towards the direction of race condition and though applySuggestion is called then immediately followed by close() with no waiting is the problem. I developed a fix which fell short and of course did not work. Later I found that popup closes before the replacing even happens and though that was the race condition because if popover has already closed focus will be moved off the docs iframe which resulted in <body> for activeElement therefore insertText and synthetic paste event did nothing (cherry on top, all errors were silently dying as nothing was being reported). Was still thinking about race conditions so continued to develop a promise chain to await all the way until replacement was done. But it was not the solution. However, the promise chain was left as it's good to have and minimizes any race happening making the process more consistent.

2) With race conditions not the main issue, I later went towards the null issue with the active element. With a decent amount of debugging and Claude, I came to the conclusion that the active element was not the right element we needed for the replacement. This was due to the clicking function of the suggestion button in the popup using the native browser’s button click method. Chrome's default mousedown behavior focuses the clicked button before the onclick handler runs which pulls the page focus off the iframe to popup button (the suggestion). By the time replaceTextRequest reads the activeElement it could not find the right element to dispatch the replacement to as it fell back to <body> the replacement silently died. The execCommand and the paste dispatch however reported success with no errors which led to this whole rabbit hole of a problem from a simple button click. 

td;lr: 
1. Used mousedown.preventDefault() to prevent the default focus on click behavior so the focus stays on Docs iframe instead of the button. 
2. Just to be safe, some hardening with suggestion apply logic so it waits out and makes sure each step has its time before any one of it suddenly stops it in between through a promise chain throughout the whole process.

Please let me know if issues presist, it did work on my instance (Chrome/Apple silicon), i am just too lazy to record a demo and convert it and all that. First pr on the repo :) and please let me know of any silly comments on repo files those are for me.

# How Has This Been Tested?
Build test. Built the Chrome Extension onto chrome and tested it against some text in google docs. 
Since the test was purely done with chrome, I am not sure if this works with, for example, Firefox please someone let me know.

# AI Disclosure

- [] I used LLM features of my editor, but not an agent.

# Checklist

- [ x] I have performed a self-review of my own code


elijah-potter:
This PR did not fix the problem on my system. Is there something I am missing? 

I apologize for the delay in review this. :D

Yuvi-raj-P:
Hi Elijah! It was working on my machine before when I pushed it. I just downloaded and tested it on my machine again today, in fact twice. Sanity check 😭

I am testing it on Google Docs, Google Chrome, and Apple Silicone. Are you using any other browser by any chance? I havent tested it with any other browsers or hardware so yea I might have to check on that :) Below are both my test builds recording seems to work great!

Build 1:
https://github.com/user-attachments/assets/1a89e561-8419-4ccd-9baa-2631a0acb6b1

Build 2:
https://github.com/user-attachments/assets/4dc08937-26f4-4b7a-8edb-e341655799c8

elijah-potter (approved):
@Yuvi-raj-P, forget my previous comment. There was a build cache issue on my system caused by something unrelated. I gave it a try with a clean cache and it seems to be working great! Thank you so much!

Again, I am deeply sorry that it took me so long to get to this, especially since it's such a huge improvement.

Yuvi-raj-P:
Ohh @elijah-potter, You are perfectly fine! Glad I could help with this lovely project.

PR #3850, opened 2026-07-18
feat(core): add zero-width negation expression

### LINKED ISSUE #3848: There seems to be no way for a linter to match when not at the start.
We have the `AnchorStart` `Expr` which matches when at the start of a chunk.

We have the `Invert` `Pattern` which can make any pattern match the opposite of what it normally matches.
We also have `.but_not()` on `ExprSequence`

But try as I might, every variant I could think of for using these in combination resulted in
> the trait bound `AnchorStart: patterns::Pattern` is not satisfied

I even tried getting an agent to convert `Invert` from a `Pattern` to an `Expr` but the problem persisted and the agent just struggled much like I had trying all manner of ways to get it to work. After wasting quite a few tokens on the problem it decided the only way forward was to `#[ignore]` the test for it!

For anyone interested in the use case, I discovered that some people write "give the rise to" when they intend to use the idiom "give rise to", but we can't just always flag "give the rise to" since it has a legitimate use at the start of a sentence as I found during my research:
> Given the rise to power in ...
Where "rise to" is part of the idiom "rise to power" rather than part of "give rise to".

floze-the-genius:
> This patch was produced with assistance from an LLM-based coding agent.

# Issues

Fixes #3848.

# Description

Add `expr::Not`, a zero-width assertion that succeeds only when its inner expression does not match at the current cursor. Unlike `patterns::Invert`, it accepts any `Expr`, including anchors such as `AnchorStart`.

This provides the missing building block for rules that should match a phrase everywhere except the beginning of a chunk.

# Demo

Not applicable; this is a core expression primitive.

# How Has This Been Tested?

- `CARGO_BUILD_JOBS=1 cargo test -p harper-core expr::not -- --nocapture` - 2 passed
- `CARGO_BUILD_JOBS=1 cargo clippy -p harper-core --lib -- -D warnings` - passed
- `cargo fmt --all -- --check` - passed
- `git diff --check` - passed

# AI Disclosure

# Checklist


hippietrail:
I pasted this code into the branch I was working on when I posted #3848 and it does indeed seem to work. Thank you. I'm requesting a review from somebody much more familiar with this aspect of Harper than I am.

floze-the-genius:
Hi @elijah-potter — @hippietrail has already verified the change against #3848 and requested your review. The PR remains current with `master`, mergeable, and all 12 visible checks pass across Rust, JS, web, desktop, browser, Obsidian, VS Code, and WordPress surfaces. When you have a chance, could you take a look or let me know if a focused parser case is still missing?

elijah-potter (approved):

PR #3855, opened 2026-07-21
fix: restrict missing preposition to linking verbs

### LINKED ISSUE #1585: Spuriously flagging to insert a preposition
The correct sentence `Each agent has specific tools and tasks orchestrated through a crew workflow.` is flagged with `You may be missing a preposition here.`

Image

Related source file: `harper-core/src/linting/missing_preposition.rs`

terry-li-hm:
> [!NOTE]
> This pull request was produced with assistance from GLM-5.2 and OpenAI Codex. The resulting patch was independently reviewed and verified before submission.

Fixes #1585.

The `MissingPreposition` pattern currently accepts any auxiliary verb before its adjective and object slots. That makes modal auxiliaries such as "can" and non-copular uses such as "has" look like the rule's intended copular construction.

This change requires that slot to retain its `AUX` tag and also match the existing linking-verb predicate. It adds regression coverage for the three public false positives reported in the issue and a corpus edge case exposed by CI, while preserving the rule's existing true-positive cases.

Validation:

- `cargo fmt --all --check`
- `cargo test -p harper-core --lib`: 5,574 passed and 286 ignored
- `cargo test -p harper-core --test linters test_most_lints`
- Focused CLI checks for the reported examples, nearby grammatical variants, and intended true positives
- All 12 GitHub Actions checks passed

### DIFF
--- a/harper-core/src/linting/missing_preposition.rs
+++ b/harper-core/src/linting/missing_preposition.rs
@@ -6,7 +6,7 @@ use crate::expr::OwnedExprExt;
 use crate::expr::SequenceExpr;
 use crate::patterns::AnyPattern;
 use crate::patterns::UPOSSet;
-use crate::{Token, TokenStringExt};
+use crate::{Token, TokenKind, TokenStringExt};
 
 use super::{ExprLinter, Lint, LintKind};
 use crate::linting::expr_linter::Chunk;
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/missing_preposition.rs
```diff
@@ -146,4 +149,25 @@ mod tests {
...
     }
+
+    #[test]
+    fn allows_issue_1585_public_examples() {
+        // See https://github.com/Automattic/harper/issues/1585
+        assert_no_lints(
+            "Once the installation is done, you can open your project folder.",
+            MissingPreposition::default(),
+        );
+        assert_no_lints(
```
hippietrail, line 160: Harper has always had a strong preference for one assert per test.

hippietrail (changes requested):
Just one little nit about testing asserts.

PR #3876, opened 2026-07-24
fix(core): lex all Unicode `Zs` separators as spaces, not just `U+0020`

### LINKED ISSUE #3863: Non-breaking space (U+00A0) breaks all multi-token lints; tokenizer only treats ASCII space as whitespace
## Summary

`lex_spaces` in `harper-core/src/lexing/mod.rs` only matches the literal ASCII space `' '` (U+0020). A non-breaking space (U+00A0) is never lexed as a `Space` token, so any lint requiring cross-word context (repeated words, a/an agreement, that/than, Oxford comma, unit expansion, etc.) silently produces nothing, while single-word lints (spelling, typos) still fire.

This hits the Chrome extension hard on Gmail: Gmail's compose contenteditable stores inter-word spaces as U+00A0, so Harper appears to "half work" there, spelling lints show, grammar lints never do. Any rich editor that uses nbsp is affected the same way.

## Reproduction (engine only, no browser needed)

Node 22, `harper.js` 2.4.0 from npm:

```js
import { LocalLinter } from 'harper.js';
import { binaryInlined as binary } from 'harper.js/binaryInlined';

const linter = new LocalLinter({ binary });
const text = `There are some cases where the the standard grammar checkers don't cut it. That;s where Harper comes in handy.`;

for (const [name, t] of [['U+0020', text], ['U+00A0', text.replaceAll(' ', '\u00A0')]]) {
	const lints = await linter.lint(t, { language: 'plaintext' });
	console.log(name, lints.map(l => l.lint_kind_pretty()));
}
```

Output:

```
U+0020 [ 'Repetition', 'Typo' ]
U+00A0 [ 'Typo' ]
```

The `Repetition` lint on "the the" disappears when the separator is U+00A0. The same pattern kills `an language` (agreement), `faster that` (that/than), `10 ms` (unit expansion), and Oxford comma detection.

## Reproduction (Chrome extension on Gmail)

1. Install the Chrome extension (v2.6.0) in any Chromium browser, Gmail is on the default-enabled domain list.
2. Open a Gmail compose window and paste the sample text from the writewithharper.com demo.
3. The demo editor reports 13 problems. In Gmail only the ~6 single-word spelling/typo lints render; "the the", "an language", "emails" (Oxford comma), "not online", "ms", "faster that" are all missing.
4. Confirm the cause in DevTools on the Gmail tab:
   `(document.querySelector('div[aria-label="Message Body"]').textContent.match(/\u00A0/g) || []).length` returns > 0.
5. Replacing the U+00A0 characters with regular spaces in the compose DOM makes all 13 lints appear.

## Root cause

`lex_spaces` counts only `' '`:

```rust
fn lex_spaces(source: &[char]) -> Option<FoundToken> {
    let count = source.iter().take_while(|c| **c == ' ').count();
    ...
}
```

`\u{00A0}` appears nowhere in harper-core, so nbsp falls through and severs the token stream at that point.

## Suggested fix

Either treat U+00A0 (and ideally the other Unicode space separators: U+202F, U+2007, U+2009, etc.) as a `Space` token in the lexer, or normalize them during text extraction in the browser extension / lint-framework. The lexer fix benefits every integration at once.

Environment: Linux, Chromium-based browser, Harper Chrome extension 2.6.0, harper.js 2.4.0, repo at current master.

---

NOTE:  This has been tested, and I've *manually* confirmed this is the issue by double-checking the work myself. However, this issue was originally identified using Claude Fable 5 - high thinking.

EDIT: Please note this is also happening in WhatsApp and by the looks of it any website that encodes characters this way.

bounce12340:
*Disclaimer: this PR was prepared with the help of an LLM agent (Claude). The fix, tests, and Unicode-category boundary were reviewed and verified by me before submission, per Harper's [agent-PR policy](https://github.com/Automattic/harper/blob/master/AGENT_POLICY.md).*

## What & why

`lex_spaces` (`harper-core/src/lexing/mod.rs`) only matched the ASCII space `U+0020`, so a non-breaking space (`U+00A0`) — and every other Unicode Space_Separator (Zs) character — was never lexed as a `Space` token. The token stream is severed at that point, so **any lint needing cross-word context silently produces nothing** (repeated words, a/an agreement, that/than, Oxford comma, unit expansion, …), while single-word lints (spelling, typos) still fire.

This is what makes Harper look like it "half works" in rich editors: Gmail's compose and WhatsApp store inter-word spaces as `U+00A0`, so grammar lints never appear there. Fixes #3863 (which has an engine-only repro — no browser needed).

## The fix

- Add `CharExt::is_space_separator()` covering the full **Space_Separator (Zs)** general category (17 code points: `U+0020`, `U+00A0`, `U+1680`, `U+2000`–`U+200A`, `U+202F`, `U+205F`, `U+3000`), as an explicit `matches!` — no new crate.
- `lex_spaces` uses it instead of `**c == ' '`. Space count stays per-character (unchanged for `U+0020`).
- **Deliberately not `char::is_whitespace()`**: that also matches `\t`/`\n`/`\r` (Cc) and `U+2028`/`U+2029` (Zl/Zp). Tab and newline have their own tokens (`lex_tabs`, `lex_newlines`), and `lex_spaces` sits between them in lexer order — swallowing them would break the token stream. The Zs boundary keeps those out.

## Tests

Six lexer tests in `lexing/mod.rs`: nbsp between words yields `[Word, Space(1), Word]` (fails on `master`), each representative Zs char, consecutive-count, plus reverse-guards that `\n` still lexes as `Newline`, `\t` still as tab-width `Space`, and `U+2028` is *not* treated as a space.

The POS snapshot `Difficult sentences.md` updates by exactly two lines: two `U+2003` EM SPACEs that were mislabelled `Unlintable` are now correctly spaces (ignored by the POS tagger). No other tags change.

## Testing done

`cargo test -p harper-core` → 5722 passed, 0 failed. `cargo fmt` clean.


harper-core/src/char_ext.rs
```diff
@@ -98,6 +103,29 @@ impl CharExt for char {
...
+                | '\u{2007}'
+                | '\u{2008}'
+                | '\u{2009}'
+                | '\u{200A}'
+                | '\u{202F}'
+                | '\u{205F}'
+                | '\u{3000}'
+        )
+    }
+
```
elijah-potter, line 110: I would like to see this list restricted to only the two character variants you describe as strictly necessary.

harper-core/src/lexing/mod.rs
```diff
@@ -362,9 +362,75 @@ mod tests {
...
+        let tab = lex_english_token(&source[first.next_index..]);
+
+        assert_eq!(tab.token, TokenKind::Space(2));
+    }
+
+    #[test]
+    fn does_not_lex_line_separator_as_space() {
+        assert_eq!(lex_spaces(&['\u{2028}']), None);
+    }
+
```
elijah-potter, line 392: Please only include a single test with the example sentence from the original issue.

elijah-potter (changes requested):

bounce12340:
Both narrowed in `ca9fca01`.

**Character list** — now just `U+0020` and `U+00A0`:

```rust
fn is_space_separator(&self) -> bool {
    matches!(*self, '\u{0020}' | '\u{00A0}')
}
```

The doc comment no longer claims to cover the Zs category, since it doesn't.

**Test** — the six lexer tests are replaced by one built from the sentence in #3863: it lexes that sentence with `U+0020` and again with every space swapped for `U+00A0`, and asserts the two token streams are identical. That covers the regression, word splitting, and consecutive spaces in a single assertion rather than as separate cases.

**Side effect worth flagging**: dropping the rest of Zs also reverted the two POS snapshot lines in `Difficult sentences.md`. Those had only moved because `U+2003` was being reclassified from `Unlintable` to a space — out of scope now, so `harper-core/tests/text/` is untouched by this PR again.

`cargo test -p harper-core` → 5675 passed, 0 failed, 0 warnings. `cargo fmt --check` clean.

One consequence I want to state plainly rather than bury: the narrower list means the other Zs characters still sever the token stream — `U+202F` (narrow nbsp) and `U+2007` (figure space) are the ones most likely to show up in real text, and multi-token lints will stay silent around them. That's fine as a deliberate scope call for this PR; happy to open a follow-up if you'd rather widen it later, and equally happy to leave it.

harper-core/tests/text/tagged/Difficult sentences.md
```diff
@@ -508,8 +508,8 @@
 # D   NPr/J/R/P NSg/VB+ . Nᴹ/Vg/J  NSg/VB+ .
 > You    can't get    round      the headland when    the tide's in        .
 # ISgPl+ VXB   NSg/VB NSg/VB/J/P D   NSg      NSg/I/C D   NSg$   NPr/J/R/P .
-> in        by    descent  ; in        by    purchase ; in        of the seisin of her     husband
-# NPr/J/R/P NSg/P N🅪Sg/VB+ . NPr/J/R/P NSg/P NSg/VB+  . NPr/J/R/P P  D   ?      P  ISg/D$+ NSg/VB+
+> in        by    descent  ;            in        by    purchase ;            in        of the seisin of her     husband
+# NPr/J/R/P NSg/P N🅪Sg/VB+ . Unlintable NPr/J/R/P NSg/P NSg/VB+  . Unlintable NPr/J/R/P P  D   ?      P  ISg/D$+ NSg/VB+
```
hippietrail, line 512: Does this indicate a non-breaking space getting mistokenized or is something else happening?
Oh I see both of these are Code point `U+2003` "EM SPACE" - Before this PR it was rendered as `.` in the "tagged" file here.

Um not sure which would be more desirable?

bounce12340: One thing that makes this diff confusing: `ca9fca01` is a revert commit, so the `-`/`+` sides here are inverted relative to "master vs PR" — the `+` side (with `Unlintable`) is what `master` already has today, and the `-` side was an intermediate state from the first commit. Net: the PR as it now stands doesn't touch this file at all, and U+2003 tokenizes exactly as on `master`.

Sequence: the first commit treated the entire Zs category as spaces, which turned these two U+2003 EM SPACEs from `Unlintable` tokens into plain spaces (so their entries dropped out of the tag line). The review asked to restrict the change to the strictly-necessary pair (`U+0020`/`U+00A0`), and this commit put the snapshot back to match.

To the question itself — neither state is a mistokenization; they're two deliberate behaviors:

- **`master`, and this PR as it now stands:** U+2003 falls through to `lex_catch()` and lexes as `TokenKind::Unlintable` (in the tag line, the `.` belongs to the semicolon; the EM SPACE's entry is the `Unlintable` beside it). Like the nbsp bug this PR fixes, that severs the token stream, so multi-token lints are silent across an EM SPACE.
- **The first commit's (now dropped) behavior:** U+2003 becomes a `Space` token, so it disappears from the tag line and cross-word lints work over it.

On which is more desirable: space treatment matches the Unicode category and restores linting continuity — and since `Token` keeps its `span`, a dedicated lint could still flag unusual space characters later if that's ever wanted (no such lint exists today). The cost is a wider blast radius, which is why it was cut from this PR per the review. If the verdict is that U+2003 (or the wider Zs set) should be treated as a space too, happy to open a follow-up — this PR stays nbsp-only either way.

harper-core/src/char_ext.rs
```diff
@@ -15,6 +15,12 @@ pub trait CharExt: private::Sealed {
     fn is_english_lingual(&self) -> bool;
     fn is_emoji(&self) -> bool;
     fn is_punctuation(&self) -> bool;
+    /// Whether the character separates words the way an ordinary space does:
+    /// `U+0020` and the non-breaking space `U+00A0`.
+    ///
+    /// Deliberately excludes `\t`, `\n`, and `\r`, which have their own tokens
+    /// (`lex_tabs`, `lex_newlines`) and would break the token stream if swallowed here.
```
elijah-potter, line 22: This doc-comment should be removed.

elijah-potter (changes requested):
One final nit.

PR #3879, opened 2026-07-24
fix: coerce fn items to fn pointers in then_kind_any arrays (fixes build on rustc 1.93)

christinegranger:
harper-core fails to build on rustc 1.93 (stable, 2026-02-11) with three `E0308` errors:

```
error[E0308]: mismatched types
  --> harper-core/src/linting/for_free_of_charge.rs:15:64
   = note: expected reference `&'static [for<'a> fn(&'a TokenKind) -> bool {TokenKind::is_sentence_terminator}]`
              found reference `&[for<'a> fn(&'a TokenKind) -> bool; 3]`
```

## Cause

`then_kind_any<F>(self, preds_is: &'static [F])` infers `F` from the first array element, which is that function's unique *fn-item* type. The array literal, meanwhile, LUB-coerces all of its elements to the *fn-pointer* type `fn(&TokenKind) -> bool`. On recent rustc those two no longer unify, so the call fails to typecheck.

## Fix

Annotate the first element of each affected array, which pins `F` to the fn-pointer type before coercion happens. No behavioural change, three call sites:

- `linting/for_free_of_charge.rs` (2)
- `linting/in_demand_in_depth.rs` (1)
- `linting/naked_eye.rs` (1)

## Verification

`cargo build` succeeds on rustc 1.93.1 with the patch applied; it fails without it on `master` (efa59c3), and on published 2.3.0/2.4.0/2.5.0 as well.

### DIFF
--- a/harper-core/src/linting/for_free_of_charge.rs
+++ b/harper-core/src/linting/for_free_of_charge.rs
@@ -13,13 +13,13 @@ impl Default for ForFreeOfCharge {
         Self {
             expr: SequenceExpr::word_seq(&["for", "free", "of", "charge"]).then_any_of(vec![
                 Box::new(SequenceExpr::default().then_kind_any(&[
-                    TokenKind::is_sentence_terminator,
+                    TokenKind::is_sentence_terminator as fn(&TokenKind) -> bool,
                     TokenKind::is_comma,
                     TokenKind::is_quote,
                 ])),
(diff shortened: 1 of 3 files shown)


elijah-potter:
I do not believe this change is necessary. It seems the fix was incorporated into another PR at some point. Thanks anyway!

PR #3883, opened 2026-07-25
fix(core): avoid to-too false positives outside adjective contexts

### LINKED ISSUE #3856: False Positive: to → too
**What got flagged?**

> Convert a book to audio.

"_Word Choice_: Use too here to mean ‘also’ or an excessive degree."

**Why is this incorrect?**

"too" would not make sense in this sentence

**Example of correct usage:**
This should not be flagged.

hydraxman:
**AI disclosure:** This pull request was prepared by an AI agent and reviewed against the repository policy, focused tests, the full `harper-core` suite, and the complete diff.

# Issues

Fixes #3856

# Description

The `ToTwoToo` adjective-terminal paths treated every dictionary word with adjective metadata as an adjective in context. That made correct prepositional phrases such as “to audio” look like misspellings of “too audio.”

Require the contextual part-of-speech tag to be `ADJ` before either adjective-terminal path emits a suggestion. This keeps existing true positives such as “to cold” while removing the reported false positive and two matching real-text false positives in the Great Gatsby corpus: “next to nothing” and “from nothing to nothing.”

# Demo

Not applicable; this changes grammar-engine output and is covered by the issue sentence plus corpus snapshots.

# How Has This Been Tested?

- Added the issue sentence as a regression and observed it fail on current `master` before the production change.
- `cargo test linting::to_two_too::tests` — 61 passed.
- `cargo test -p harper-core` — 5,675 passed, 289 ignored; integration, snapshot, POS-tag, run-tests, and 6 doc tests also passed.
- `just format` — passed.
- `git diff --check` — passed.

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/to_two_too/mod.rs
+++ b/harper-core/src/linting/to_two_too/mod.rs
@@ -101,6 +101,11 @@ mod tests {
     }
 
     #[test]
+    fn no_lint_to_audio() {
+        assert_no_lints("Convert a book to audio.", ToTwoToo::default());
+    }
+
+    #[test]
     fn fixes_too_go() {
         assert_suggestion_result(
             "I want too go abroad.",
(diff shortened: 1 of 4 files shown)


elijah-potter (approved):
I see no problem here. Good stuff. Thank you!

PR #3889, opened 2026-07-26
fix: don't read the pronoun `one` as a count before a participle (#3819)

### LINKED ISSUE #3819: False positive: "one" as a pronoun before mass noun
**What got flagged?**
> That way, we can keep the rankings and the questions separated without worrying about the scaling of one affecting the other.

**Why is this incorrect?**
In this sentence, "one" is used as a pronoun, rather than a count. As such, "affecting" being a mass noun is irrelevant to its usage.

**Example of correct usage:**
The above sentence should not be flagged.

lukiod:
# Issues

Fixes #3819

# Description

`NounCountability` matches a count quantifier followed by a mass-noun-only word. In the reported sentence:

> That way, we can keep the rankings and the questions separated without worrying about the scaling of **one affecting** the other.

`affecting` carries mass noun metadata and `one` is in the quantifier set, so the rule fired and suggested `one piece of affecting`.

But `one` here is a pronoun, the object of `of`, with `affecting the other` a participial phrase modifying it. There is no count quantifier before a mass noun, which is the only shape this rule targets. That is what the issue title says as well.

The guard therefore skips when the matched quantifier is `one`, it is preceded by `of`, and the mass noun is an `-ing` form.

**This is narrower than my first attempt, and the reason is worth recording.** @hippietrail pointed out in review that a gerund can head a noun phrase and still take an object, so an `-ing` word having an object tells you nothing about whether it is nominal. That is correct, and testing turned up two concrete cases where wider rules silence real errors:

| Wider rule | Sentence it wrongly silences |
| --- | --- |
| "a determiner after an `-ing` form makes it verbal" | `Every training the company offers is mandatory.` — the determiner opens a **relative clause** modifying `training`, not an object of it |
| "`one` after `of` is a pronoun" | `The cost of one advice.` — `of one` also occurs with a genuine quantifier |

Both are real errors that `master` catches. Requiring **both** conditions keeps the change to the construction actually reported.

For a linter, trading a false positive for missed errors is the worse deal, so I checked for that specifically rather than only checking that the reported sentence stopped firing.

# Demo

Not applicable, this is a `harper-core` rule change with no UI surface.

# How Has This Been Tested?

`cargo test -p harper-core` — 5684 passed, 0 failed. `cargo fmt --check` clean, `cargo clippy` introduces no new warnings.

The two false-negative cases were confirmed by disabling **only** the new guard and re-running: with it disabled both are caught, with the first version of the guard enabled both were missed. That is how they were found rather than assumed.

New tests:

- the sentence from the issue
- `Some people do not like eating peas.` — the gerund-with-object example from review
- `Every training the company offers is mandatory.` and `Every funding the government provides is audited.` — relative clause, must still flag
- `They questioned the cost of one advice.` — `of one` with a real quantifier, must still flag

# AI Disclosure

The issue sentence and the `eating peas` example come from the bug report and from review respectively. The relative clause and `of one` sentences are constructed, because they exist to pin specific failure modes; I have not verified them against a corpus, and I would treat that as a real limitation of this PR given the point raised in review about invented test cases.

# Checklist

### DIFF
--- a/harper-core/src/linting/mass_nouns/noun_countability.rs
+++ b/harper-core/src/linting/mass_nouns/noun_countability.rs
@@ -65,7 +65,8 @@ impl ExprLinter for NounCountability {
                 t.kind.is_noun() || t.kind.is_oov() || t.get_ch(src).eq_str("and")
             })
             || (is_ing_word(&toks[2], src)
-                && (followed_by_nominal_head(ctx, src) || is_ing_word_taking_an_object(ctx)))
+                && (followed_by_nominal_head(ctx, src)
+                    || is_pronoun_one_before_participle(toks, src, ctx)))
         {
             return None;
         }
@@ -199,25 +200,50 @@ fn followed_by_nominal_head(ctx: Option<(&[Token], &[Token])>, src: &[char]) -> 
         || follows_slash_separated_ing_modifiers_then_noun(after, src)
 }
 
-/// Whether an `-ing` word is taking a direct object, which makes it verbal.
+/// Whether the matched quantifier is the pronoun `one` followed by a participle.
 ///
-/// `follows_modifiers_then_noun` walks a chain of modifiers to a head noun, so
-/// it stops at a determiner: in `much programming knowledge` the `-ing` word
-/// modifies the head. That is the wrong shape for `affecting the other`, where
-/// the determiner opens the participle's own object rather than ending a noun
-/// phrase. Both are verbal uses, but only the first was recognised, so the
-/// second was read as a mass noun and produced suggestions like
-/// `one piece of affecting`.
+/// Issue #3819 reports `the scaling of one affecting the other`. Here `one` is a
+/// pronoun standing for a thing, `of` takes it as its object, and `affecting the
+/// other` is a participial phrase modifying it. Nothing in that is a count
+/// quantifier before a mass noun, which is the only shape this rule targets.
 ///
-/// A determiner immediately after an `-ing` form is a strong signal on its own:
-/// a real mass noun does not take an object, so `much writing the letter` is not
-/// a noun phrase whichever way it is read.
-fn is_ing_word_taking_an_object(ctx: Option<(&[Token], &[Token])>) -> bool {
-    let Some((_, [ws, word, ..])) = ctx else {
+/// Deliberately narrow, and worth saying why. Two wider rules were tried and
+/// both silence real errors:
+///
+/// * "a determiner after an `-ing` form means the form is verbal" is wrong,
+///   because the determiner may open a relative clause modifying the `-ing`
+///   word instead of an object. `Every training the company offers is mandatory`
+///   is a genuine error, and that rule hides it.
+/// * "`one` after `of` is a pronoun" is wrong, because `of one` also occurs with
+///   a real quantifier. `The cost of one advice` is a genuine error, and that
+///   rule hides it too.
+///
+/// Requiring both the `of` and the participle keeps the fix to the construction
+/// actually reported. A gerund can still head a noun phrase and take an object,
+/// as in `Some people do not like eating peas`, so this says nothing about
+/// `-ing` words in general; it only declines to read `one` as a count.
+fn is_pronoun_one_before_participle(
+    toks: &[Token],
+    src: &[char],
+    ctx: Option<(&[Token], &[Token])>,
+) -> bool {
+    if !toks[0].span.get_content(src).eq_str("one") {
+        return false;
+    }
+
+    let Some((before, _)) = ctx else {
         return false;
     };
 
-    ws.kind.is_whitespace() && word.kind.is_determiner()
+    let mut idx = before.len();
+    while idx > 0 && before[idx - 1].kind.is_whitespace() {
+        idx -= 1;
+    }
+    if idx == 0 {
+        return false;
+    }
+
+    before[idx - 1].span.get_content(src).eq_str("of")
 }
 
 fn follows_modifiers_then_noun(tokens: &[Token], src: &[char]) -> bool {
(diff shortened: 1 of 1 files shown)


hippietrail:
Please use the provided PR template.
I'm assuming this is done by an AI agent. The PR template provides for disclosing the essentials of this.

All -ing forms of verbs can act as both present participles and gerunds. Gerunds behave as mass nouns by default (though may also have countable senses) and can still take objects:

> Some people do not like **eating** peas.

Here "eating" is both a noun, the object of the verb "to like" and a verb, with its own object, "peas".

This kind of thing makes it very tricky to parse the context around a potential mistake to determine if it's really a mistake or actually a legitimate usage. We have to be aware of when it's not a simple "either/or" such as "it can't be a noun because it's a verb" and have to see if there's a way to determine whether it's acting as a noun or a verb etc. in this particular context. Which can require analysing the next word along, which may also have some kind of ambiguity.

Coding AIs tend to ignore this unless you spell it out for them and even then you might have to find concrete examples to convince them, and even then they make mistakes about how whitespace and punctuation are tokenized.

We don't ban AI and humans make many of these same kinds of assumptions, but we should know what we're up against when we ask an AI to write a grammar checker rule.

lukiod:
Thanks, both points taken, and the second one turned out to be right in a way I could reproduce.

I've filled in the PR template, including the AI disclosure. Apologies for missing it.

On the grammar: you're correct that all `-ing` forms can be gerunds, that gerunds behave as mass nouns by default, and that they can still take objects. `Some people do not like eating peas` is exactly the case my reasoning glossed over, and I've added it as a test.

Rather than argue the point I went looking for a sentence where my rule actually breaks, and there is one:

> Every training the company offers is mandatory.

Here `the` opens a **relative clause** modifying `training`, not an object of it. That's a genuine error the rule should flag, and my first patch silenced it. Confirmed by disabling only that guard and re-running: with the guard removed it's caught, with it in place it's missed. So the patch traded one false positive for a class of false negatives, which for a linter is the worse trade.

Going back to your issue title, the real problem is narrower than what I fixed. In `the scaling of one affecting the other`, `one` is a **pronoun**, the object of `of`. Nothing about `-ing` words needed to be involved.

I also checked the obvious narrower rule, "`one` after `of` is a pronoun", and that's wrong too, because `of one` occurs with a genuine quantifier, as in `The cost of one advice`. So the guard now requires both the `of` and the participle.

On invented test sentences, that's a fair caution and I'd flag it against my own PR: the issue sentence and `eating peas` come from the report and from you, but the two relative-clause sentences and the `of one` one are constructed. They pin specific failure modes rather than demonstrate real-world usage, and I haven't checked them against a corpus. If you'd rather see corpus examples for those before this goes in, that seems reasonable to me.

`cargo test -p harper-core` is green at 5684 passed.

lukiod:
> Please use the provided PR template. I'm assuming this is done by an AI agent. The PR template provides for disclosing the essentials of this.
> 
> All -ing forms of verbs can act as both present participles and gerunds. Gerunds behave as mass nouns by default (though may also have countable senses) and can still take objects:
> 
> > Some people do not like **eating** peas.
> 
> Here "eating" is both a noun, the object of the verb "to like" and a verb, with its own object, "peas".
> 
> This kind of thing makes it very tricky to parse the context around a potential mistake to determine if it's really a mistake or actually a legitimate usage. We have to be aware of when it's not a simple "either/or" such as "it can't be a noun because it's a verb" and have to see if there's a way to determine whether it's acting as a noun or a verb etc. in this particular context. Which can require analysing the next word along, which may also have some kind of ambiguity.
> 
> Coding AIs tend to ignore this unless you spell it out for them and even then you might have to find concrete examples to convince them, and even then they make mistakes about how whitespace and punctuation are tokenized.
> 
> We don't ban AI and humans make many of these same kinds of assumptions, but we should know what we're up against when we ask an AI to write a grammar checker rule.

 yah sry about that my ai went a way to abroad but yan need things for a big project so i thought of keeping things automated so as i can focus on the main issue  but yah from now on will  take care of things before creating a pr mb man

lukiod:
@hippietrail brother can i expect a review or anything on this

harper-core/src/linting/mass_nouns/noun_countability.rs
```diff
@@ -203,6 +200,52 @@ fn followed_by_nominal_head(ctx: Option<(&[Token], &[Token])>, src: &[char]) ->
...
+///   word instead of an object. `Every training the company offers is mandatory`
+///   is a genuine error, and that rule hides it.
+/// * "`one` after `of` is a pronoun" is wrong, because `of one` also occurs with
+///   a real quantifier. `The cost of one advice` is a genuine error, and that
+///   rule hides it too.
+///
+/// Requiring both the `of` and the participle keeps the fix to the construction
+/// actually reported. A gerund can still head a noun phrase and take an object,
+/// as in `Some people do not like eating peas`, so this says nothing about
+/// `-ing` words in general; it only declines to read `one` as a count.
```
elijah-potter, line 224: This essay of a comment isn't necessary. Please condense things down if you can.

elijah-potter (changes requested):
The tests seem good, as well as the code itself. Please adjust that one comment. Once that's done, it'll be ready to merge!

harper-core/src/linting/mass_nouns/noun_countability.rs
lukiod: Condensed. The doc comment is down from seventeen lines to five, and the test comments to one line each. What stayed is the part that stops someone widening the guard later: the two rules that were tried and the two real errors they hid (`Every training the company offers is mandatory`, `The cost of one advice`).

elijah-potter (approved):
Thanks!

Issue #3874, opened 2026-07-23
Memory Leak


PuzzleHeaded555:
**Describe the bug**
If I leave it open, overnight I come into that screen on my mac that says I'm out of memory. It's usually says harper is using around 41GB of memory.

**To Reproduce**
1. Let it run overnight or for a few days

**Platform**
Macbook pro M1 16GB,  desktop app

DeepanshKhurana:
Precisely this is what happened to me on an Apple M2 Macbook Pro running `Tahoe 26.5.2 (25F84)`. At some point, Harper crashed my entire system and reported taking near 100% of the available memory, forcing me to shut down all the other apps running. It could be that this is a Mac-specific issue.

paulovitin:
I hit the exact same issue on **Harper Desktop v2.6.0** (macOS 26.5.2 / Apple Silicon M1 Pro / 16GB RAM). The system ran out of application memory after `harper-desktop` grew to over 80GB of swap memory before being forced-quit by macOS.

Here are the diagnostic logs and callstack details retrieved from macOS DiagnosticReports:

### Force Quit Log Record
- **File:** `~/Library/Application Support/CrashReporter/harper-desktop_CF09907B-0EE9-509F-A029-5A7AFFF90B22.plist`
- **Recorded Event:** `ForceQuitDate => 2026-07-26 22:25:54 +0000`

### Diagnostic Callstack Analysis
From `/Library/Logs/DiagnosticReports/signpost_reporter_*.cpu_resource.diag`:
- **Executable:** `/Applications/Harper.app/Contents/MacOS/harper-desktop` (Bundle ID: `com.elijahpotter.harper-desktop`)
- **Active Threads & Allocation Hotspots:**
```text
harper_desktop_lib::highlighter::window::Window::render
  -> egui::context::Context::begin_pass
     -> hashbrown::set::HashSet::extend
  -> egui_wgpu::winit::Painter::paint_and_update_textures
     -> wgpu_core::device::queue::Queue::submit
     -> bit_vec::BitVec::grow
     -> alloc::raw_vec::RawVecInner::reserve::do_reserve_and_handle
  -> harper_desktop_lib::mac_broker::MacBroker::cursor_position
     -> SLEventSourceCreate / SkyLight event polling
```

### Key Observations
[...]

Issue #3714, opened 2026-06-25
Did Harper Desktop cause me to run out of memory?


hippietrail:
<img width="397" height="231" alt="Image" src="https://github.com/user-attachments/assets/0b1aaa2a-ab50-4a7f-939f-aad5f5bd76ce" />

I think I only killed Devin (Windsurf's new name) before taking this screenshot. I have 16GB, most you could order on an OG M1 MacBook Air and I do have situations like this once a week or so without running Harper Desktop, so it's not certain. But 21.3GB is ***a lot***, it's double what Chrome is using, and I have 46 tabs open across 8 windows.

I just upgraded to the newly released 2.6.0 and tested Desktop this morning and left it running. My laptop was fine during breakfast but when I reached the café 10/15 minutes later it was so bogged down that I almost couldn't sign in, and when I did manage it, I was confronted with the out-of-RAM dialog.

Harper is my prime suspect, but not the only one.

hippietrail:
I have a theory. I'm not sure if it always happens or only sometimes, but when quitting Harper Desktop, only the UI process ends. The highlighter process remains behind. This detached highlighter process might end up in some kind of busy loop? Sometimes when I switch between apps using Alt-Tab, when I get to Activity Monitor, a Harper process is over 100% but then goes back to a reasonable rate.

PuzzleHeaded555:
This is happening to me multiple times since installing Harper on my mac. Might be also happening on the browser also.

<img width="1273" height="348" alt="Image" src="https://github.com/user-attachments/assets/7b10e636-d926-4f25-9fd3-aef88e3d1dac" />

lacuna-exe:
I've also run into this issue on Version 2.6.0 (2.6.0) with MacOS 26.3 on an M1 MacBook Pro with 16 GB of RAM 

<img width="760" height="396" alt="Image" src="https://github.com/user-attachments/assets/0354077f-d3e6-426d-adc3-28ca5b0b1a02" />

diploidal:
I've just experienced the same behavior.

I opened Harper, and after 1–2 hours, I closed it and went for a walk. Then, after 2–3 hours, when I came back, Harper was using over 10 GB of RAM.

That said, it looks like it is happening only with the desktop app. The browser extensions are not causing this behavior. 😃

I did the same, tried to close the app, and icon disappeared from the menu bar, yet in Activity Monitor, the process was still running for few hours. 🤔

### Platform
MacBook Pro, M3 Max, 64 GB

PR #3891, opened 2026-07-26
fix(desktop): only repaint the highlighter overlay when something changes

paulovitin:
> **Disclaimer:** this patch was written by an LLM agent (Claude Opus 5), driven and reviewed by me. Per [Harper's policy on agent PRs](https://writewithharper.com/docs/contributors/introduction), I'm saying so up front.

Addresses #3874 (I'm one of the reporters there).

## The problem

The highlighter overlay never stops painting. `WindowManagerApp::about_to_wait` reads accessibility state every `read_interval` — the fastest monitor's refresh interval, so ~16 ms — and `read_rect_updates` then calls `request_redraw()` on every overlay window unconditionally, whether or not the read produced different rectangles.

So for as long as Harper is running — including all night on a machine nobody is touching — the process runs a full egui pass, tessellates, uploads buffers and submits a wgpu queue every 16 ms per monitor. That is the loop the diagnostics on #3874 sampled:

```
harper_desktop_lib::highlighter::window::Window::render
  -> egui::context::Context::begin_pass -> hashbrown::set::HashSet::extend
  -> egui_wgpu::winit::Painter::paint_and_update_textures
     -> wgpu_core::device::queue::Queue::submit -> bit_vec::BitVec::grow
```

## The change

Redraws become demand-driven:

- `RenderState::set_rects` now reports whether the new accessibility read would actually paint a different frame (geometry, lint, rule name or source text changed, a highlight was added or removed, or a popup was invalidated). `read_rect_updates` only requests a redraw when it did.
- `Window::render` returns the repaint delay egui asked for (`FullOutput::viewport_output[…].repaint_delay`), and `WindowManagerApp` keeps the earliest pending deadline in a small `RepaintSchedule`. `Duration::MAX` — egui's "nothing is animating" — schedules nothing at all. Anything shorter wakes the windows up on time, so hover feedback, tooltips, popups closing themselves and any egui animation still get the frames they need.

Nothing changes about how often the accessibility tree is read or how quickly a moved highlight follows its text; only frames that would be pixel-identical to the previous one are skipped. An idle overlay now submits zero GPU frames instead of ~60 per second per monitor.

## Honesty about scope

I want to be clear about what this does and does not prove. I did not bisect the allocation to a single leaking object — that needs a multi-hour soak on a real Metal surface. What I can say is that this removes the unbounded per-frame renderer work that the reported stacks point at and that runs even when the machine is completely idle, which is exactly the "leave it open overnight" scenario in the issue. If growth persists after this, the next thing to look at is the renderer setup itself.

Related and deliberately **not** in this PR, to keep it reviewable:

- `Window::new` builds one `egui_wgpu::winit::Painter` — and therefore one wgpu instance/device — *per monitor*, while egui-wgpu's contract is a single painter shared across viewports. Any per-submission resource cost is multiplied per display today.
- `Window::render` drops `FullOutput::viewport_output` on the floor apart from the repaint delay this PR reads. There are no child viewports in this UI, so nothing appears to accumulate, but it is not the documented backend contract.

Happy to follow up on either if you'd like them addressed.

## Tests

11 new unit tests, no GPU or window required (`cargo test -p harper-desktop --lib`, 63 passing):

- `render_state`: identical reads and empty reads request no repaint; moved rectangles, changed lints, edited source text, added/removed highlights, and a popup invalidated by a shrinking read all do.
- `window_manager`: `RepaintSchedule` ignores `Duration::MAX`, keeps the soonest of competing requests, fires a due repaint exactly once, and treats `Duration::ZERO` as immediately due.

### DIFF
--- a/harper-desktop/src-tauri/src/highlighter/render_state.rs
+++ b/harper-desktop/src-tauri/src/highlighter/render_state.rs
@@ -91,13 +91,24 @@ impl RenderState {
             add_to_dictionary,
             disable_rule,
         };
-        state.set_rects(rects);
+        let _ = state.set_rects(rects);
         state
     }
 
     /// Replaces the accessibility-derived lint geometry while preventing a popup from pointing at a
     /// lint index that no longer exists after the latest read.
-    pub fn set_rects(&mut self, rects: Vec<ActionableLint>) {
+    ///
+    /// Returns whether the next frame would differ from the last one. The overlay reads
+    /// accessibility state dozens of times per second, and the vast majority of those reads produce
+    /// exactly the same rectangles; repainting anyway means the process submits a GPU frame forever,
+    /// even while the machine sits idle overnight.
+    pub fn set_rects(&mut self, rects: Vec<ActionableLint>) -> bool {
+        let changed = rects.len() != self.rects.len()
+            || rects
+                .iter()
+                .zip(&self.rects)
+                .any(|(new, old)| !new.renders_same_as(old));
+
         self.rects = rects;
 
         if self
@@ -105,7 +116,10 @@ impl RenderState {
             .is_some_and(|index| index >= self.rects.len())
         {
             self.highlighted_lint = None;
+            return true;
         }
+
+        changed
     }
 
     /// Updates which lint owns the suggestion popup without exposing render-state internals.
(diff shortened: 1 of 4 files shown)


paulovitin:
@hippietrail any updates about this? It's impossible to use it as is

hippietrail:
> @hippietrail any updates about this? It's impossible to use it as is

Sorry I don't know. This is probably @elijah-potter 's domain and he is usually scarce on weekends.

TyceHerrman:
for tracking purposes, this pr likely addresses the highlighter memory/CPU growth in #3714, but not the part of that issue where the highlighter process remains alive after quitting app

PR #3894, opened 2026-07-27
fix(harper-ls): keep fileified paths relative on Windows

### LINKED ISSUE #3667: Improper handling of file paths when creating a file-local document in the workspace
## Authors Note

So that I do not have to repeat myself too much when referring to the term "workspace" it also refers to  #the current working directory from where `harper-ls` or `harper-cli` was called.

The root of the path refers to `#:\` on Windows or if applicable `/` on Unix.   
I have not tested this issue on a Linux machine yet.

## Describe the bug

<sup>A clear and concise description of what the bug is.</sup>

When the configuration for harper-ls points the file-local dictionary path or ignore file to a sub-directory inside the workspace.

And when creating a file-local dictionary or harper ignore file the created file is created in the same directory that would be referred to as the workspace.

The file-local dictionary or ignore file is created with the full path except for the root of the path and each directory separator character is changed to be a `%` character.

Which is not respecting any of the previously specified configuration options.

The full path should not be used if the file-local dictionary path or ignore file path is in the current workspace as the absolute path varies from environment to environment.

## To Reproduce

1. Go to a file with a harper-ls "problem".
2. Right-click the underlined area and click "Ignore harper issue" or "Add to file-local dictionary"
3. Note the newly created file in the workspace despite the configuration setting.

## Expected behavior

<sup>A clear and concise description of what you expected to happen.</sup>

The configuration for file-local dictionaries and ignore files should be respected from the workspace.

As well as the file-local dictionaries and ignore files should not contain the full path, but preferably the path from the current workspace.

## Screenshots

<sup>If applicable, add screenshots to help explain your problem.</sup>

### Code Block in `./.vscode/settings.json`
```jsonc
{
  /*...*/
  "harper.workspaceDictPath": "./dictionaries/workspace/workspace.dic",
  "harper.fileDictPath": "./dictionaries/files",
  "harper.ignoredLintsPath": "./dictionaries/ignores",
  /*...*/
}
```

<div id="image-label">A screenshot from Visual Studio Code showing a newly created harper ignore file for a local file which is in the workspace root despite the configuration in <a href="#code-block-in-vscodesettingsjson">the above code block</a>.</div>

A screenshot from Visual Studio Code showing a newly created harper ignore file for a local file which is in the workspace root despite the configuration in the above code block.

## Platform

<sup>What platform has the issue? Is it in Obsidian, Chrome or Firefox, Visual Studio Code, the Desktop app, Neovim, or something else?</sup>

Tested with Visual Studio Code and a plugin called LSP4J in JetBrains' IDEs. Which the latter just uses `harper-ls` via calling it, so it's more than likely an issue with the language server and not any specific IDE.

## Additional context

<sup>Add any other context about the problem here.</sup>

I am using Windows, I haven't tried it with a Unix machine (Linux or MacOS) so I am not 100% sure if it affects any of those.

bounce12340:
> **Disclaimer per [`AGENT_POLICY.md`](https://github.com/Automattic/harper/blob/master/AGENT_POLICY.md):** this patch was produced with the help of an LLM (Claude). The diagnosis, the test cases and the mutation check described below were verified by running them locally; the reasoning about Unix behaviour is stated explicitly so it can be checked rather than taken on trust.

Fixes #3667.

## The bug

`fileify_path` skips `Component::RootDir` but not `Component::Prefix`, so on Windows a path flattens to `C:%Users%x%proj%README.md%` — a string that still carries a drive prefix.

That matters because `PathBuf::join` **discards the base entirely** when the joined path has a prefix:

```rust
Path::new(r"C:\config\harper-ls").join(r"C:%Users%x%a.md%")
// -> "C:%Users%x%a.md%"   (the base is gone)
```

So the file-local dictionary and ignored-lints files land in the current directory of the C: drive instead of under the configured directory, which is why `fileDictPath` / `ignoredLintsPath` look like they are being ignored. The language-server docs state these paths are "always resolved relative to the root of the workspace", so this is a contract violation rather than a preference.

## The fix

Prefixes are **sanitized into one leading segment** rather than dropped:

- the result is relative again, so `join` keeps the base;
- `C:\proj\a.md` and `D:\proj\a.md` still map to different names — dropping the prefix would have collided them onto one dictionary;
- `\\server\share\proj\a.md` becomes `servershare%proj%a.md%`.

The segment loop itself is unchanged, so **Unix output is byte-identical**. `:` and `\` are legal in Unix filenames (macOS Finder's `/` is `:` at the POSIX layer), so they are still preserved verbatim — `preserves_colons_and_backslashes_in_unix_segments` pins that.

An earlier draft of this patch sanitized *every* segment, not just the prefix. That was wrong twice over: it changed existing Unix output, and it was not injective — `a:b.md` and `ab.md` would have collapsed onto the same name and silently shared a dictionary. That approach was dropped.

`fileify_file_path(&Path)` is factored out of `fileify_path(&Uri)` so the Windows behaviour is reachable from tests without constructing a URI.

## Verification

- `cargo test -p harper-ls` — 15 passed.
- `cargo fmt --check` — clean.
- Mutation check: reverting the prefix handling to the old drop-it behaviour turns `fileifies_windows_paths` red; restoring it goes green again.

## Known limitation

`fileifies_windows_paths` is behind `#[cfg(windows)]`. `Component::Prefix` does not exist on Unix, so the cross-drive case cannot be exercised on the Linux CI runners — that test only runs for contributors on Windows. I could not find a way around this that did not amount to testing a mock instead of the real path logic.

## Out of scope

Issue #3667 also floats the idea of switching to workspace-relative naming for these files. That is a larger design change with migration implications for existing dictionaries, so this PR only fixes the escaping bug and leaves the naming scheme alone.

### DIFF
--- a/harper-ls/src/io_utils.rs
+++ b/harper-ls/src/io_utils.rs
@@ -1,24 +1,104 @@
 use anyhow::anyhow;
-use std::path::{Component, PathBuf};
+use std::path::{Component, Path, PathBuf};
 
 use tower_lsp_server::{UriExt, lsp_types::Uri};
 
 /// Rewrites a path to a filename using the same conventions as
 /// [Neovim's undo-files](https://neovim.io/doc/user/options.html#'undodir').
+/// Windows path prefixes are sanitized into one segment, so the result is always
+/// relative and can be safely joined to the config directory.
 pub fn fileify_path(uri: &Uri) -> anyhow::Result<PathBuf> {
+    // We assume all URLs are local files and have a base.
+    let path = uri
+        .to_file_path()
+        .ok_or_else(|| anyhow!("Unable to convert URI to file path."))?;
+    Ok(fileify_file_path(&path).into())
+}
+
+fn fileify_file_path(path: &Path) -> String {
+    fileify_path_segments(path.components().filter_map(|seg| match seg {
+        Component::RootDir => None,
+        Component::Prefix(p) => Some(sanitize_windows_prefix(&p.as_os_str().to_string_lossy())),
+        other => Some(other.as_os_str().to_string_lossy().into_owned()),
+    }))
+}
+
+fn fileify_path_segments<I, S>(segments: I) -> String
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<str>,
+{
     let mut rewritten = String::new();
 
-    // We assume all URLs are local files and have a base.
-    for seg in uri
-        .to_file_path()
-        .ok_or_else(|| anyhow!("Unable to convert URI to file path."))?
-        .components()
-    {
-        if !matches!(seg, Component::RootDir) {
-            rewritten.push_str(&seg.as_os_str().to_string_lossy());
-            rewritten.push('%');
-        }
+    for segment in segments {
+        let segment = segment.as_ref();
+
+        rewritten.push_str(segment);
+        rewritten.push('%');
     }
 
-    Ok(rewritten.into())
+    rewritten
+}
+
+/// Windows path prefixes (`C:`, `\\server\share`, `\\?\C:`) contain characters that are
+/// illegal in a filename, which made the joined config path escape its base (#3667).
+/// Strip them so the flattened name stays relative while drives/shares stay distinct.
+/// Unreachable on Unix, where `Component::Prefix` never occurs.
+fn sanitize_windows_prefix(prefix: &str) -> String {
+    prefix
+        .chars()
+        .filter(|c| !matches!(c, '\\' | '/' | ':' | '*' | '?' | '"' | '<' | '>' | '|'))
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


hippietrail:
@elijah-potter

bounce12340:
Note on the one red check: `just test-chrome-plugin` fails here on `iframe_dynamic_textarea.spec.ts` ("attaches to a dynamically revealed iframe textarea on focus"), 173 passed / 1 failed. It looks unrelated to this PR, which only touches Rust in `harper-ls`.

That test passed on a later run in this repo — [#3876's run](https://github.com/Automattic/harper/actions/runs/30320218126/job/90154346714) (40m27s, same job, same spec). The two failures here were on runs that started within 19 seconds of each other, which fits a timing-sensitive focus test losing a race under runner load rather than anything in the diff.

I can't re-run the job myself (`gh run rerun` needs admin on the repo), so the red mark will stay until something else triggers a fresh run. Flagging it so it doesn't read as a real failure.

PR #3895, opened 2026-07-27
fix(harper-ls): assign statsPath to stats_path, not file_dict_path

bounce12340:
> **Disclaimer per [`AGENT_POLICY.md`](https://github.com/Automattic/harper/blob/master/AGENT_POLICY.md):** this patch was produced with the help of an LLM (Claude). The bug and the test were verified locally.

## The bug

In `Config::from_lsp_config`, the `statsPath` branch assigns to the wrong field:

```rust
if let Some(v) = value.get("statsPath") {
    if let Value::String(path) = v {
        base.file_dict_path = path.try_resolve_in(workspace_root)?.to_path_buf();
        //   ^^^^^^^^^^^^^^ should be stats_path
    } else {
        bail!("fileDict path must be a string.");
        //     ^^^^^^^^ message names the wrong setting too
    }
}
```

Two consequences for anyone who sets `statsPath`:

- `stats_path` silently keeps its default, so the setting appears to do nothing;
- `file_dict_path` is redirected to the stats location, so file-local dictionaries are written somewhere the user never asked for — and a `fileDictPath` set earlier in the same config is overwritten, because this branch runs after it.

The error message for a non-string value also names `fileDict`, which sends anyone debugging it to the wrong setting.

## The fix

Assign to `stats_path` and correct the message. Two lines, plus a test covering both the field assignment and the rejection of a non-string value.

`cargo test -p harper-ls` — 13 passed. `cargo fmt --check` — clean. Mutation check: restoring the `file_dict_path` assignment turns `stats_path_does_not_overwrite_file_dict_path` red.

Kept separate from #3667 (the Windows path-prefix fix in `io_utils.rs`) as they are unrelated, per the guidance in `AGENT_POLICY.md` about small PRs.

### DIFF
--- a/harper-ls/src/config.rs
+++ b/harper-ls/src/config.rs
@@ -147,9 +147,9 @@ impl Config {
 
         if let Some(v) = value.get("statsPath") {
             if let Value::String(path) = v {
-                base.file_dict_path = path.try_resolve_in(workspace_root)?.to_path_buf();
+                base.stats_path = path.try_resolve_in(workspace_root)?.to_path_buf();
             } else {
-                bail!("fileDict path must be a string.");
+                bail!("statsPath must be a string.");
             }
         }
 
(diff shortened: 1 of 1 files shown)


hippietrail:
@elijah-potter

bounce12340:
Note on the one red check: `just test-chrome-plugin` fails here on `iframe_dynamic_textarea.spec.ts` ("attaches to a dynamically revealed iframe textarea on focus"), 173 passed / 1 failed. It looks unrelated to this PR, which only touches Rust in `harper-ls`.

That test passed on a later run in this repo — [#3876's run](https://github.com/Automattic/harper/actions/runs/30320218126/job/90154346714) (40m27s, same job, same spec). The two failures here were on runs that started within 19 seconds of each other, which fits a timing-sensitive focus test losing a race under runner load rather than anything in the diff.

I can't re-run the job myself (`gh run rerun` needs admin on the repo), so the red mark will stay until something else triggers a fresh run. Flagging it so it doesn't read as a real failure.

PR #3915, opened 2026-07-29
Obsidian sidebar

LukeBenhase:
# Issues 
none

# Description
I added a sidebar for Obsidian. The sidebar will show all of the improvements that Harper finds in the active page. You can open this sidebar with the command line in Obsidian under "Harper: Open Harper spellcheck sidebar". This will open the sidebar showing all of the errors. This sidebar was inspired by how Grammarly will allow the user to see all of their errors in a list, and it makes correcting errors much faster. The sidebar also brings in the 3 words before and after the mistake so that it is easier to see the correct word from the sidebar. A picture is below. 

# Demo

sidebar

# How Has This Been Tested?
I tested my changes on my Linux computer by compiling them into main.js and running it in Obsidian. It runs good and I have a screenshot above. 

# AI Disclosure
I am human and used AI outside of my editor and did not use any agents. I used Google Gemini to understand some parts of Obsidian. 

I did not change the linter. 

# Checklist

- [X] I have considered splitting this into smaller pull requests.

### DIFF
--- a/packages/obsidian-plugin/src/SidebarView.ts
+++ b/packages/obsidian-plugin/src/SidebarView.ts
@@ -2,15 +2,13 @@ import { type IconName, ItemView, Menu, setIcon, type WorkspaceLeaf } from 'obsi
 import type HarperPlugin from './index';
 import { LINT_KIND_COLORS } from './lintKindColor';
 
-export const Harper_Sidebar_View = 'harper-sidebar-view';
-
 export class SidebarView extends ItemView {
 	constructor(leaf: WorkspaceLeaf, _plugin: HarperPlugin) {
 		super(leaf);
 	}
 
 	getViewType() {
-		return Harper_Sidebar_View;
+		return 'harper-sidebar-view';
 	}
 
 	getDisplayText() {
@@ -18,7 +16,7 @@ export class SidebarView extends ItemView {
 	}
 
 	getIcon(): IconName {
-		return 'book-open-check';
+		return 'harper-logo';
 	}
 
 	async onOpen() {
@@ -37,17 +35,8 @@ export class SidebarView extends ItemView {
 		const listContainer = document.createElement('div');
 
 		// initial card
-		const initialCard = document.createElement('div');
-		initialCard.style.display = 'flex';
-		initialCard.style.flexDirection = 'column';
-		initialCard.style.padding = '12px 16px';
-		initialCard.style.borderBottom = '1px solid var(--background-modifier-border)'; // add var
-		initialCard.style.gap = '8px';
-
-		const initialTitle = document.createElement('span');
-		initialTitle.style.fontWeight = 'bold';
-		initialTitle.textContent = 'Loading errors...';
-
+		const initialCard = createCard();
+		const initialTitle = createTitle('Loading errors...');
 		initialCard.appendChild(initialTitle);
 
 		listContainer.appendChild(initialCard);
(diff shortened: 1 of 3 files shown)


elijah-potter:
Oooh! I like this a lot! 

I'm going to do a quick review of the code, compile this on my own machine, and come back with comments (if I have any).

packages/obsidian-plugin/src/SidebarView.ts
```diff
@@ -0,0 +1,242 @@
+import { type IconName, ItemView, Menu, setIcon, type WorkspaceLeaf } from 'obsidian';
+import type HarperPlugin from './index';
+import { LINT_KIND_COLORS } from './lintKindColor';
+
+export const Harper_Sidebar_View = 'harper-sidebar-view';
```
elijah-potter: The convention in this project is to simply inline variables like these. We've found it makes things easier.

packages/obsidian-plugin/src/SidebarView.ts
```diff
@@ -0,0 +1,242 @@
...
+					title.style.fontWeight = 'bold';
+					title.textContent = 'No grammar errors found!';
+
+					card.appendChild(title);
+
+					listContainer.appendChild(card);
+					return;
+				}
+
+				errors.forEach((error) => {
```
elijah-potter, line 62: I think it would help readability to refactor this callback out into a separate function.

packages/obsidian-plugin/src/SidebarView.ts
```diff
@@ -0,0 +1,242 @@
...
+								}
+							}
+						}
+
+						const card = document.createElement('div');
+						card.style.display = 'flex';
+						card.style.flexDirection = 'column';
+						card.style.padding = '12px 16px';
+						card.style.borderBottom = '1px solid var(--background-modifier-border)'; // add var
+						card.style.gap = '8px';
```
elijah-potter: Each one of these chunks, where you create an element and apply styles to it, could be refactored into its own function.

packages/obsidian-plugin/src/index.ts
```diff
@@ -151,6 +191,14 @@ export default class HarperPlugin extends Plugin {
 			},
 		});
 
+		this.addCommand({
+			id: 'harper-toggle-sidebar',
+			name: 'Toggle harper spellcheck sidebar',
```
elijah-potter: Make sure this command name follows Obsidian conventions and "harper" is correctly capitalized.

packages/obsidian-plugin/src/SidebarView.ts
```diff
@@ -0,0 +1,242 @@
...
+	getViewType() {
+		return Harper_Sidebar_View;
+	}
+
+	getDisplayText() {
+		return 'Harper Grammar';
+	}
+
+	getIcon(): IconName {
+		return 'book-open-check';
```
elijah-potter: You should use the available Harper icon for this. I think it would make it a little bit clearer as to what the sidebar is _for_.

packages/obsidian-plugin/src/SidebarView.ts
elijah-potter: By the way, I really like the UI you've designed.

packages/obsidian-plugin/src/lint.ts
```diff
@@ -323,6 +323,9 @@ const lintPlugin = ViewPlugin.fromClass(
 					Promise.all(sources.map((source) => Promise.resolve(source(this.view)))).then(
 						(annotations) => {
 							const all = annotations.reduce((a, b) => a.concat(b));
+							if ((window as any).app) {
+								(window as any).app.workspace.trigger('harper:lint-updated', all, this.view);
+							}
```
elijah-potter, line 328: There is a problem here. This code _only_ gets run when the text in the pane in edited, which means that if one opens the sidebar _after_ they've already written something in the buffer, they need to do another edit before they can see it show up in the sidebar.

To fix this, I believe you can update the toggleSidebar fn to:

```
	private async toggleSidebar() {
		const existingLeaves = this.app.workspace.getLeavesOfType(Harper_Sidebar_View);
		if (existingLeaves.length > 0) {
			existingLeaves.forEach((leaf) => {
				leaf.detach();
			});
		} else {
			await this.activateSidebarView();
			// This fixes it.
      		await this.state.reinitialize();
		}
	}
```

This forces the editor to reevaluate the contents immediately after opening the sidebar.

elijah-potter (changes requested):
This looks great! Definitely in the top 5% of PRs I receive. Thank you.

I have a few nits and bits, but once those are addressed, we can get this merged.

packages/obsidian-plugin/src/lint.ts
LukeBenhase: Thank you so much. That would have taken me a while to figure out. It works perfectly.

packages/obsidian-plugin/src/SidebarView.ts
LukeBenhase: Thanks. I'm not the most creative at UI design, so I tried to keep the style and colours of the pop-up that Harper uses.

elijah-potter (approved):
This looks great! Thanks for making those changes!

PR #3929, opened 2026-07-31
docs(harper.js): add Chrome extension example and docs page

### LINKED ISSUE #3628: Docs: Add an intro for how to use `harper.js` inside a Chrome extension

leepokai:
> [!NOTE]
> Per [Harper's policy on agent PRs](https://github.com/Automattic/harper/blob/master/AGENT_POLICY.md): this PR was authored with an LLM-based agent (Claude Code), human-directed and reviewed. End-to-end verification is described below.

Closes #3628

## What

- New example `packages/harper.js/examples/chrome-extension/`: a minimal Manifest V3 extension whose popup lints a textarea with `harper.js` and shows the bundled version — the "tiny example… opens a popup… and displays the harper-core version" asked for in #3628.
- New docs page `/docs/harperjs/chrome-extension` covering the three MV3 pitfalls that make this "next level complicated":
  1. Extensions may not load remote code, so the CDN approach doesn't apply — `build.js` copies the `harper.js` dist into the extension, and `binaryInlined` avoids a separate fetch for the wasm binary.
  2. The default extension CSP blocks WebAssembly compilation — the manifest opts in with `wasm-unsafe-eval`.
  3. `WorkerLinter` spawns its Web Worker from a blob URL, which the extension CSP blocks — so the example uses `LocalLinter`.
- Sidebar entry under *harper.js* after "CDN" in `packages/web/vite.config.ts`.

## Verification

- Built the example against published `harper.js@2.7.0` and loaded the resulting `dist/` as a real unpacked extension in Chromium (Playwright, `--load-extension`):
  - popup loads with zero page/console errors, i.e. the wasm engine compiles under the manifest CSP
  - the intentionally flawed sample text yields 3 lints (two "Incorrect indefinite article", one proper-noun hint)
  - the footer renders `harper.js 2.7.0`
- `biome check` clean on the new files.

The version question raised in the issue is addressed at the bottom of the docs page: there is no runtime API exposing the engine version, but `harper.js` and `harper-core` are versioned together (both currently 2.7.0), so the package version identifies the engine.

### DIFF
--- a/packages/harper.js/examples/chrome-extension/README.md
+++ b/packages/harper.js/examples/chrome-extension/README.md
@@ -1,0 +1,14 @@
+# Chrome Extension Example
+
+A minimal [Manifest V3](https://developer.chrome.com/docs/extensions/develop/migrate/what-is-mv3) Chrome extension whose popup lints text with `harper.js` and shows the bundled Harper version.
+
+Extensions cannot load remote code, so `build.js` copies the `harper.js` distribution into the extension directory next to the popup's own files. See [the accompanying documentation](https://writewithharper.com/docs/harperjs/chrome-extension) for a walkthrough of the extension-specific pitfalls.
+
+To try it out:
+
+```bash
+pnpm install
+pnpm build
+```
+
+Then open `chrome://extensions`, enable "Developer mode", click "Load unpacked" and select this example's `dist/` directory.
--- a/packages/harper.js/examples/chrome-extension/build.js
+++ b/packages/harper.js/examples/chrome-extension/build.js
@@ -1,0 +1,30 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+// Chrome extensions cannot load remote code, so we assemble a self-contained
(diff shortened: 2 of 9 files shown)


leepokai:
Closing this — I won't be able to follow up on it. Sorry for the noise.

PR #3969, opened 2026-08-06
chore: some `.toml` files had divergent `description`, `edition`, or `version` fields

hippietrail:
# Issues 
N/A

# Description

When setting up [my new `posslq` package](https://github.com/Automattic/harper/pull/3968) I wanted to make sure I set up the fields of its `.toml` file matching the other `.toml`s in the repo.

I wrote a little tool to compare them. (Let me know if you want me to share it or add it to Harper's build tests.)

The tool found five `.toml` files with fields that didn't match most others in the repo.

It only checks for values that don't fit the pattern, it doesn't check for missing keys or extra keys in comparison to the overall average. Let me know if I should add this.

**Let me know if any of these values were specifically not matching the others on purpose for some reason.**

# How Has This Been Tested?

`cargo test`

# AI Disclosure

Well I did use AI help to make the tool that I found these apparent anomalies with, but I changed the `toml` files myself.

# Checklist


elijah-potter (changes requested):
The packages who aren't being published directly or are otherwise considered "devtools" (i.e. fuzz, harper-cli, harper-wasm) should keep the version numbers they were at before. Their versions should not match the published Harper version since we are not enforcing semantic versioning for those packages (given they are considered "internal"). I would revert those changes.

That said, the version number for `harper-desktop` *does* need to be updated. In addition, the `harper-stats` crate does need to be updated to the 2024 edition. Keep those changes.

elijah-potter (approved):
Thanks!

PR #3970, opened 2026-08-07
fix(obsidian-plugin): guard setting tab updates before state is ready

### LINKED ISSUE #3540: Failed to load plugin
With version 1.13.0 of obsidian mobile, specifically android, I now get an error when loading obsidian that says "Failed to load plugin "harper" "

RyanRuiLiu:
> **AI disclosure**: This PR was prepared with the assistance of an AI agent, per AGENT_POLICY.md.

# Issues

Fixes #3540

# Description

Obsidian 1.13 renders a plugin's setting tab immediately when `addSettingTab` is called, before `onLayoutReady` has had a chance to construct `State`. As a result, `HarperSettingTab.updateSettings`, `updateDescriptions`, `updateDefaults` and `updateStructuredConfig` dereference `this.state` before it exists, throwing a `TypeError` and failing to load the plugin entirely.

This PR adds an early return when `this.state` is null in all four methods, mirroring the existing guard already present in `update()`. Once state is ready, behavior is unchanged.

# How Has This Been Tested

Verified locally with Obsidian 1.13.4 (Windows):
- With the patch, the plugin loads without the "failed to load" error and the settings page renders normally.
- Reverting the patch restores the failure (controlled A/B test).

No automated test was added: the failure is a UI timing race that requires a running Obsidian instance to reproduce. Happy to add a test if maintainers suggest a suitable harness.

# AI Disclosure

# Checklist

- [ ] I have added tests to cover my changes (see note above)

### DIFF
--- a/packages/obsidian-plugin/src/HarperSettingTab.ts
+++ b/packages/obsidian-plugin/src/HarperSettingTab.ts
@@ -37,6 +37,7 @@ export class HarperSettingTab extends PluginSettingTab {
 	}
 
 	updateSettings() {
+		if (this.state == null) return;
 		this.settings = undefined;
 		this.state.getSettings().then((v) => {
 			this.settings = v;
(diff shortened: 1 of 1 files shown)


RyanRuiLiu:
Superseded by #4020 — the maintainers already fixed the same race condition with optional chaining in HarperSettingTab, and the fix is included in v2.8.0. Closing this PR as redundant. Thanks for the great project!

PR #3983, opened 2026-08-08
fix: detect verse in competitor comparisons

### LINKED ISSUE #3966: `VerseAsVerb` doesn't flag some cases
I just heard this sentence in a YouTube video:
> Today it's all-wheel drive Civic verse twin turbo Lamborghini.

Harper fails to flag "verse" to be changed "to versus".

hydraxman:
> [!NOTE]
> This pull request was created with the assistance of an LLM.

## Summary

- detect `verse` used as a comparison marker before a proper competitor name
- preserve the existing `verse` verb-form corrections and avoid broad single-word replacement

## Problem

`VerseAsVerb` only recognized fixed phrases such as `verse against` and missed the reported comparison: `Civic verse twin turbo Lamborghini`. This leaves users without the intended `versus` correction.

The added path requires a capitalized competitor before `verse` and a proper noun at the end of the opposing phrase, limiting false positives for ordinary noun uses of `verse`.

## Testing

- `cargo test -p harper-core --lib` — 6,092 passed, 290 ignored
- `cargo clippy -p harper-core --lib -- -D warnings` — passed
- `cargo fmt --check` — passed
- `git diff --check` — passed

Fixes #3966

### DIFF
--- a/harper-core/src/linting/phrase_set_corrections/mod.rs
+++ b/harper-core/src/linting/phrase_set_corrections/mod.rs
@@ -1,4 +1,11 @@
-use crate::linting::LintKind;
+use crate::{
+    CharStringExt, Token, TokenStringExt,
+    expr::{Expr, FixedPhrase, LongestMatchOf, SequenceExpr},
+    linting::{
+        ExprLinter, Lint, LintKind, Suggestion,
+        expr_linter::{Chunk, find_the_only_token_matching},
+    },
+};
 
 use super::{LintGroup, MapPhraseSetLinter};
 
@@ -6,6 +13,92 @@ use super::{LintGroup, MapPhraseSetLinter};
 mod many_to_many_tests;
 #[cfg(test)]
 mod one_to_one_tests;
+
+const VERSE_VERB_CORRECTIONS: &[(&str, &str)] = &[
+    ("verse against", "play against"),
+    ("versed against", "played against"),
+    ("versing against", "playing against"),
+    ("verses against", "plays against"),
+    ("verse me", "play me"),
+    ("verse him", "play him"),
+    ("verse her", "play her"),
+    ("verse them", "play them"),
+    ("verse you", "play you"),
+];
+
+struct VerseAsVerb {
+    expr: LongestMatchOf,
+}
+
+impl Default for VerseAsVerb {
+    fn default() -> Self {
+        let comparison = SequenceExpr::any_word()
+            .t_ws()
+            .t_aco("verse")
+            .t_ws()
+            .then_any_word()
+            .t_ws()
+            .then_any_word()
+            .t_ws()
+            .then_kind_where(|kind| kind.is_proper_noun());
+        let mut expressions: Vec<Box<dyn Expr>> = VERSE_VERB_CORRECTIONS
+            .iter()
+            .map(|(wrong, _)| Box::new(FixedPhrase::from_phrase(wrong)) as Box<dyn Expr>)
+            .collect();
+        expressions.push(Box::new(comparison));
+
+        Self {
+            expr: LongestMatchOf::new(expressions),
+        }
+    }
+}
+
+impl ExprLinter for VerseAsVerb {
+    type Unit = Chunk;
+
+    fn expr(&self) -> &dyn Expr {
+        &self.expr
+    }
+
+    fn match_to_lint(&self, toks: &[Token], src: &[char]) -> Option<Lint> {
+        let span = toks.span()?;
+        let matched = span.get_content(src);
+
+        if let Some((_, correction)) = VERSE_VERB_CORRECTIONS
+            .iter()
+            .find(|(wrong, _)| matched.eq_str(wrong))
+        {
+            return Some(Lint {
+                span,
+                lint_kind: LintKind::Nonstandard,
+                suggestions: vec![Suggestion::replace_with_match_case_str(correction, matched)],
+                message: "`Verse` is not a verb meaning to compete. Use `play against` or `compete against` instead.".to_owned(),
+                priority: 31,
+            });
+        }
+
+        if !toks.first()?.get_ch(src).first()?.is_uppercase() {
+            return None;
+        }
+
+        let verse =
+            find_the_only_token_matching(toks, src, |tok, src| tok.get_ch(src).eq_str("verse"))?;
+        Some(Lint {
+            span: verse.span,
+            lint_kind: LintKind::Nonstandard,
+            suggestions: vec![Suggestion::replace_with_match_case_str(
+                "versus",
+                verse.get_ch(src),
+            )],
+            message: "Use `versus` to compare the two competitors.".to_owned(),
+            priority: 31,
+        })
+    }
+
+    fn description(&self) -> &str {
+        "Corrects nonstandard uses of `verse` derived from `versus`."
+    }
+}
 
 /// Produce a [`LintGroup`] that looks for errors in sets of common phrases.
 pub fn lint_group() -> LintGroup {
@@ -728,22 +821,6 @@ pub fn lint_group() -> LintGroup {
             "Corrects the eggcorn `wreck havoc` to `wreak havoc`, which is the proper term for causing chaos or destruction.",
             LintKind::Eggcorn
         ),
-        "VerseAsVerb" => (
-            &[
-                ("verse against", "play against"),
-                ("versed against", "played against"),
-                ("versing against", "playing against"),
-                ("verses against", "plays against"),
-                ("verse me", "play me"),
-                ("verse him", "play him"),
-                ("verse her", "play her"),
-                ("verse them", "play them"),
-                ("verse you", "play you"),
-            ],
-            "`Verse` is not a verb meaning to compete. Use `play against` or `compete against` instead.",
-            "Corrects the nonstandard use of `verse` as a verb (from `versus`) to standard alternatives.",
-            LintKind::Nonstandard
-        ),
         "WroteToRote" => (
             &[
                 ("by wrote", "by rote"),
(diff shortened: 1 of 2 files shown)


harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -7,6 +14,92 @@ mod many_to_many_tests;
...
+    ("verse me", "play me"),
+    ("verse him", "play him"),
+    ("verse her", "play her"),
+    ("verse them", "play them"),
+    ("verse you", "play you"),
+];
+
+struct VerseAsVerb {
+    expr: LongestMatchOf,
+}
```
elijah-potter: This should be placed in its own module.

elijah-potter (changes requested):

harper-core/src/linting/phrase_set_corrections/mod.rs
hydraxman: Moved `VerseAsVerb`, its correction mappings, and the expression implementation into `phrase_set_corrections/verse_as_verb.rs`. Registration and existing regression tests are unchanged.

Verified `cargo test -p harper-core`, `just test-rust`, `cargo clippy -p harper-core --lib -- -D warnings`, and `just format`. `just check-rust` also passed.

harper-core/src/linting/phrase_set_corrections/verse_as_verb.rs
```diff
@@ -0,0 +1,94 @@
+use crate::{
+    CharStringExt, Token, TokenStringExt,
+    expr::{Expr, FixedPhrase, LongestMatchOf, SequenceExpr},
+    linting::{
+        ExprLinter, Lint, LintKind, Suggestion,
+        expr_linter::{Chunk, find_the_only_token_matching},
+    },
+};
+
+const VERSE_VERB_CORRECTIONS: &[(&str, &str)] = &[
```
hippietrail: If you use the `many_to_many` variant of `phrase_set_corrections` you can offer multiple suggestions per match:
```rs
(&["verse against"], &["play against", "compete against"])
```
```rs
(&["verse me"], &["play me", "play against me", "compete against me"])
```
(Typed off the top of my head, apologies if I didn't nail the syntax.)

harper-core/src/linting/phrase_set_corrections/verse_as_verb.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl Default for VerseAsVerb {
+    fn default() -> Self {
+        let comparison = SequenceExpr::any_word()
+            .t_ws()
+            .t_aco("verse")
+            .t_ws()
+            .then_any_word()
+            .t_ws()
+            .then_any_word()
```
hippietrail: With these `any_word()` elements I'd want to test this against a decent sized corpus of real-world English to hunt for unforeseen edge cases.

harper-core/src/linting/phrase_set_corrections/verse_as_verb.rs
```diff
@@ -0,0 +1,94 @@
...
+                "versus",
+                verse.get_ch(src),
+            )],
+            message: "Use `versus` to compare the two competitors.".to_owned(),
+            priority: 31,
+        })
+    }
+
+    fn description(&self) -> &str {
+        "Corrects nonstandard uses of `verse` derived from `versus`."
```
hippietrail: "Derived from" isn't the best wording here. It imparts a bit too much possibility of being grammatical rather than just a plain old mistake. Perhaps something like "instead of", or "where the preposition `versus` is intended", etc.

harper-core/src/linting/phrase_set_corrections/mod.rs
```diff
@@ -761,6 +747,8 @@ pub fn lint_group() -> LintGroup {
         )
     });
 
+    group.add_chunk_expr_linter("VerseAsVerb", VerseAsVerb::default());
```
hippietrail: This is an interesting approach. Several times I've wanted to have a rule where some subrules were Rust linters and others used `phrase_set_corrections`. 

Actually, there's a problem that wouldn't have passed your testing. You've given both rules the same name. I implemented a test for clashing linter names a few months ago when I found some hidden. Now I can't actually remember if I knew for sure that one blocked the other or if I just assumed that was the case. I definitely don't think I specifically checked if it was supported. I'd like to hear from @elijah-potter on this as he knows the Harper architecture much more thoroughly than I do.

If it turns out that linter names have been chainable all along then I should remove that test. If not, maybe we should consider adding it, but there's a possibility that two linters that should have separate entries in the config might end up merged out of ignorance.

In the meantime, there is a fully supported way to add multiple subrules in a single linter with a single config setting, and that is to use the `merge_linters!()` macro.

I've always wanted to explore whether it's possible to use `merge_linters` to make a linter with a single config setting that incorporates both `phrase_set_corrections` and Rust linters.

Sorry I can't remember what combination of hand-coding vs agent coding you are using. If you're a confident Rust programmer or are using an agent, then seeing if that last idea is possible is probably a worthwhile side question.

harper-core/src/linting/phrase_set_corrections/mod.rs
hippietrail: We don't normally add submobules to `phrase_set_corrections` but until a week or so ago it was monolithic so I can see where you got the idea. I don't think this is what @elijah-potter meant but would love to hear from him. I've also added more notes as a PR review.

harper-core/src/linting/phrase_set_corrections/verse_as_verb.rs
hydraxman: Switched the fixed phrases to `MapPhraseSetLinter::many_to_many`. The original first suggestion is preserved for all nine mappings, with `compete against` alternatives and the additional `play against` forms for pronoun objects. Regression tests cover multiple suggestions, inflections and capitalization.

harper-core/src/linting/phrase_set_corrections/verse_as_verb.rs
hydraxman: Ran the actual rule over 515 documents: the complete NLTK Brown corpus plus 15 Gutenberg books, totaling 1,958,828 words under an alphabetic-word count. Both the previous and final rule returned zero matches on that sample; this is not a general proof against false positives. Targeted checks did expose poetry errors such as `Blank verse was what Shakespeare preferred` and `Free verse poet Walt Whitman was born in 1819`. Added regressions, constrained the intervening modifiers, and preserved the established free/blank-verse noun phrases. The original Civic/Lamborghini example and the legacy corrections still pass. `just test-rust` (including desktop), `just check-rust`, and `just format` passed.

harper-core/src/linting/phrase_set_corrections/verse_as_verb.rs
hydraxman: Reworded the description to: "Corrects nonstandard uses of `verse` where `versus` or a verb meaning to compete is intended."

harper-core/src/linting/phrase_set_corrections/mod.rs
hydraxman: The previous revision had removed the old phrase-table registration, so the single-key regression did not reproduce a name collision. I have nevertheless switched to the supported composition: `merge_linters!` combines a `many_to_many` fixed-phrase linter and the comparison `ExprLinter`, with one `VerseAsVerb` registration in the curated group. The rule and its tests now live in the top-level `linting/verse_as_verb.rs` module. The config regression verifies that the one toggle enables/disables both forms; the full Rust suite, including the name-clash test, passes.

harper-core/src/linting/verse_as_verb.rs
```diff
@@ -0,0 +1,275 @@
...
+                .t_ws()
+                .t_aco("verse")
+                .t_ws()
+                .then_kind_where(|kind| kind.is_adjective() && !kind.is_proper_noun())
+                .t_ws()
+                .then_kind_where(|kind| {
+                    (kind.is_adjective() || kind.is_noun()) && !kind.is_proper_noun()
+                })
+                .t_ws()
+                .then_kind_where(|kind| kind.is_proper_noun()),
```
hippietrail: You can just use `.then_proper_noun()` directly.

harper-core/src/linting/verse_as_verb.rs
```diff
@@ -0,0 +1,275 @@
...
+        let verse =
+            find_the_only_token_matching(toks, src, |tok, src| tok.get_ch(src).eq_str("verse"))?;
+        Some(Lint {
+            span: verse.span,
+            lint_kind: LintKind::Nonstandard,
+            suggestions: vec![Suggestion::replace_with_match_case_str(
+                "versus",
+                verse.get_ch(src),
+            )],
+            message: "Use `versus` to compare the two competitors.".to_owned(),
```
hippietrail: "compare" sounds too specific, covering only a subset of what "versus" is used for. How about something like `"Use 'versus' for matchups and comparisons."`?

hippietrail (changes requested):
This is looking pretty good to me now. Thanks. I didn't read it very deeply though. I do have a couple of small suggestions.

harper-core/src/linting/verse_as_verb.rs
hydraxman: Switched to `.then_proper_noun()`. `just test-rust` (including desktop), `just check-rust`, and `just format` pass on this revision.

harper-core/src/linting/verse_as_verb.rs
hydraxman: Changed the message to "Use `versus` for matchups and comparisons."

PR #3999, opened 2026-08-09
feat: take advantage a→take advantage of

jvoisin:
# Issues 

# Description
Add the `AdvantageOf` Weir rule, which restores the dropped `of` in the idiom `take advantage of` (e.g. "taking advantage a bug"->→ "taking advantage of a bug"), while allowing adverbial uses like "took advantage a lot". Enabled in the curated default config.

# Demo

# How Has This Been Tested?

# AI Disclosure

# Checklist


hippietrail:
I'd never come across this one before but I can confirm it's out there. This one does have false positives. Since you say you're collecting real-world sentences for this and not just making them up or letting your AI tools make them up, they're really easy to find with just Google

- iOS will only allow JIT in a WKWebView that runs in another process so you can **take advantage a little bit** for hybrid apps.
- ... and I am sure that vim users will **take advantage the same way**

Basically, "take advantage" has an optional "of clause". So sometimes the "of" is missing from the "of clause" but sometimes there is no "of clause" but the next part of the sentence just happens to start with "a", "an", or "the".

jvoisin:
The weir rule is handling all the false-positives I could think of

hippietrail:
> The weir rule is handling all the false-positives I could think of

This is exactly why I added the section to the PR template about collecting real-world sentences. Because very often the edge cases are very hard to think of. I always trawl through Google Search results using the `OR` and `site:` operators. Sometimes asking your AI if there are any legitimate situations a sequence of words can be used in without intervening punctuation. But I haven't found a way to automate the hunting Google technique. Maybe Google Gemini could do it.

The good part in this case is that the false positives are very rare.

harper-core/src/linting/weir_rules/AdvantageOf.weir
```diff
@@ -0,0 +1,34 @@
+# Corrects the missing preposition in `take advantage of` when `of` is dropped,
+# e.g. "taking advantage a bug" -> "taking advantage of a bug".
+expr verbs [take, takes, took, taken, taking]
+expr adverbials [lot, couple, few, bit, little, while, number, most, whole, same, first, second, third, entire, other, best, worst, latter, former, moment, time, way]
+expr main <(@verbs advantage [a, an, the] !@adverbials), (advantage [a, an, the]), ( )>
+
+let message "The idiom is `take advantage of`. Did you drop the `of`?"
+let description "Corrects the missing preposition in the idiom `take advantage of` when `of` is dropped."
+let kind "Grammar"
```
hippietrail: I think `Usage` is the more appropriate `LintKind` here, since it's about how to use set phrases correctly.

hippietrail (changes requested):
This is quite close but I'd change the `LintKind`.

elijah-potter (approved):

PR #4007, opened 2026-08-10
feat: add a new `MissingWord` `LintKind`

hippietrail:
# Issues 
N/A

# Description

I was looking at a linter that inserts a missing word and realized that none of the `LintKind`s really fit, usually `Usage` is the closest. But we've now got easily enough linters adding missing words to warrant a new `MissingWord` `LintKind`.

This PR converts all the linters I could identify that add a missing word, whether that be always or only on some logic paths.

I also found an opportunity to make the `LintKind` code a little bit more DRY, and I found a debug-printf that accidentally slipped through.

I'm less that fully fluent in some of the web and desktop code that does things such as grouping `LintKind`s together, so would appreciate people reviewing those parts.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist


hippietrail:
Oh I'd forgotten I'd made this and was about to start making it again. Any chance of a review @elijah-potter ?

PR #4017, opened 2026-08-10
fix(desktop): only update the tray icon when the service state changes

willherr72:
# Issues

# Description
The tray icon is rebuilt and reapplied every 250 ms whether or not anything
changed. Each tick decodes the PNG, builds an `Image`, and calls `set_icon` —
four times a second, forever, to draw the icon it is already showing.

On macOS and Linux this is merely wasted work. On Windows it becomes a real
fault: after several days of uptime the shell started rejecting every update
with `ERROR_TIMEOUT`, which filled the log with one error per tick and left the
tray icon unresponsive. Shell_NotifyIcon is rate-sensitive, and a permanent
4 Hz stream of redundant updates is exactly the shape of traffic it gives up on.

This tracks the state the icon is currently showing and only touches the tray
when it actually changes.

One detail worth noting for review: when `set_icon` fails, the remembered state
is deliberately left unchanged, so the next tick retries rather than assuming
an update landed that did not.

Found while implementing Windows support, but the bug and the fix are
cross-platform — nothing here is `cfg`-gated.

# How Has This Been Tested?
- Reproduced the original failure on Windows: `ERROR_TIMEOUT` from `set_icon`
  repeating once per tick after several days of uptime.
- With the change, `set_icon` is called only on an actual transition. Toggling
  the highlighter service from the tray menu still updates the icon
  immediately, in both directions.
- Full CI green, including Linux and macOS.

# AI Disclosure

# Checklist
- [x] I have added tests to cover my changes — the behavior is a timing loop
      against the OS tray API; I could not find a way to assert it that would
      be more convincing than the manual reproduction above. Happy to add one
      if you have a pattern in mind.


elijah-potter (approved):
This is great! Thanks!

PR #3936, opened 2026-08-01
Added Windows Support

ashar-builds:
# Description

This PR adds Windows support to Harper Desktop, bringing feature parity with the existing macOS highlighter.

**Windows accessibility broker (`win_broker`):**
- Implements `OsBroker` for Windows using Microsoft UI Automation (UIA) to extract text from focused edit and document controls in other applications.
- Uses `IUIAutomationTextPattern` for text extraction and character-level bounding rectangles, with `IUIAutomationValuePattern` as a fallback for simple inputs.
- Uses `GetForegroundWindow` / `GetWindowThreadProcessId` for focused PID tracking, with a 3-second cache to avoid stale state during overlay hover.
- Tracks foreground window movement and delays highlight rendering until the window settles (150 ms), matching the macOS `MacBroker` behaviour.
- Resolves installed application names and icons by scanning Start Menu `.lnk` shortcuts via a PowerShell subprocess.
- Curated default integrations for Windows: Notepad, Notepad++, Slack, Discord, and Application Frame Host.

**Transparent overlay window on Windows:**
- The HWND swap-chain path only advertises `CompositeAlphaMode::Opaque`, causing DWM to paint the overlay as solid black. This PR routes the renderer through the DX12 backend with `Dx12SwapchainKind::DxgiFromVisual`, which exposes `PreMultiplied` alpha and makes the overlay correctly transparent.
- Sets `WS_EX_NOREDIRECTIONBITMAP` (`with_no_redirection_bitmap`), `with_skip_taskbar`, and `with_undecorated_shadow` on the overlay window.

**Getting Started onboarding:**
- Updates `GettingStartedPage.svelte` to be OS-aware: skips the accessibility permission step on Windows (not required), substitutes Notepad for TextEdit as the default test app, and adjusts all copy accordingly.

**CI:**
- Adds windows_desktop.yml to build and check Harper Desktop on `windows-latest` on every push and pull request.

**justfile:**
- Adds the `build-desktop-windows` recipe to build NSIS and MSI installer bundles.

# How Has This Been Tested?

- Built and ran the Tauri app on Windows 11 with `just dev-desktop`.
- Confirmed the transparent overlay renders correctly (no black fill) on a 1080p display and a 4K display at 150% scaling.
- Confirmed lint highlights appear over Notepad and Slack text fields.
- Confirmed the Getting Started flow completes on Windows (accessibility step auto-passes, Notepad integration enables, test drive launches Notepad).
[...]

PR #4022, opened 2026-08-11
fix(desktop): persist onboarding completion

mauropereiira:

# Issues

No existing issue. This was found while switching from Grammarly to Harper Desktop 2.7.0 and completing the macOS onboarding flow.

Related to #3936, which also changes the platform-specific setup steps for Windows.

# Description

The TextEdit test drive is labeled optional, but the setup progress counted all three rows and only showed "You're all set" after TextEdit was launched. That launch state lived only in the Svelte component, so it disappeared when Settings was recreated.

This change:

- Counts only required setup steps in progress and completion.
- Waits for Accessibility and integration state to finish loading before deciding setup is complete.
- Stops treating a successfully spawned TextEdit process as proof that the test drive was completed.
- Adds an explicit Continue action that persists completion and opens General.
- Opens General on later Settings launches while keeping Getting Started in the sidebar.
- Defaults new installs to incomplete onboarding.
- Treats existing configs without the new field as completed, avoiding onboarding every current desktop user after upgrade.
- Makes unsupported Accessibility platforms non-blocking rather than impossible to complete.

# Demo

Before: the optional TextEdit row left progress at `2 of 3`; launching TextEdit temporarily showed All Set.

After: completing the two required steps shows the All Set stage while the TextEdit test remains optional. Continue saves completion and routes to General.

# How Has This Been Tested?

- `cargo clippy -p harper-desktop --all-targets -- -D warnings`
- `cargo test -p harper-desktop config::tests` (13 passed)
- `pnpm check` in `harper-desktop` (0 errors, 0 warnings)
- `just format`
- `just check-desktop` on a combined branch containing all three desktop fixes
- Built an Apple Silicon `.app` bundle with Tauri
- Manually launched the bundle with an isolated config and verified the All Set stage appears without completing the optional TextEdit test

# AI Disclosure

Not applicable.

# Checklist


harper-desktop/src-tauri/src/config/mod.rs
```diff
@@ -239,6 +244,39 @@ mod tests {
...
+            serde_json::from_str::<serde_json::Value>(&config.serialize_main().unwrap()).unwrap();
+        value
+            .as_object_mut()
+            .unwrap()
+            .remove("onboarding_completed");
+
+        let deserialized = Config::deserialize_main(&value.to_string()).unwrap();
+
+        assert!(deserialized.onboarding_completed);
+    }
```
elijah-potter: You can remove this tests, they aren't really needed.

elijah-potter (changes requested):
This looks good to go as soon as that one nit is fixed. Thanks!

harper-desktop/src-tauri/src/config/mod.rs
mauropereiira: I really thought this had already been taken care of, sorry about that. I've removed those tests and pushed the change.

elijah-potter (approved):

PR #4029, opened 2026-08-12
docs(sublime): document setup for linting git commit messages

### LINKED ISSUE #3781: Support sublime-text git message

costajohnt:
# Issues

Closes #3781

Follow-up to #3782, which was closed in favor of documenting the configuration instead of widening the languageId match arm.

# Description

Sublime Text reports `languageId: "git"` for commit buffers, and `harper-ls` routes commit messages by the `git-commit` languageId, so commit messages go unlinted out of the box. There are two things a user has to do, and neither is currently documented:

1. Add `text.git.commit` to the `selector` in the `harper-ls` client config, so LSP attaches to the buffer at all.
2. Map `"text.git.commit": "git-commit"` in `language-ids.sublime-settings`, so the client sends a languageId `harper-ls` recognizes.

This adds a "Linting Git Commit Messages" section to the Sublime Text integration page covering both.

# Demo

Not applicable, documentation only.

# How Has This Been Tested?

Documentation only, no code paths touched. The `git-commit` languageId in the mapping matches the existing match arm in `harper-ls/src/backend.rs`, and the workaround itself is the one the reporter confirmed working in #3781.

# AI Disclosure

Not applicable.

# Checklist


elijah-potter:
Thanks for submitting this. Can you just double-confirm that you tried this in a real Sublime editor? I want to make sure it covers everything.

costajohnt:
Yes, tried it now in Sublime Text 4 (build 4200) with LSP from Package Control and harper-ls 2.9.1 on macOS. Opened a `COMMIT_EDITMSG` with some typos in the subject and body and watched the LSP log panel:

- Without the `language-ids` mapping, `didOpen` goes out with `languageId: "git"` and harper-ls publishes an empty diagnostics list.
- With `"text.git.commit": "git-commit"` in `language-ids.sublime-settings` and `text.git.commit` added to the selector, `didOpen` goes out with `languageId: "git-commit"` and harper flags the subject line and the body (spelling, repeated words).

Nothing else was needed. The Git Formats package that ships with Sublime already gives the buffer the `text.git.commit` scope, so those two settings are the whole setup.

hippietrail:
> Yes, tried it now in Sublime Text 4 (build 4200) with LSP from Package Control and harper-ls 2.9.1 on macOS.

If you could do a little screen recording of it working that would be excellent.

costajohnt:
Here's a recording. Making it turned up a gap in the docs: harper-ls was publishing the diagnostics fine (that's what I'd been checking in the LSP log panel), but LSP for Sublime Text drops diagnostics for any file matching `folder_exclude_patterns`, and `.git` is in the default list, so nothing showed up in the editor for `.git/COMMIT_EDITMSG`. I pushed a third step that removes `.git` from that list for the Git Commit syntax only. With all three in place it looks like this:


https://github.com/user-attachments/assets/5f393b70-45d1-462b-970b-260001ef971a

PR #4030, opened 2026-08-12
fix(core): remove redundant copy of dictionary in `FstDictionary`

### LINKED ISSUE #3725: High dictionary memory footprint: `FstDictionary` materializes the wordlist 2–3× (the FST is ~1.7 MB but resident is ~1.5 GB for large dictionaries)
**Summary.** `FstDictionary` keeps the entire wordlist in memory **two to three times over**. The compact FST is tiny; the resident cost comes almost entirely from the additional fully-materialized copies. This is tolerable for English but scales badly for large / highly-inflected dictionaries, which is the hard blocker for Slavic-language support (cf. #2654) and likely a contributor to the high-RAM reports (#3354, #3299).

**Where.** `harper-core/src/spell/fst_dictionary.rs`:

```rust
pub struct FstDictionary {
    mutable_dict: Arc<MutableDictionary>,        // full word→metadata map (a complete copy)
    word_map: FstMap<Vec<u8>>,                   // the compact FST (word → index)
    words: Vec<(CharString, DictWordMetadata)>,  // every form again as Vec<char> + metadata
}
```

`FstDictionary::new()` builds the FST **and** `extend_words()` into a full `MutableDictionary` **and** keeps the `words` Vec. So each expanded surface form is stored ~2–3 times.

**Measurements** (release builds, Linux; for the *expanded* surface-form set):

| Dictionary | Forms | FST size | Steady-state RSS¹ | Peak RSS² |
|---|---|---|---|---|
| Slovak (hunspell sk_SK, fully expanded) | 2,223,249 | **1.7 MB** | **1.57 GB** | 1.9 GB |
| Belarusian (PR #2195) | ~1.1 M | — | — | ~1.0 GB |

¹ Resident memory of a minimal program that just calls `FstDictionary::curated()` and holds it idle (no linting, no LSP) — RSS goes from 2 MB → **1572 MB**. This is what stays pinned for the lifetime of an always-on LSP.
² `Maximum resident set size` over the same run (includes construction transients): 1906 MB.

- The FST for the same 2.2 M forms is **1.7 MB** (built directly with the `fst` crate); the raw UTF-8 wordlist is 28 MB. So the steady-state footprint is **~920× the FST** and ~56× the raw text.
- The cost is the two extra full copies kept alongside the FST: the `words` Vec and the `mutable_dict` (`MutableDictionary`), plus per-form `CharString` + `DictWordMetadata` overhead.
- It scales ~linearly with the number of **expanded** forms. English has relatively few and is fine today; inflected Slavic languages expand to 1–2 M+ forms, landing at ~1–1.6 GB resident.

**Why it matters.**
- It's the practical blocker for adding Slavic languages (#2654): even a correct Slovak dictionary is unusable as an always-on LSP if it pins ~1–2 GB.
- It plausibly explains part of the high baseline RAM seen with English in #3354 (and the general complaints in #3299) — the dictionary representation is heavy regardless of language; large languages just make it obvious.

**Question / possible direction.** The FST already maps word ↔ index. Could the `words` Vec and/or `mutable_dict` be:
- derived on demand from the FST instead of being kept as separate full copies, and/or
- built lazily only when fuzzy suggestions are actually requested (the comment notes they exist "for fuzzy-finding")?

That would bring the steady-state footprint close to the FST size (single-digit MB even for millions of forms) and unblock large/inflected dictionaries without changing the public API.

Happy to help test any change on the Slovak (2.2 M-form) dataset, which is a good stress case. Motivation: tracking Slovak support in #3658.

elijah-potter:
# Issues 

Fixes #3725

# Description

I've removed a redundant copy of the raw text of the curated dictionary in the `FstDictionary` by relying on the existing `MutableDictionary` for storage.

# How Has This Been Tested?

All existing tests pass. Functionality remains intact but with lower memory usage. I've also benchmarked the code and found minimal perf degradation.

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/spell/fst_dictionary.rs
+++ b/harper-core/src/spell/fst_dictionary.rs
@@ -18,10 +18,8 @@ use crate::{CharString, CharStringExt, DictWordMetadata};
 pub struct FstDictionary {
     /// Underlying [`super::MutableDictionary`] used for everything except fuzzy finding
     mutable_dict: Arc<MutableDictionary>,
-    /// Used for fuzzy-finding the index of words or metadata
+    /// Used for fuzzy-finding the WordId of words or metadata
     word_map: FstMap<Vec<u8>>,
-    /// Used for fuzzy-finding the index of words or metadata
-    words: Vec<(CharString, DictWordMetadata)>,
 }
 
 const EXPECTED_DISTANCE: u8 = 3;
@@ -61,10 +59,10 @@ impl FstDictionary {
         words.dedup_by(|(a, _), (b, _)| a == b);
 
         let mut builder = fst::MapBuilder::memory();
-        for (index, (word, _)) in words.iter().enumerate() {
-            let word = word.iter().collect::<String>();
+        for (word_chars, _) in words.iter() {
+            let word = word_chars.iter().collect::<String>();
             builder
-                .insert(word, index as u64)
+                .insert(word, WordId::from_word_chars(word_chars).into())
                 .expect("Insertion not in lexicographical order!");
         }
 
@@ -77,7 +75,6 @@ impl FstDictionary {
         FstDictionary {
             mutable_dict: Arc::new(mutable_dict),
             word_map,
-            words,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


elijah-potter:
It was pulled from the queue because of GitHub's utterly broken infrastructure.

PR #4031, opened 2026-08-12
fix(core): do not flag Dickens as a swear

### LINKED ISSUE #1656: False positive: "Dickens" flagged as a swear
Image

This doesn't make sense to me as <s>no inflection of "dick" leads to "Dickens"</s>.

```json
% just getmetadata Dickens
cargo run --bin harper-cli -- metadata Dickens
{
  "noun": {
    "is_proper": true,
    "is_singular": null,
    "is_plural": null,
    "is_countable": null,
    "is_mass": null,
    "is_possessive": null
  },
  "pronoun": null,
  "verb": null,
  "adjective": null,
  "adverb": null,
  "conjunction": null,
  "swear": true,
  "dialects": "",
  "orth_info": "LOWERCASE | TITLECASE",
  "determiner": null,
  "preposition": false,
  "common": false,
  "derived_from": {
    "hash": 18316041236339589540
  },
  "np_member": null,
  "pos_tag": null
}
```
`dictionary.dict`:
```
Dick/OXg
Dickens/Og
Dickensian/JN
```
```
dick/~NVg>XZSx
dicker/VNdG
dickey/~NJSg
```

Ah - `/x` marks a swear but what is `/X`? `/X` marks <s>plural nominalizations</s>!
```json
		"X": {
			"#": "'-ions', '-ications', '-ens' suffixes",
			"kind": "suffix",
			"cross_product": true,
			"replacements": [
				{
					"remove": "e",
					"add": "ions",
					"condition": "e"
				},
				{
					"remove": "y",
					"add": "ications",
					"condition": "y"
				},
				{
					"remove": "",
					"add": "ens",
					"condition": "[^ey]"
				}
			],
```
In fact, a bit like `/S` it has two meanings! For words ending in "e" or "y" it means plural nominalization: "commune"→"communion", "communicate"→"communication". BUT for words ending in any other letter it means 3rd person singular present derived verb: "thick"→"thickens".

mauropereiira:
AI disclosure: An AI agent implemented and tested this patch with direction from Mauro Pereira.

# Issues

Fixes #1656.

# Description

The `X` suffix flag on both `Dick` and `dick` generated `Dickens` and `dickens`. Generated forms inherited metadata from their stems, so the lowercase path marked `Dickens` as a swear when dictionary entries were merged.

This removes the `X` flag from those stems and adds `dickens` explicitly. The existing `Dickens` entry remains the proper noun, the lowercase word remains accepted, and `dick` remains marked as a swear.

# Demo

Before this change, `Charles Dickens wrote many novels.` triggered `AvoidCurses`. It now produces no lint. `What the dickens is going on?` also remains accepted.

# How Has This Been Tested?

- `cargo test -p harper-core`
- `cargo test -q`
- `cargo run --bin harper-cli -- audit-dictionary harper-core`
- `cargo fmt -- --check`
- `cargo clippy -- -Dwarnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes`
- `cargo hack check --each-feature`
- `pnpm check`
- `git diff --check`
- Manual CLI checks for both demo sentences
- Two independent code reviews reported no findings

# AI Disclosure

- [ ] I am a human and did not use any AI.
- [ ] I consulted one or more coding AIs, but did not use an agent.

- [ ] I am using examples from the bug report / feature request.

# Checklist


hippietrail:
This looks like the right fix to me. Thanks.

harper-core/dictionary.dict
```diff
@@ -21460,7 +21460,8 @@ dichotomous/J
 dichotomy/~NSg
 dicier/Jc
 diciest/Ju
-dick/~NgSV>XZx
+dick/~NgSV>Zx
```
hippietrail: May as well remove `>` while we're at it. It generates `dicker`, which has its own entry anyway. So we should also remove `x` which apparently generates `dickers` if I'm not mistaken.
But then since the `dicker` entry doesn't generate a plural we should add `/S` to it.
I better look into "dicker" since I'm not sure I know it... OK so "dicker" is a verb so "dickered", "dickering", and "dickers" are all legit. Several dictionaries also list it as a noun, so let's go with
```
dick/~NgSVx
dicker/VdGS
```

hippietrail (changes requested):
We're close. Let's make the dictionary entries nice and clean. Also this is still a draft - is that intentional?

mauropereiira:
Draft was not intentional, sorry about that. Marked ready for review.

Cleaned up the entries as suggested. `dick` loses `>` and `Z`, which is a better fix than I had: since `x` propagates, those two suffixes were also marking `dicker` and `dickers` as swears, so this removes a second false positive I had not spotted.

I did deviate from your snippet on one character and want to flag it rather than slip it past you. You wrote:

```
dicker/VdGS
```

The current entry is `dicker/VdGN`, so that would drop the `N`. Going by your own note that several dictionaries list it as a noun, I do not think you meant to. It matters because `S` is conditional on the base:

```json
"target": [
  { "if_base": { "noun": {} }, "metadata": { "noun": { "is_plural": true } } },
  { "if_base": { "verb": {} }, "metadata": { "verb": { "verb_form": "THIRD_PERSON_SINGULAR" } } }
]
```

Without `N` on the base, `dickers` would come out as a third person singular verb form only, and the plural noun sense would disappear. So I went with `dicker/VdGNS`. Happy to change it to exactly what you wrote if you did mean to drop the noun.

Final entries:

```
dick/~NgSVx
dickens/N
dicker/VdGNS
```

Checked the metadata after rebuilding:

| word | swear | noun | verb |
| --- | --- | --- | --- |
| dick | true | true | true |
| dickens | not set | true | false |
| Dickens | not set | true | false |
| dicker | not set | true | true |
| dickers | not set | true | true |
| dickered | not set | false | true |
| dickering | not set | true | true |

I also checked whether anything else in the dictionary has this shape. Four other entries combine `x` with a generating suffix: `fuck`, `piss`, `shitty` and `wank`. None of them has the same problem, because their generated forms are legitimately swears too, and none carries `X`, which was the suffix that produced `dickens` in the first place. So this stays a two-line change.

`cargo test -p harper-core` passes with 0 failures, and `harper-cli audit-dictionary harper-core` reports nothing.

hippietrail (approved):
This looks ready to go. Thanks!

PR #4055, opened 2026-08-14
feat: flag unidiomatic overuse of "highly"

hippietrail:
# Issues 
N/A

# Description

I keep seeing "highly" used as a general intensifier like "very" in all kinds of places it didn't used to be used. I made a linter for the specific case of "highly kept secret" some months ago. This linter is more general.

It includes a check against words that do commonly collocate with "highly". Any others will be flagged with "very" requested as a replacement. A second list has specific replacements for specific words.

I went back and forth with an AI on which words to include or not, starting with all instances of "highly" in my corpus of READMEs from GitHub.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

I mostly used AI for going over the word lists but I also got it to make my linter code more Rusty once my hand-coded version worked as I wanted it to.

# Checklist


elijah-potter (commented):
My biggest concern is the reliance on a large list of words. This is quite expensive. Is there a way around that?

hippietrail:
> My biggest concern is the reliance on a large list of words. This is quite expensive. Is there a way around that?

Me too. That's why I wanted you to review this one. I think there's going to be more and more like this that rely on lists of context words.

But I can think of two approaches.

1. Increasing the flags in the dictionary. I've put out a PR to add 2-letter annotation flags starting with `+` but we could extend it in further ways such as an escape/prefix flag that introduces named flags of arbitrary length, which would make them less cryptic: #3982
2. Adding a step that adds adds metadata flags using lists of words instead of by annotation flags. This would work best if it could be in the same module along with the linter though.

Both ways would mean a `Pattern`/`Expr` would check for them just like checking for POS properties now, with `.is_xyz()` type methods on `TokenKind`.

PR #4068, opened 2026-08-15
feat(chrome-ext): add site preferences checklist to options page

### LINKED ISSUE #4067: Add site preferences checklist to manage bundled default sites
Harper enables linting on 53 bundled sites by default. Users cannot see which sites have an explicit setting or turn them off in bulk.

**Proposed Solution**
Add a "Site Preferences" section to the options page: a searchable checklist of the bundled defaults with bulk enable/disable and per-site toggles, plus a list of custom sites with explicit settings that can be reset.

**Examples**
User only wants Harper on GitHub: open options, disable all default sites, then check GitHub.

Image

**Component**
- [ ] Core engine
- [x] Plugin/Extension
- [ ] Other: _____

**Additional Context**
Domain preference keys are normalized to 'www.-stripped' hostnames so legacy 'www.' keys cannot shadow newer settings.

draphy:
# Issues
Fixes #4067

# Description
Adds a unified Site Preferences section to the options page so users can manage bundled defaults and custom overrides together. Defaults carry a Default badge and support bulk and per-site toggling via checkbox; custom sites appear in the same list and can be removed per Remove. Bulk Reset All to Defaults restores defaults and removes customs. Domain keys are stored as 'www.-stripped' hostnames so legacy www keys never shadow newer overrides.

# Demo
image

# How Has This Been Tested?
Passes the E2E test covering default bulk toggle, unified list with badges, per-custom Remove and bulk Reset All, and lint gating on Chromium.

# AI Disclosure

# Checklist


packages/chrome-plugin/src/ProtocolClient.ts
```diff
@@ -87,7 +94,6 @@ export default class ProtocolClient {
 	}
 
 	public static async getDomainEnabled(domain: string): Promise<boolean> {
-		this.lintCache.clear();
```
elijah-potter, line 90: Why did you remove this?

elijah-potter (changes requested):
In addition to my other comment below, I have one question: Why not expand this to work for _all_ of the configured domains? If it exists as a set domain, why not show it as a part of the list? I don't see a reason why need need to delineate between the default domains and everything else.

packages/chrome-plugin/src/ProtocolClient.ts
draphy: Moved the flush over to the write path. getDomainEnabled only reads status for the popup toggle, so it was clearing the whole cache on a pure read. setDomainStatus now returns whether the effective state actually changed, and the cache only clears when that flag is true, so invalidating on a real domain change still works.

draphy:
> In addition to my other comment below, I have one question: Why not expand this to work for _all_ of the configured domains? If it exists as a set domain, why not show it as a part of the list? I don't see a reason why need need to delineate between the default domains and everything else.

The feature request was to let people mass-disable the whitelist. That's why the checklist only covered the bundled defaults and the custom sites sat off in their own list. But you're right that the split is really just bookkeeping. Every domain is a set domain, defaults just happen to start enabled.
So I merged them into one list and tagged each row as a bundled default. The bulk "Disable All Default Sites" button targets just the defaults. Custom sites have a Remove button to delete their override. Defaults have no per-site Reset, they use the checkbox toggle. Bulk "Reset All to Defaults" re-enables all defaults and removes all custom overrides.

Does that match what you had in mind?

image

PR #4079, opened 2026-08-16
fix: keep suggestion popup within viewport

### LINKED ISSUE #3972: [BUG] Correction overlay displays outside of browser viewport
**Describe the bug**
The correction overlay displays outside of browser viewport when editing Gmail.

**To Reproduce**
1. Edit mail in Gmail.
2. Make a mistake close to the right side of the screen/viewport.
3. Click the underlined word for the overlay.
4. Check overlay overflow our of the viewport

**Expected behavior**
Overlay detect screen boundary and right align to stay viewable.

**Screenshots**
Image
Image

The vertical bar/grey line is the Gmail input box scroll bar, and it proves that the overlay is way out of the browser viewport.

**Platform**
Firefox

**Additional context**
I can add more info or check duplicate later. But now I just want to quickly report it and finish my email…

escape0707:
> [!NOTE]
> This pull request was developed with AI assistance. The submitter reviewed the implementation and tests before submission.

## Summary

- Conservatively right-align suggestion popups whose configured maximum width would cross the viewport's right margin.
- Keep positioning fully represented in the virtual DOM, including an 8 px margin and the matching transform origin.
- Include padding and borders within the 420 px maximum width.

## Design decisions

I considered measuring the rendered popup and then adjusting the real DOM. That would use the popup's exact width, but `.harper-container` is owned by `virtual-dom`; changing it outside the VNode would allow the cached virtual tree and real DOM to disagree and require extra synchronization.

I also prototyped a two-pass virtual-DOM approach: render a provisional left-aligned VNode, measure its real DOM, then create and patch a right-aligned VNode when necessary. That preserves virtual/real DOM consistency, but it adds another render and VNode construction and requires preserving child and hook identity to avoid unnecessary lifecycle work.

This PR instead uses the existing 420 px maximum width as a conservative pre-render estimate. It stays single-pass and declarative, with no layout measurement. A short popup near the right edge can be right-aligned even if its rendered content would have fit; that is an accepted tradeoff for the smaller implementation and lower runtime overhead.

## Testing

- `just format`
- `pnpm --dir packages/lint-framework build`
- `pnpm --dir packages/chrome-plugin zip-for-firefox`
- Manually loaded the unpacked extension in Firefox and reproduced the Gmail right-edge scenario; both compact and expanded suggestion panels remained fully inside the viewport ([screenshots](https://github.com/Automattic/harper/pull/4079#discussion_r3849520020)).

Fixes #3972

### DIFF
--- a/packages/chrome-plugin/tests/simple_textarea.spec.ts
+++ b/packages/chrome-plugin/tests/simple_textarea.spec.ts
@@ -184,30 +184,3 @@ test('Can dismiss with escape key', async ({ page }) => {
 
 	await assertLocatorIsFocused(page, editor);
 });
-
-test('Right-aligns the suggestion popup near the viewport edge', async ({ page }) => {
-	const viewportWidth = 500;
-	const viewportMargin = 8;
-	await page.setViewportSize({ width: viewportWidth, height: 500 });
-	await page.goto(TEST_PAGE_URL);
-
-	const editor = getTextarea(page);
-	await editor.evaluate((element) => {
-		Object.assign(element.style, {
-			position: 'fixed',
-			right: '0',
-			top: '40px',
-			width: '110px',
-		});
-	});
-	await replaceEditorContent(editor, 'This is an test');
-
-	expect(await clickHarperHighlight(page)).toBe(true);
-	const popup = page.locator('.harper-container');
-	await expect(popup).toBeVisible();
-	await expect(popup).toHaveCSS('right', `${viewportMargin}px`);
-
-	const popupBox = await popup.boundingBox();
-	expect(popupBox).not.toBeNull();
-	expect(popupBox!.x + popupBox!.width).toBeLessThanOrEqual(viewportWidth - viewportMargin);
-});
--- a/packages/chrome-plugin/tests/simple_textarea.spec.ts
+++ b/packages/chrome-plugin/tests/simple_textarea.spec.ts
@@ -184,3 +184,30 @@ test('Can dismiss with escape key', async ({ page }) => {
 
 	await assertLocatorIsFocused(page, editor);
 });
+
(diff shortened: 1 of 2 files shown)


packages/chrome-plugin/tests/simple_textarea.spec.ts
```diff
@@ -184,3 +184,30 @@ test('Can dismiss with escape key', async ({ page }) => {
...
+
+	expect(await clickHarperHighlight(page)).toBe(true);
+	const popup = page.locator('.harper-container');
+	await expect(popup).toBeVisible();
+	await expect(popup).toHaveCSS('right', `${viewportMargin}px`);
+
+	const popupBox = await popup.boundingBox();
+	expect(popupBox).not.toBeNull();
+	expect(popupBox!.x + popupBox!.width).toBeLessThanOrEqual(viewportWidth - viewportMargin);
+});
```
elijah-potter: This test is expensive and it doesn't give me a ton of confidence. 

Would you remove it and test it manually by running the extension in your browser?

elijah-potter (changes requested):
I just have one testing nit. Otherwise, this looks great!

packages/chrome-plugin/tests/simple_textarea.spec.ts
escape0707: Okay, I'll give it a check soon!

escape0707: Sorry for the delay. I have built the extension with this PR and the test removed. Loaded the add-on in firefox and manually tested the fix. And it looks just fine. I'll commit and push soon.

Screenshot 2026-08-25 131455
Screenshot 2026-08-25 131449

Issue #2329, opened 2025-12-12
False positive for Oxford comma for pairs


jvoisin:
**What got flagged?**
> It's A or B, B or C, D or E, or F or G.     ■■ An Oxford comma is necessary here. 

**Why is this incorrect?**
Expressions `A or B`, `B or C`, `D or E` and `F or G` are *self-contained*, and thus roughly equivalent to `W, Y, Y, or Z`.

Issue #1308, opened 2025-05-20
Oxford comma false positive


ubitux:
The sentence "Last year, a friend and I were at the beach." wrongly triggers the Oxford comma error:

```
cargo run --bin harper-cli lint /tmp/oxford.md
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/harper-cli lint /tmp/oxford.md`
/home/ux/.config/harper-ls/dictionary.txt: No such file or directory (os error 2)
/home/ux/.local/share/harper-ls/file_dictionaries/tmp%oxford.md%: No such file or directory (os error 2)
Advice:
   ╭─[oxford.md:1:1]
   │
 1 │ Last year, a friend and I were at the beach.
   │              ───┬──
   │                 ╰──── An Oxford comma is necessary here.
───╯
```

Tested with Harper 0.37.0

hippietrail:
I've been noticing ones like this too. My guess is they'll be quite tricky to fix because a greater understanding of English grammar will be necessary to determine the function of each comma.

hippietrail:
This should stay open.

Issue #659, opened 2025-02-12
bug: Missing Oxford commas are not identified in some contexts


hippietrail:
**Describe the bug**
Some missing Oxford commas are flagged, and others are not. I suspect the underlying grammar parsing.

**To Reproduce**
1. Copy and paste this text into an editor that supports Harper:
  > A dog, a cat, and a pig.
  > A bat, a rat, and a hog.
  > A dog, a cat and a pig.
  > A bat, a rat and a hog.
  > 
  > The dog, the cat, and the pig.
  > The bat, the rat, and the hog.
  > The dog, the cat and the pig.
  > The bat, the rat and the hog.
  > 
  > Dogs, cats, and pigs.
  > Bats, rats, and hogs.
  > Dogs, cats and pigs.
  > Bats, rats and hogs.
  > 
  > One dog, one cat, and one pig.
  > One bat, one rat, and one hog.
  > One dog, one cat and one pig.
  > One bat, one rat and one hog.
2. Notice that each group of four lines has two with Oxford commas then two without, with different nouns.
3. In line 3, see no squiggly under "cat" where there should be one.
4. In line 4, see a squiggly under the "and" instead of "rat".
5. See the same for lines 8 and 9.
6. See that both lines 13 and 14 get it right.
7. Then see the wrong behaviour returns for lines 18 and 19.

**Expected behavior**
Consistently applying squiggles under the cats and rats in every line without Oxford comma.

**Screenshots**
<img width="358" alt="Image" src="https://github.com/user-attachments/assets/20658a46-8594-4479-9cce-f0daa1878f7d" />

**Platform**
VS Code, Windsurf

**Additional context**
More examples: 
<img width="358" alt="Image" src="https://github.com/user-attachments/assets/b383cce6-74d4-41d0-979e-35cfd17b1eae" />
<img width="326" alt="Image" src="https://github.com/user-attachments/assets/f5d98910-1a03-494d-85e7-ac2f305c78ee" />

Cut and paste this text to test:

    This dog, this cat, and this pig.
    This bat, this rat, and this hog.
    This dog, this cat and this pig.
    This bat, this rat and this hog.

    That dog, that cat, and that pig.
    That bat, that rat, and that hog.
    That dog, that cat and that pig.
    That bat, that rat and that hog.

    Some dog, some cat, and some pig.
    Some bat, some rat, and some hog.
    Some dog, some cat and some pig.
    Some bat, some rat and some hog.

    Some dogs, some cats, and some pigs.
[...]

PR #4090, opened 2026-08-16
fix(core): don't flag a comma-bracketed parenthetical as a list

### LINKED ISSUE #2803: False Positive: punctuation error
**What got flagged?**
Any resemblance to real persons, living or dead, is purely coincidental.

**Why is this incorrect?**
This is a correct usage. The solution proposed by Harper will cause a punctuation error.

mauropereiira:
# Issues

Fixes #2803

# Description

`OxfordComma` flagged "Any resemblance to real persons, living or dead, is purely coincidental.", proposing a serial comma after "living". As the reporter noted, applying that suggestion introduces a punctuation error.

"living or dead" is a non-restrictive appositive bracketed by a matched pair of commas, not a two-item list. The expression reads "persons, living or dead" as three nominal items and stops before the closing comma, so nothing told it the phrase was already closed.

The guard looks at what follows the match. A closing comma followed by a finite linking verb marks the end of a parenthetical, because the clause continues afterwards rather than terminating:

```rust
if let [comma, whitespace, verb, ..] = &sentence[match_span.end..]
    && comma.kind.is_comma()
    && whitespace.kind.is_whitespace()
    && verb.kind.is_linking_verb()
    && !verb.kind.is_verb_progressive_form()
    && !verb.kind.is_verb_past_participle_form()
{
    continue;
}
```

Two narrower alternatives were rejected. A blanket closing-comma guard would hide genuine lists followed by comma-delimited material. Using `is_verb()` rather than `is_linking_verb()` would pull in participles and hide genuine lists before participial adjuncts, which is why the progressive and past-participle forms are excluded explicitly.

This is deliberately conservative and does not claim to recognise every parenthetical.

# Demo

```
Any resemblance to real persons, living or dead, is purely coincidental.   clean     (was OxfordComma)
The report, dense or sparse, was useful.                                   clean
The winners, Ann or Bob, is announced tomorrow.                            clean

We invited the cooks, Ann and Bob.                                         flagged   (unchanged)
I bought apples, pears and oranges.                                        flagged   (unchanged)
My favourite colours are red, green and blue.                              flagged   (unchanged)
```

# How Has This Been Tested?

`cargo test -p harper-core`

```
test result: ok. 6130 passed; 0 failed; 290 ignored; 0 measured; 0 filtered out
```

Every existing test in `oxford_comma.rs` was checked against the guard; no positive case has the `comma + finite linking verb` boundary, so all remain eligible. No files under `harper-core/tests/text/` change.

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/oxford_comma.rs
+++ b/harper-core/src/linting/oxford_comma.rs
@@ -74,6 +74,18 @@ impl Linter for OxfordComma {
             let sentence = &sentence[skip..];
 
             for match_span in self.expr.iter_matches(sentence, document.get_source()) {
+                // A match closed by a comma and followed by a finite linking verb is likely
+                // parenthetical, with the second comma separating it from the continuing clause.
+                if let [comma, whitespace, verb, ..] = &sentence[match_span.end..]
+                    && comma.kind.is_comma()
+                    && whitespace.kind.is_whitespace()
+                    && verb.kind.is_linking_verb()
+                    && !verb.kind.is_verb_progressive_form()
+                    && !verb.kind.is_verb_past_participle_form()
+                {
+                    continue;
+                }
+
                 let lint = self.match_to_lint(
                     &sentence[match_span.start..match_span.end],
                     document.get_source(),
@@ -208,6 +220,15 @@ mod tests {
     fn allow_standoff() {
         assert_lint_count(
             "In a tense standoff, Alex and his reflection engage in a battle of wills.",
+            OxfordComma::default(),
+            0,
+        );
+    }
+
+    #[test]
+    fn issue_2803_parenthetical() {
+        assert_lint_count(
+            "Any resemblance to real persons, living or dead, is purely coincidental.",
             OxfordComma::default(),
             0,
         );


hippietrail:
I'm pretty sure there are a number of issues related to Oxford comma rules getting it wrong. Best to see how many can be tackled in one PR. Or at least looking at them for common patterns or antipatterns etc.

mauropereiira:
Good prompt, I went and looked. There are four open Oxford comma issues, and this PR closes one of them. Tested against this branch:

| issue | example | on this branch |
| --- | --- | --- |
| #2803 | Any resemblance to real persons, living or dead, is purely coincidental. | fixed |
| #1308 | Last year, a friend and I were at the beach. | still flags `friend` |
| #2329 | It's A or B, B or C, D or E, or F or G. | still flags `D` |
| #659 | A dog, a cat and a pig. | still missed |

**#659 turned out to be two separate bugs, and neither is the tagger.** The POS tags are identical for the working and the broken case:

```
A dog, a cat and a pig.          DET NOUN DET NOUN CCONJ DET NOUN   missed
The dog, the cat and the pig.    DET NOUN DET NOUN CCONJ DET NOUN   flagged
```

The first cause is the `skip` block at the top of `lint`:

```rust
if let (Some(first), Some(second)) = (words.next(), words.next())
    && first.preposition
    && second.is_likely_homograph()
{
    skip = sentence.iter().position(|t| t.kind.is_comma())...
}
```

`a` and `an` both carry `preposition: true` in the dictionary, so a sentence opening with either one followed by a noun that is also a verb has everything up to the first comma discarded, taking the first list item with it. Minimal pair, same words, order swapped:

```
A pear, a dog and a cat.   flags `dog`
A dog, a pear and a cat.   missed
```

`pear` is noun only, `dog` is noun and verb, and `is_likely_homograph` needs more than one category. That is the entire difference.

The second cause is `one`, which tags as `NUM` rather than `DET`, so `then_determiner()` never matches it:

```
One dog, one cat and one pig.    NUM NOUN NUM NOUN CCONJ NUM NOUN   missed
```

Between them those two explain every line in your original report: `the` and bare plurals work, `a` and `one` do not.

**On tackling them together.** The false positives do share a root, I think. The expression treats every comma as a list separator and never asks what that comma is doing. #2803 is a closing parenthetical, #1308 is a fronted adverbial, #2329 is items that contain their own coordination. Three surface shapes, one gap underneath.

I am less sure one PR is the right container. The guard here keys on what follows the match, #1308 would need to look left at a sentence-initial adverbial, and #2329 needs the notion that a single item can hold its own conjunction. #659 is unrelated to all three, since those are false negatives coming from the guards rather than from the matcher.

If it suits you, my suggestion would be to keep this one scoped to #2803 while it is small and tested, let me take #659 separately since both its causes are contained and independent of this work, and then look at #1308 and #2329 together, since both are really about what counts as one item.

Happy either way though, including folding the lot into this PR if you would rather review it once.

hippietrail:
> Good prompt, I went and looked. There are four open Oxford comma issues, and this PR closes one of them. Tested against this branch:
> 
> issue	example	on this branch
> #2803	Any resemblance to real persons, living or dead, is purely coincidental.	fixed
> #1308	Last year, a friend and I were at the beach.	still flags `friend`
> #2329	It's A or B, B or C, D or E, or F or G.	still flags `D`
> #659	A dog, a cat and a pig.	still missed
> **#659 turned out to be two separate bugs, and neither is the tagger.** The POS tags are identical for the working and the broken case:
> 
> ```
> A dog, a cat and a pig.          DET NOUN DET NOUN CCONJ DET NOUN   missed
> The dog, the cat and the pig.    DET NOUN DET NOUN CCONJ DET NOUN   flagged
> ```
> 
> The first cause is the `skip` block at the top of `lint`:
> 
> ```rust
> if let (Some(first), Some(second)) = (words.next(), words.next())
>     && first.preposition
>     && second.is_likely_homograph()
> {
>     skip = sentence.iter().position(|t| t.kind.is_comma())...
> }
> ```
> 
> `a` and `an` both carry `preposition: true` in the dictionary, so a sentence opening with either one followed by a noun that is also a verb has everything up to the first comma discarded, taking the first list item with it. Minimal pair, same words, order swapped:
> 
> ```
> A pear, a dog and a cat.   flags `dog`
> A dog, a pear and a cat.   missed
> ```
> 
> `pear` is noun only, `dog` is noun and verb, and `is_likely_homograph` needs more than one category. That is the entire difference.
> 
> The second cause is `one`, which tags as `NUM` rather than `DET`, so `then_determiner()` never matches it:
> 
> ```
> One dog, one cat and one pig.    NUM NOUN NUM NOUN CCONJ NUM NOUN   missed
> ```
> 
> Between them those two explain every line in your original report: `the` and bare plurals work, `a` and `one` do not.
> 
> **On tackling them together.** The false positives do share a root, I think. The expression treats every comma as a list separator and never asks what that comma is doing. #2803 is a closing parenthetical, #1308 is a fronted adverbial, #2329 is items that contain their own coordination. Three surface shapes, one gap underneath.
> 
> I am less sure one PR is the right container. The guard here keys on what follows the match, #1308 would need to look left at a sentence-initial adverbial, and #2329 needs the notion that a single item can hold its own conjunction. #659 is unrelated to all three, since those are false negatives coming from the guards rather than from the matcher.

Good analysis. Maybe look into combining multiple linters. Let me look up how to do that... `merge_linters.rs` - basically you make two or more small linters that each tackle closely related problems, or different aspects of the same problem, then link them under a single lint rule/name. (You can even make hierarchies/trees - I think there's one three-level one in the repo.)

> If it suits you, my suggestion would be to keep this one scoped to #2803 while it is small and tested, let me take #659 separately since both its causes are contained and independent of this work, and then look at #1308 and #2329 together, since both are really about what counts as one item.
> 
> Happy either way though, including folding the lot into this PR if you would rather review it once.

One or multiple PRs or one PR with multiple commits. I don't feel comfortable approving AI PRs that are not ultra-simple or that I didn't rewrite by hand anyway so when it's at the right stage you'd want to request Elijah's review I think.

mauropereiira:
I went and read `merge_linters.rs`. The three-level one you're thinking of is `NounVerbConfusion`, which merges `NounInsteadOfVerb` and `VerbInsteadOfNoun`, and `NounInsteadOfVerb` itself merges `GeneralNounInsteadOfVerb` and `EffectAffect`.

Having read it, I don't think it's the right container for this family, and the reason is in the macro body. A merged linter runs each member and concatenates:

```rust
$(
    lints.extend(self.[< $linter:snake >].lint(document));
)*

remove_overlaps(&mut lints);
```

`remove_overlaps` sorts by priority and position and drops any lint whose span starts inside the previous one. So a member can only ever add lints. It has no way to cancel one another member produced.

That matters here because of what the four issues actually are. Three of them are false positives, and each one needs `OxfordComma` to stop emitting in a context it currently emits in. That's a guard inside the linter that emits, and no amount of merging reaches it. Only #659 is a false negative, and it's the one that could in principle live in its own detector.

The other thing pushing against it is that all four live in a single `OxfordComma`, 215 lines with one expression and one `lint`. Splitting it into disjoint detectors by list shape so that the guards become separate members is a real refactor, and I don't think three suppressions justify it. Say the word if you disagree, since you know the codebase far better than I do and this is the sort of thing I'd get wrong from the outside.

There are already two registered rules in this area, `OxfordComma` and `NoOxfordComma`. Merging those two under one name would be the obvious reading of "combine the Oxford comma linters", but they detect opposite faults, so I don't think that's what you meant.

So my plan stays what I proposed, unless you'd rather it didn't: keep this one on #2803, and take the others separately.

On that, **#659 is done and ready to go up as its own PR**. Two independent causes, both contained:

- `a` and `an` carry `preposition: true` statically, so a sentence opening `A dog, a cat and a pig.` was skipped to the first comma and lost its first item. Guarded by excluding first words that are contextually tagged `DET`, which leaves `While` and friends behaving as before since they tag `SCONJ`.
- Numerals quantify list items but carry no determiner metadata, so `one dog, one cat and one pig` never matched. Added a `NUM` branch alongside the determiner one.

It moves `The Great Gatsby` by one lint in each direction: it gains a genuine missing Oxford comma in a three-item list, and it loses a false positive on a sentence opening with a fronted `first,`. I'll open it once this one has landed or been reviewed, so they don't collide in the same file.

And understood on the approval, that's fair and I'd rather you said it than sat on it. I'll request Elijah's review here.

PR #4096, opened 2026-08-17
fix(core): six panics reachable from document text

PrazwalR:
> Disclaimer: this patch was produced with the assistance of an LLM agent, per
> AGENT_POLICY.md. Every crash below was reproduced against `harper-cli` before
> and after the fix.

## Problem

`harper-core` aborts on ordinary text. The release profile sets `panic = "abort"`,
so each of these is a SIGABRT (exit 134), not a catchable error. In `harper-ls`
that kills the language server mid-edit; in WASM the module traps and stays
unusable for the rest of the page session. There is no `catch_unwind` anywhere in
`harper-core`, `harper-ls`, `harper-cli` or `harper-wasm`.

| Input | Site |
|---|---|
| `barely b` | `char_string.rs` — `then_some` evaluates its argument eagerly |
| `both 2` | `expr/step.rs` — cursor past the end of the slice |
| `There'S cats` | `there_is_agreement.rs` — case-sensitive dispatch on a case-insensitive match |
| `would \t of` | `modal_of.rs` — `unreachable!()` on a reachable match length |
| `One \n criteria is essential.` | `singular_criteria_phenomena.rs` — fixed index past a variable-width whitespace step |
| `Damages` | `damages.rs` — `usize` underflow |

Two are also silent correctness bugs in release, not only aborts. `eq_ch` and
`eq_str` lowercase only their receiver, so `MultipleFrequencyAdverbs` reported
`always ALWAYS` as "redundant or contradictory" despite being one repeated word,
and `RedundantAcronyms` built `eq_str("Partys")` from the display-cased
`ACRONYMS` table — a lowercased receiver can never equal that, so the branch was
unmatchable rather than merely wrong.

## Root cause behind two of them

`condense_spaces` incremented its cursor twice per merge — once at the top of the
inner loop and again after pushing to `remove_these` — so it skipped the token
after each merge and left runs of three or more only partly condensed:

```
a<SP><TAB>b  ->  [Space(3), Space(1)]
a<TAB><SP>b  ->  [Space(3)]
```

`condense_newlines` had the identical structure. `then_whitespace()` consumes a
whole run, so rules indexing a fixed offset past it read a whitespace token
instead of the word they expected.

`Space` and `Newline` are distinct kinds that never merge, so whitespace crossing
a line break is still legitimately several tokens. The two rules that crashed are
hardened accordingly rather than relying on the condensing fix alone.

## Notes for review

Six commits, one per root cause, so they can be taken or dropped separately.

Roughly 264 `matched_tokens[i]` sites in `src/linting` share the fixed-width
whitespace assumption. I fixed the two that crash; the assumption is latent
elsewhere and probably deserves its own issue.

## Testing

- `cargo test` — 6375 passed, 0 failed, no snapshot changes
- `cargo clippy -- -Dwarnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes` — exit 0
- `cargo fmt --check` — clean
- Each input above verified to abort before the fix and lint cleanly after

## Breaking changes

None.

### DIFF
--- a/harper-core/src/char_string.rs
+++ b/harper-core/src/char_string.rs
@@ -305,7 +305,10 @@ mod tests {
     /// through `BarelyUn`.
     #[test]
     fn strip_prefix_handles_prefix_longer_than_haystack() {
-        assert_eq!(['b'].strip_prefix_ignore_ascii_case_chars(&['u', 'n']), None);
+        assert_eq!(
+            ['b'].strip_prefix_ignore_ascii_case_chars(&['u', 'n']),
+            None
+        );
         assert_eq!([].strip_prefix_ignore_ascii_case_chars(&['u', 'n']), None);
     }
 
--- a/harper-core/src/char_string.rs
+++ b/harper-core/src/char_string.rs
@@ -210,12 +210,14 @@ impl CharStringExt for [char] {
     }
 
     fn strip_prefix_ignore_ascii_case_chars(&self, prefix: &[char]) -> Option<&[char]> {
+        // `then` rather than `then_some`: the latter takes its argument by value and
+        // would index the slice even when the guard is false.
         (self.len() >= prefix.len()
             && self
                 .iter()
                 .zip(prefix)
                 .all(|(a, b)| a.eq_ignore_ascii_case(b)))
-        .then_some(&self[prefix.len()..])
+        .then(|| &self[prefix.len()..])
     }
 }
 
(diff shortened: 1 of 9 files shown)


hippietrail:
It came up some months ago that panicking should never happen in Harper when I put an `unreachable!` in somewhere related to hyphens and spellcheck I think, but it did turn out to be reachable and crashed `harper-ls`.

From memory this applied to anything the LSP could hit, so anything in `harper-core`. It might not apply everywhere, such as in `harper-cli`, which is a dev tool, and other places I couldn't really say.

I do know that I had spotted places before and after that where I either knew a panic could result or felt that one could but wasn't sure enough of my Rust knowledge to be sure.

At least in linters I know just `return None` for anything that should never happen. For other parts of the code I'm not sure how to handle unexpected situations in a way that can bring our attention to them but not panic.

@elijah-potter

PR #4149, opened 2026-08-20
fix(core): don't merge compound nouns across an open compound

### LINKED ISSUE #3950: `CompoundNouns` does not look left: "high school teacher" → "high schoolteacher"
**What got flagged?**

"high school teacher" → "high schoolteacher"

**Why is this incorrect?**

The join rule (`CompoundNouns`) corrects `school teacher` → `schoolteacher`. Inside the noun
phrase `high school teacher` this produces `high schoolteacher`, which is not
correct — the bigram is matched without considering the word to its left.

Control that shows it is context-blindness rather than a dictionary gap:
`high school student` is clean, because `school student` is not a closed
compound — the rule never consults `high` in either case.

**Example of correct usage:**

When the preceding token forms an established open compound with the
first word of the bigram (`high school`), the join should not fire. 

"high school teacher"

Reproduces at default configuration (harper.js 2.4.0)

mauropereiira:
# Issues

Fixes #3950

# Description

`CompoundNouns` corrects `school teacher` → `schoolteacher`, and inside "high school teacher" that produces "high schoolteacher".

The rule does consult the token to its left. `CompoundNounAfterDetAdj` needs a determiner or adjective there in order to fire at all, so "school teacher" alone is clean and "high school teacher" is not. What it never asks is what that left token is doing. When the left token forms an open compound with the first word of the pair, the compound boundary falls to the left and the pair spans two compounds rather than being one.

The guard is that question, asked once, against the dictionary handle `MergeableWords` already holds:

```rust
if self
    .split_expr
    .is_open_compound(&matched_tokens[0], &matched_tokens[2], source)
{
    return None;
}
```

`is_open_compound` joins the two tokens with a space and looks the result up in `FstDictionary::curated()`. That is the same lookup `get_merged_word` already performs for the pair being merged, applied one token to the left.

The guard needs dictionary coverage to do anything, so this also adds eight open compounds, chosen because each one can actually reach the rule (only determiners and adjectives get through the context expression) and each is an ordinary dictionary lexeme:

```
boarding school   coffee shop   front door      high school
middle school     open house    primary school  public school
```

Alternatives considered and rejected. A blanket "skip when the left token is an adjective" guard would disable the rule for its main context. Keying on specific words such as "high" would fake generality. An open-compound lexicon covering every such phrase is a much larger project, and it turns out not to be needed: only phrases whose left token is a determiner or adjective can reach this rule at all, which is a far shorter list.

# Demo

Fixed:

```
high school teacher      was: high schoolteacher
primary school teacher   was: primary schoolteacher
public school teacher    was: public schoolteacher
middle school teacher    was: middle schoolteacher
boarding school girl     was: boarding schoolgirl
high school work         was: high schoolwork
high school yard         was: high schoolyard
front door bell          was: front doorbell
back door man            was: back doorman
open house plant         was: open houseplant
coffee shop keeper       was: coffee shopkeeper
white paper work         was: white paperwork
first time keeper        was: first timekeeper
a public key server      was: a public keyserver
```

`back door`, `white paper`, `first time` and `public key` were already in the dictionary, so those four are fixed by the guard alone.

Unchanged, and still flagged:

```
The web cam captured a stunning image.          → webcam
The device features a responsive touch screen.  → touchscreen
A reliable back plane is essential.             → backplane
robust fire wall                                → firewall
a note book                                     → notebook
light house keeper                              → housekeeper
```

`light house keeper` stays flagged on purpose. `lighthouse` is the closed form, so `light house` is not a lexeme to add.

# How Has This Been Tested?

`cargo test -p harper-core`

```
test result: ok. 6165 passed; 0 failed; 290 ignored; 0 measured; 0 filtered out
```

No files under `harper-core/tests/text/` change. `audit-dictionary`, `cargo fmt --check` and `cargo clippy -- -Dwarnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes` are clean.

Because a guard like this can only suppress, the useful question is what it suppresses in text nobody wrote for the test suite. Two corpora, whole rule set rather than just `CompoundNouns`, before and after:

959 README files pulled through GitHub search (top-starred overall, plus recently created and recently pushed repos, and Python and Rust slices), 2,489,920 words:

```
base    127,224 lints
guard   127,223 lints
```

144 random Wikipedia articles, 121,035 words, since READMEs contain almost none of this phrase shape:

```
base    7,919 lints
guard   7,917 lints
```

Three lints removed across 2.6M words, none added. All three are this bug. The README one is in `nodejs/node`, "Alternatively, you can import them from a public key server", flagged as `public keyserver`. The two prose ones are "He was a high school teacher and later lectured at the University of Tasmania" in a biography.

The cost is on the other side, suppressing a suggestion someone wanted, and it is bounded by the roughly 500 two-word lexemes in the dictionary. Generating candidates from those, the ones that look genuinely arguable are `the sound wave length` (was `wavelength`), `a real time step` (was `timestep`) and `the white paper work` (was `paperwork`). None occurred in either corpus. I lean towards clean being right for all three, since "the length of the sound wave" is at least as likely a reading, but that is a judgement call worth flagging.

# AI Disclosure

The `public key server` test sentence is the line from the Node.js README that the corpus run surfaced.

# Checklist


harper-core/src/expr/mergeable_words.rs
```diff
@@ -31,6 +31,18 @@ impl MergeableWords {
...
+    ///
+    /// Used to tell "high school | teacher" from "high | school teacher": if the left pair is
+    /// already a compound, the boundary falls between them and the right pair must not merge.
+    pub fn is_open_compound(&self, word_a: &Token, word_b: &Token, source: &[char]) -> bool {
+        let mut compound: CharString = word_a.get_ch(source).into();
+        compound.push(' ');
+        compound.extend_from_slice(word_b.get_ch(source));
+
+        self.dict.get_word_metadata(&compound).is_some()
+    }
```
elijah-potter, line 44: This method can be private, assuming you execute my other comment.

harper-core/src/linting/compound_nouns/compound_noun_after_det_adj.rs
```diff
@@ -65,6 +65,15 @@ impl ExprLinter for CompoundNounAfterDetAdj {
...
         }
 
+        // "high school teacher": when the preceding word forms an open compound with the first
+        // word of the pair, the compound boundary falls to the left and the pair is not one noun.
+        if self
+            .split_expr
+            .is_open_compound(&matched_tokens[0], &matched_tokens[2], source)
+        {
+            return None;
+        }
```
elijah-potter: This check should be performed within the `MergeableWords` expression.

Doing so would retain the division of responsibility.

elijah-potter (changes requested):

mauropereiira:
Both done in d48f2afb.

The check now runs inside `MergeableWords::run`, right after the inner match and before the merge lookup, so it looks left from the cursor rather than being handed the tokens by the linter:

```rust
if cursor >= 2
    && tokens[cursor - 1].kind.is_whitespace()
    && tokens[cursor - 2].kind.is_word()
    && self.is_open_compound(&tokens[cursor - 2], &tokens[cursor], source)
{
    return None;
}
```

`is_open_compound` is private now, and `compound_noun_after_det_adj.rs` is untouched by this PR again.

Two things worth flagging from the move. It now covers `CompoundNounAfterPossessive` and `CompoundNounBeforeAuxVerb` as well, which is the point of putting it here but does widen the change. And output is unaffected: same 959 READMEs and 144 Wikipedia articles, whole rule set, byte-identical to the version you reviewed. 6,165 tests pass, `cargo fmt --check` and `cargo clippy -- -Dwarnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes` are clean.

PR #4181, opened 2026-08-21
feat: experimental `FastWordSet` based on `WordId`

hippietrail:
# Issues 

Related to several discussions of `WordSet` being expensive and problems with linters using large sets of words.

# Description

While looking into another idea using Google Search's built-in AI that knows a bit about Harper's internals, I stumbled across an idea on how to improve `WordSet` `Pattern` by storing `WordId`s instead of `char` `Vec`s.

This is an isolated `FastWordSet` `Expr` that is not integrated into any linter. Trying to do so turned out to be very tricky. But this idea might be helpful for exploring other ways to handle big lists of words in linters.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

I discussed the idea with Google Search's AI, then I sketched out an `Expr` module and sketched out an idea for an approach on how to use it, when the free AI agent built in to the Devin editor (formerly Windsurf) turned into a working example.

# Checklist


harper-core/src/expr/fast_word_set.rs
```diff
@@ -0,0 +1,161 @@
...
+    word_ids: HashSet<WordId>,
+    // We only use the inefficient character comparison for words not in the dictionary
+    oov_words: SmallVec<[CharString; 4]>,
+}
+
+impl FastWordSet {
+    /// Create a new FastWordSet from a list of words using any dictionary reference.
+    /// Words that exist in the dictionary are stored as WordIds for fast lookup.
+    /// Words that don't exist in the dictionary are stored as CharStrings for fallback matching.
+    pub fn new<D: Dictionary + ?Sized>(dict: &Arc<D>, words: &[&str]) -> Self {
```
elijah-potter, line 27: An initial version could just consume the curated dictionary.

elijah-potter (commented):
I like this! 

I actually tried a version of it a while ago, and I found that it wasn't much faster for the number of chars we already have in our rules. The cache efficience in the conventional `WordSet` is hard to overstate. I would love to see how this implementation performs on a larger list, though. Maybe we could call it `LargeWordSet` as an implementation that works faster on larger lists of words.

Overall, this is good work!

PR #4194, opened 2026-08-22
feat: might have not done→might not have done

### LINKED ISSUE #4182: The `ModalVerbs` `Pattern` includes nonsense words
I was starting work on a new linter involving modal verbs and I noticed the logic in `patterns::modal_verbs.rs` looked wrong where it creates negative forms. I made up a quick `Display` method with a dictionary lookup:

Image

That was probably my fault. I remember working on logic like that but don't remember it looking so broken at the time. I promise I wasn't drinking!

hippietrail:
# Issues 

Fixes #4182

# Description

Fixes the wrong word order with modal verbs in negative constructions like "should have not" to "should not have" etc.

I also rewrote the `ModalVerb` `Pattern` to fix #4182 and make it generally more robust.
I now knows about the various kinds of negatives and exceptions to the standard patterns and has much more thorough testing.

The pseudo-modal verb "need"/"needn't" was missing so I added it which required updating the `DoubleModal` linter, which was checking against "need" and passing by fluke.

I also exposed another `TokenKind`/`DictWordMetadata` method to `SequenceExpr` to test if regular verbs are in the common preterite+past participle form since they aren't always marked as past participles as needed in this linter.

----
I've revisited this since my changes to `WordSet` to work with iterators rather than only `&[&str]`, which makes it simpler IMHO.

# Demo
# How Has This Been Tested?

`cargo test`

# AI Disclosure

I asked AIs to refactor code I had working to make it more DRY, then iterated between me tweaking it and an AI tweaking it.

# Checklist


elijah-potter (commented):
This seems good, my only concern is the level of testing. Do you believe all of the tests in the `model_verb` module are necessary?

hippietrail:
> This seems good, my only concern is the level of testing. Do you believe all of the tests in the `model_verb` module are necessary?

I'll have a look. Generally the ones with clear rules only need a few tests but the hairy ones need a bunch of real-world tests to make it all data driven. Also the ones with specific paradigms I like to make sure each variant gets checked as I've seen things slip through before.

I don't remember which case this was but if it was the former I should be able to trim it right down.

PR #4203, opened 2026-08-23
feat(harper-core): add GetAccessTo lint rule

### LINKED ISSUE #3309: Should correct "get access at" to "get access to"

vjymisal0:
﻿> Disclaimer: This pull request was prepared with the assistance of an AI coding assistant (Antigravity).

Closes #3309

### Summary
Adds the `GetAccessTo` lint rule in `harper-core` to correct nonstandard usage of `access at` (such as `get access at`, `gain access at`, `have access at`, `obtain access at`, etc.) to `access to`.

### Details
- Matches combinations of access verbs (`get`, `gain`, `have`, `obtain`, `request`, `grant`, `give`, `provide`, `seek`, `secure`, `allow`, `deny`, `lose`) + optional modifiers/adjectives + `access at`.
- Replaces `at` with `to` while preserving case matching.
- Guards against common false positives:
  - Adverbial phrases such as `at all`, `at least`, `at once`, `at present`, `at will`, `at night`, etc.
  - Temporal adverbial constructs like `at this time`, `at that moment`, `at any point`, `at that stage`.
  - Scope expressions like `at the disk level`, `at the system level`.
  - URLs, domains, hostnames, and email addresses following `at`.
- Registered in `harper-core/src/linting/mod.rs`, `harper-core/src/linting/lint_group/mod.rs`, and synced with `harper-core/default_config.json`.
- Comprehensive unit tests added covering positive test cases and false-positive exceptions.

### DIFF
--- a/harper-core/src/linting/get_access_to.rs
+++ b/harper-core/src/linting/get_access_to.rs
@@ -127,9 +127,7 @@ impl Default for GetAccessTo {
                 .t_ws()
                 .then_optional(optional_modifiers)
                 .then_optional(optional_adjectives)
-                .t_aco("access")
-                .t_ws()
-                .t_aco("at"),
+                .then_word_seq(&["access", "at"]),
         }
     }
 }
@@ -152,13 +150,16 @@ impl ExprLinter for GetAccessTo {
         ctx: Option<(&[Token], &[Token])>,
     ) -> Option<Lint> {
         if let Some((_, after)) = ctx {
-            let following_words: Vec<&Token> = after
-                .iter()
-                .filter(|t| !t.kind.is_whitespace())
-                .take(4)
-                .collect();
+            let mut following_words = [None; 4];
+            for (slot, tok) in following_words
+                .iter_mut()
+                .zip(after.iter().filter(|t| !t.kind.is_whitespace()))
+            {
+                *slot = Some(tok);
+            }
+            let following_words_len = following_words.iter().flatten().count();
 
-            if let Some(first_tok) = following_words.first() {
+            if let Some(first_tok) = following_words[0] {
                 // If followed by URL, hostname, email, or unlintable token
                 if matches!(
                     first_tok.kind,
@@ -178,8 +179,8 @@ impl ExprLinter for GetAccessTo {
                 }
 
                 // Check for temporal adverbials like "at this time", "at any point", "at that moment"
-                if following_words.len() >= 2 {
-                    let second_chars = following_words[1].span.get_content(src);
+                if following_words_len >= 2 {
+                    let second_chars = following_words[1].unwrap().span.get_content(src);
                     if first_chars.eq_any_ignore_ascii_case_str(TEMPORAL_DETERMINERS)
                         && second_chars.eq_any_ignore_ascii_case_str(TEMPORAL_NOUNS)
                     {
@@ -189,7 +190,7 @@ impl ExprLinter for GetAccessTo {
 
                 // Check for level expressions like "at the disk level", "at the system level"
                 if first_chars.eq_any_ignore_ascii_case_str(TEMPORAL_DETERMINERS) {
-                    for following_tok in &following_words[1..] {
+                    for following_tok in following_words[1..].iter().flatten() {
                         let word_chars = following_tok.span.get_content(src);
                         if word_chars.eq_any_ignore_ascii_case_str(&["level", "levels"]) {
                             return None;
--- a/harper-core/src/linting/get_access_to.rs
+++ b/harper-core/src/linting/get_access_to.rs
@@ -1,0 +1,350 @@
+use crate::{
+    CharStringExt, Lint, Token, TokenKind,
+    expr::{Expr, SequenceExpr},
+    linting::{ExprLinter, LintKind, Suggestion, expr_linter::Chunk},
+};
+
+static ACCESS_VERBS: &[&str] = &[
+    "get",
+    "gets",
+    "got",
+    "gotten",
+    "getting",
+    "gain",
+    "gains",
+    "gained",
+    "gaining",
+    "have",
+    "has",
+    "had",
+    "having",
+    "obtain",
+    "obtains",
+    "obtained",
+    "obtaining",
+    "grant",
+    "grants",
+    "granted",
+    "granting",
+    "give",
+    "gives",
+    "gave",
+    "given",
+    "giving",
+    "request",
+    "requests",
+    "requested",
+    "requesting",
+    "provide",
+    "provides",
+    "provided",
+    "providing",
+    "seek",
+    "seeks",
+    "sought",
+    "seeking",
+    "secure",
+    "secures",
+    "secured",
+    "securing",
+    "deny",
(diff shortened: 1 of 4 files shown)


harper-core/src/linting/get_access_to.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+impl Default for GetAccessTo {
+    fn default() -> Self {
+        let access_verbs = SequenceExpr::word_set(ACCESS_VERBS);
+
+        let optional_modifiers = SequenceExpr::any_of(vec![
+            Box::new(SequenceExpr::default().then_determiner()),
+            Box::new(SequenceExpr::default().then_possessive_determiner()),
+            Box::new(SequenceExpr::default().then_quantifier()),
```
hippietrail, line 119: You can probably simplify this with `.then_kind_any()` ... oh actually the second two are just specializations of the first which is the generalization - you can test this if it gives you more confidence.

harper-core/src/linting/get_access_to.rs
```diff
@@ -0,0 +1,350 @@
...
+        ])
+        .t_ws();
+
+        let optional_adjectives = SequenceExpr::default().then_one_or_more_adjectives().t_ws();
```
hippietrail, line 123: Many adjectives are also other POSes so you'd want to think up some edge cases or run against a large corpus of non-contrived text to see if it trips over any adjective+ words.

harper-core/src/linting/get_access_to.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        Self {
+            expr: access_verbs
+                .t_ws()
+                .then_optional(optional_modifiers)
+                .then_optional(optional_adjectives)
+                .t_aco("access")
+                .t_ws()
+                .t_aco("at"),
```
hippietrail: You can combine these three lines with `.then_word_seq(&["access", "at"])`

harper-core/src/linting/get_access_to.rs
```diff
@@ -0,0 +1,350 @@
...
+        toks: &[Token],
+        src: &[char],
+        ctx: Option<(&[Token], &[Token])>,
+    ) -> Option<Lint> {
+        if let Some((_, after)) = ctx {
+            let following_words: Vec<&Token> = after
+                .iter()
+                .filter(|t| !t.kind.is_whitespace())
+                .take(4)
+                .collect();
```
hippietrail: We've got a range of methods available for avoiding allocating unless really necessary. Is there a reason you can't use a slice or iterator instead of allocating a new `Vec`?

harper-core/src/linting/get_access_to.rs
```diff
@@ -0,0 +1,350 @@
...
+                    || next_str.starts_with("https://")
+                    || next_str.starts_with("www.")
+                    || next_str.contains(".com")
+                    || next_str.contains(".org")
+                    || next_str.contains(".io")
+                    || next_str.contains(".net")
+                    || next_str.contains(".edu")
+                    || next_str.contains(".gov")
+                    || next_str.contains(".dev")
+                    || next_str.contains(".app")
```
hippietrail, line 212: Harper should be tokenizing URLs as their own token. If you're getting some as text that may be a bug to file.

hippietrail (commented):
Just some quick thought from looking at the code without digging right in to fully grok it.

harper-core/src/linting/get_access_to.rs
vjymisal0: combined via then_word_seq(&["access", "at"]) in 8e16c3f

harper-core/src/linting/get_access_to.rs
vjymisal0: switched to a fixed [Option<&Token>; 4] array instead of the Vec in 8e16c3f

vjymisal0:
Thanks, these were right. I updated the matcher to use `then_word_seq(&["access", "at"])`, removed the extra special cases, and avoided the allocation in the URL/false-positive guard. I also left the URL-token check in place and kept the string fallback only as a defensive guard for cases the tokenizer misses.

CI is green on the latest branch.

harper-core/src/linting/get_access_to.rs
vjymisal0: Good point! 	hen_determiner() indeed encompasses possessive determiners and quantifiers under the POS taxonomy. We kept it explicit during initial rule construction to be completely certain about quantifier matching, but simplifying to .then_determiner() keeps the sequence expression graph much cleaner. Happy to streamline this.

harper-core/src/linting/get_access_to.rs
vjymisal0: Thanks for pointing that out! To mitigate POS tagging ambiguities where adjectives overlap with nouns or particles, we added guards against common false-positive patterns like temporal adverbials ('at this time/moment') and scope qualifiers ('at the system/disk level'). Running test sweeps against documentation and real prose has kept false-positives at zero so far, and we will monitor closely for any other overlapping POS edge cases.

harper-core/src/linting/get_access_to.rs
vjymisal0: Good catch! We noticed during testing that certain bare URLs or links without scheme prefixes were occasionally tokenized as plain words rather than TokenKind::Url. The check here acts as a defensive guard so we don't emit spurious false positives for URLs. I'll file a separate issue with an isolated repro for Harper's tokenizer so that can be tracked and fixed upstream.

PR #4204, opened 2026-08-23
fix(missing_to): avoid false positive on participial verbs in prepositional phrases

### LINKED ISSUE #4132: Style: insert "to" to complete the infinitive wrong
**What got flagged?**
Image

“Ishethra,” says one who wears those ludicrous sunglasses and swim trunks, his mundane mortal flesh rune-carved, abs lined with ridges. “You’re under arrest for racketeering and attempted murder of Alek Suvor.”

"attempted" was flagged as a style issue.

**Why is this incorrect?**
- The text is in dialog
- The text is still gramatically connect. Maybe a "the" could be added before attempted, but still fine

**Example of correct usage:**
See text below screenshot

vjymisal0:
﻿> Disclaimer: This pull request was prepared with the assistance of an AI coding assistant (Antigravity).

Closes #4132

### Summary
Fixes a false positive in the `MissingTo` lint rule where participial controllers ending in `-d` or `-en` (e.g., `attempted`, `prepared`, `arranged`) inside prepositional phrases (such as `for racketeering and attempted murder`, `convicted of attempted robbery`, `with nicely arranged and prepared flowers`) were incorrectly flagged as missing `to`.

### Details
- Introduces `preposition_or_determiner_within_four` to check preceding context across conjunctions (`and`, `or`, `but`, `nor`) and participial nouns/gerunds (e.g. `racketeering`) for prepositions (`of`, `for`, `in`, `with`, `without`, `during`, `after`, `before`, `by`, `about`, `against`, `from`, `on`, `at`, `into`, `through`) as well as determiners.
- Suppresses false-positive `MissingTo` lints on past-participle controller words occurring within prepositional and determiner noun phrases.
- Added test cases covering:
  - `"You’re under arrest for racketeering and attempted murder of Alek Suvor."`
  - `"He was convicted of attempted robbery."`
  - Positive test verification: `"They attempted solve the problem without assistance."` -> correctly suggests `"They attempted to solve the problem without assistance."`

### DIFF
--- a/harper-core/src/linting/missing_to.rs
+++ b/harper-core/src/linting/missing_to.rs
@@ -242,28 +242,17 @@ impl MissingTo {
                     break;
                 }
 
+                if tok.kind.is_punctuation() {
+                    break;
+                }
+
                 let word = tok.get_str(source).to_lowercase();
                 let word = word.as_str();
 
-                if tok.kind.is_conjunction()
-                    || matches!(word, "and" | "or" | "but" | "nor")
-                    || tok.kind.is_verb_progressive_form()
-                {
-                    continue;
-                }
-
-                if tok.kind.is_determiner()
-                    || tok.kind.is_upos(UPOS::ADP)
+                if tok.kind.is_upos(UPOS::ADP)
                     || matches!(
                         word,
-                        "a" | "an"
-                            | "the"
-                            | "this"
-                            | "that"
-                            | "these"
-                            | "those"
-                            | "of"
-                            | "for"
+                        "of" | "for"
                             | "in"
                             | "with"
                             | "without"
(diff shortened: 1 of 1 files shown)


harper-core/src/linting/missing_to.rs
```diff
@@ -225,6 +225,57 @@ impl MissingTo {
...
+    ) -> bool {
+        let mut scan_cursor = controller_span_start;
+
+        for _ in 0..4 {
+            let Some((word, start)) = Self::previous_word_with_span(source, scan_cursor) else {
+                break;
+            };
+            let word = word.as_str();
+
+            if matches!(word, "and" | "or" | "but" | "nor") || word.ends_with("ing") {
```
hippietrail: Surely you'd want to use `.is_verb_progressive_form()` on the `Token`'s `TokenKind` and not just match "thing" and "sing" and "fling"?

hippietrail (commented):
Something jumps out as not quite right to me.

harper-core/src/linting/missing_to.rs
vjymisal0: Updated in b019ab8 — switched to inspecting preceding tokens from context and checking 	ok.kind.is_verb_progressive_form() instead of suffix-matching, plus added a regression test for words like 	hing.

vjymisal0:
Good call. I switched that check over to the token kind instead of matching a few  words by text. CI is green on the latest commit.

vjymisal0:
Small correction: the shell ate the inline code in my previous comment.

Good call. I switched that check over to the token kind (`tok.kind.is_verb_progressive_form()`) instead of matching a few `-ing` words by text. CI is green on the latest commit.

elijah-potter (approved):
Looks good. I have no issues.

PR #4211, opened 2026-08-24
refactor: convert two linter pairs to use `OneOfMany`

### LINKED ISSUE #4209: "Sneaked" and "Snuck" are currently both flagged
I think this is a regression resulting from the duplication of prefs logic between `lint_group/mod.rs` and `default_config.json` and choosing to keep the latter as the single source of truth.

It could be that the old way was able to set "prefer sneaked" to "enabled" and "prefer snuck" to "disabled" where the JSON file just sets `SneakedSnuck` to "true".

Or it could be a bug in how I implemented `SneakedSnuck` at the time and didn't spot and might not have been affected by the changes in how linter prefs were handled.

I do seem to recall noticing somewhere along the line that some kind of support got added for mutually exclusive linters but that it wasn't very thoroughly discussed or documented at the time and isn't really used. But I could be misremembering this too.

hippietrail:
# Issues 

Fixes #4209

# Description

This converts the `SneakedSnuck` linter and the `OxfordComma` & `NoOxfordComma` linter pair to use the `OneOfMany` config type where only one linter in a set will be enabled.

This fixes the problem I discovered earlier where both "sneaked" and "snuck" were being flagged to change to the other.

The solution involved not using `merge_linters` for sets of linters that use `OneOfMany`, which I didn't initially foresee, and might have some cons. Please feel free to analyse this solution and provide feedback.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

I wrote all the code but I used the AI agent built into the Devin editor to help track down parts of the problem, and suggest and critique approaches to fixing its various aspects.

I added tests to enforce that only half of each pair of linters is enabled by default.

# Checklist


LunaMeerkats:
Autonomous OpenAI Codex agent here. A heads-up on the new `to_flat_config` change at 6e5a8bb:

- exact-head `just test-rust` currently fails because the loop variable `name` is unused; and
- the indexed one-pass update can overwrite a selected duplicate name. For validator-valid `names = ["Same", "Same"]` with `choice = 0`, it sets `Same` to `true` and then back to `false`.

PR #4220 uses a two-phase update (initialize every name to `false`, then restore the selected name to `true`) and includes no-choice and duplicate-name regressions. Its 12 exact-head checks pass. This is just coordination evidence; either PR can be integrated in the way maintainers prefer.

elijah-potter (approved):
Looks great! Thanks!

PR #4213, opened 2026-08-24
docs: document the `OneOfMany` linter preference type

### LINKED ISSUE #4210: The `OneOfMany` linter preference type doesn't seem to be documented anywhere
I was planning to add more linters that handle verbs with multiple past participles like "hanged" vs "hung", "dived" vs "dove", "pleaded" vs "pled", and started looking at how I did the "sneaked" vs "snuck" one.

This led me to discover that both "sneaked" and "snuck" are now flagged: 

Image

Digging further in #4209 I discovered that this is due to the change in linter configuration from `lint_group/mod.rs` to `default_configuration.json` and the fact that there is a `OneOfMany` type of linter preference that was introduced.

I've looked manually, Googled in both `site:github.com/Automattic/harper` and `site:writewithharper.com`, and asked the built-in coding agent in the Devin fork of VS Code if it can find documentation for `OneOfMany` and we can't find any.

We could still be missing something but it seems to be undocumented and so far not used for any linter.

I have not dug into whether anything needs to be documented downstream for `harper.js` or any of the integrations.

MannXo:
> Per `AGENT_POLICY.md`, up front: an agent drafted this PR, including the
> documentation prose, working from the source in
> `harper-core/src/linting/lint_group/structured_config/`. Checkboxes below.

# Issues

Fixes #4210

# Description

`OneOfMany` was undocumented because the file it belongs to was undocumented.
`harper-core/default_config.json` had no reference page anywhere on the site, so
the only way to learn its three setting types was to read
`structured_config/mod.rs`. `Bool` and `Group` are guessable from 6000 lines of
examples. `OneOfMany` is not, since nothing in the file uses it yet.

This adds the missing reference and fills in the two other places `OneOfMany`
surfaces.

**New page, `/docs/contributors/default-configuration`.** Documents `Group`,
`Bool` and `OneOfMany`, with a field table for each and which keys may be
omitted. The `OneOfMany` section covers the parts that are not visible from the
type alone.

- Flattening enables only the chosen rule. The unchosen names get no entry in
  the `FlatConfig`, which reads as disabled, so there is no "two of these three"
  and no per-alternative fallback.
- Reading back the other way takes the *first* enabled name in `names`, so the
  order is meaningful rather than cosmetic.
- The two invariants (`labels` length, `name` membership) fail closed through
  `StructuredConfig::curated().unwrap()`, which is a panic rather than a
  diagnostic.
- Each alternative needs its own registered name, which rules out
  `merge_linters!`. This one cost time in #4209, so it seemed worth stating in
  the docs rather than leaving it to be rediscovered.

**`/docs/harperjs/configurerules`.** Adds a section on the structured view,
which had no documentation at all despite being exported and used by the Chrome
and Obsidian settings UIs. Covers `getStructuredLintConfig`, the fact that there
is no structured setter, and the write-back pattern for `OneOfMany`, where
leaving the unselected names untouched is how you end up with two conflicting
rules both reporting.

**Doc comments.** `Setting::OneOfMany` and `HumanReadableSetting::OneOfMany` now
describe the semantics and the field relationships instead of only naming the
variant.

**`/docs/contributors/author-a-rule`.** The "Update `default_config.json`"
section now points at the new page and says which setting type to reach for. Its
`docs.rs` link was also dead, since `lint_group` is a private module and the
`StructuredConfig` page is published under `harper_core::linting`.

**`AGENTS.md`.** New page added to the sidebar listing, and the
`configurerules` line updated, so the docs map stays accurate.

The examples use placeholder rule names on purpose, so this does not depend on
#4211 landing first. Happy to add a worked `OxfordComma` example on top once it
merges, or to fold that in here if you would rather the docs ship with a real
one.

# Demo

Prose only, no UI change beyond a new sidebar entry, so there is nothing much to
screenshot. Say the word if you want a render of the page.

# How Has This Been Tested?

```
cargo test -p harper-core --lib structured_config
    19 passed; 0 failed; 6526 filtered out

cargo clippy -p harper-core -- -Dwarnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes
    clean

cargo fmt --check
    clean

cargo doc -p harper-core --no-deps
    3 warnings, identical to master (verified by rerunning on a stashed tree).
    No new intra-doc link warnings, so the new links resolve.

biome check packages/web/vite.config.ts   (v2.3.3, the pinned version)
    Checked 1 file. No fixes applied.
```

Every behavioural claim on the new page was checked against the code rather than
inferred. Two were wrong on the first pass and were corrected.

- I assumed serde required the `name` key on `OneOfMany` because it has no
  `#[serde(default)]`. It does not. Missing `Option` fields deserialize as
  `None`, so `name` and `labels` are both genuinely optional. Verified against a
  standalone reproduction of the enum before writing it down.
- The 15 top-level entries in `default_config.json` are all `Group`, and the 849
  leaf settings are all `Bool`, checked by walking the JSON rather than assumed.

Not run locally: `just precommit` and `just build-web`. Both need `wasm-pack`,
and `pnpm install` cannot resolve `harper-wasm@workspace:*` until
`harper-wasm/pkg` has been generated. I did not want to install toolchain into
this machine unprompted, so those two gates are unverified by me and CI will be
the first to see them. The only non-Markdown file in the diff is
`packages/web/vite.config.ts`, and biome accepts it as formatted.

# AI Disclosure

I reviewed and approved the plan before any of it was written, and reviewed the
result. The words are the agent's.

Not applicable. No linter and no test sentences.

# Checklist

No tests, since the change is documentation and two doc comments. The existing
`structured_config` tests already pin the behaviour being described.

On splitting, I did consider it. The obvious cut is contributor docs in one PR
and the `harper.js` section in another, but both describe the same type and the
two pages cross-link, so splitting would put half a description in each. It is
about 200 lines of prose in five files. Say so if you would still rather have it
in two.


hippietrail:
Just to let you know, my PR to update sneaked vs snuck and Oxford comma linters are failing due to some problem with the Obsidian plugin that I'm having trouble understanding. This is despite `cargo test` succeeding for Rust.

So this may be something common that ought to be documented, but I'm not sure yet. Still trying to figure it out...

MannXo:
Refreshed onto current master and corrected one thing the refresh made necessary.

**The red check was not this branch.** `just test-firefox-plugin` failed on `Can ignore suggestion` in the chrome-plugin Playwright suite, expecting one `#harper-highlight` and finding none. That run is from 2026-08-27 against a base that is now 90 commits old, this branch contains no code change that can affect linting (the only Rust in it is doc comments, and the only other non-Markdown file is a nav entry in `packages/web/vite.config.ts`), and master's own Chrome Plugin workflow is green today. The push should re-run it against a current base.

**#4211 changed the behaviour this PR documents, while the PR was open.** `to_flat_config` used to write only the chosen rule; it now writes an entry for every name in the set, the chosen one `true` and the rest `false`. The merge was textually clean, so nothing flagged that two paragraphs of the new contributors page and one doc comment were now describing the old behaviour. Corrected in `bc53e71` against the current source rather than against the diff, including the no-selection case, which now writes every name as `false` instead of writing nothing.

The harper.js page is deliberately unchanged. Its advice to set the unselected names to `false` is about what a JavaScript caller should write, not about what flattening produces, and that is still correct.

Local checks on the new head: `cargo clippy -- -Dwarnings -D clippy::dbg_macro -D clippy::needless_raw_string_hashes` exits 0, `cargo fmt --check` exits 0, and `cargo test -p harper-core -- structured_config` is 19 passed.

hippietrail (approved):
This looks good to me but there are parts I do not have experience with so this requires approval from @elijah-potter before it can be merged.

PR #4238, opened 2026-08-27
fix(title-case): keep lowercase `x` as a connector word

pyrooka:
# Issues 

I haven't found any related issue and decided to open a PR straight. Feel free to close if you think this is not the expected behavior.

I ran into this problem while I was working on a Markdown document that's related to 2 different projects. I think it would make sense to treat single `x` characters as "conjunctions" since they usually act as connectors between 2 things.

# Description

Standalone lowercase 'x' between two words (e.g. 'Project 1 x Project 2')
is used as a connector, similar to 'by' or 'cross', and should not
be capitalized by the title-case rule.

The fix is case-sensitive: uppercase 'X' (e.g. 'Chapter X') is an
intentional label and remains unaffected.

# How Has This Been Tested?

Added new test cases and used `cargo test` to ensure everything works fine.

# AI Disclosure

# Checklist


elijah-potter (approved):
I don't have any immediate objections to this. Thanks!

PR #4253, opened 2026-08-30
Add a linter for confusing `wary` and `weary`

### LINKED ISSUE #4234: Need linter to catch mixing up "wary" and "weary"
"Wary" is related to "beware" and means to be cautious of something. "Weary" means "tired".

I'm willing to bet using "weary" when the meaning of "wary" is intended is common and the reverse is not.

I don't think this can be detected by either POS context or keywords in the context. And semantics are difficult.

But a gentle `message` guiding the user to make sure they're using the word they think they're using for every instance of "weary" might be OK.

Examples:
- I'm still a little bit **weary** of the fact that I haven't seen any contracts...

Jorge-Polanco-Roque:
# Issues
Fixes #4234

# Description
Adds the `WaryWeary` rule, mirroring the existing `HopHope` linter. `wary`/`weary` overlap in some contexts (both `wary of` and `weary of` are valid), so instead of guessing from POS/semantics it flags only **high-confidence fixed idioms** in both directions:

| Input | Suggestion |
|---|---|
| `weary eye` / `weary eyes` | `wary eye` (a cautious, watchful eye) |
| `bone wary` / `world wary` | `bone weary` / `world weary` (exhausted) |

Ambiguous uses such as `weary of` (tired of) are intentionally left untouched. Registered in the lint group and added to `default_config.json` (required by `curated_default_config_lists_every_registered_rule`) and the VS Code plugin settings.

# Demo
N/A — CLI linter rule; behavior is covered by the unit tests below.

# How Has This Been Tested?
8 unit tests: 5 corrections (both directions, mixed case) and 3 negatives (correct `wary eye`, correct `bone weary`, and the ambiguous `weary of`). The full `harper-core` suite (6323) passes; `cargo fmt` and `clippy` are clean.

# AI Disclosure

Thanks @hippietrail for the context-word data — really useful. I kept this first pass deliberately conservative (only bulletproof idioms) to avoid false positives, since `wary of`/`weary of` genuinely overlap. Happy to broaden the pattern set from your differentiator tool if you'd like more coverage in a follow-up.

### DIFF
--- a/harper-core/src/linting/wary_weary/mod.rs
+++ b/harper-core/src/linting/wary_weary/mod.rs
@@ -22,11 +22,14 @@ mod tests {
     }
 
     #[test]
-    fn corrects_weary_eyes_to_wary() {
+    fn does_not_touch_correct_weary_eyes() {
+        // "weary eyes" (plural) is a common, correct collocation for "tired
+        // eyes" and must not be flagged. Real-world testing showed this is the
+        // dominant usage, so the linter only targets the singular "weary eye".
         assert_suggestion_result(
-            "They watched with weary eyes for any danger.",
+            "She rubbed her weary eyes after the long shift.",
             WaryWeary::default(),
-            "They watched with wary eyes for any danger.",
+            "She rubbed her weary eyes after the long shift.",
         );
     }
 
--- a/harper-core/src/linting/wary_weary/mod.rs
+++ b/harper-core/src/linting/wary_weary/mod.rs
@@ -5,7 +5,7 @@ mod to_weary;
 use to_wary::ToWary;
 use to_weary::ToWeary;
 
-merge_linters!(WaryWeary => ToWary, ToWeary => "Handles common confusions between `wary` (cautious) and `weary` (tired), such as `weary eye` (should be `wary eye`) and `bone wary` (should be `bone weary`).");
+merge_linters!(WaryWeary => ToWary, ToWeary => "Handles common confusions between `wary` (cautious) and `weary` (tired).");
 
 #[cfg(test)]
 mod tests {
--- a/harper-core/src/linting/wary_weary/mod.rs
+++ b/harper-core/src/linting/wary_weary/mod.rs
@@ -1,0 +1,87 @@
+use super::merge_linters::merge_linters;
+
+mod to_wary;
+mod to_weary;
+use to_wary::ToWary;
+use to_weary::ToWeary;
+
+merge_linters!(WaryWeary => ToWary, ToWeary => "Handles common confusions between `wary` (cautious) and `weary` (tired), such as `weary eye` (should be `wary eye`) and `bone wary` (should be `bone weary`).");
+
+#[cfg(test)]
+mod tests {
+    use super::WaryWeary;
+    use crate::linting::tests::assert_suggestion_result;
+
+    #[test]
+    fn corrects_weary_eye_to_wary() {
+        assert_suggestion_result(
+            "She kept a weary eye on the exits.",
+            WaryWeary::default(),
+            "She kept a wary eye on the exits.",
+        );
+    }
+
+    #[test]
+    fn corrects_weary_eyes_to_wary() {
+        assert_suggestion_result(
+            "They watched with weary eyes for any danger.",
+            WaryWeary::default(),
+            "They watched with wary eyes for any danger.",
+        );
+    }
+
+    #[test]
+    fn corrects_mixed_case_weary_eye() {
+        assert_suggestion_result(
+            "He cast a Weary eye over the crowd.",
+            WaryWeary::default(),
+            "He cast a Wary eye over the crowd.",
+        );
+    }
+
+    #[test]
+    fn corrects_bone_wary_to_weary() {
+        assert_suggestion_result(
+            "After the march they were bone wary.",
+            WaryWeary::default(),
+            "After the march they were bone weary.",
+        );
+    }
+
+    #[test]
+    fn corrects_world_wary_to_weary() {
+        assert_suggestion_result(
+            "The old sailor had a world wary look.",
+            WaryWeary::default(),
+            "The old sailor had a world weary look.",
+        );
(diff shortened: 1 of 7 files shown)


hippietrail:
Welcome and thanks for this! Please use our PR template though and note the AI disclosure fields. I'm working on a tool to find context words that differentiate two confusable terms. For now it's limited to the top ten context words, and only one word of context on either side, so can only be a guide and not pasted directly of course. See if any of this fits:
image

Jorge-Polanco-Roque:
Thanks for the differentiator data — really helpful, and it actually reinforces the conservative scope here. Most of the surrounding context works for **both** words: `very/so/more/still/grow + weary` and `+ wary` are all valid, and `wary of` (cautious) vs `weary of` (tired) genuinely overlap, so those don't disambiguate. The clearest signal in your data is the right-side `eye`/`eyes`, which is exactly the `wary eye` idiom the rule already covers.

Because false positives in a grammar checker are costly, I kept this first pass to only near-bulletproof idioms:
- `weary eye`/`weary eyes` → `wary eye` (a cautious, watchful eye)
- `bone wary`/`world wary` → `bone weary`/`world weary` (exhausted)

If you'd like broader coverage, I'm happy to add specific patterns you consider safe from the tool's output in a follow-up — I'd rather grow it deliberately than risk flagging valid usages. Also updated the PR to use the template + AI disclosure, thanks for the nudge.

harper-core/src/linting/wary_weary/mod.rs
```diff
@@ -0,0 +1,87 @@
+use super::merge_linters::merge_linters;
+
+mod to_wary;
+mod to_weary;
+use to_wary::ToWary;
+use to_weary::ToWeary;
+
+merge_linters!(WaryWeary => ToWary, ToWeary => "Handles common confusions between `wary` (cautious) and `weary` (tired), such as `weary eye` (should be `wary eye`) and `bone wary` (should be `bone weary`).");
```
hippietrail: There's no need to include the examples in the `description`. This seems to be a thing AIs like to do as I've seen it often in my own work. Just the glosses will do.

hippietrail (changes requested):
I'd leave just the intent and the helpful glosses in the `description` and omit the current details/restrictions.

Jorge-Polanco-Roque:
Done — trimmed the `description` to just the intent and the glosses, and dropped the inline examples. Good call.

Jorge-Polanco-Roque:
Thanks for the review and the context-words tool — that's a neat approach.

I simplified both linter descriptions to just the intent and the glosses, dropping the idiom examples:
- `ToWary` → *"Detects `weary` (tired) used where `wary` (cautious) is intended."*
- `ToWeary` → *"Detects `wary` (cautious) used where `weary` (tired) is intended."*

I also synced the generated `vscode-plugin` `package.json` description, which still carried the old examples.

On the unit-test sentences: they're currently constructed rather than collected from real usage. If it'd strengthen the linter, I'm happy to replace them with real-world examples drawn from the differentiating context words (e.g. `wary eye` / `bone weary`) in a follow-up — just say the word.

hippietrail:
My only concern now is that it flags mistakes that are easy to spot but for which we have no clue if they're frequent or not.

Currently all the test sentences are contrived. To know if they are useful in the real world we need to test them against real world data.

In my experience agents are not good at Googling for examples, probably proper access to Google with `OR` and `site:` operators is only available with certain skills at certain price points. I've only tried with free agents. We have had one contributor who was able to get their agent to build a corpus to test against, so I guess it's possible depending what you/your agent has access to.

For my manually-crafted linters I manually hunt using Google especially on domains that are coding-relevant and have a high proportion of English errors. I try `site:github.com` first, then I try `site:slashdot.org OR site:news.ycombinator.com OR site:stackoverflow.com` and then I try `site:stackexchange.com OR site:medium.com OR site:reddit.com`

I also have a tool that scrapes README files of the trending repos on GitHub. It purposefully doesn't update them since they are likely to be edited over time to correct errors.

If you do this kind of real-world testing either manually or get your agent to do it, and add real-world examples, that would prove the utility of this linter. It may of course also help you improve the heuristics and logic to find more rear errors while avoiding false positives.

Jorge-Polanco-Roque:
Thanks @hippietrail — fair point, so I did the real-world testing instead of adding more contrived sentences.

**Findings (GitHub code search + web):**
- The confusion is real and documented (Merriam-Webster and QuillBot both have dedicated "weary vs wary" entries), so the linter is worth having.
- But the specific patterns are rare or risky in real data:
  - `weary eye(s) on` (the "watchful" mistake): ~0 hits in code search; the correct idiom `wary eye on` dominates.
  - `weary eyes` (plural): actually a common, correct collocation for "tired eyes" — one vocabulary corpus even lists it as a textbook "adjective + eyes" example. The old pattern flagged it, which would be a frequent false positive.
  - `bone wary` / `world wary`: ~0 hits, genuinely almost nonexistent in public text.

So rather than claim a frequency I cannot back up, I tightened the linter to cut false positives:
- `ToWary` now matches only the **singular** `weary eye` (the watchful idiom is always singular, "a wary eye"). Dropped `eyes`, since `weary eyes` = tired eyes is correct and common. Added a regression test asserting `weary eyes` is left untouched.
- Kept `bone/world wary` — high precision, harmless even if rare.

Honest take: this is a high-precision, low-frequency linter. Happy to drop it entirely if you'd rather only ship linters with demonstrated real-world frequency — no attachment either way.

Jorge-Polanco-Roque:
Hey @hippietrail — no rush at all, just circling back since it's been a couple of weeks. After your point about the test sentences being contrived, I went and checked real usage and ended up narrowing the rule to only the singular `weary eye` case; the plural/idiomatic ones that were risky are excluded now.

If you're still not convinced it trips people up often enough to be worth it, I'm genuinely happy to scope it down more or drop it — your call. Just didn't want it to quietly fall off the radar. Thanks for the thoughtful review either way.

PR #3857, opened 2026-07-21
fix(core): major issues with `SplitWords`

elijah-potter:
# Issues 

N/A

# Description

`SplitWords` is the single most commonly reported rule, besides `SpellCheck`. I've gone through the reports and addressed several of the most common issues. 

Those include:

1. Single letters are sometimes emitted as a whole word (i.e. `c ommitted`).
2. Cases that _should_ be handled by `SpellCheck` are overridden by `SplitWords`.
3. Missing or improperly tagged words in the dictionary.

# How Has This Been Tested?

Additional unit tests.

# AI Disclosure

- [x] I am a human and didn't use any AI.
- [ ] I used LLM features of my editor, but not an agent.
- [ ] I used an AI agent interactively.
- [ ] I am an agent or I got an agent to do the work autonomously.

# If Your PR Implements or Enhances a Linter

- [x] I made up the sentences in the unit tests.
- [ ] The sentences in the unit tests were generated by an AI.
- [x] I'm using examples from the bug report / feature request.
- [ ] I collected real-world sentences for the unit tests.

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests to cover my changes
- [x] I have considered splitting this into smaller pull requests.

PR #4255, opened 2026-08-30
feat: publish some linter priorities that other linters refer to

hippietrail:
# Issues 
N/A

# Description

We already had a couple of linters that declared their priorities had to be higher or lower than a certain other linter. This was only noted in comments.

While investigating why "unvisible" was flagged by `SpellCheck` instead of `WrongNegative`, I wanted to do this again. So instead I've made doing this a bit more robust by publishing the priorities of two linters and the other linters can bump those up or down in logic rather than just in comments.

One closely related problem is that lower priority numbers mean higher priority. So one of the existing ones may have had this backward. I kept its logic as it was but added a comment stating this possibility.

Update: I added a test to make sure `WrongNegative` gets a shot at linting an error before `SpellCheck`. In doing so I discovered that `assert_lint_message` only failed if there was a lint message but it didn't match. I now also fails if there's no lint message at all. This revealed that two long-standing tests have been silently failing: `ves_the_right_message_for_canadian_english` in `out_of_the_window.rs` and `not_confident_proc_should_be_pro_c` in `split_words`.

I removed the one from `split_words` which tested that "proc" got corrected to "pro c" since splits resulting in single letters was specifically removed in #3857 - so I moved "proc" to that test.

The one in `out_of_the_window` had silently been backwards and is now correct.

One gotcha was that to test multiple linters on the same mistake, you have to manually call `remove_overlaps()`, which was not obvious to me.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist


hippietrail:
@elijah-potter I just noticed I've merged this by accident somehow but I also requested your review since I wasn't sure it was definitely a desirable change. Feel free to revert.

PR #4283, opened 2026-09-03
fix: add missing "themself" to the curated dictionary

### LINKED ISSUE #4269: Themself false positive
**What got flagged?**
"themself"

**Why is this incorrect?**
https://dictionary.cambridge.org/dictionary/english/themself

**Example of correct usage:**
Avery caught themself staring at the stranger.

rzaman8677:
## Summary
Per #4269, "themself" (a standard singular gender-neutral reflexive pronoun — [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/themself)) was flagged as a spelling error, since it had no entry in `harper-core/dictionary.dict`.

- Added `themself/~Ia3F.` to `dictionary.dict`, alphabetically before `themselves`, using the same flag pattern as `itself`/`myself`/`yourself` (3rd-person **singular** reflexive — `themselves` uses the plural `:` flag instead of `.`).
- `harper-core/src/expr/reflexive_pronoun.rs` explicitly listed `"themself"` in `BAD_REFLEXIVE_PRONOUNS` with the comment "not in dictionary.dict but commonly used anyway", and that module's own tests assert bad pronouns must **not** be in the dictionary. Moved it to `GOOD_REFLEXIVE_PRONOUNS` instead, which is exactly what the existing test comments say to do "if this changes."
- `harper-core/src/dict_word_metadata.rs` already had a test asserting `themself` has pronoun metadata, marked `#[ignore = "not in dictionary"]` — removed the `#[ignore]` now that it's true. That test also checked `"y'all'"` (a stray trailing apostrophe, not the real dictionary entry `"y'all"`), which would have kept the un-ignored test failing for an unrelated reason — fixed that typo too since it's a one-character fix directly blocking the same test.
- Added `dont_flag_themself` in `harper-core/src/linting/spell_check.rs` using the exact example from the issue, following the existing `dont_flag_pr`-style pattern.

## Testing
- `cargo test -p harper-core --lib`: 6365 passed, 0 failed, 303 ignored (one fewer than before, since the stale `#[ignore]` is now gone).
- Verified the new `dont_flag_themself` test fails on `master` (confirms it's exercising the fix, not a tautology) and passes with the dictionary entry added.
- `cargo fmt --check -p harper-core` and `cargo clippy -p harper-core --lib -- -D warnings`: clean.

Closes #4269

### DIFF
--- a/harper-core/dictionary.dict
+++ b/harper-core/dictionary.dict
@@ -49242,6 +49242,7 @@ them/~Ia3o:g        # I~pronoun a~personal 3~person :~plural o~object (g=them's 
 thematic/~JNQ
 thematically/Ry
 theme/~NSgVdG
+themself/~Ia3F.     # I~pronoun a~personal 3~person .~singular F~reflexive
 themselves/~Ia3F:   # I~pronoun a~personal 3~person :~plural F~reflexive
 then/~RJNgC
 thence/~
--- a/harper-core/src/dict_word_metadata.rs
+++ b/harper-core/src/dict_word_metadata.rs
@@ -1853,10 +1853,9 @@ pub mod tests {
 
         // nonstandard pronouns
         #[test]
-        #[ignore = "not in dictionary"]
         fn nonstandard_pronouns() {
             assert!(md("themself").pronoun.is_some());
-            assert!(md("y'all'").pronoun.is_some());
+            assert!(md("y'all").pronoun.is_some());
         }
     }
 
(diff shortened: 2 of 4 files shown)


hippietrail:
Similar changes to this are already pending in a dictionary curation PR: #4248

elijah-potter:
Closing this in favor of @hippietrail's PR.

PR #4347, opened 2026-09-09
fix: give `MassPlurals` higher priority than `SplitWords`

### LINKED ISSUE #4346: "advices" is flagged by `SplitWords` instead of by `MassPlurals`
"advices" is a common mistake made by nonnative speakers who have no intuition of mass vs count nouns.

We have a linter, `MassPlurals` that deals with this but instead of flagging this error, another linter, `SplitWords` suggests splitting it into "adv" + "ices" or "ad + vices":

Image

hippietrail:
# Issues 
Fixes #4346

# Description

Prevents `SplitWords` from making a bad suggestion for "advices" when `MassPlurals` has the right suggestion.

This is done by exporting `SplitWords`'s lint priority which `MassPlurals` can then import to make sure it has a slightly higher priority (lower number).

I have a hunch that sooner or later we'll find cycles where a linter A needs higher priority than linter B to fix one mistake but linter B needs higher priority than linter A to fix another. At that stage a proper fix that fixes clashes between `SpellCheck` and other linters will need to be tackled.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/mass_nouns/mass_plurals.rs
+++ b/harper-core/src/linting/mass_nouns/mass_plurals.rs
@@ -3,7 +3,10 @@ use hashbrown::HashSet;
 use crate::{
     CharStringExt, Token, TokenStringExt,
     expr::{All, Expr, FirstMatchOf, FixedPhrase, SequenceExpr},
-    linting::{ExprLinter, Lint, LintKind, Suggestion, expr_linter::Chunk},
+    linting::{
+        ExprLinter, Lint, LintKind, Suggestion, expr_linter::Chunk,
+        split_words::SPLIT_WORDS_PRIORITY,
+    },
     spell::Dictionary,
 };
 
@@ -132,7 +135,7 @@ impl<D: Dictionary> ExprLinter for MassPlurals<D> {
             lint_kind: LintKind::Grammar,
             suggestions,
             message,
-            ..Default::default()
+            priority: SPLIT_WORDS_PRIORITY - 1, // higher priority (lower number) than split words
         })
     }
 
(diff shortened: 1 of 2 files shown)


elijah-potter (changes requested):
Would you mind inlining that new constant and setting the priority of `MassNouns` to a literal? I've found that this kind of import/export from independent parallel modules is kind of a code smell.

You can also add a test to `linting/lint_group.rs` to check that the specific text you're worried about gets corrected properly. 

Thanks for submitting this. The `SplitWords` rule has been the cause of the plurality of real complaints in the Chrome Extension reports (see table below).

Image

Most of the `SpellCheck` complaints are about rare words we don't include in the dictionary.

hippietrail:
> Would you mind inlining that new constant and setting the priority of `MassNouns` to a literal? I've found that this kind of import/export from independent parallel modules is kind of a code smell.

You mean each linter whose priority needs to be relative to another linter's priority should just have a local constant and maybe a comment about which other one/ones it needs to be higher/lower than?

Would putting them all in a central place also be a solution?

> You can also add a test to `linting/lint_group.rs` to check that the specific text you're worried about gets corrected properly.

I've done this in other linters that have a priority clash with another linter. Can do it here as well.

> Thanks for submitting this. The `SplitWords` rule has been the cause of the plurality of real complaints in the Chrome Extension reports (see table below).

Yes it's the most visible one but it's not the only one, and `SpellCheck` isn't the only other side of clashing pairs either. Though I've been leaning more and more towards the proper solution being to consider the ones that clash with `SpellCheck` to be like contributors of suggestions to `SpellCheck`. Basically `SpellCheck` is for words not in the dictionary, but so is `SplitWords` and a couple of others that come up a lot less often.

So I think some kind of rearchitecting of `SpellCheck` is probably the way to go. But if doing so we also want to look into multi-word corrections as in #2879.

elijah-potter (approved):
This looks great! Thanks for making those changes. I'll take a look at that other PR soon.

PR #4354, opened 2026-09-10
feat: add a debug printf to the linter asserts using `cfg!(debug_assertions)`

hippietrail:
# Issues 
N/A

# Description

I keep adding a debug printf in `linting::search_for_suggestion()` to print attempts that don't match the desired result.

I just discovered `cfg!(debug_assertions)`, which lets us have it there permanently without it affecting release builds.

# How Has This Been Tested?

`cargo test`

# AI Disclosure

# Checklist

### DIFF
--- a/harper-core/src/linting/mod.rs
+++ b/harper-core/src/linting/mod.rs
@@ -710,6 +710,8 @@ pub mod tests {
         // Check if we've reached the expected result
         if text == needle {
             return true;
+        } else if cfg!(debug_assertions) {
+            eprintln!(" 🔎 \"{text}\"");
         }
 
         // Lint current text and try each suggestion branch


harper-core/src/linting/mod.rs
```diff
@@ -710,6 +710,8 @@ pub mod tests {
         // Check if we've reached the expected result
         if text == needle {
             return true;
+        } else if cfg!(debug_assertions) {
+            eprintln!(" 🔎 \"{text}\"");
```
elijah-potter: My only feedback: Could we make it more obvious (to new contributors) what this means by replacing the emoji with some text? Something like, "Checking…".

hippietrail: > My only feedback: Could we make it more obvious (to new contributors) what this means by replacing the emoji with some text? Something like, "Checking…".

Of course. I just find that the majority of the time I see walls of output text amongst which it's hard to spot the things that might be of interest/important. (I used to hate emoji.)

----
I've added the text but also retained the emoji pending your re-review. Happy to change it again as you prefer...

PR #4433, opened 2026-09-21
feat: warn when a linter test has the same text for the error and the fix

hippietrail:
# Issues 
N/A

# Descriptiton

Yesterday when working on a fairly big `phrase_set_linter` I noticed that all my new tests were passing on the first try even though I thought there was going to be at least mistake.

A very common way to make your linting tests is with `assert_suggestion_result()` to enter the text with the mistake in it, copy it, paste it after the linter reference, and edit it to correct the mistake. When you're creating a bunch of similar tests for variants, it can be quite easy to overlook one.

It turned out that since `assert_suggestion_method()` and friends became recursive using the same engine that they checked on each iteration if the result so far matched the desired correction. But this started on the 0th iteration, so if the result text was still a direct copypaste of the error text, it would spuriously match and pass.

With this PR it will still pass but it will emit a ⚠️ warning in the console too. Let me know if I should change this to a fail.

# Demo

```
⚠️  Input and expected are both 'misstake' - is the test correct?
```

# How Has This Been Tested?

`cargo test`

I also manually tested it with a test linter and a correct matching test, a spurious matching test as mentioned, and a correct failing test. I did not keep these tests as they were `eprintln` only. Let me know if I should reinstate them as proper `assert` tests.

# AI Disclosure

I originally thought this bug was lurking in `phrase_set_corrections` testing logic. I used an AI to help me track it down to its true location but I then fixed it by hand.

# Checklist


elijah-potter (approved):
Looks good!

